PR #69, opened 2021-01-15
Simulation tests

sebgrijalva:
This PR includes testing of `simulation.py` and `simresults.py`, as well as some minor edits in both files.

Coverage is 98% for both, although it should be checked to see the testing has been done reasonably well.


pulser/tests/test_simresults.py
```diff
@@ -0,0 +1,99 @@
...
+def test_sample_final_state():
+    with pytest.raises(ValueError, match="undefined measurement basis"):
+        sim_no_meas = Simulation(seq_no_meas)
+        results_no_meas = sim_no_meas.run()
+        results_no_meas.sample_final_state()
+    with pytest.raises(ValueError, match="can only be"):
+        results_no_meas.sample_final_state('wrong_measurement_basis')
+    with pytest.raises(NotImplementedError, match="dimension > 3"):
+        results_large_dim = deepcopy(results)
+        results_large_dim.dim = 7
```
HGSilveri, line 82: This reminds me, should we make `SimulationResults` private? I guess otherwise the user might think this is okay to do. We can make properties for the ones we want to show, like `states`.

HGSilveri (commented):
Nice job on these! I've reviewed  the results `simulresults.py`, I have nothing to point. There was only one uncovered line and I've made a commit that solves that. 
As for `simulation.py`, there are currently three uncovered lines. Here are my comments on them:
 - line 166: We control what `coeffs` is, and we always make it a `ndarray`, so perhaps we can just surpress lines 165 and 166, as they represent an unnecessary check.
 - lines 264 and 268: Add tests for runs where you give an `initial_state` as a `QObjEvo` and as an array.

I'll review `simulation.py` after you take care of these.

pulser/simresults.py
```diff
@@ -63,6 +68,8 @@ def expect(self, obs_list):
 
         qobj_list = []
         for obs in obs_list:
+            if not hasattr(obs, "shape"):
```
lhenriet, line 71: I would have personnally checked for the possible types of obs.

lhenriet (approved):
Changes on simresults & simulation and tests look good

pulser/tests/test_simresults.py
sebgrijalva: `dim` and `size` are now private attributes. Added `states` as a property.

sebgrijalva: This is the `__init__()` so far:

```python
self._states = run_output
        self._dim = dim
        self._size = size
        if basis_name not in {'ground-rydberg', 'digital', 'all'}:
            raise ValueError(
                "`basis_name` must be 'ground-rydberg', 'digital' or 'all'."
                )
        self._basis_name = basis_name
        if meas_basis:
            if meas_basis not in {'ground-rydberg', 'digital'}:
                raise ValueError(
                    "`meas_basis` must be 'ground-rydberg' or 'digital'."
                    )
        self._meas_basis = meas_basis

        @property
        def states(self):
            return self._states
```

Is something like this what you had in mind?

HGSilveri: Yes, pretty much. There's only one small detail, which as to do `self._states` being a list and the `property` does not  protect its contents. Just return `list(self._states)` instead and use `self._states` instead of `self.states` internally.

HGSilveri:
@sebgrijalva Thanks for the updates, I'll finish the review.

pulser/simresults.py
```diff
@@ -39,20 +39,24 @@ def __init__(self, run_output, dim, size, basis_name, meas_basis=None):
...
             if meas_basis not in {'ground-rydberg', 'digital'}:
                 raise ValueError(
                     "`meas_basis` must be 'ground-rydberg' or 'digital'."
                     )
-        self.meas_basis = meas_basis
+        self._meas_basis = meas_basis
+
+    @property
+    def states(self):
+        return list(self._states)
```
HGSilveri, line 59: Add docstring (single-line is enough).

HGSilveri (approved):
Looks good to me overall, good job. One thing though: running `test_simulation.py` is taking quite a while, clocking in at around 11 seconds. Since we don't care about the result, it would be good if you could bring this down, either by simplifying the sequence , reducing the time steps or whatever else works.

pulser/simresults.py
sebgrijalva: Made this check explicit for `qutip.Qobj`and `numpy.ndarray`

sebgrijalva:
Thanks for all your comments, I'm merging this and will now begin the big merge to master.

PR #70, opened 2021-01-18
Simulation and SimulationResults Classes.

sebgrijalva:
This PR includes the `Simulation` class that simulates the time-evolution of a Pulser sequence and allows for some basic processing of the results (as an instance of a `SimulationResults` class). The numerics are facilitated by using `QuTiP` for solving the Schrödinger Equation.  It also includes tests for the corresponding files and a set of notebooks that show several example usecases using `Simulation`


HGSilveri (approved):
Thanks to the way we've been doing this, I don't think there's a file here that hasn't been already reviewed. On my part, I think this is ready to be merged!

HGSilveri:
By the way, there are a lot of commits here, I think it would be cleaner to "Squash and merge".

lhenriet:
Looks good to me ! Let's keep the commits, so that Lucas and Constantin also appear as contributors of master.

lhenriet (approved):

PR #77, opened 2021-02-01
Removing _chirps from Pulse

### LINKED ISSUE #75: Removing _chirps from Pulse?
I realise that the `Pulser._chirps()` method is a bit out of place, considering we don't use it anywhere. I need it for the QM translation, but I can also just have it there instead. 
The argument I have in favour of leaving it in Pulser is that it contains the information on how the detuning waveform will actually be done, which we might want to use for a more realistic display of the Sequence or even its simulation, but I don't know if this is something we want anyway. 

What's your take on this @lhenriet @sebgrijalva ? Do we keep it or not?

HGSilveri:
Following our discussion, this PR serves to eliminate the `_chirps` method from the `Pulse` class, as it doesn't have any use inside Pulser. 

Closes #75 .

### DIFF
--- a/pulser/pulse.py
+++ b/pulser/pulse.py
@@ -15,7 +15,7 @@
 import matplotlib.pyplot as plt
 import numpy as np
 
-from pulser.waveforms import Waveform, ConstantWaveform, RampWaveform
+from pulser.waveforms import Waveform, ConstantWaveform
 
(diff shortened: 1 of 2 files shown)


sebgrijalva (approved):
LGTM

PR #80, opened 2021-02-08
Variable retarget time and detuning adjustment

### LINKED ISSUE #78: Variable retarget time
To more closely mimic the actual implementation, the retarget times should vary according to the last time a "target" command was added to a channel. The only criteria is that two "target" commands have to be spaced 220 ns. A retarget time is only added to enforce this rule. This will mean, in almost every case, that the target times will go down to 0.

It will have an effect on some tutorials, other than that it should be fine.

@sebgrijalva @lhenriet The only question is: do we want it in Pulser v0 or can we leave it for the next release?

HGSilveri:
This change implements the variable retarget time, which now depends on when the last `target` command occurred. The `retarget_time` parameter specified on the device now represents the closest together two `target` commands can be in time - a delay is only introduced when its necessary to idle the channel for this condition to be met.

This change implied going through the tutorials to see if everything was working properly. In order to avoid doing this again later, I adjusted the value of the detuning in the `Chadoq2` channels. This last change turned out not to affect any notebook.

Relevant changes to notebooks:
- `creating_sequences` was adapted, since the `retarget_times` mostly disappeared.
- The CZ gate implementation no longer needs two parallel Rydberg local channels, so it was adapted to feature only one.

Closes #78 .


tutorials/applications/Control-Z Gate Sequence.ipynb
```diff
@@ -83,7 +83,7 @@
    "outputs": [
```
sebgrijalva, line 83: <p>Nice!</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pasqal-io/Pulser/pull/80/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='36'/>

sebgrijalva (approved):
This looks fine, It's good that the machine takes care of "resetting" the targeting while the sequence continues acting. Thanks for updating the simulation tests and the CZ gate tutorial.

PR #84, opened 2021-02-10
v0.1.0 release

HGSilveri:
These changes serve to prepare `master` for the release v0.1.0. They are simple modifications to: `setup.py`, the `README` and some docs pages.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,27 +16,24 @@ from setuptools import setup, find_packages
 
 setup(
     name="pulser",
-    version="0.0.1",
+    version="0.1.0",
     install_requires=[
         "matplotlib",
         "numpy",
         "scipy",
+        "qutip",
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -16,27 +16,24 @@
 
 setup(
     name="pulser",
-    version="0.0.1",
+    version="0.1.0",
```
sebgrijalva, line 19: I'm seeing `0.1.1` on TestPypi , is this relevant? I.e. does this version numbering correspond to the real Pypi numbering?

HGSilveri: This is the right one, I had to bump it up in TestPypi because I had to upload more than once and it doesn't let you repeat a version.

sebgrijalva (approved):

PR #87, opened 2021-02-17
Drawing blockade radius and connectivity graph

### LINKED ISSUE #74: Drawing the Rydberg radius
Just an idea that might be useful: we could add the option to draw the Rydberg blockade radius around the atoms in the Register, in typical unit-sphere fashion.

Further on, we could even combine this with a routine to extract the maximum Rabi frequency from a sequence, to automatically determine the Rydberg blockade radius. 

Another idea would be to take in the Rydberg radius and spit out the corresponding graph, like we do in MIS problems. Something like a `Register.to_graph(blockade_r)` method.

What do you think @lhenriet , @sebgrijalva , @cdalyac ?

HGSilveri:
Here's the implementation of the drawing of the blockade radius and the corresponding connectivity graph. I implemented it in some of the tutorials, which I would ask you to review as well to see if they benefit from it.

As a bonus, I added the method `PasqalDevice.rabi_from_blockade`  to get the Rabi frequency that gives a certain Rydberg blockade radius because I felt it could be useful when the user has a given array and is trying to figure out their Sequence's parameters.

Closes #74 .


sebgrijalva (approved):
Looks good, thanks for the addition!

PR #88, opened 2021-02-18
Create local operators by list comprehension.

sebgrijalva:
I changed the way the local operator builder works by. using list comprehension. This runs faster than using a loop after first creating a list of identities. No further changes in notebooks, although one can run the nb again and see the speed improves slightly (~5-10%)


HGSilveri (approved):
Looking good, nice job!

PR #90, opened 2021-02-25
Sequence Builder, Parametrized Objects and Variables

HGSilveri:
This PR introduces new features to the library, specifically:
- The `SequenceBuilder` class, which serves as a factory that can build Sequence objects 
- The `Variable` class, to be used as a variable inside a `SequenceBuilder` that conditions the return `Sequence`. For now, it can hold values of type `int`, `float` or `str` and can be a 1-D array with multiple entries (in which case features like item retrieval and iteration are also available). 
- The `ParamObj` class, which enables the parametrization of objects related to sequence building like `Waveform` and `Pulse`. Additionally, it adds support for conventional numeric operations involving variables or parametrized objects.

The only class that the user should have direct access to (i.e. should instantiate) is `SequenceBuilder`. All others are "hidden away" inside the new `pulser.parametrized` module.

The `SequenceBuilder` class is designed so that it's usage resembles the creation of a standard Sequence as closely as possible. A tutorial and documentation support on how to use it will come in a future PR.


HGSilveri:
@sebgrijalva The failed test is from SimResults, which is unrelated to these changes but rather a consequence of the latest version of QuTiP (4.5.3). Can you please look into it and push the fix to this branch?

sebgrijalva (approved):
I have been using the builder in a couple of applications and it's working very well. From my side, this is a good enhancement to Pulser, thanks a lot!

HGSilveri:
Thanks for the review! I'm questioning whether it is a good idea to merge it into the master branch right away, mainly because I don't want to release it in the next version like this but there's no going back after we merge. I'm thinking of temporarily closing the PR and we use the branch like we used `simulation` in the past. Then, when it's time to actually merge to master, I'll just reopen it or make a new one. What do you think?

By the way, I feel like I should say that this was not a waste of time for you. Your review was still quite needed so that we're sure that we'll be building on a solid foundation moving forward.

sebgrijalva:
> use the branch like we used `simulation` in the past.

Ok, let's keep it in a branch so that we may keep tinkering and possibly suggest other changes.

PR #92, opened 2021-02-26
Fixing compatibility issue with latest version of QuTiP

HGSilveri:
The latest version of `qutip` (4.5.3) does not allow a direct call of `np.abs` on a `QObj`, so this is a fix where the data is explicitly retrieved and turned into a `np.ndarray` beforehand.

@sebgrijalva : Before we merge this, can you think of anywhere else where something like this might be happening?

### DIFF
--- a/pulser/simresults.py
+++ b/pulser/simresults.py
@@ -125,7 +125,7 @@ class SimulationResults:
 
         N = self._size
         self.N_samples = N_samples
-        probs = np.abs(self._states[-1])**2
+        probs = np.abs(self._states[-1].data.toarray())**2
         if self._dim == 2:
             if meas_basis == self._basis_name:
                 # State vector ordered with r first for 'ground_rydberg'


pulser/simresults.py
```diff
@@ -125,7 +125,7 @@ def sample_final_state(self, meas_basis=None, N_samples=1000):
 
         N = self._size
         self.N_samples = N_samples
-        probs = np.abs(self._states[-1])**2
+        probs = np.abs(self._states[-1].data.toarray())**2
```
sebgrijalva, line 128: I'd suggest using `Qobj.full()` to access the ndarray, to comply with QuTiP's preferred method.

sebgrijalva (approved):

sebgrijalva:
Seems like the only place where we have such operations on `Qobj` . There is a lot of information regarding this change (which was triggered by a compatibility upgrade to Numpy 1.20) in qutip/qutip#1433 .

pulser/simresults.py
HGSilveri: Damn, I just saw the approval and didn't notice this suggestion before merging. I'll sneakily add this in when I bump the version.

PR #95, opened 2021-03-01
Increasing robustness of Register initialisation

HGSilveri:
There we no checks on the atom coordinates given to a Register upon initialisation and although these were verified by `PasqalDevice.validate_register()`, it still gave rise to bugs when calling the register methods e.g. when drawing a register with 3D coordinates or even 2D coordinates of type `int`.

These changes look out for invalid inputs, standardise how the information is stored and make the whole class less prone to spurious errors.


sebgrijalva (approved):
LGTM

PR #96, opened 2021-03-03
Method to get a simulation results' final state

HGSilveri:
By calling `SimulationResults.get_final_state()`, one can access the final state of a simulation. 
- The option `ignore_global_phase` modifies the global phase such that the largest term (in absolute value) is real. 
- The option `reduce_to_basis` let's the user reduce the state to the chosen basis by eliminating the states that don't belong (e.g. `reduce_to_basis=digital` will eliminate all terms in which one of the atoms is in |r>), as long as all the eliminated states have negligible populations.


pulser/simresults.py
```diff
@@ -59,6 +59,53 @@ def states(self):
...
+                ex_state = "0"
+            else:
+                raise ValueError("'reduce_to_basis' must be 'ground-rydberg' "
+                                 + f"or 'digital', not '{reduce_to_basis}'.")
+            ex_inds = [i for i in range(3**self._size) if ex_state in
+                       np.base_repr(i, base=3).zfill(self._size)]
+            ex_probs = np.abs(final_state.extract_states(ex_inds).full()) ** 2
+            if not np.all(np.isclose(ex_probs, 0)):
+                raise TypeError(
+                    "Can't reduce to chosen basis because the population of a "
```
sebgrijalva, line 102: This is OK, however, my perspective would be to leave it as a warning and let the user still obtain the reduction of the basis if she wishes so (there might be a particular reason to want to extract this information --or even keep the neglected part!).

HGSilveri: Hum... I don't know, I think a warning might be too weak here. Let's discuss this at the meeting tomorrow.

sebgrijalva (approved):
LGTM

PR #97, opened 2021-03-05
Allowing Blackman waveforms with negative area

HGSilveri:
When trying to program an Hadamard, where `\Omega=-\delta`, with Blackman waveforms, I noticed that we were forbidding Blackman waveforms with negative area. This is too strict of a requirement, so I'm relaxing it with this PR.


sebgrijalva (approved):
Looking good, thanks!

PR #101, opened 2021-03-15
Method for Blackman waveform creation from a maximum allowed value

### LINKED ISSUE #99: Blackman Waveform from maximum reachable value
The duration of a Blackman waveform is often irrelevant and what matters in fact is the maximum value it reaches, for it can influence the Rydberg radius on amplitude waveforms for Rydberg channels. 

The proposal here is a class method for the `BlackmanWaveform` where the defining arguments are the area and the maximum value reached, so the duration of the pulse adapts accordingly.

HGSilveri:
An alternative way of creating a BlackmanWaveform in which the user provides a `max_val` instead of a `duration`. The resulting waveform minimizes the duration while ensuring that `max_val` is not overcome. 

Closes #99 .

### DIFF
--- a/pulser/tests/test_waveforms.py
+++ b/pulser/tests/test_waveforms.py
@@ -138,3 +138,14 @@ def test_blackman():
     assert np.isclose(wf.integral, -2)
     assert np.all(wf.samples <= 0)
     assert wf == BlackmanWaveform(100, np.array([-2]))
+
+    with pytest.raises(ValueError, match="matching signs"):
+        BlackmanWaveform.from_max_val(-10, np.pi)
+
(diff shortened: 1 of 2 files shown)


Louis-PaulHenry (approved):
Seems good to me. 
Can there really be a case where the rounding errors surpasses `max_val` ?

HGSilveri:
Thanks for the review!

> Can there really be a case where the rounding errors surpasses max_val ?

Yes, it actually happens. The greatest source of deviation comes from the discretisation of the waveform, which makes its area be slightly below `0.42 * duration`. This makes the initial guess for `duration` be slightly smaller than it should, so we have to check with the actual waveform that we're not going over the `max_val`.

PR #103, opened 2021-03-16
Draw the graph by default, when given a blockade radius

HGSilveri:
As discussed, the graph is now drawn by default when given the blockade radius.

@sebgrijalva : I didn't remove the `draw_` from the args because, without them, it seemed to me that it would imply that `blockade_radius` was actually `draw_blockade_radius`. Do you see what I mean?


sebgrijalva (approved):
Looks good, thanks!

PR #105, opened 2021-03-16
Implement multiplication by scalar and negation on waveforms

### LINKED ISSUE #98: Special operators for waveforms
I think it would be a nice improvement to have some special operators for waveforms. I came up with these ones:
 - Negation: Doing `-wf` would return a Waveform with all it's samples' sign flipped
 - Multiplication: Doing `wf * 2` would multiply all it's values by 2. It doesn't make sense to multiply two waveforms.
 
 Not so sure these make sense:
 - Addition: Doing `wf + 2` would add 2 to all values. `wf1 + wf2` would concatenate two waveforms.
 - Absolute value
 - Power
 - Modulo

Feel free to discuss the pertinence of each or add more suggestions.

HGSilveri:
Implements negation, scalar multiplication and scalar division of waveforms. 

All operations change the waveform's samples, not the duration.

Closes #98 .

### DIFF
--- a/pulser/tests/test_waveforms.py
+++ b/pulser/tests/test_waveforms.py
@@ -122,9 +122,10 @@ def test_composite():
 
 
 def test_custom():
-    wf = CustomWaveform(np.arange(16))
+    data = np.arange(16, dtype=float)
+    wf = CustomWaveform(data)
     assert wf.__str__() == 'Custom'
-    assert wf.__repr__() == f'CustomWaveform(16 ns, {np.arange(16)!r})'
+    assert wf.__repr__() == f'CustomWaveform(16 ns, {data!r})'
 
 
 def test_ramp():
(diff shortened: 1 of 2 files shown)


HGSilveri:
@Louis-PaulHenry , are you up for another review?

Louis-PaulHenry:
Sure, I'll have a look

Louis-PaulHenry (approved):
Seems to be doing what we discussed (simple enough I guess!). Now that it's easy to rescale waveforms, would there be a use for a `renormalize`method that sets the total area to `area=1` (or a user defined value?

HGSilveri:
> Seems to be doing what we discussed (simple enough I guess!). Now that it's easy to rescale waveforms, would there be a use for a `renormalize`method that sets the total area to `area=1` (or a user defined value?

Thanks for the review! I think that could be useful, but it might not make sense for every waveform (i.e. a `RampWaveform` going from negative to positive can have an area of 0). Do you want to open an issue laying out what you had in mind?

PR #110, opened 2021-03-22
Getter For Hamiltonian 

### LINKED ISSUE #108: Hamiltonian from Simulation Object
It will be useful to have access to the Hamiltonian constructed from the `Simulation` object, specially since the `qutip.QobjEvo` object gives access to H(t) via its `__call__()` function. This could take the form of a getter from the `Simulation._hamiltonian` attribute.

sebgrijalva:
Creates a property for the Hamiltonian object, so that it doesn't get mixed up with the internal attribute which is used during the initialization.

Closes #108

### DIFF
--- a/pulser/simulation.py
+++ b/pulser/simulation.py
@@ -74,6 +74,11 @@ class Simulation:
         self._build_basis_and_op_matrices()
         self._construct_hamiltonian()
 
+    @property
+    def hamiltonian(self):
+        '''Get the Hamiltonian QObjEvo created from the sequence'''
+        return self._hamiltonian
+
     def _extract_samples(self):
         """Populate samples dictionary with every pulse in the sequence."""
 
--- a/pulser/simulation.py
+++ b/pulser/simulation.py
@@ -244,12 +244,17 @@ class Simulation:
         Args:
             time (float): The specific time in which we want to extract the
                     Hamiltonian (in ns).
+
+        Returns:
+            Qutip.Qobj: A new Qobj for the Hamiltonian with coefficients
+                    extracted from the effective sequence (determined by
+                    `self.sampling_rate`) at the specified time.
         """
         if time > 1000 * self._times[-1]:
             raise ValueError("Provided time is larger than sequence duration.")
         if time < 0:
             raise ValueError("Provided time is negative.")
-        return self._hamiltonian(time/1000).copy()
+        return self._hamiltonian(time/1000)  # Creates new Qutip.Qobj
 
     # Run Simulation Evolution using Qutip
     def run(self, initial_state=None, progress_bar=None, **options):
--- a/pulser/simulation.py
+++ b/pulser/simulation.py
@@ -74,11 +74,6 @@ class Simulation:
         self._build_basis_and_op_matrices()
         self._construct_hamiltonian()
 
-    @property
-    def hamiltonian(self):
-        '''Get the Hamiltonian QObjEvo created from the sequence'''
-        return self._hamiltonian
-
     def _extract_samples(self):
         """Populate samples dictionary with every pulse in the sequence."""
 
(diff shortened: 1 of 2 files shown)


sebgrijalva:
I will update later on with tests, I forgot to add them.

pulser/simulation.py
```diff
@@ -74,6 +74,11 @@ def __init__(self, sequence, sampling_rate=1.0):
         self._build_basis_and_op_matrices()
         self._construct_hamiltonian()
 
+    @property
+    def hamiltonian(self):
+        '''Get the Hamiltonian QObjEvo created from the sequence'''
```
HGSilveri, line 79: nit: Use `"""` instead of `'''`

pulser/simulation.py
```diff
@@ -74,6 +74,11 @@ def __init__(self, sequence, sampling_rate=1.0):
...
+        return self._hamiltonian
```
HGSilveri, line 80: Bear in mind that you will be giving direct access to the stored hamiltonian anyway, because it is a mutable object. If you want to protect the object stored in `self._hamiltonian`, you should return a copy instead (I believe `.copy()` works).

HGSilveri (changes requested):

HGSilveri:
By the way, this is returning the `QObjEvo`, right? Didn't we want to return the Hamiltonian at a specific time `t`? Maybe I misinterpreted that

sebgrijalva:
The Hamiltonian is now a method that returns a Qobj at a specific time (which is relative* to ns, according to the `Sequence` time units --for simulation we were using µs).  I am restricting the times to be within the bounds of the original sequence. QuTiP does an interpolation for returning intermediate values between the `sampling_rate` samples, so there is a dependance of the "quality" of the Hamiltonian at any time t with the `sampling_rate`.

*) Because of the interpolation, in principle any float value can be given, I didn't see a reason to enforce an `int`

pulser/simulation.py
```diff
@@ -238,6 +238,19 @@ def build_coeffs_ops(basis, addr):
...
+
+        Args:
+            time (float): The specific time in which we want to extract the
+                    Hamiltonian (in ns).
+        """
+        if time > 1000 * self._times[-1]:
+            raise ValueError("Provided time is larger than sequence duration.")
+        if time < 0:
+            raise ValueError("Provided time is negative.")
+        return self._hamiltonian(time/1000).copy()
```
HGSilveri, line 252: Did you check if the `copy` is necessary? It might be returning a new `QObjEvo` when you call it.

pulser/simulation.py
```diff
@@ -238,6 +238,19 @@ def build_coeffs_ops(basis, addr):
 
         self._hamiltonian = ham
 
+    def get_hamiltonian(self, time):
+        """Get the Hamiltonian created from the sequence at a fixed time.
+
+        Args:
+            time (float): The specific time in which we want to extract the
+                    Hamiltonian (in ns).
```
HGSilveri, line 246: Add `Returns` section

HGSilveri (approved):
Apart from these minor comments, this LGTM, good job!

pulser/simulation.py
sebgrijalva: Removed the `copy` method, since a new `Qobj` is indeed constructed from the `QobjEvo`

pulser/tests/test_simulation.py
```diff
@@ -196,6 +196,24 @@ def test_empty_sequences():
...
+    simple_seq.add(rise, 'ising')
+
+    simple_sim = Simulation(simple_seq, sampling_rate=0.01)
+    with pytest.raises(ValueError, match='larger than'):
+        simple_sim.get_hamiltonian(1650)
+    with pytest.raises(ValueError, match='negative'):
+        simple_sim.get_hamiltonian(-10)
+    # Constant detuning, so |rr><rr| term is C_6/r^6 - 2*detuning for any time
+    simple_ham = simple_sim.get_hamiltonian(143)
+    assert (simple_ham[0, 0] == Chadoq2.interaction_coeff / 10**6 - 2 * detun)
```
Louis-PaulHenry, line 214: Is `==` enough, or would `isclose` be more stable there ? (just asking, I'm not yet so familiar with the behavior of assert with respect to float precision)

Louis-PaulHenry:
This LGTM (I guess that my remark on precision is not really relevant)

PR #111, opened 2021-03-24
Incorporating SequenceBuilder into Sequence

HGSilveri:
This PR is the incorporation of `SequenceBuilder` into `Sequence`, following the realisation that having a different object for parametrized sequences was not necessary after all. 

The process is virtually identical to what it was with `SequenceBuilder`. What happens is that, once a variable is declared, the `Sequence` becomes flagged as "parametrized" and starts storing the build calls for deferred execution. Naturally, things like sequence drawing or some input validation is no longer possible (although more checks are possible now). Upon building, it returns a new, "regular" sequence, with the specified values attributed to the variables. 

"Regular" building calls are also being stored now, which will be useful for serialisation later on.

This renders `SequenceBuilder` obsolete, so it will be removed.

@sebgrijalva Once this go through, can you review the Sequence Builder tutorial? Here, I made only the necessary changes to make it run, but I guess the text will need a make over too.


pulser/sequence.py
```diff
@@ -258,6 +383,16 @@ def target(self, qubits, channel):
                 f"{self._channels[channel].max_targets} qubits at a time"
             )
 
+        if not self._building:
+            for q in qs:
+                if q not in self._qids and not isinstance(q, Parametrized):
+                    raise ValueError("All non-variable qubits must belong to "
```
bejito, line 389: There are variable qubits ?

pulser/sequence.py
```diff
@@ -375,14 +528,72 @@ def align(self, *channels):
...
+            missing_vars = all_keys - given_keys
+            if missing_vars:
+                raise TypeError("Did not receive values for variables: "
+                                + ", ".join(missing_vars))
+
+        for name, value in vars.items():
+            self._variables[name]._assign(value)
+
+        # Can't deepcopy with stored parametrized objects due to recursiveness
+        seq = copy.copy(self)
```
bejito, line 583: Why do you need to copy here since you are already deepcopying below ? Wouldn't a dummy `seq = Sequence()` be enough here ?

pulser/sequence.py
```diff
@@ -452,6 +674,14 @@ def _validate_pulse(self, pulse, channel):
             raise ValueError("The pulse's detuning values go out of the range "
                              "allowed for the chosen channel.")
 
+    def _reset_parametrized(self):
+        """Resets all attributes related to parametrization."""
+        self._building = True
```
bejito, line 679: What is this `_building` for exactly ? Could you add a small comment ?

bejito (commented):
Looks neat ! 
A few comments on the code. Overall, I feel that there are two things going in parallel and I'm confused on what is used for what:  `_save` vs `_store`, `self._calls` vs `self._to_build_calls`, ...

pulser/sequence.py
```diff
@@ -13,20 +13,70 @@
...
 from pulser._seq_drawer import draw_sequence
 from pulser.utils import validate_duration
 
 # Auxiliary class to store the information in the schedule
 _TimeSlot = namedtuple('_TimeSlot', ['type', 'ti', 'tf', 'targets'])
+# Encodes a sequence building calls
+_Call = namedtuple("_Call", ['name', 'args', 'kwargs'])
+
+
+def _save(func):
```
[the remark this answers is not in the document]
HGSilveri: You're absolutely right, we actually don't. I was trying to distinguish calls that you will always execute right away, which are saved always in `_calls`, and the ones who's location (`_calls` or `_to_build_calls`) depend on whether you're actively building the sequence or not (`self._building` is `True` or `False`). 

Once I moved on to the serialization, I realized that the cases where you use `_save` are actually better served with a manual addition to `_calls`, so I got rid of `_save` altogether. I have that ready in another branch, which I rebased from this one. I guess we can either let this be, knowing it will be modified on the next PR, or perhaps I merge the other branch with this one and we turn it into a bigger PR with `serialization` already included? Let me know what you prefer.

pulser/sequence.py
```diff
@@ -150,7 +234,43 @@ def declare_channel(self, name, channel_id, initial_target=None):
...
+
+            To avoid confusion, it is recommended to store the returned
+            Variable instance in a Python variable with the same name.
+        """
+        if name in self._variables:
+            raise ValueError("Name for variable is already being used.")
+        var = Variable(name, dtype, size=size)
+        self._variables[name] = var
+        self._building = False
+        self._var_calls.append(
```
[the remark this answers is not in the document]
HGSilveri: You're right again. I got rid of it once I did the serialization...

pulser/sequence.py
```diff
@@ -204,6 +330,7 @@ def add(self, pulse, channel, protocol='min-delay'):
         tf = ti + pulse.duration
         if ti > t0:
             self.delay(ti-t0, channel)
+            self._calls.pop()
```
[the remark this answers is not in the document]
HGSilveri: I don't mind option #1, I don't like popping the call either.

pulser/sequence.py
```diff
@@ -227,7 +354,9 @@ def add(self, pulse, channel, protocol='min-delay'):
         if pulse.post_phase_shift:
             self.phase_shift(pulse.post_phase_shift, *last.targets,
                              basis=basis)
+            self._calls.pop()
 
+    @_store
```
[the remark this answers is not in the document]
HGSilveri: It can be used directly, yes. On `Local` channels, before start adding pulses, you use the `target` command to select/change your target.

pulser/sequence.py
HGSilveri: I was thinking you can use a `Variable`, it works.

pulser/sequence.py
HGSilveri: Good question. I'm copying so that I start from how the sequence was before the first `Variable` was declared (when `self._building` switches to `False`). Also, channel declarations are not stored, so they always come along here. 

However, there's also the possibility of starting from the empty `Sequence` and executing the calls stored in `_calls`. This would amount to recreating the object from the start, the same way it'll be done when you deserialize it.

pulser/sequence.py
HGSilveri: It signals whether you're actively building the sequence or if your storing the calls for deferred execution. I'll add a comment.

Also, right now I'm flipping `_building` to `False` at the first variable declaration, which is makes a clear separation from when the `Sequence` is "regular" and when it becomes "parametrized". However, I realize this might be a bit premature, my other option would be to flip it at the first call that has a `Parametrized` object as an argument. Do you have any thoughts on this?

HGSilveri:
@bejito Thanks a lot for the review! As you'll see in my response to your first comment, I've since implemented the serialization on a different branch, where some of your remarks here were already addressed. I'm sorry about this, I should probably have waited for this branch to be merged, but I wanted to continue working on the serialization right away. 
Would you be in favour of closing this PR and opening a new one with these changes + the serialization? I would address all the unsolved issues you raised over there, naturally.

HGSilveri:
This PR was incorporated in PR #113 . All relevant comments and suggestions have been addressed there.

PR #113, opened 2021-03-31
Incorporating SequenceBuilder into Sequence + Serialization

### LINKED ISSUE #67: Serializing a Sequence
In order to send a `Sequence` over to the QPU, there will need to be some sort of serialization. Although JSON is the obvious choice (mainly because it's the only one I know...), I'm sure there are other options, which might be better suited for the job. What's your take on this, @bejito ?

Another thing to decide here is whether or not we want to deserialize back to a Pulser Sequence. Currently, the translation from Pulser to the QM program is built to take in the `Sequence` object, so that would be a point in favour of having a mechanism for deserialization. Apart from the work involved (which I would risk saying is not tremendous), do you find any points against building this?

HGSilveri:
This PR is an extension of #111 , in which `SequenceBuilder` capabilities are incorporated into `Sequence`, where serialization/deserialization support to JSON is added as well. This makes for a big PR, but the two aspects are connected in a few important ways, so a split is not very natural.

**SequenceBuilder into Sequence**

The `Sequence` behaves the way same as before, but now allows for the declaration of variables. Once a variable (or a `Parametrized` object created from a variable) is given as an argument to a `Sequence`-building call, the `Sequence` becomes flagged as parametrized, from which point onwards the calls are stored for deferred execution.

**Serialization/Deserialization**

JSON serialization is built around the serialization of a `Sequence`, so only the relevant objects support it. Each class that might require serialization has a `_to_dict()` method that tells the encoder how to serialize it. All encoding protocols for objects that don't support JSON serialization by the default are defined in `PulserEncoder`.

On the other hand, `PulserDecoder` specifies how to deserialize a `Sequence`, and all objects within it. The `Sequence` is serialized as a list of calls that fully define it, which are executed upon deserialization to retrieve an identical `Sequence` to that which was previously serialized.

Usage:
- Serialization: For a `Sequence` instance `seq`, call `s = seq.serialize()`, where `s` is the resulting JSON-formatted string.
- Deserialization: From `s`, call `Sequence.deserialize(s)` to get the encoded `Sequence` instance.

Closes #67 .


pulser/_json_coders.py
```diff
@@ -0,0 +1,83 @@
...
+            return obj
+
+        if (obj_name == "Variable" and
+           module_str == "pulser.parametrized.variable"):
+            var_name = obj["name"]
+            try:
+                var = self.vars[var_name]
+                assert var.name == var_name
+                assert var.dtype == obj["dtype"]
+                assert var.size == obj["size"]
```
bejito, line 58: The error raised by assert is not very explicit if I remember correctly. Might be worth having a more detailed error message.
edit: I seems you can add a custom message to assert :)
```python
assert var.dtype == obj["dtype"], f"variable {var_name} already defined with different parameters"
```

pulser/utils.py
```diff
@@ -15,6 +15,36 @@
...
+        args: If the object is to be built, the arguments to give on creation.
+        kwargs: If the object is to be built, the keyword arguments to give on
+            creation.
+
+    Returns:
+        dict: The dictionary encoding the object.
+    """
+
+    d = {
+        "_build": _build,
```
bejito, line 37: Just a thought, but since we are rebuilding all the object's internals just from its arguments, would it be worth having the version of pulser used ? Otherwise deserializing an object from one version to another version of pulser might lead to unexpected behaviours.
Having the version on each obj might way waay to redundant though, I'm sure we can think of something smarter, at the root level of the object. The root level is always a sequence, right ?

pulser/sequence.py
```diff
@@ -15,18 +15,71 @@
...
+                for y in x:
+                    verify_variable(y)
+
+        if self._is_measured and self.is_parametrized():
+            raise SystemError("The sequence has been measured, no further "
+                              "changes are allowed.")
+        # Check if all Parametrized inputs stem from declared variables
+        for x in chain(args, kwargs.values()):
+            verify_variable(x)
+        storage = self._calls if self._building else self._to_build_calls
```
bejito, line 79: I am litlle confused between `self._calls` and `self._to_build_calls`. What is each one used for ? They both seem to be used at the same time. Do we need both ?

bejito: If it is just about keeping track of which calls need to be build, maybe it would simpler to have an option in the _Call namedtuple ?
I see that when a sequence is parametrized, we use both `self._calls` and `self._to_build_calls` (e.g. with `declare_channel`). Splitting the calls between those two lists might become an issue with teh order of calls at some point no ? Since we don't have a way to keep the order of calls once they are split. I might be missing something though

pulser/_json_coders.py
HGSilveri: Cool! I didn't know this either, I will use it. I used `assert` statements more as a safety check that anything else, it seemed excessive to be raising a bunch of special errors. I agree the lack of an error message is a downside, but this solves that!

pulser/utils.py
HGSilveri: It's well thought. Indeed, the root level is always a Sequence, so we probably can "teach" the decoder to extract the version from there. From this point, if the version there does not match the running program's, can we change it and construct the object accordingly, or do we just raise an error?

pulser/sequence.py
HGSilveri: I think you got the main idea, but just to clarify: `self._calls` stores calls that are executed right when they're called, while `self._to_build_calls` are stored to be executed only upon building. 

In general, all calls start out being stored in `_calls` but once the sequence becomes parametrized, the calls are not executed anymore and go instead to `_to_build_calls`. The only exception here is `declare_channel` (`declare_variable` is a different exception since it is never stored), which does not need to be stored in `_to_build_calls` because:
1. It does not accept parametrized parameters.
2. The order at which it occurs doesn't really matter, because the channels are filled independently. 

You'll see that the saving of the `declare_channel` call is not done through the `_store` decorator, but rather "manually". So, in fact, it never ends up in `self._to_build_calls`, always in `_calls`. 

Therefore, the relevant order of the calls, as far as reconstructing an exact replica of the original `Sequence` is concerned, is always well-defined: 
1. Run the `_calls` 
2. Declare the variables 
3. Run the `_to_build_calls`

pulser/utils.py
bejito: I'm not sure we can change the version like that.
I think we might have to keep a list of versions compatible with the current one, and if the client uses another version we might have to throw an error. Maintaining a list sounds a bit cumbersome... There must be some better way

pulser/sequence.py
bejito: Ok, this works now because declare channels is always done before any parametrized call right ? 
Anyway, I think I got it :+1:

HGSilveri: Yup, that's it!

pulser/utils.py
bejito: Yeah, works for me :+1:

bejito (approved):

lhenriet (approved):

HGSilveri:
@sebgrijalva Now that this is approved, can you review the SequenceBuilder tutorial please?

bejito (approved):

PR #115, opened 2021-04-06
Effective pulse

Louis-PaulHenry:
Added the _sim_drawer.py file, to allow for the drawing of the effective pulse similarly to the sequence drawing, where the sampling rate is taken into account


HGSilveri:
Hey @Louis-PaulHenry ! It seems your code is failing the "style" test, which we enforce with `flake8`. You can install and run it to see which lines need changing (see [here](https://flake8.pycqa.org/en/latest/)) . I also highly recommend you install a linter in your text editor, which will tell you which lines are "wrong" in real-time, as you are coding.

PR #117, opened 2021-04-09
Reaching and enforcing 100% coverage

HGSilveri:
Since we'll start receiving outside contributions, I figured it's a good idea to ask for 100% unit test coverage. This PR reaches 100% coverage on the existing code and enforces 100% coverage on the CI tests.

There is still the option to use the `# pragma: no cover` comment on lines that can't be tested, although it should only be used under special circumstances.

@lhenriet @sebgrijalva There is not much to review, I'm requesting a review from your side as more of an acknowledgement of these changes.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,10 @@
+# .coveragerc to control coverage.py
+[report]
+# Regexes for lines to exclude from consideration
+exclude_lines =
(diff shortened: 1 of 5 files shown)


lhenriet (approved):
Good idea !

sebgrijalva (approved):
Agreed!

PR #122, opened 2021-04-09
Drawing of effective sequence

Louis-PaulHenry:
The draw_sequence method was modified to allow for the specification of an array of times_slices to be used in the Qutip solver


pulser/simulation.py
```diff
@@ -74,6 +75,16 @@ def __init__(self, sequence, sampling_rate=1.0):
...
 
+    def draw(self):
+        """Draws the simulation, showing both the input sequence and
+        the effective, constant-by-part (sampled) sequence used in QuTip.
+        """
+
+        # The effective time steps
+        time_slices = (1000*self._times).astype(int)
+
+        draw_sequence(self, time_slices)
```
HGSilveri, line 86: I think what you want here is 
```suggestion
        draw_sequence(self._seq, time_slices)
```
The testing error comes from there.

Louis-PaulHenry: The change is already made, the error comes from the test. I get 
`index 5062 is out of bounds for axis 0 with size 5011` that I'm still trying to understand

HGSilveri: I'll make a full review. In the meantime, use `pytest` locally to make sure all tests pass (or fix what's wrong) before pushing the changes.

Louis-PaulHenry: yes, I figured it's less spammy this way

pulser/_seq_drawer.py
```diff
@@ -74,11 +74,16 @@ def gather_data(seq):
     return data
 
 
-def draw_sequence(seq):
+def draw_sequence(seq, time_slices=np.array([])):
```
HGSilveri, line 77: It's a bit weird to put an empty array as a default. I'd rather have `None`. Then you can switch the `if len(time_slices) > 0` by just `if time_slices`

pulser/_seq_drawer.py
```diff
@@ -74,11 +74,16 @@ def gather_data(seq):
...
     """Draw the entire sequence.
 
     Args:
         seq (pulser.Sequence): The input sequence of operations on a device.
+
+    Keyword args:
+        time_slices(numpy.ndarray): The time slices at wich the sequence is
+        to be sampled. If this is provided, the constant-by-part Sequence
+        defined on these slices is also drawn.
```
HGSilveri, line 86: A typo and indentation:
```suggestion
    Keyword args:
        time_slices(numpy.ndarray): The time slices at which the sequence is
            to be sampled. If this is provided, the constant-by-part Sequence
            defined on these slices is also drawn.
```

pulser/_seq_drawer.py
```diff
@@ -138,6 +143,16 @@ def phase_str(phi):
...
+        if len(time_slices) > 0:
+            teff = []
+            yaeff = []
+            ybeff = []
+
+            for ti, tf in zip(time_slices, time_slices[1:]):
+                teff += [t[ti], t[tf]]
+                yaeff += [ya[ti], ya[ti]]
+                ybeff += [yb[ti], yb[ti]]
+
```
HGSilveri, line 155: I'm a bit confused about this. Doesn't this work?
```suggestion
            teff = t[time_slices]
            yaeff = np.array(ya)[time_slices]
            ybeff = np.array(yb)[time_slices]
```
Perhaps you can even convert `ya` and `yb` into `np.arrays` when they're first defined, not sure without trying.

pulser/simulation.py
```diff
@@ -74,6 +75,16 @@ def __init__(self, sequence, sampling_rate=1.0):
...
 
+    def draw(self):
+        """Draws the simulation, showing both the input sequence and
+        the effective, constant-by-part (sampled) sequence used in QuTip.
+        """
+
+        # The effective time steps
+        time_slices = (1000*self._times).astype(int)
+
+        draw_sequence(self._seq, time_slices)
```
HGSilveri, line 86: Does this fit?
```suggestion
        draw_sequence(self._seq, (1000*self._times).astype(int))
```
You would avoid defining `time_slices`, one less variable I guess.

HGSilveri (changes requested):
A couple issues here and there. I think the errors you're getting come from how you're creating `teff`, `yaeff` and `ybeff`.

pulser/_seq_drawer.py
Louis-PaulHenry: There was indeed an issue there, that the test discovered : if the time steps in t are not regularly spaced, it wouldn't work. I had to change it and it now seems to be working fine on my machine

pulser/_seq_drawer.py
Louis-PaulHenry: that was my first version, but it felt wrong somehow. I can definitely change it back though, less code is always better

pulser/simulation.py
Louis-PaulHenry: yes, I would just remove the comment on top then

pulser/_seq_drawer.py
HGSilveri: That's were you get me, are you sure that it's a "constant-by-part" shape we're doing? Is it not piecewise-linear instead?

Louis-PaulHenry: That's the impression I had after discussing with Sebastian, I'll check again, to be sure

Louis-PaulHenry: I've found the answer to that question in the `QobjEvo` documentation: 

> By default, a cubic spline interpolation will be used for the coefficient 
> at time t. If the coefficients are to be treated as step functions, use the arguments 'args = {"_step_func_coeff": True}'.

In the end, I think the "effective" pulse will always be close to the target one, since the `sampling_rate` only affect measurements, not the time evolution itself

Louis-PaulHenry: The question I have then is how our device behaves within this 4ns time window. If the user wants to compare

1. The pulse they input
2. The pulse the device would be _actually_ creating
3. The pulse Qutip is used in the simulation

HGSilveri: Even though the duration of the pulses has to be a multiple of 4ns, the device can define the amplitude at each 1ns. I would say that, at this level, it is irrelevant whether it changes in steps or ramps between each ns time step. For the detuning, it is defined as piecewise-linear ramps of 4ns. 

In any case, what happens on the device is not what we're looking for here, is it? This is supposed to be a realistic representation of the input to the simulation, not the device controller's output.

HGSilveri: Bear in mind that this is not terribly relevant. We are discretely modulating the amplitude of a wave of frequency in the tens of MHz with a sampling rate of 1 GHz. For the controller, all that really matters is the amplitude value at each time step, not what's in between.

Louis-PaulHenry: I agree. I have a strange issue raised by pytest when testing this using sthe spline interpolation :
apparently in one of the pulses used in the test, `np.array(data[ch]['time'])` is not an increasing array - namely there are three points like this one `3999.0 4000.0 3999.0` around 4000, 5000, 6000 and 7000.
Is it coming from the addition of pulses in the sequence creation?

pulser/_seq_drawer.py
```diff
@@ -132,11 +138,42 @@ def phase_str(phi):
...
+            ya2 = []
+            yb2 = []
+            for t_solv in solver_time:
+                while t_solv > t[t2]:
+                    t2 += 1
+                    if t2+1 < nt:
+                        if t[t2+1] == t[t2]:
+                            t2 += 1
+                ya2.append(ya[t2])
+                yb2.append(yb[t2])
```
HGSilveri, line 172: This block is a bit confusing to me, I don't really get why you're not just taking the values at `t_solv`. Can you elaborate on that, please? Maybe add a comment in the code too

HGSilveri (changes requested):
Nice job, this is looking good! Don't get scared by the amount of comments, they are mostly minor stuff

pulser/_seq_drawer.py
```diff
@@ -132,11 +138,42 @@ def phase_str(phi):
...
+            cs_amp[ch] = CubicSpline(solver_time, ya2)
+            cs_detuning[ch] = CubicSpline(solver_time, yb2)
```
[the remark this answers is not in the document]
Louis-PaulHenry: Good remark, there's no particular reason at this point, indeed

pulser/simulation.py
```diff
@@ -36,10 +37,15 @@ class Simulation:
     Keyword Args:
         sampling_rate (float): The fraction of samples that we wish to
             extract from the pulse sequence to simulate. Has to be a
-            value between 0.05 and 1.0.
+            value between 0.05 and 1.0
```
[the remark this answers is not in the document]
Louis-PaulHenry: Oop, my bad

pulser/simulation.py
```diff
@@ -74,6 +80,46 @@ def __init__(self, sequence, sampling_rate=1.0):
...
+                                 "negative values.")
+            # Ensure the list of times is sorted
+            eval_times = np.array(np.sort(evaluation_times))
+            if t_min > 0:
+                eval_times = np.insert(eval_times, 0, 0.)
+            if t_max < self._times[-1]:
+                eval_times = np.append(eval_times, self._times[-1])
+            self.eval_times = eval_times
+            # always include initial and final times
+        elif isinstance(evaluation_times, type(None)):
```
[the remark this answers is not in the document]
Louis-PaulHenry: That's true. Now that you mention it, I don't see why I would do it the way I did!

pulser/_seq_drawer.py
Louis-PaulHenry: This dates back to when there was a double counting at the jonction between two pulses. You've corrected it since then. I've just simply removed lines 168-170 now. I guess it's now clearer.

Louis-PaulHenry:
There are apparently some conflicts, but I don't have write access to resolve them

HGSilveri:
> There are apparently some conflicts, but I don't have write access to resolve them

You can do it locally and push. Just pull `develop`, merge with this branch and solve the existing conflicts.

HGSilveri:
After making sure that all changes to `pulser/simulation.py` were carried to `pulser/simulation/simulation.py`, you should delete it from this branch.

pulser/simulation/simulation.py
```diff
@@ -92,12 +93,12 @@ def __init__(self, sequence, sampling_rate=1.0, evaluation_times=None):
...
             else:
-                raise ValueError("Wrong evaluation time label ")
+                raise ValueError("Wrong evaluation time label. It should "
+                                 "be `Full` or `Minimal`")
         elif isinstance(evaluation_times, (list, tuple, np.ndarray)):
-            #            all(isinstance(n, int) for n in lst)
             t_max = np.max(evaluation_times)
             t_min = np.min(evaluation_times)
-            if t_max > 1000 * self._times[-1]:
+            if t_max > self._times[-1]:
```
HGSilveri, line 101: If `t_max` is supposed to be in ns, why remove the `1000 *`? Isn't `self._times` in µs?

HGSilveri (commented):
I'm confused about the units of `evaluation_times`, the docstring says it's in nanoseconds but then it is compared or copied from `self._times`, which I believe is in µs

pulser/simulation/simulation.py
Louis-PaulHenry: You're right. I started off as ns and then thought it was better for it to be closer to `self._times`. I've now corrected it to be in μs

pulser/simulation/simulation.py
```diff
@@ -36,10 +37,16 @@ class Simulation:
     Keyword Args:
         sampling_rate (float): The fraction of samples that we wish to
             extract from the pulse sequence to simulate. Has to be a
-            value between 0.05 and 1.0.
+            value between 0.05 and 1.0
+        evaluation_times (str,list): The list of times at which the quantum
+            state should be evaluated, in μs. If 'Full' is provided, this list
+            is set to be the one used to define the Hamiltonian to the solver.
+            The inital and final times are always included, so that f 'Minimal'
```
HGSilveri, line 44:
```suggestion
            The initial and final times are always included, so that if 'Minimal'
```

pulser/tests/test_simresults.py
```diff
@@ -105,7 +106,8 @@ def test_expect():
         results.expect([np.array(3)])
     op = [qutip.tensor(qutip.qeye(2),
                        qutip.basis(2, 1)*qutip.basis(2, 0).dag())]
-    assert len(results.expect(op)[0]) == duration
+    # only initial and final times are evaluated
```
HGSilveri, line 109: This comment doesn't make sense, does it? Isn't `number_of_meas = 1000`?

HGSilveri (approved):
Alright, apart from these nits, LGTM!

pulser/tests/test_simresults.py
Louis-PaulHenry: I don't remember editing these lines (ever), that's strange... @sebgrijalva ?

Louis-PaulHenry: It might be an unfortunate copy-paste in the wrong window on my side... I'd just delete the comment anyway

PR #125, opened 2021-04-14
Decorator to enable parametrization +  Parametrized Sequence tutorial

HGSilveri:
Following #124 , I realized that it's actually a good idea to have a decorator that enables support for `Parametrized` arguments in a function. Right now, I'm only using it in `classmethods`, but it can be used with any function. It's not supposed to be open to the general public though, it's just for internal use for now. @bejito , since you're the most familiar with this part, could you please review it? It should be fairly quick, there's not much code.

As for the tutorial, it's supposed to be included in the docs. @sebgrijalva , do you have time for a quick review?


HGSilveri:
Forgot to adapat the tests, will reopen when that's done.

HGSilveri:
This took longer than I thought it would, the serialization protocols for the parametrized class methods were a bit tricky. It should all be working now, sorry for the delay guys.

pulser/parametrized/paramobj.py
```diff
@@ -99,12 +100,30 @@ def build(self):
...
+                               _module=cls.__module__)
+
+        args = list(self.args)
         if isinstance(self.cls, Parametrized):
             cls_dict = self.cls._to_dict()
+        elif (hasattr(args[0], self.cls.__name__)
+              and inspect.isfunction(self.cls)):
+            # Check for parametrized methods
+            if inspect.isclass(self.args[0]):
+                # staticmethod or classmethod
```
bejito, line 114: In the case of static methods, are you sure the first arg is also a cls ?

bejito (approved):
Minor comment, otherwise LGTM

pulser/parametrized/paramobj.py
HGSilveri: You're right, I was thinking. This will only work for `classmethods`, I've fixed this.

HGSilveri:
Thanks a lot for the speedy review @bejito !

sebgrijalva (approved):
Tutorial looks nice!

PR #131, opened 2021-04-22
Removing Pasqal mentions + Relaxing restrictions on allowed devices for Sequences

HGSilveri:
- `PasqalDevice` was renamed `Device`
- Other mentions of Pasqal were omitted
- A `Sequence` no longer enforces the device to be imported from `pasqal.devices`, it just demands it be a `Device` instance. A warning is still issued in cases where the user makes their own device.


sebgrijalva (approved):
LGTM

PR #142, opened 2021-05-07
Relocating pulse restrictions and checks

### LINKED ISSUE #133: Making the restrictions on durations Device specific
Currently, the QM restrictions on durations of waveforms, pulses and delays are enforced every time. Since not all devices might necessarily need to obey such restrictions, we should move them to the `Device` class. In this way, the default duration restrictions will just be that they are `int`, seeing as we don't plan on going below 1 ns precision.

HGSilveri:
Encompasses multiple changes, most notably:

- Waveform, pulse and delay restrictions on duration are now an attribute of the `Channel` instances, and thus are only enforced upon addition to a channel. Consequently:
    - `validate_duration` was turned into a method of `Channel` (and `utils.py` was eliminated, with the remaining `obj_to_dict` function being moved to a new `pulser.json` module)
    - Upon creation, `Waveforms` now enforce only that their duration is an `int`, not a multiple of 4
    - Select `Waveform` subclasses now have a `change_duration` method, which returns a new waveform of the same type but with the new duration. This is only defined in cases where the change is unambiguous (ie when it is an argument upon initialization), like `BlackmanWaveform`, `ConstantWaveform` and `RampWaveform`. Inversely, `CustomWaveform` and `CompositeWaveform` do not have this method.
    - Upon addition of a pulse of non-conforming duration, the `add` method will attempt to adapt the waveforms through the `change_duration` method or raise an exception asking the user to do so manually when it fails.
- Pulse validation now happens inside a device, through the new `Device.validate_pulse()` method. When adding a pulse to a `Sequence`, this is what's called to validate it.

Regarding tutorials, in principle these changes shouldn't break them, but they will definitely change the warnings. Also, since I made the warnings more detailed (eg by including the values to which a duration was rounded to), they will appear much more often. It is probably worth it to revisit the tutorials after this, but for now, I feel like the PR is already too long.

Closes #133.


Louis-PaulHenry (approved):
LGTM!
I like a lot the change from multiple of 4 ns to simply integer, it's way cleaner this way.
What abaout the max duration of 67108864 ns? Is it just an arbitrary large number or is there anything deeper about it?

HGSilveri:
> LGTM!
> I like a lot the change from multiple of 4 ns to simply integer, it's way cleaner this way.
> What abaout the max duration of 67108864 ns? Is it just an arbitrary large number or is there anything deeper about it?

Thanks for the review! That max duration is the value enforced by the Quantum Machines controller.

PR #144, opened 2021-05-10
Adding the Microwave channel and operation mode

### LINKED ISSUE #72: Adding the MW channel and specific functionality
Adding the MW channel type for implementation of the XY Hamiltonian. Note that, unlike with the Rydberg and Raman channels, the MW channel cannot coexist with other channel types on the same Sequence, as it requires a specific initialisation procedure and works on a basis that is incompatible with the other channels. This will demand changes to the `Sequence` and `Simulation` classes.

HGSilveri:
- The `Microwave` channel is added as new type of `Channel`
- The `'xy'` basis is introduced (**this name is up for debate**)
- `Sequence` identifies whether it is working with `Microwave` channels or the other types. It decides based on the first added channel and forbids mixing channels that have different modes of operation.
- For now, raises errors in `Simulation` to clarify that a `Sequence` working in the XY Hamiltonian cannot be simulated.

Closes #72 .

### DIFF
--- a/pulser/channels.py
+++ b/pulser/channels.py
@@ -146,7 +146,7 @@ class Microwave(Channel):
     """Microwave adressing channel.
 
     Channel targeting the transition between two rydberg states, thus encoding
-    the 'xy' basis. See base class.
+    the 'XY' basis. See base class.
     """
     name: ClassVar[str] = 'Microwave'
-    basis: ClassVar[str] = 'xy'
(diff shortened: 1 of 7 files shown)


sebgrijalva (changes requested):
Thanks a lot for this addition! I left a few remarks, but otherwise this looks good for merging.

sebgrijalva (approved):
LGTM

PR #158, opened 2021-05-15
Small changes in the documentation (rectangle vs. square) and typo on "neutral"

LaurentAjdnik:
Uses `square(N_side, ...)` instead of `rectangle(N_side, N_side, ...)`

Corrects typo: `neutral` instead of `netural`


HGSilveri (approved):
LGTM, thanks for spotting these!

Issue #152, opened 2021-05-12
Reorganizing simulation class


julien-bremont:
Hi,

With @sebgrijalva, we are reorganizing the way `simulation.py` works to make noisy simulations possible (this is related to #151 ). Its `run` method now returns, for noiseless simulations, a `CleanResults` object, which is the analog of the previous `SimulationResults` in `v0.2` : it contains a list of `Qobj`. When adding noise, it will return a `NoisyResults` object, which contains no `Qobj` list but a probability distribution of the bitstrings at a given time. 

We added a `_config` field to the object in order to customize the simulation. You may choose the time at which the result is to be returned, the number of times the sequence is run (this is important for noisy simulations where noise is different each run), and the number of samples per run -this is solely to get more data in less computing time, at the cost of accuracy and realism.

HGSilveri:
Solved with #197 and #199 .

PR #161, opened 2021-05-17
Noise simulation

### LINKED ISSUE #151: Adding several noise models in the simulation method
Hi,

We're currently working with @sebgrijalva on noise models related to  
- SPAM (State Preparation And Measurement) errors
- Doppler-induced frequency shifting that affects the laser's detuning 
- Amplitude noise on gaussian lasers

We're also adding a possibility to simulate a laser sequence several times and average the results, since noise parameters are different at each run, to better simulate experiments in the lab. You may also change several noise parameters, be it the Doppler shift standard deviation, false-positive (negative) probabilities, etc.

julien-bremont:
Add SPAM errors, Doppler noise, and amplitude noise.
Modify `simulation` and `simresults`.
Add a Jupyter tutorial in tutorials/application.
Add a new `simulation` and `simresults` test.

Closes #151 and #152


tutorials/applications/Simulating Sequences with Errors and Noise.ipynb
```diff
@@ -0,0 +1,1437 @@
+{
```
sebgrijalva, line 1: <p>This must not appear in the final version.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pasqal-io/Pulser/pull/161/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

sebgrijalva, line 1: <p>Remains to add an "Introduction" section explaining what this notebook will show. What does "SPAM" mean (briefly)? Same for each type of noise</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pasqal-io/Pulser/pull/161/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

sebgrijalva, line 1: <p>Add a <code>;</code>  after the <code>plot</code>  function to avoid the output line from matplotlib</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pasqal-io/Pulser/pull/161/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

sebgrijalva, line 1: <p>Add correct citation</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pasqal-io/Pulser/pull/161/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='28'/>

sebgrijalva, line 1: <p>Explain noisy result from the amount of runs/samples per run</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pasqal-io/Pulser/pull/161/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='29'/>

sebgrijalva, line 1: <p>Comment the results (orange bars vs blue bars)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pasqal-io/Pulser/pull/161/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='53'/>

pulser/simulation/simulation.py
```diff
@@ -256,57 +296,214 @@ def get_hamiltonian(self, time):
...
+            self._prepare_spam_detune()
+        if noise_type == 'doppler':
+            self.init_doppler_sigma()
+        if noise_type == 'dephasing':
+            if self.basis_name == 'digital' or self.basis_name == 'all':
+                raise ValueError("Cannot include dephasing noise in digital "
+                                 "basis.")
+            self.init_dephasing()
+        # Register added noise(s)
+        self._noise.append(noise_type)
```
sebgrijalva, line 343: This list should not have repeated elements if one calls `self.add_noise('noise_type')` multiple times . Perhaps just throw a Warning `'Noise type already in place. Skipping instruction.'` and compare to `set(self._noise)` before updating

pulser/simulation/simresults.py
```diff
@@ -162,22 +355,23 @@ def sample_final_state(self, meas_basis=None, N_samples=1000):
...
+                "`meas_basis` can only be 'ground-rydberg' or 'digital'."
                 )
 
         N = self._size
         self.N_samples = N_samples
-        probs = np.abs(self._states[-1].full())**2
+        final_state = self._states[t]
+        # Case of a density matrix
+        # Don't take the modulus square in this case !
+        # if final_state.type != "ket":
```
sebgrijalva, line 370: In the case of the `'dephasing'` noise, the Lindblad Master Equation will be called, so there will be non-zero off-diagonal elements in the resulting density matrix. In that case, the diagonal should be taken into account (commented lines). This corresponds to the projective measurements in the computational basis.

tutorials/applications/Simulating Sequences with Errors and Noise.ipynb
julien-bremont: Added a more thorough introduction.

julien-bremont: Added some comments at the end of the tutorial

pulser/simulation/simulation.py
julien-bremont: Simply prints `'Noise type already in place. Skipping instruction.'` and does nothing.

pulser/simulation/simresults.py
julien-bremont: Indeed, now uncommented.

pulser/simulation/simulation.py
```diff
@@ -297,57 +337,222 @@ def get_hamiltonian(self, time):
...
+            time_indices = range(len(self.eval_times))
+            total_count = np.array([Counter() for _ in time_indices])
+            for _ in range(self._config['runs']):
+                # At each run, new random noise
+                if 'SPAM' in self._noise:
+                    self._prepare_spam_detune()
+                self._build_hamiltonian()
+                meas_basis = _assign_meas_basis()
+                # Get CleanResults instance from sequence with added noise:
+                res_with_noise = _run_solver(as_subroutine=True,
```
sebgrijalva, line 533: It is a bit awkward to call this `res_with_noise` although it is a `CleanResults` instance. Perhaps something like `clean_res_noisy_seq` would be more explicit to read (and would make clear the comment above)

pulser/simulation/simulation.py
```diff
@@ -297,57 +337,222 @@ def get_hamiltonian(self, time):
...
+                            N_samples=self._config['samples_per_run'])
+                         for t in time_indices]
+                else:
+                    total_count += \
+                        [res_with_noise.sample_state(
+                            t=t,
+                            meas_basis=meas_basis,
+                            N_samples=self._config['samples_per_run'])
+                         for t in time_indices]
+            prob = [Counter({k: v / (self._config['runs']
```
sebgrijalva, line 551: This could also be more explicitly named. What about instead of `total_count` above we call it `single_run_count` and instead of `prob` we call this `total_run_count`?

HGSilveri:
@julien-bremont All type annotations have been added in the `develop` branch, so you can now deal with the merge conflicts and add type hints to your own code as well. 

When you feel like the module is ready for review, let me know.

tutorials/applications/Simulating Sequences with Errors and Noise.ipynb
```diff
@@ -0,0 +1,899 @@
...
```
LucasGitQ, line 1: <p>Just comments on the text.</p><p>I would say "During the imaging process" rather than "During the simulation" in the eps' description. </p><p>Not sure about what you mean by "Amplitude noise". I would use this term for the random variation of Omega and another for the fact that the waist of the laser is finite. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pasqal-io/Pulser/pull/161/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

LucasGitQ, line 1: <pre class="ql-syntax" spellcheck="false">Line #4.    pulse = Pulse.ConstantPulse(duration, 1.5*2*np.pi, 0.*2*np.pi, 0)
</pre><p>Why is there a 1.5 in front of 2pi ? </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pasqal-io/Pulser/pull/161/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

LucasGitQ, line 1: <pre class="ql-syntax" spellcheck="false">Line #1.    obs = qutip.tensor(qutip.qeye(2),
</pre><p>Is this definition still accurate for 1 atom ? </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pasqal-io/Pulser/pull/161/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

LucasGitQ, line 1: <p>I would plot from 0 to 1 to actually see the impact of eta </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pasqal-io/Pulser/pull/161/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='31'/>

LucasGitQ, line 1: <p>Only an esthetic comment. Reduce 0.3333333333 to 0.33 as you did with the next graph</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pasqal-io/Pulser/pull/161/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='35'/>

LucasGitQ, line 1: <p>You may want to define more precisely what "not well-prepared" means. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pasqal-io/Pulser/pull/161/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='36'/>

LucasGitQ, line 1: <p>"the norm of the effective wavevector" but I'm being picky here</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pasqal-io/Pulser/pull/161/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='48'/>

julien-bremont:
@HGSilveri I merged latest type hints. I'll now add type hints to my own code and add some more tests.

tutorials/applications/Simulating Sequences with Errors and Noise.ipynb
julien-bremont: <p>Yup, this is much clearer.</p>

pulser/simulation/simulation.py
```diff
@@ -145,34 +154,73 @@ def prepare_dict() -> dict[str, np.ndarray]:
...
+                if noise == 'doppler':
+                    noise_det += np.random.normal(0, self._doppler_sigma)
+                # qubit qid badly prepared
+                if noise == 'SPAM' and self.spam_detune[qid[0]]:
+                    noise_det += det_spam
+                # gaussian beam for global pulses
+                if noise == 'amplitude':
+                    position = self._qdict[qid[0]]
+                    r = np.linalg.norm(position)
+                    w0 = 175.
```
lhenriet, line 178: How do I do if I want to change w0 ? Or if it depends on the device ?

pulser/simulation/simulation.py
```diff
@@ -314,60 +369,265 @@ def get_hamiltonian(self, time: float) -> qutip.Qobj:
...
+        # Reset any previous sequence:
+        self.reset_sequence()
+
+    def remove_all_noise(self) -> None:
+        """Removes noise from simulation"""
+        self._noise = []
+        # Reset any previous sequence:
+        self.reset_sequence()
+        self._collapse_ops = []
+        self.init_spam()
```
lhenriet, line 464: Why do you init_spam() here ?

pulser/simulation/simulation.py
```diff
@@ -44,28 +46,25 @@ class Simulation:
...
             state should be evaluated, in μs. If 'Full' is provided, this list
             is set to be the one used to define the Hamiltonian to the solver.
             The initial and final times are always included, so that if
             'Minimal' is provided, the list is set to only contain the initial
-            and the final times.
+            and the final times. Use a float to act as a sampling rate for
+            the resulting state.
     """
 
     def __init__(self, sequence: Sequence, sampling_rate: float = 1.0,
```
lhenriet, line 59: Just an idea here, open for discussion. 

The Simulation class is getting big. What about separating from the class all the configs elements related to noise ? Then, we could provide the necessary elements at initialization. Something like my_sim = Simulation(sequence, noise_config).

lhenriet (changes requested):
Hello Julien. Great job :) I made some comments.

pulser/simulation/simresults.py
```diff
@@ -60,6 +67,205 @@ def __init__(self, run_output: Sequence[qutip.Qobj], dim: int, size: int,
...
+        qobj_list = []
+        for obs in obs_list:
+            if not (isinstance(obs, np.ndarray)
+                    or isinstance(obs, qutip.Qobj)):
+                raise TypeError("Incompatible type of observable.")
+            if obs.shape != (2**self._size, 2**self._size):
+                raise ValueError("Incompatible shape of observable.")
+            qobj_list.append(qutip.Qobj(obs))
+
+        return qutip.expect(qobj_list, density_matrices)
```
HGSilveri, line 186: Try casting the output to `ArrayLike`:
```suggestion
        return cast(ArrayLike, qutip.expect(qobj_list, density_matrices))
```

pulser/simulation/simresults.py
```diff
@@ -60,6 +67,205 @@ def __init__(self, run_output: Sequence[qutip.Qobj], dim: int, size: int,
...
+        return Counter(
+               {np.binary_repr(i, N): dist[i] for i in np.nonzero(dist)[0]})
+
+    def sample_final_state(self, N_samples: int = 1000) -> Counter:
+        return self.sample_state(N_samples=N_samples)
+
+    def _standard_dev(self, op: qutip.Qobj) -> ArrayLike:
+        """Returns the square root of the variance of operator op."""
+        density_mats = [self.get_state(t) for t in range(len(self._states))]
+        return np.sqrt(qutip.variance(op, density_mats) / self.N_measures)
```
HGSilveri, line 216: Cast this to `ArrayLike` too.

pulser/simulation/simulation.py
```diff
@@ -314,60 +369,265 @@ def get_hamiltonian(self, time: float) -> qutip.Qobj:
...
+
+    @temperature.setter
+    def temperature(self, value: float):
+        # value set in microkelvin
+        self._temperature = value * 10**-6
+        self._calc_sigma_doppler()
+
+    @property
+    def spam_dict(self):
+        return self._spam_dict
```
HGSilveri, line 407: I would say the `property`s have a specific return type, while the `setter`s should return `None`. You tried this and it didn't work?

HGSilveri (commented):
A few things that might help with your `mypy` errors

pulser/simulation/simulation.py
julien-bremont: I added a new `SimConfig` class in an attempt to be clearer. It encapsulates noise types, temperature field, number of runs, and solver options.

pulser/simulation/simulation.py
julien-bremont: It is now done through the `SimConfig` class.

pulser/simulation/simulation.py
```diff
@@ -75,34 +74,44 @@ def __init__(self, sequence: Sequence, sampling_rate: float = 1.0,
...
         }
         self.operators = deepcopy(self.samples)
-
-        self._extract_samples()
-        self._build_basis_and_op_matrices()
-        self._construct_hamiltonian()
-
+        self._noise: List[str] = []
+        self._collapse_ops: List[qutip.Qobj] = []
+        self._temperature = 50e-6
```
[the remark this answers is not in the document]
julien-bremont: `temperature` now belongs to `SimConfig`, and is set regardless of the noise.

pulser/simulation/simulation.py
```diff
@@ -145,34 +154,73 @@ def prepare_dict() -> dict[str, np.ndarray]:
...
+            """constructs hamiltonian coefficients, taking into account, if
+                necessary, noise errors, which are local and depend
+                on the qubit's id qid"""
             _pulse = cast(Pulse, slot.type)
-            samples_dict['amp'][slot.ti:slot.tf] += _pulse.amplitude.samples
-            samples_dict['det'][slot.ti:slot.tf] += _pulse.detuning.samples
+            noise_det = 0
+            noise_amp = 1
+            # detuning offset related to bad preparation
+            # not too large, or else ODE integration errors
```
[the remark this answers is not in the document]
julien-bremont: I actually have no idea, but will try running some tests.

pulser/simulation/simulation.py
julien-bremont: Outdated.

pulser/simulation/simconfig.py
```diff
@@ -0,0 +1,129 @@
...
+            noise_types (string list): types of noises to be used in the
+                simulation.
+            samples_per_run (int value): number of samples per noisy run.
+                Useful for cutting down on computing time, but unrealistic.
+            runs (int value): number of runs needed : each run
+                draws a new random noise.
+            laser_waist (float): waist of the gaussian laser in global pulses.
+            solver_options (qutip.Options): options for the qutip solver.
+        """
+    def __init__(self, noise_types: List[str] = [], runs: int = 15,
```
HGSilveri, line 38: Don't forget it's best to avoid using `list` in the args, use `collections.abc.Sequence` instead. To avoid confusions with `pulser.Sequence`, I suggest import it like this:
```
import collections.abc as abc
```

and then use `abc.Sequence`.

pulser/simulation/simconfig.py
```diff
@@ -0,0 +1,129 @@
...
+                 samples_per_run: int = 5, temperature: float = 50.,
+                 laser_waist: float = 175.,
+                 solver_options: qutip.Options = qutip.Options(max_step=5)):
+        self.noise = noise_types
+        self.temperature = temperature
+        self.runs = runs
+        self.samples_per_run = samples_per_run
+        self.laser_waist = laser_waist
+        self.solver_options = solver_options
```
HGSilveri, line 47: This type of construction is quite suitable for a `dataclass`, I think it would be more elegant to use it. We use it already on `Device` and `Channel`, you should take a look

pulser/simulation/simconfig.py
```diff
@@ -0,0 +1,129 @@
...
+    def __str__(self) -> str:
+        s = ""
+        s += "Options:\n"
+        s += "----------       \n"
+        s += "noise types:       " + str(self.noise) + "\n"
+        s += "spam dictionary:   " + str(self.spam_dict) + "\n"
+        s += "temperature:       " + str(self.temperature) + "K" + "\n"
+        s += "number of runs:    " + str(self.runs) + "\n"
+        s += "samples per runs:  " + str(self.samples_per_run) + "\n"
+        s += "laser waist:       " + str(self.laser_waist) + "μm \n"
```
HGSilveri, line 59: nit: Each item would look better if starting with a capital letter (e.g. `Noise types`)

nit #2: It's best to avoid accumulating a string with `+` or `+=` (see why [here](https://google.github.io/styleguide/pyguide.html#310-strings))

pulser/simulation/simconfig.py
```diff
@@ -0,0 +1,129 @@
...
+
+    @noise.setter
+    def noise(self, noise_types: List[str]) -> None:
+        self._noise: List[str] = []
+        for noise_type in noise_types:
+            self._add_noise(noise_type)
+
+    def _add_noise(self, noise_type: str) -> None:
+        """Adds a noise model to the SimConfig instance, to be used
+            in Simulation.
```
HGSilveri, line 75: Don't forget, this needs to fit in a single line and be followed by a blank line?

pulser/simulation/simconfig.py
```diff
@@ -0,0 +1,129 @@
...
+            Args:
+                noise_type (str): Choose among:
+                    'dephasing': random phase (Z) flip
+                    'doppler': Noisy doppler runs
+                    'amplitude': Noisy gaussian beam
+                    'SPAM': SPAM errors. Adds:
```
HGSilveri, line 81: If these are the only options, `noise_type` would be better served if typed as a `Literal`. Look for a similar case with `PROTOCOLS` in `pulser.Sequence`.

pulser/simulation/simconfig.py
```diff
@@ -0,0 +1,129 @@
...
+        # Check proper input:
+        if noise_type not in noise_dict_set:
+            raise ValueError('Not a valid noise type')
+        self._noise.append(noise_type)
+
+    @property
+    def spam_dict(self) -> dict[str, float]:
+        return self._spam_dict
+
+    def set_spam(self, **values: float) -> None:
```
HGSilveri, line 95: Instead of `**values`, perhaps it would be more informative to specify that they are `eta`, `epsilon` and `epsilon_prime`. Use `Optional` type hints to let the user know they don't have to specify all three

pulser/simulation/simconfig.py
```diff
@@ -0,0 +1,129 @@
...
+            {'eta': 0.005, 'epsilon': 0.01, 'epsilon_prime': 0.05}
+
+    @property
+    def temperature(self) -> float:
+        return self._temperature
+
+    @temperature.setter
+    def temperature(self, value: float) -> None:
+        # value set in microkelvin
+        self._temperature = value * 10.**-6
```
HGSilveri, line 115: Use `1e-6` instead of `10.**-6`

pulser/simulation/simconfig.py
```diff
@@ -0,0 +1,129 @@
...
+            if values[param] > 1 or values[param] < 0:
+                raise ValueError('Invalid value : must be between 0 and 1')
+            self._spam_dict[param] = values[param]
+
+    def init_spam(self) -> None:
+        self._spam_dict: dict[str, float] = \
+            {'eta': 0.005, 'epsilon': 0.01, 'epsilon_prime': 0.05}
+
+    @property
+    def temperature(self) -> float:
```
HGSilveri, line 109: I was not sure if `dataclass` would work with a `setter`, but I tested it and it works

pulser/simulation/simconfig.py
```diff
@@ -0,0 +1,129 @@
...
+        self._calc_sigma_doppler()
+
+    @property
+    def doppler_sigma(self) -> float:
+        return self._doppler_sigma
+
+    def _calc_sigma_doppler(self) -> None:
+        # sigma = keff Deltav, keff = 8.7mum^-1, Deltav = sqrt(kB T / m)
+        self._doppler_sigma: float = \
+            8.7 * np.sqrt(1.38e-23 * self._temperature / 1.45e-25)
```
HGSilveri, line 125: nit: I usually dislike using `\`, so I suggest this instead:
```suggestion
        self._doppler_sigma: float = 8.7 * np.sqrt(
            1.38e-23 * self._temperature / 1.45e-25)
```
Maybe the indentation on the last line is wrong, see if you can fit it

HGSilveri (commented):
Hi Julien,

See here my comments on the new `SimConfig`. They mostly have to do with the style conventions we use throughout the code, along with some functional suggestion. I just highlighted the first conflicts I saw, so please take your time to apply these rules throughout the PR

pulser/simulation/simconfig.py
julien-bremont: It doesn't work for me : when I try to set the `temperature` field in a dataclass, I get : 
`    101     def temperature(self, value: float) -> None:
    102         # value set in microkelvin
--> 103         self.temperature = value * 1e-6
    104         self._calc_sigma_doppler()
    105 

TypeError: unsupported operand type(s) for *: 'property' and 'float' `

HGSilveri: I think that's because you have `self.temperature` instead of `self._temperature`

julien-bremont: None seem to work. I'm not all keen on using `properties` with `dataclasses`, it seems it's quite a hassle from what I've been researching. Maybe we should stick to a regular class.

HGSilveri: I find that odd, a `dataclass` is just a regular class with a default `__init__` function, using a `property` shouldn't be any different. Can you point me to your research on this?

julien-bremont: https://florimond.dev/en/posts/2018/10/reconciling-dataclasses-and-properties-in-python/
I could try doing this but it kinda feels like a weird workaround.

pulser/simulation/simconfig.py
```diff
@@ -0,0 +1,129 @@
...
+                 samples_per_run: int = 5, temperature: float = 50.,
+                 laser_waist: float = 175.,
+                 solver_options: qutip.Options = qutip.Options(max_step=5)):
+        self.noise = noise_types
+        self.temperature = temperature
+        self.runs = runs
+        self.samples_per_run = samples_per_run
+        self.laser_waist = laser_waist
+        self.solver_options = solver_options
+        self.init_spam()
```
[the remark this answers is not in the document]
julien-bremont: Irrelevant now that we agreed not to use a `dataclass`.

pulser/simulation/simulation.py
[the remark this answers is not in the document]
julien-bremont: It does slow it down a bit when choosing a large `eta`, i.e., more state preparation errors. 
For `eta = 0.5`, we have ~3s of running time when simulating 100 SPAM runs on a single qubit for Rabi oscillations with `det_spam = 150`, versus ~2.8s with `det_spam = 50`, which does essentially the same thing as an effective implementation of state preparation errors, seeing the plots I have.
Here for `det_spam=50` :
det_spam50
Here for `det_spam=150` :
det_spam150
That said, `eta = 0.005` is so small normally that even with a large number of runs and several atoms, the added cost wouldn't be that big. It is exaggerated here with such a large `eta` and still is pretty small. 
I still think we should go with `det_spam = 50.`, it seems to be the sweet spot.

HGSilveri:
@julien-bremont Just checking, are you done with the changes to this PR? Is this ready for review?

julien-bremont:
> @julien-bremont Just checking, are you done with the changes to this PR? Is this ready for review?

Yes, we've added what we wanted to add with @sebgrijalva. This is ready for review. I'll just have to update my branch regarding latest PRs.

HGSilveri:
@julien-bremont This is quite a huge PR, so I will probably do it in batches. After every review, I would ask you to see if the comments made also apply to other parts of the code and change those too.

pulser/__init__.py
```diff
@@ -22,4 +22,4 @@
 
 from pulser.sequence import Sequence
 
-from pulser.simulation import Simulation
+from pulser.simulation import Simulation, SimConfig
```
HGSilveri, line 25: I would rather not have `SimConfig` importable from `pulser`, but rather from `pulser.simulation` only.

pulser/simulation/simconfig.py
```diff
@@ -14,32 +14,35 @@
...
 
 
 class SimConfig:
-    """Include additional parameters to simulation. Will be necessary for
-        noisy simulations.
+    """Include additional parameters to simulation.
 
         Keyword arguments:
-            noise_types (string list): types of noises to be used in the
+            noise (list[NOISE_TYPES]): Types of noises to be used in the
```
HGSilveri, line 32: You should specify what the possible noise types are.

pulser/simulation/simconfig.py
```diff
@@ -51,55 +54,59 @@ def __str__(self) -> str:
...
     @property
-    def noise(self) -> List[str]:
+    def noise(self) -> list[NOISE_TYPES]:
         return self._noise
 
     @noise.setter
-    def noise(self, noise_types: List[str]) -> None:
-        self._noise: List[str] = []
+    def noise(self, noise_types: abc.Sequence[NOISE_TYPES]) -> None:
+        self._noise: list[NOISE_TYPES] = []
```
HGSilveri, line 72: How are you supposed to use this? It is written as if you can split the adding of noise in two commands, in which case a `setter` is not the best option. Let's say I want to add `SPAM` and then `doppler`:
```
sim_config.noise = ["SPAM"]
sim_config.noise = ["doppler"]
```
See how this can be misleading? Intuitively, I would expect `noise` to be only `["doppler"]` after this.

pulser/simulation/simconfig.py
```diff
@@ -51,55 +54,59 @@ def __str__(self) -> str:
...
-                        epsilon: false positives
-                        epsilon_prime: false negatives
+                noise_type (NOISE_TYPES): Choose among:
+                -   'dephasing': Random phase (Z) flip
+                -   'doppler': Noisy doppler runs
+                -   'amplitude': Noisy gaussian beam
+                -   'SPAM': SPAM errors. Adds:
+                    --  eta: Probability of each atom to be badly prepared
+                    --  epsilon: Probability of false positives
+                    --  epsilon_prime: Probability of false negatives.
```
HGSilveri, line 87: It would also be good to given the user access to this information. In a private method, they are hidden away.

pulser/simulation/simconfig.py
```diff
@@ -0,0 +1,136 @@
...
+        """
+        # Check proper input:
+        if noise_type not in get_args(NOISE_TYPES):
+            raise ValueError('Not a valid noise type')
```
HGSilveri, line 91: Try to be a bit more informative. Say something like: "Noise type ... is not a valid noise type, must be: ..."

pulser/simulation/simconfig.py
```diff
@@ -0,0 +1,136 @@
...
+                 epsilon: Optional[float] = None,
+                 epsilon_prime: Optional[float] = None) -> None:
+        """Allows the user to change SPAM parameters in dictionary"""
+        values = {'eta': eta, 'epsilon': epsilon,
+                  'epsilon_prime': epsilon_prime}
+        for param in values:
+            val = values[param]
+            if val is not None:
+                if val > 1. or val < 0.:
+                    raise ValueError('Invalid value : must be between 0 and 1')
```
HGSilveri, line 108: In trying to keep the error messages more uniform and informative, try to follow the guidelines laid out in #185.

pulser/simulation/simconfig.py
```diff
@@ -0,0 +1,136 @@
...
+        return self._spam_dict
+
+    def set_spam(self, eta: Optional[float] = None,
+                 epsilon: Optional[float] = None,
+                 epsilon_prime: Optional[float] = None) -> None:
+        """Allows the user to change SPAM parameters in dictionary"""
+        values = {'eta': eta, 'epsilon': epsilon,
+                  'epsilon_prime': epsilon_prime}
+        for param in values:
+            val = values[param]
```
HGSilveri, line 105: Do this instead:
```suggestion
        for param, val in values.items():
```

pulser/simulation/simconfig.py
```diff
@@ -0,0 +1,136 @@
...
+        self._noise.append(noise_type)
+
+    @property
+    def spam_dict(self) -> dict[str, float]:
+        return self._spam_dict
+
+    def set_spam(self, eta: Optional[float] = None,
+                 epsilon: Optional[float] = None,
+                 epsilon_prime: Optional[float] = None) -> None:
+        """Allows the user to change SPAM parameters in dictionary"""
```
HGSilveri, line 101: Specify the arguments in the docstring.

pulser/simulation/simconfig.py
```diff
@@ -0,0 +1,136 @@
...
+        for param in values:
+            val = values[param]
+            if val is not None:
+                if val > 1. or val < 0.:
+                    raise ValueError('Invalid value : must be between 0 and 1')
+                self._spam_dict[param] = val
+
+    def init_spam(self) -> None:
+        self._spam_dict: dict[str, float] = \
+            {'eta': 0.005, 'epsilon': 0.01, 'epsilon_prime': 0.05}
```
HGSilveri, line 113: Still think this method is unnecessary, would rather have it's contents in `__init__`

pulser/simulation/simconfig.py
```diff
@@ -0,0 +1,136 @@
...
+
+    @property
+    def temperature(self) -> float:
+        return self._temperature
+
+    @temperature.setter
+    def temperature(self, value: float) -> None:
+        # value set in microkelvin
+        self._temperature = value * 1e-6
```
HGSilveri, line 122: You should validate the value beforehand

pulser/simulation/simconfig.py
```diff
@@ -0,0 +1,136 @@
...
+        self._calc_sigma_doppler()
+
+    @property
+    def doppler_sigma(self) -> float:
+        return self._doppler_sigma
+
+    def _calc_sigma_doppler(self) -> None:
+        # sigma = keff Deltav, keff = 8.7mum^-1, Deltav = sqrt(kB T / m)
+        self._doppler_sigma: float = 8.7 * np.sqrt(
+            1.38e-23 * self._temperature / 1.45e-25)
```
HGSilveri, line 132: Instead of using this numbers here, define them as constants at the top of the file

pulser/simulation/simconfig.py
```diff
@@ -0,0 +1,136 @@
...
+        s += "Options:\n"
+        s += "----------       \n"
+        s += "Noise types:       " + str(self.noise) + "\n"
+        s += "Spam dictionary:   " + str(self.spam_dict) + "\n"
+        s += "Temperature:       " + str(self.temperature) + "K" + "\n"
+        s += "Number of runs:    " + str(self.runs) + "\n"
+        s += "Samples per runs:  " + str(self.samples_per_run) + "\n"
+        s += "Laser waist:       " + str(self.laser_waist) + "μm \n"
+        s += "\n" + "Solver Options: \n"+str(self.solver_options)[10:-1]+"\n"
+        return s
```
HGSilveri, line 64: Just a reminder of a previous comment concerning the poor performance of string concatenation with `+` and `+=`

HGSilveri (changes requested):
Here goes round 2 of `SimConfig` comments

pulser/simulation/simconfig.py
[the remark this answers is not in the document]
HGSilveri: Actually, I still think it's relevant to omit the `self.init_spam()` method and move it's contents here explicitly instead

pulser/simulation/simconfig.py
julien-bremont: I kept it because the user might want to reset values after tinkering with them.

HGSilveri: I see... But they should never go change the source code. If you want to give that option, you should specify the default of `spam_dict` in `__init__`.

pulser/simulation/simconfig.py
julien-bremont: Well, this is what happens in this case ? If you want to add both `SPAM` and `doppler` you would do 
`sim_config.noise = ['SPAM', 'doppler]`.
In the above example, `sim_config.noise = ['doppler']`. The `add_noise` method isn't supposed to be used by the user, maybe I should add an underscore in front of it.

HGSilveri: You're right, I missed `self.noise` being reset to `[]` at the beginning. Still, this is an awkward interface, `setters` are more suited for setting immutable objects.
I need to clarify how the user is supposed to use this class. From what I can tell, they initialize it and then add all the noise at once, is that it? If that's the case, you won't need the setter, because they can just provide the noise types to `__init__`.

julien-bremont: Yes, the user can initialize the noise all at once, but I also wanted to keep some flexibility and allow for changes. That said, I use `setters` in my code for convenience for the user and to run some code while modifying a field. For example, using `SimConfig.temperature = 500` will also modify the `doppler_sigma` accordingly, since the latter depends on the temperature.

julien-bremont:
Marked this as draft since I'm bored of writing tests, but still awaiting reviews.

pulser/simulation/simulation.py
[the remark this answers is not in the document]
lhenriet: I'm not sure about this. What about just setting all the Hamiltonian connecting this atom to zero ? We either (i) go with the realistic 150 MHz detuning thing, or (ii) we just consider that the atom is uncoupled to the rest. As the results will  be the same (up to many digits), I suggest going for (ii) for performances issues.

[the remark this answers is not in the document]
julien-bremont: That might actually work way better than removing the atom from the `Register`. I'll do this and run tests.

pulser/simulation/simconfig.py
```diff
@@ -0,0 +1,112 @@
...
+            runs (int): Number of runs needed : each run draws a new random
+                noise.
+            samples_per_run (int): Number of samples per noisy run.
+                Useful for cutting down on computing time, but unrealistic.
+            temperature (float): Temperature, set in muK, of the Rydberg array.
+                Also sets the standard deviation of the speed of the atoms.
+            laser_waist (float): Waist of the gaussian laser in global pulses.
+            solver_options (qutip.Options): options for the qutip solver.
+        """
+    noise: list[NOISE_TYPES] = field(default_factory=list)
```
HGSilveri, line 56: Make it a `tuple` instead of a `list`, to give the sense of immutability.

pulser/simulation/simconfig.py
```diff
@@ -0,0 +1,112 @@
...
+                -   'amplitude': Noisy gaussian beam
+                -   'SPAM': SPAM errors. Adds:
+                    --  eta: Probability of each atom to be badly prepared
+                    --  epsilon: Probability of false positives
+                    --  epsilon_prime: Probability of false negatives.
+            runs (int): Number of runs needed : each run draws a new random
+                noise.
+            samples_per_run (int): Number of samples per noisy run.
+                Useful for cutting down on computing time, but unrealistic.
+            temperature (float): Temperature, set in muK, of the Rydberg array.
```
HGSilveri, line 51:
```suggestion
            temperature (float): Temperature, set in µK, of the Rydberg array.
```

pulser/simulation/simconfig.py
```diff
@@ -0,0 +1,112 @@
...
+                Also sets the standard deviation of the speed of the atoms.
+            laser_waist (float): Waist of the gaussian laser in global pulses.
+            solver_options (qutip.Options): options for the qutip solver.
```
HGSilveri, line 54: nit: 
```suggestion
            solver_options (qutip.Options): Options for the qutip solver.
```

pulser/simulation/simconfig.py
```diff
@@ -0,0 +1,112 @@
...
+----------
+Noise types:         {self.noise}
+Spam dictionary:     {self.spam_dict}
+Temperature:         {self.temperature}K
+Number of runs:      {self.runs}
+Samples per runs:    {self.samples_per_run}
+Laser waist:         {self.laser_waist}μm
+
+Solver Options:
+{str(self.solver_options)[10:-1]}"""
```
HGSilveri, line 85: nit: The indentation looks weird like this... I would rather have each line on a list and then joined. See an example  in `Device._specs()`.

pulser/simulation/simconfig.py
```diff
@@ -0,0 +1,112 @@
...
+import qutip
+
+
+NOISE_TYPES = Literal['doppler', 'amplitude', 'SPAM', 'dephasing']
+mass = 1.45e-25
+kb = 1.38e-23
+keff = 8.7
+
+
+@dataclass
```
HGSilveri, line 30: I think we should push for having the dataclass with `frozen=True`. It will make things a bit ugly, but only inside then hidden methods. When you make it frozen, you will no longer be able to change attributes, so, for example, you can't do `self.temperature *= 1.e-6`. However, you can cheat by doing `self.__dict__["temperature"] *= 1e-6` instead. It will look uglier, but it has the added benefit of not letting the user change the config without cheating themselves. See how we do this in `Variable`, for example.

HGSilveri (commented):
Here are some more comments (still only on `SimConfig`, but that's because I want to have it ready before moving on)

pulser/simulation/simconfig.py
julien-bremont: I found an uglier solution on stackexchange and thought it was ugly, but yours is way better. We'll use that one with `frozen=True`.

julien-bremont:
Classes merged. A tutorial is to come, see #216.

julien-bremont:
Classes merged. A tutorial is to come, see #216.

PR #162, opened 2021-05-17
Change instances of `.format` to f-strings [unitaryHACK]

Yash-10:
Hello,

This PR tries to make a small change by making strings consistent by replacing the `.format` syntax with f-strings.

Thanks!

### DIFF
--- a/pulser/_seq_drawer.py
+++ b/pulser/_seq_drawer.py
@@ -95,7 +95,7 @@ def draw_sequence(seq, sampling_rate=None):
         elif value == 0:
             return "0"      # pragma: no cover - just for safety
         else:
-            return r"{:.2g}$\pi$".format(value)
+            return fr"{value:.2g}$\pi$"
 
     n_channels = len(seq._channels)
     if not n_channels:
--- a/pulser/channels.py
+++ b/pulser/channels.py
@@ -109,8 +109,11 @@ class Channel:
         return _duration
 
     def __repr__(self):
-        s = ".{}(Max Absolute Detuning: {} rad/µs, Max Amplitude: {} rad/µs"
-        config = s.format(self.addressing, self.max_abs_detuning, self.max_amp)
+        config = (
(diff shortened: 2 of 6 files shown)


HGSilveri:
Hi @Yash-10 ! Thanks for the contribution! Are you coming from UnitaryHack?

Yash-10:
Thanks for the opportunity! Yes, I am coming from UnitaryHack. I am hoping to solve issue(s) labelled for UnitaryHack too! I thought of this small change while I was going through the source code.

Could you let me know your thoughts? :)

HGSilveri:
> Thanks for the opportunity! Yes, I am coming from UnitaryHack.

Then don't forget to write [unitaryHACK] on the title!

> I am hoping to solve issue(s) labelled for UnitaryHack too! I thought of this small change while I was going through the source code.
> 
> Could you let me know your thoughts? :)

That will depend on your physics and quantum computing knowledge... DM me in Discord (HGSilveri) if you want to discuss further.

Yash-10:
> Then don't forget to write [unitaryHACK] on the title!

Oh, thanks for that. I have added it in the title.


> That will depend on your physics and quantum computing knowledge... DM me in Discord (HGSilveri) if you want to discuss further.

Thanks, I shall do so after picking up an issue.

HGSilveri:
While we're at it, we should get rid of all unnecessary `format` cases. I looked for them in the code and found two more:

- pulser/sequence.py, line 373
- pulser/_seq_drawer.py, line 98

Do you think you could change these too? Thanks!

Yash-10:
Thanks, I think the change in`sequence.py` got reverted after I rebased earlier. Now it seems to be addressed?

HGSilveri (approved):
LGTM, thanks for the contribution!

HGSilveri:
> Thanks, I think the change in`sequence.py` got reverted after I rebased earlier. Now it seems to be addressed?

All looks good, thanks!

Issue #16, opened 2020-11-10
Add type hints, and use mypy to enforce in CI


karalekas:
Python is a dynamically-typed language (see: [duck typing](https://en.wikipedia.org/wiki/Duck_typing)), but Python 3.5 and [pep-484](https://www.python.org/dev/peps/pep-0484/) introduced type hints as a way to annotate Python code with types. These are not enforced by the language at runtime, but they can be checked by third-party libraries such as [mypy](https://mypy.readthedocs.io/en/stable/). The impetus for adding type hints and using mypy is that they generally encourage developers to write saner, simpler code (the flexibility of Python is a double-edged sword). This issue is mostly about good code style and therefore is optional -- it doesn't block CI setup (but once it is complete, the CI can be used to do this type checking automatically).

TripleR47:
Hi @HGSilveri  @karalekas, I went through the issue and would like to work on it. I have a few questions, I found this [GitHub Action](https://github.com/marketplace/actions/lint-action) which can be used to implement the type hint ci, so can it be used? or do it need to be implemented using shell script?

LaurentAjdnik:
Hi @TripleR47, I've done it for the `Register` class (I will PR #132 tomorrow) but I did it manually 🤨, while adding "Enhanced ErrorValue messages" (#148) at the same time.

HGSilveri:
Hi @TripleR47, no preference regarding how the mypy checks are occur, as long as they are programmed into the existing workflow, if that's what you're asking. But bear in mind that is not the crux of the PR, but rather going through the existing code and adding the appropriate type hints to each function and method. After that's done, then you can incorporate `mypy` into the CI routine. 
I realize it's quite a bit of work, but that's why there's a bounty on it!

TripleR47:
Okay, I'll ask in the unitaryHack's server if anyone wants to collaborate on this issue. Can it be assigned to me?

TripleR47:
Hi @LaurentAjdnik, wanna collaborate on this issue?

LaurentAjdnik:
> Hi @LaurentAjdnik, wanna collaborate on this issue?

👍

TripleR47:
> > Hi @LaurentAjdnik, wanna collaborate on this issue?
> 
> 👍

Are you in unitaryfund's discord server? If you are then DM me on discord, we can discuss further on this.

HGSilveri:
@TripleR47 @LaurentAjdnik So, can I assign it to you both?

[...]

PR #163, opened 2021-05-19
[unitaryHACK] Add type hints to `waveforms.py` (#16)

LaurentAjdnik:
MyPy checked, with the `--ignore-missing-imports` option.

I have doubts about:
* Line 146: `np.all()` returns a `np.bool_` which is not a `bool`. So i checked with `== np.True_` but it might not be the best way to do it.
* Line 504: Introducing `f: Callable` throws a lot of `"Callable[..., Any]" has no attribute "XXX"` errors and I don't know how to handle that.

Feedback more than welcome 😊.

### DIFF
--- a/pulser/waveforms.py
+++ b/pulser/waveforms.py
@@ -12,6 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from __future__ import annotations
+
+from typing import Callable, Optional
+
 from abc import ABC, abstractmethod
 import functools
 import inspect
@@ -20,6 +24,7 @@ import sys
 import types
 import warnings
 
+from matplotlib.axes import Axes
 import matplotlib.pyplot as plt
 import numpy as np
 
@@ -38,7 +43,7 @@ class Waveform(ABC):
         else:
             return object.__new__(cls)
 
-    def __init__(self, duration):
+    def __init__(self, duration: int):
         """Initializes a waveform with a given duration.
 
         Args:
@@ -61,13 +66,13 @@ class Waveform(ABC):
 
     @property
     @abstractmethod
-    def duration(self):
+    def duration(self) -> int:
         """The duration of the pulse (in ns)."""
         pass
 
     @property
     @abstractmethod
-    def samples(self):
+    def samples(self) -> np.ndarray:
         """The value at each time step that describes the waveform.
 
         Returns:
@@ -76,21 +81,21 @@ class Waveform(ABC):
         pass
 
     @property
-    def first_value(self):
+    def first_value(self) -> float:
         """The first value in the waveform."""
         return self.samples[0]
 
     @property
-    def last_value(self):
+    def last_value(self) -> float:
         """The last value in the waveform."""
         return self.samples[-1]
 
     @property
-    def integral(self):
+    def integral(self) -> float:
         """Integral of the waveform (time in ns, value in rad/µs)."""
-        return np.sum(self.samples) * 1e-3  # ns * rad/µs = 1e-3
+        return float(np.sum(self.samples)) * 1e-3  # ns * rad/µs = 1e-3
 
-    def draw(self):
+    def draw(self) -> None:
         """Draws the waveform."""
 
         fig, ax = plt.subplots()
@@ -98,7 +103,7 @@ class Waveform(ABC):
 
         plt.show()
 
-    def change_duration(self, new_duration):
+    def change_duration(self, new_duration: int) -> Waveform:
         """Returns a new waveform with modified duration.
 
         Args:
@@ -108,42 +113,43 @@ class Waveform(ABC):
                                   " modifications to its duration.")
 
     @abstractmethod
-    def _to_dict():
+    def _to_dict(self) -> dict:
         pass
 
     @abstractmethod
-    def __str__(self):
+    def __str__(self) -> str:
         pass
 
     @abstractmethod
-    def __repr__(self):
+    def __repr__(self) -> str:
         pass
 
     @abstractmethod
-    def __mul__(self, other):
+    def __mul__(self, other: float) -> Waveform:
         pass
 
-    def __neg__(self):
+    def __neg__(self) -> Waveform:
         return self.__mul__(-1)
 
-    def __truediv__(self, other):
+    def __truediv__(self, other: float) -> Waveform:
         if other == 0:
             raise ZeroDivisionError("Can't divide a waveform by zero.")
         else:
             return self.__mul__(1/other)
 
-    def __eq__(self, other):
+    def __eq__(self, other: object) -> bool:
         if not isinstance(other, Waveform):
             return False
         elif self.duration != other.duration:
             return False
         else:
-            return np.all(np.isclose(self.samples, other.samples))
+            return np.all(np.isclose(self.samples, other.samples)) == np.True_
 
-    def __hash__(self):
+    def __hash__(self) -> int:
         return hash(tuple(self.samples))
 
-    def _plot(self, ax, ylabel, color=None):
+    def _plot(self, ax: Axes, ylabel: str,
+              color: Optional[str] = None) -> None:
         ax.set_xlabel('t (ns)')
         ts = np.arange(self.duration)
         if color:
@@ -175,7 +181,7 @@ class CompositeWaveform(Waveform):
         self._waveforms = list(waveforms)
 
     @property
-    def duration(self):
+    def duration(self) -> int:
         """The duration of the pulse (in ns)."""
         duration = 0
         for wf in self._waveforms:
@@ -183,7 +189,7 @@ class CompositeWaveform(Waveform):
         return duration
 
     @property
-    def samples(self):
+    def samples(self) -> np.ndarray:
         """The value at each time step that describes the waveform.
 
         Returns:
@@ -192,38 +198,38 @@ class CompositeWaveform(Waveform):
         return np.concatenate([wf.samples for wf in self._waveforms])
 
     @property
-    def first_value(self):
+    def first_value(self) -> float:
         """The first value in the waveform."""
         return self._waveforms[0].first_value
 
     @property
-    def last_value(self):
+    def last_value(self) -> float:
         """The last value in the waveform."""
         return self._waveforms[-1].last_value
 
     @property
-    def waveforms(self):
+    def waveforms(self) -> list:
         """The waveforms encapsulated in the composite waveform."""
         return list(self._waveforms)
 
-    def _validate(self, waveform):
+    def _validate(self, waveform: Waveform) -> None:
         if not isinstance(waveform, Waveform):
             raise TypeError(f"{waveform!r} is not a valid waveform. "
                             "Please provide a valid Waveform.")
 
-    def _to_dict(self):
+    def _to_dict(self) -> dict:
         return obj_to_dict(self, *self._waveforms)
 
-    def __str__(self):
-        contents = ["{!r}"] * len(self._waveforms)
-        contents = ", ".join(contents)
+    def __str__(self) -> str:
+        contents_list = ["{!r}"] * len(self._waveforms)
+        contents = ", ".join(contents_list)
         contents = contents.format(*self._waveforms)
         return f'Composite({contents})'
 
-    def __repr__(self):
+    def __repr__(self) -> str:
         return f'CompositeWaveform({self.duration} ns, {self._waveforms!r})'
 
-    def __mul__(self, other):
+    def __mul__(self, other: float) -> CompositeWaveform:
         return CompositeWaveform(*(wf * other for wf in self._waveforms))
 
 
@@ -235,19 +241,19 @@ class CustomWaveform(Waveform):
             (in rad/µs). The number of samples dictates the duration, in ns.
     """
 
-    def __init__(self, samples):
+    def __init__(self, samples: np.ndarray):
         """Initializes a custom waveform."""
         samples_arr = np.array(samples, dtype=float)
         self._samples = samples_arr
         super().__init__(len(samples))
 
     @property
-    def duration(self):
+    def duration(self) -> int:
         """The duration of the pulse (in ns)."""
         return self._duration
 
     @property
-    def samples(self):
+    def samples(self) -> np.ndarray:
         """The value at each time step that describes the waveform.
 
         Returns:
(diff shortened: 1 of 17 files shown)


HGSilveri (commented):
Looking good! I noticed you skipped the cases where the number of arguments is variable. I think we can still add some typing, in some cases. Here are two examples
- In `Waveform.__new__()`, try typing the return as `Union[ParamObj, Waveform]`
- In `CompositeWaveform.__init__()`, you can type the args as `*waveforms: Waveform`

I think these should work, give it a shot!

LaurentAjdnik:
> * In `Waveform.__new__()`, try typing the return as `Union[ParamObj, Waveform]`

MyPy threw this error: `"__new__" must return a class instance (got "Union[ParamObj, Waveform]")` ☹️.

> * In `CompositeWaveform.__init__()`, you can type the args as `*waveforms: Waveform`

Done :+1:.

> Since the argument of __mul__ is float, perhaps change this to:

Done :+1:.

> What about this instead?

Done :+1:.

> Missed this one.

Done :+1:.

Any advice on how to handle `f: Callable` on line 504 ?

HGSilveri:
> Any advice on how to handle `f: Callable` on line 504 ?

Use `FunctionType` instead of `Callable`, for both the arg and the return type. Also, please change the `import types` to `from types import FunctionType`.

pulser/waveforms.py
```diff
@@ -235,36 +240,36 @@ class CustomWaveform(Waveform):
             (in rad/µs). The number of samples dictates the duration, in ns.
     """
 
-    def __init__(self, samples):
+    def __init__(self, samples: np.ndarray):
```
HGSilveri, line 243: From looking at this, I think that sometimes it's a bit too restrictive to ask for a numpy array specifically. I think it's a good place to test using `ArrayLike` from `numpy.testing` (see [here](https://numpy.org/devdocs/reference/typing.html#numpy.typing.ArrayLike)). If it works, we can start doing that throughout the code.

pulser/waveforms.py
```diff
@@ -12,14 +12,18 @@
...
+from __future__ import annotations
+
 from abc import ABC, abstractmethod
 import functools
 import inspect
 import itertools
 import sys
-import types
+from types import FunctionType
+from typing import Optional
```
HGSilveri, line 23: I agree with the placement of the import here instead

pulser/waveforms.py
```diff
@@ -192,38 +197,38 @@ def samples(self):
...
 
     @property
-    def last_value(self):
+    def last_value(self) -> float:
         """The last value in the waveform."""
         return self._waveforms[-1].last_value
 
     @property
-    def waveforms(self):
+    def waveforms(self) -> list:
```
HGSilveri, line 210: This one would be even better as:
```suggestion
    def waveforms(self) -> List[Waveform]:
```

HGSilveri (commented):
Ok, I think these are my last comments

pulser/waveforms.py
```diff
@@ -207,7 +207,7 @@ def last_value(self) -> float:
         return self._waveforms[-1].last_value
 
     @property
-    def waveforms(self) -> list:
+    def waveforms(self) -> list[Waveform]:
```
HGSilveri, line 210: Please use `List[Waveform]` instead (you'll have to import `List` from `typing`), as this syntax is only available since Python 3.9.

HGSilveri:
Alright, one more thing missing (and I apologize for only remembering about this now): 

`Waveform` and `Pulse` classes accept another type of argument upon creation, of the so-called `Parametrized` type. Thus, an object of type `Parametrized` is a valid argument for each argument of the `__init__` methods and for the `classmethods`, so the types you currently have (eg `int`) should be turn into a `Union` (eg `Union[int, Parametrized]`).

Also, I'm writing a Mypy config file and will add `Mypy` to the CI tests before I merge this PR. This should make your life easier moving forward.

HGSilveri:
> `Waveform` and `Pulse` classes accept another type of argument upon creation, of the so-called `Parametrized` type. Thus, an object of type `Parametrized` is a valid argument for each argument of the `__init__` methods and for the `classmethods`, so the types you currently have (eg `int`) should be turn into a `Union` (eg `Union[int, Parametrized]`).

@LaurentAjdnik Actually, doing it like this will cause a bunch of problems, because the code after the `__init__` is not actually run when an argument is `Parametrized`, but Mypy will think it is (this is related to the `__new__` method of these particular classes).

There is a way to overcome this, I'll get back to you on it, please standby for now.

HGSilveri:
@LaurentAjdnik Alright, so here's the plan: 

You'll do the `Union` as previously mentioned, but then, when necessary, you'll use the `cast` from `typing` (see [here](https://docs.python.org/3/library/typing.html#typing.cast)) to cast the variable to the type that is not `Parametrized`. 

What this will do is tell the type checker that the type of that variable is supposed to be the one you tell it. It sounds a bit like cheating, but in this case it applies because **the code block doesn't run when the arg is `Parametrized`**.

So, as an example, in  `Waveform.__init__(self, duration: Union[int, Parametrized])`, you'll add the line:
```
duration = cast(int, duration)
```

This will tell mypy that, inside the function definition, `duration` is only `int`, not `Parametrized`. (make sure you import `cast` from `typing`)

I hope this is clear, let me know if you have any questions.

LaurentAjdnik:
> `Waveform` and `Pulse` classes accept another type of argument upon creation, of the so-called `Parametrized` type.

@HGSilveri:  What is its use and how does it work? Thanks!

HGSilveri:
> > `Waveform` and `Pulse` classes accept another type of argument upon creation, of the so-called `Parametrized` type.
> 
> @HGSilveri: What is its use and how does it work? Thanks!

Essentially, you need it to create parametrized sequences, where you use variables instead of fixed values. It is like a factory of a regular sequence, to which you attribute values to the variables only upon building. You can check out [this tutorial](https://pulser.readthedocs.io/en/stable/paramseqs.html) to see how it works.

pulser/waveforms.py
HGSilveri: Just a reminder to look into this

HGSilveri: Also, `samples` can be `Parametrized` too.

pulser/waveforms.py
```diff
@@ -164,7 +170,7 @@ class CompositeWaveform(Waveform):
         waveforms(Waveform): Two or more waveforms to combine.
     """
 
-    def __init__(self, *waveforms):
+    def __init__(self, *waveforms: Waveform):
```
HGSilveri, line 173: Needs to be `Union` with `Parametrized` and `cast` to `Waveform` too. You should put the `cast` above line 179.

pulser/waveforms.py
```diff
@@ -276,18 +282,18 @@ class ConstantWaveform(Waveform):
         value (float): The modulation value (in rad/µs).
     """
 
-    def __init__(self, duration, value):
+    def __init__(self, duration: int, value: float):
```
HGSilveri, line 285: These can be `Parametrized` too.

pulser/waveforms.py
```diff
@@ -339,19 +345,19 @@ class RampWaveform(Waveform):
         stop (float): The final value (in rad/µs).
     """
 
-    def __init__(self, duration, start, stop):
+    def __init__(self, duration: int, start: float, stop: float):
```
HGSilveri, line 348: These can also be `Parametrized`.

pulser/waveforms.py
```diff
@@ -409,7 +415,8 @@ class BlackmanWaveform(Waveform):
             case it takes the positive waveform and changes the sign of all its
             values.
     """
-    def __init__(self, duration, area):
+
+    def __init__(self, duration: int, area: float):
```
HGSilveri, line 419: `Parametrize` these args too.

HGSilveri (changes requested):
I think you got the right idea, just need to do it throughout.

pulser/waveforms.py
```diff
@@ -83,12 +84,12 @@ def samples(self) -> np.ndarray:
     @property
     def first_value(self) -> float:
         """The first value in the waveform."""
-        return self.samples[0]
+        return cast(float, self.samples[0])
```
HGSilveri, line 87: Instead of casting, really make it a `float`. The same for `last_value`.

HGSilveri (changes requested):
I see it's putting up a fight... Here are some more comments

pulser/waveforms.py
```diff
@@ -108,42 +114,43 @@ def change_duration(self, new_duration):
                                   " modifications to its duration.")
 
     @abstractmethod
-    def _to_dict(self):
+    def _to_dict(self) -> dict:
```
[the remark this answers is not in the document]
HGSilveri: Can you do it for the rest of the `_to_dict` definitions too, please?

pulser/waveforms.py
```diff
@@ -409,16 +424,22 @@ class BlackmanWaveform(Waveform):
...
+        duration = cast(int, duration)
         super().__init__(duration)
-        self._area = float(area)
-        self._norm_samples = np.clip(np.blackman(self._duration), 0, np.inf)
-        self._scaling = self._area / np.sum(self._norm_samples) / 1e-3
+        self._area: float = cast(float, area)
+        self._norm_samples: np.ndarray = np.clip(
+            np.blackman(self._duration), 0, np.inf)
+        self._scaling: float = self._area / \
+            float(np.sum(self._norm_samples)) / 1e-3
```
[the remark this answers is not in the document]
HGSilveri: Don't forget this one, it will solve the failed CI tests.

LaurentAjdnik:
Now, tests fail with `AttributeError: 'BlackmanWaveform' object has no attribute '_norm_samples'`.

Every time I try to fix on attribute, it just gets worse...

This is getting way too tricky for me.

pulser/waveforms.py
```diff
@@ -409,16 +424,22 @@ class BlackmanWaveform(Waveform):
...
-    def __init__(self, duration, area):
+
+    def __init__(self, duration: Union[int, Parametrized],
+                 area: Union[float, Parametrized]):
         """Initializes a Blackman waveform."""
         super().__init__(duration)
-        self._area = float(area)
-        self._norm_samples = np.clip(np.blackman(self._duration), 0, np.inf)
-        self._scaling = self._area / np.sum(self._norm_samples) / 1e-3
+        if isinstance(area, float):
```
HGSilveri, line 432: Sorry, maybe I wasn't clear here. What I meant you should do here was: 
```
if not isinstance(area, float):
    raise TypeError(...)
```

HGSilveri:
> Now, tests fail with `AttributeError: 'BlackmanWaveform' object has no attribute '_norm_samples'`.
> 
> Every time I try to fix on attribute, it just gets worse...
> 
> This is getting way too tricky for me.

I'm sorry, maybe I wasn't clear. See my last comment for a clarification.

LaurentAjdnik:
I finally went for:

```
        try:
            self._area: float = float(cast(float, area))
        except (TypeError, ValueError):
            raise TypeError("area needs to be castable to an float but "
                            "type %s was provided" % type(area))
```
just like for the parent class.

Otherwise, using `isinstance(area, float)` raised an error when called with an int, like it currently is in test functions.

HGSilveri:
@LaurentAjdnik Great, you did it! I'll add some more comments, which should be the last

pulser/waveforms.py
```diff
@@ -429,12 +429,16 @@ def __init__(self, duration: Union[int, Parametrized],
...
         """Initializes a Blackman waveform."""
         super().__init__(duration)
-        if isinstance(area, float):
-            self._area: float = area
-            self._norm_samples: np.ndarray = np.clip(
-                np.blackman(self._duration), 0, np.inf)
-            self._scaling: float = self._area / \
-                float(np.sum(self._norm_samples)) / 1e-3
+        try:
+            self._area: float = float(cast(float, area))
```
HGSilveri, line 433: Do you need the `cast` here?

pulser/waveforms.py
```diff
@@ -429,12 +429,16 @@ def __init__(self, duration: Union[int, Parametrized],
...
+        except (TypeError, ValueError):
+            raise TypeError("area needs to be castable to an float but "
+                            "type %s was provided" % type(area))
```
HGSilveri, line 436:
```suggestion
            raise TypeError("Area needs to be castable to a float but "
                            f"type {type(area)} was provided.")
```

pulser/waveforms.py
```diff
@@ -429,12 +429,16 @@ def __init__(self, duration: Union[int, Parametrized],
...
+
+        self._norm_samples: np.ndarray = np.clip(
+            np.blackman(self._duration), 0, np.inf)
+        self._scaling: float = self._area / \
```
HGSilveri, line 440: nit: I prefer parenthesis for line breaks.

HGSilveri (commented):
Alright, final stretch!

pulser/waveforms.py
HGSilveri: Nevermind, I can see you do.

LaurentAjdnik: > Do you need the `cast` here?

According to MyPy, I do: `pulser\waveforms.py:433: error: Argument 1 to "float" has incompatible type "Union[float, Parametrized]"; expected "Union[SupportsFloat, _SupportsIndex, str]"`.

pulser/waveforms.py
LaurentAjdnik: Copied from the parent class. I'll update both.

pulser/waveforms.py
LaurentAjdnik: My IDE did it! 😆

pulser/waveforms.py
```diff
@@ -409,16 +424,26 @@ class BlackmanWaveform(Waveform):
...
+                 area: Union[float, Parametrized]):
         """Initializes a Blackman waveform."""
         super().__init__(duration)
-        self._area = float(area)
-        self._norm_samples = np.clip(np.blackman(self._duration), 0, np.inf)
-        self._scaling = self._area / np.sum(self._norm_samples) / 1e-3
+        try:
+            self._area: float = float(cast(float, area))
+        except (TypeError, ValueError):
+            raise TypeError("area needs to be castable to an float but "
```
HGSilveri, line 435: Arg, only a typo left!

```suggestion
            raise TypeError("area needs to be castable to a float but "
```

HGSilveri (approved):
LGTM, nice job!

PR #165, opened 2021-05-19
Expand `Register.max_connectivity()` for few atoms.

sebgrijalva:
I thought the `max_connectivity()` was nice enough that expanding its use for less than 7 atoms would be useful. This PR includes these cases, occupying one by one the sites of an underlying hexagon coordinate list.

### DIFF
--- a/pulser/register.py
+++ b/pulser/register.py
@@ -336,9 +336,12 @@ class Register:
             raise ValueError(f"Spacing ({spacing}) for this device must be"
                              f" {device.min_atom_distance} or above.")
 
-        if n_qubits == 1:
-            return cls.from_coordinates(np.array([(0.0, 0.0)], dtype=float),
-                                        center=False, prefix=prefix)
+        if n_qubits < 7:
+            hex_coords = np.array([(0.0, 0.0), (1.0, 0.0), (0.5, np.sqrt(3/4)),
+                                   (1.5, np.sqrt(3/4)), (2.0, 0.0),
+                                   (0.5, -np.sqrt(3/4))])
+            return cls.from_coordinates(spacing * hex_coords[:n_qubits],
+                                        prefix=prefix)
 
         full_layers = int((-3.0 + np.sqrt(9 + 12 * (n_qubits - 1))) / 6.0)
         atoms_left = n_qubits - 1 - (full_layers**2 + full_layers) * 3
--- a/pulser/tests/test_register.py
+++ b/pulser/tests/test_register.py
@@ -167,6 +167,19 @@ def test_max_connectivity():
     atoms = list(reg.qubits.values())
     assert(np.all(np.isclose(atoms[0], [0.0, 0.0])))
 
+    # Check for less than 7 atoms:
+    for i in range(1, 6):
+        hex_coords = np.array([(0.0, 0.0), (1.0, 0.0), (0.5, np.sqrt(3/4)),
+                               (1.5, np.sqrt(3/4)), (2.0, 0.0),
+                               (0.5, -np.sqrt(3/4))])
+        reg = Register.max_connectivity(i, device)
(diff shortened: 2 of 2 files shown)


HGSilveri (approved):
You're right, and it was a lapse on my part not to have noticed this. Thanks for the addition, LGTM

LaurentAjdnik:
Thanks @sebgrijalva for the addition and @HGSilveri for the merge!

I had indeed ignored the case when the number of atoms was less than 7, because C3 and C6 symmetries did not make much sense then.

I afford a few questions, to possibly improve `max_connectivity()` even further. In which case I will pull a request.

1. Tests

```
    # Check for less than 7 atoms:
    for i in range(1, 6):
```

If I'm not mistaken, doesn't it skip the case when i = 6 atoms, which should be tested too?

```
        reg2 = Register.from_coordinates(4 * hex_coords[:i])  # Use min spacing
```

Shouldn't we use `spacing` that is set to `device.min_atom_distance` rather than the constant `4`, in case the `Chadoq2` device has this value changed some day?

2. Hole

Here's the result for 6 atoms:

H06

Is there a reason why you left a "hole" between atoms 4 and 5?


3. Order of atoms

When n >= 7, I enforced the same numbering pattern for all layers:
* Atom 0 is always the innermost atom
* The first atom on each layer is "one step after the far left"

It results in a spiral-like pattern:

H19

4. Centering

I purposedly did not center arrays of atoms, so that the innermost atom always lies at (0,0). Can we keep this behavior when n < 7 atoms?

Thanks for your feedback!

HGSilveri:
Hi @LaurentAjdnik ,

> If I'm not mistaken, doesn't it skip the case when i = 6 atoms, which should be tested too?

That is correct, it should be included for completeness.

> Shouldn't we use `spacing` that is set to `device.min_atom_distance` rather than the constant `4`, in case the `Chadoq2` device has this value changed some day?

Personally, I prefer to have things explicit in the tests, because I see them as a way to catch mistakes. If someone wants to change the `min_atom_distance` on `Chadoq2`, they can easily do so in the tests too. More importantly, if someone indadvertedly changes this value, the tests will catch that, which is the point.

I'll let @sebgrijalva explain his choices for the other topics.

sebgrijalva:
Hello @LaurentAjdnik ,

Your comments about the layout of the atoms for N<7 are valid observations, thanks a lot! How do you feel about implementing these ideas?

LaurentAjdnik:
> Hello @LaurentAjdnik ,
> 
> Your comments about the layout of the atoms for N<7 are valid observations, thanks a lot! How do you feel about implementing these ideas?

Well... I will PR something soon. 😁

LaurentAjdnik:
> Personally, I prefer to have things explicit in the tests, because I see them as a way to catch mistakes. If someone wants to change the `min_atom_distance` on `Chadoq2`, they can easily do so in the tests too. More importantly, if someone indadvertedly changes this value, the tests will catch that, which is the point.

Sorry to disagree on that one. 😜

If `Chaqod2.min_atom_distance` were to be changed for a good reason, the test in `test_register.py` would fail, while it should not.

IMHO, our tests should focus only and precisely on what is being tested and not be dependent on things that can be made modular.

Otherwise, we could use constants in many places. But that would make our tests extremely fragile, failing often for unjustified reasons.

If we care deeply about the value inside `Chadoq2` not being changed, it should be a test by itself in `test_devices.py`.

HGSilveri:
@LaurentAjdnik  I can tell you `Chadoq2.min_atom_distance` is not a value that should be changed any time soon, even less more than once. I see your point, although I don't think this case in particular will cause an issue. I still see a benefit in having something checking for changes on the device classes, though I agree that a concrete test would be more appropriate. 

I will not insist on my point of view, so if you still want to change it, I will not oppose.

Issue #149, opened 2021-05-11
Display area and phase of pulse on it


romainfd:
As the area of a pulse is a key element of its effect, displaying the area on the pulse when drawing the sequence could make it easier to interpret.

For example, Figure 8 of the [recent Pulser paper](https://arxiv.org/pdf/2104.15044.pdf) would be more self-explanatory with such a feature as section "3.2.5 Composing the Sequence" is focused on building pulses of specific areas.
![image](https://user-images.githubusercontent.com/32517197/117844392-eeaf7900-b27f-11eb-9674-b17a7dbe8f04.png)

However, this could make the drawing quite crowded. Perhaps a flag, a formatting option or/and a bit of logic to better display pi, 2.pi, ... values could help with that?

HGSilveri:
This would indeed be quite useful, and I have considered it before. However, as you point out, it will get quite crowded. Not only that, the space available for writing this information will decrease as the sequence increases, eventually reaching a point where you no longer have enough space to write down the area and it will just pile on, potentially making a mess. 

I don't have a great idea of how to handle this robustly and automatically, but perhaps we don't have to if we make this optional, with the default being not to do it. In that way, the user has to purposefully choose to do it and will understand why it looks bad if that ends up happening.

HGSilveri:
A clarification of this issue's objective:

When drawing a `Sequence`, add the option of **displaying the area and phase**, as multiples of π, of the various amplitude waveforms. The placement of such labels is not set in stone, but should be as robust to different sizes of the pulses as possible.

**Utility**:
This will allow the identification of the pulses as rotations around a Bloch sphere axis (cos(phase), -sin(phase), 0), by an angle = area of the pulse, just from looking at the sequence plot.

Yash-10:
I am finding this issue interesting to solve. Hence, I am going to attempt to solve it.

Thanks!

HGSilveri:
Successfully solved by @Yash-10

PR #166, opened 2021-05-19
Add phase and area text to pulse plot [unitaryHACK]

Yash-10:
This pull request attempts to address #149.

**Details**

- It adds a condition `draw_phase_area` (default to `False`) in the `draw` method of a `Sequence` object. The `draw_sequence` function also has this argument.
- Based on this argument, it would allow to plot phase and area for each of the waveforms, where the values are in multiples of `pi`.
- Instead of writing "Phase" and "Area", it is written once in the title and then the symbols are used throughout to avoid consuming large space on the plot.

Since this is currently a draft, there is still some work to do regarding the placement of the text.

The output of running:

```
seq.draw(draw_phase_area=True)
```

on a sample sequence (the same sequence as shown in the issue) currently looks like this:

sequence

I would love to get any thoughts on this.

Thanks!

### DIFF
--- a/pulser/_seq_drawer.py
+++ b/pulser/_seq_drawer.py
@@ -106,8 +106,9 @@ def draw_sequence(seq, sampling_rate=None, draw_phase_area=False):
     time_scale = 1e3 if seq._total_duration > 1e4 else 1
 
     # Boxes for qubit and phase text
-    q_box = dict(boxstyle="round", facecolor='orange', alpha=0.7)
-    ph_box = dict(boxstyle="round", facecolor='ghostwhite', alpha=0.7)
+    q_box = dict(boxstyle="round", facecolor='orange')
+    ph_box = dict(boxstyle="round", facecolor='ghostwhite')
+    area_ph_box = dict(boxstyle='round', facecolor='ghostwhite')
 
     fig = plt.figure(constrained_layout=False, figsize=(20, 4.5*n_channels))
     gs = fig.add_gridspec(n_channels, 1, hspace=0.075)
@@ -202,6 +203,7 @@ def draw_sequence(seq, sampling_rate=None, draw_phase_area=False):
         b.set_ylabel(r'$\delta$ (rad/µs)', fontsize=14)
 
         if draw_phase_area:
+            top = False  # Variable to track position of box, top or center.
             for pulse_num, seq_ in enumerate(seq._schedule[ch]):
                 # Select only `Pulse` objects
                 if isinstance(seq_.type, Pulse):
@@ -210,11 +212,13 @@ def draw_sequence(seq, sampling_rate=None, draw_phase_area=False):
                     x_plot = (seq_.ti + seq_.tf) / 2
                     if (
                         seq._schedule[ch][pulse_num-1].type == "target"
-                        or pulse_num % 2 == 0
+                        or not top
                     ):
                         y_plot = np.max(seq_.type.amplitude.samples) / 2
-                    else:
+                        top = True  # Next box at the top.
+                    elif top:
                         y_plot = np.max(seq_.type.amplitude.samples)
+                        top = False  # Next box at the center.
                     if phase_val == 0:
                         txt = fr"A: {area_val:.2g}$\pi$"
                     else:
@@ -223,7 +227,7 @@ def draw_sequence(seq, sampling_rate=None, draw_phase_area=False):
                         txt = "\n".join([phase_fmt, area_fmt])
                     a.text(
                         x_plot, y_plot, txt, fontsize=10,
-                        ha="center", va="center", bbox=ph_box,
+                        ha="center", va="center", bbox=area_ph_box,
                     )
 
         target_regions = []     # [[start1, [targets1], end1],...]
--- a/pulser/_seq_drawer.py
+++ b/pulser/_seq_drawer.py
@@ -108,7 +108,7 @@ def draw_sequence(seq, sampling_rate=None, draw_phase_area=False):
     # Boxes for qubit and phase text
     q_box = dict(boxstyle="round", facecolor='orange')
     ph_box = dict(boxstyle="round", facecolor='ghostwhite')
-    area_ph_box = dict(boxstyle='round', facecolor='ghostwhite')
+    area_ph_box = dict(boxstyle='round', facecolor='ghostwhite', alpha=0.7)
 
     fig = plt.figure(constrained_layout=False, figsize=(20, 4.5*n_channels))
     gs = fig.add_gridspec(n_channels, 1, hspace=0.075)
@@ -212,9 +212,9 @@ def draw_sequence(seq, sampling_rate=None, draw_phase_area=False):
                 # Select only `Pulse` objects
                 if isinstance(seq_.type, Pulse):
                     if sampling_rate:
-                        area_val = np.sum(
-                            np.abs(yaeff[int(seq_.ti):int(seq_.tf)])
-                        ) / 1e3 / np.pi
+                        area_val = np.sum(cs_amp(
+                            np.arange(seq_.ti, seq_.tf)
+                        )) * 1e-3 / np.pi
                     else:
                         area_val = seq_.type.amplitude.integral / np.pi
                     phase_val = seq_.type.phase / np.pi
@@ -228,7 +228,8 @@ def draw_sequence(seq, sampling_rate=None, draw_phase_area=False):
                     elif top:
                         y_plot = np.max(seq_.type.amplitude.samples)
                         top = False  # Next box at the center.
-                    area_fmt = fr"A: {area_val:.2g}$\pi$"
+                    area_fmt = (r"A: $\pi$" if round(area_val, 2) == 1
+                                else fr"A: {area_val:.2g}$\pi$")
                     if not draw_phase:
                         txt = area_fmt
                     else:
--- a/pulser/_seq_drawer.py
+++ b/pulser/_seq_drawer.py
@@ -16,6 +16,7 @@ import matplotlib.pyplot as plt
 import numpy as np
 
 from pulser.waveforms import ConstantWaveform
+from pulser.pulse import Pulse
 from scipy.interpolate import CubicSpline
 
 
@@ -107,10 +108,13 @@ def draw_sequence(seq, sampling_rate=None, draw_phase_area=False):
 
     # Boxes for qubit and phase text
     q_box = dict(boxstyle="round", facecolor='orange')
-    ph_box = dict(boxstyle="round", facecolor='ghostwhite')
+    ph_box = dict(boxstyle="round", facecolor='ghostwhite', alpha=0.5)
 
     fig = plt.figure(constrained_layout=False, figsize=(20, 4.5*n_channels))
-    gs = fig.add_gridspec(n_channels, 1, hspace=0.1)
+    gs = fig.add_gridspec(n_channels, 1, hspace=0.075)
+
+    if draw_phase_area:
+        fig.suptitle(r"$\phi$: Phase, A: Area", fontsize=14)
 
     ch_axes = {}
     for i, (ch, gs_) in enumerate(zip(seq._channels, gs)):
@@ -202,22 +206,18 @@ def draw_sequence(seq, sampling_rate=None, draw_phase_area=False):
         b.set_ylabel(r'$\delta$ (rad/µs)', fontsize=14)
 
         if draw_phase_area:
-            a.set_title(r"Phase: $\phi$, Area: A", fontsize=16)
             for seq_ in seq._schedule[ch]:
                 # Select only `Pulse` objects
-                if not isinstance(seq_.type, str):
+                if isinstance(seq_.type, Pulse):
                     phase_val = seq_.type.phase / np.pi
                     area_val = seq_.type.amplitude.integral / np.pi
                     # X and Y coordinates for placing text
                     x_plot = (seq_.ti + seq_.tf) / 2
-                    y_plot = np.max(seq_.type.amplitude.samples)
+                    y_plot = np.max(seq_.type.amplitude.samples) / 2
                     a.text(
-                        x_plot, y_plot+8, fr"$\phi$ = {phase_val:.3g}$\pi$",
-                        fontsize=14, ha="center", va="center",
-                    )
-                    a.text(
-                        x_plot, y_plot+2, fr"A = {area_val:3g}$\pi$",
-                        fontsize=14, ha="center", va="center",
+                        x_plot, y_plot, fr"$\phi$: {phase_str(phase_val)}"
+                        + "\n" + fr"A: {area_val:.2g}$\pi$", fontsize=10,
+                        ha="center", va="center", bbox=ph_box,
                     )
 
         target_regions = []     # [[start1, [targets1], end1],...]
--- a/pulser/_seq_drawer.py
+++ b/pulser/_seq_drawer.py
@@ -212,9 +212,11 @@ def draw_sequence(seq, sampling_rate=None, draw_phase_area=False):
                 # Select only `Pulse` objects
                 if isinstance(seq_.type, Pulse):
                     if sampling_rate:
-                        area_val = np.sum(cs_amp(
-                            np.arange(seq_.ti, seq_.tf)
-                        )) * 1e-3 / np.pi
+                        area_val = (
+                            np.sum(
+                                cs_amp(np.arange(seq_.ti, seq_.tf)/time_scale)
+                            ) * 1e-3 / np.pi
+                        )
                     else:
                         area_val = seq_.type.amplitude.integral / np.pi
                     phase_val = seq_.type.phase / np.pi
--- a/pulser/_seq_drawer.py
+++ b/pulser/_seq_drawer.py
@@ -112,9 +112,6 @@ def draw_sequence(seq, sampling_rate=None, draw_phase_area=False):
     fig = plt.figure(constrained_layout=False, figsize=(20, 4.5*n_channels))
     gs = fig.add_gridspec(n_channels, 1, hspace=0.075)
 
-    if draw_phase_area:
-        fig.suptitle(r"$\phi$: Phase, A: Area", fontsize=14)
-
     ch_axes = {}
     for i, (ch, gs_) in enumerate(zip(seq._channels, gs)):
         ax = fig.add_subplot(gs_)
@@ -211,13 +208,21 @@ def draw_sequence(seq, sampling_rate=None, draw_phase_area=False):
                     phase_val = seq_.type.phase / np.pi
                     area_val = seq_.type.amplitude.integral / np.pi
                     x_plot = (seq_.ti + seq_.tf) / 2
-                    if pulse_num % 2 == 0:
+                    if (
+                        seq._schedule[ch][pulse_num-1].type == "target"
+                        or pulse_num % 2 == 0
+                    ):
                         y_plot = np.max(seq_.type.amplitude.samples) / 2
                     else:
                         y_plot = np.max(seq_.type.amplitude.samples)
+                    if phase_val == 0:
+                        txt = fr"A: {area_val:.2g}$\pi$"
+                    else:
+                        phase_fmt = fr"$\phi$: {phase_str(phase_val)}"
+                        area_fmt = fr"A: {area_val:.2g}$\pi$"
+                        txt = "\n".join([phase_fmt, area_fmt])
                     a.text(
-                        x_plot, y_plot, fr"$\phi$: {phase_str(phase_val)}"
-                        + "\n" + fr"A: {area_val:.2g}$\pi$", fontsize=10,
+                        x_plot, y_plot, txt, fontsize=10,
                         ha="center", va="center", bbox=ph_box,
                     )
 
--- a/pulser/_seq_drawer.py
+++ b/pulser/_seq_drawer.py
@@ -213,7 +213,7 @@ def draw_sequence(seq, sampling_rate=None, draw_phase_area=False):
                 if isinstance(seq_.type, Pulse):
                     if sampling_rate:
                         area_val = np.sum(
-                            yaeff[int(seq_.ti*1e-3):int(seq_.tf*1e-3)]
+                            np.abs(yaeff[int(seq_.ti):int(seq_.tf)])
                         ) / 1e3 / np.pi
                     else:
                         area_val = seq_.type.amplitude.integral / np.pi
--- a/pulser/_seq_drawer.py
+++ b/pulser/_seq_drawer.py
@@ -204,6 +204,10 @@ def draw_sequence(seq, sampling_rate=None, draw_phase_area=False):
 
         if draw_phase_area:
             top = False  # Variable to track position of box, top or center.
+            draw_phase = any(
+                seq_.type.phase != 0 for seq_ in seq._schedule[ch]
+                if isinstance(seq_.type, Pulse)
+            )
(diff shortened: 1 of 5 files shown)


HGSilveri:
Hey, thanks for the contribution! Here are a few suggestion:
- Try to make the text occupy less space, maybe bring the lines closer together and reduce the font size
- Put a box around the text
- Filter the values for specific cases (i.e. when it's 0, don't put the π, and when it's 1π omit the 1). The function `phase_str` should do that for you

Yash-10:
Thanks for your suggestions @HGSilveri !

I tried to occupy lesser space for the text by decreasing their font-size and put a box around the text. 

There seems to be atleast three things pending: 

- As done for the phase, the area value would also need formatting (i.e. writing `π` instead of `1π` and `0` instead of `0π`.  I would also do that.
- The title seems to be way above the graph using the default arguments. I would try to fix it too.
- Include the changes in the tests to avoid coverage failures.

I changed the placement of the text box to be in the center of the pulse as shown below:

new_sequence

This could be a better choice than the earlier implementation since otherwise the text has a high chance to go out of the plot. Could you let me know your thoughts on this placement?

Thanks!

HGSilveri:
@Yash-10 looks better! I like the placement in the center, but I'm curious to see how it would look if you had a lot of shorter pulses. How would the boxes overlap, then? Can you test for that?

Yash-10:
For pulses having lesser area and higher amplitude, the boxes do overlap (for example seen below):

Screenshot from 2021-05-20 21-17-44

Moreover, in cases where the pulse has a very low amplitude and short duration, the box almost covers the whole pulse.

For such extreme cases, we could do something like:

```python
text = a.text(...)
```
and then add some conditions on `text._bbox_patch.get_height()` and `text._bbox_patch.get_width()` to check if it's area is an extreme case (i.e. one of the situations from above) and handle the placement of the box as per the situation.

If this idea doesn't look like an overkill, we could proceed with that?

Yash-10:
@HGSilveri I am trying to handle the cases shown above but it seems more trickier than expected, since it requires finding coordinates of the adjacent bounding boxes and adjust them if they overlap. I am working to find better ways to handle such overlaps.

Additionally, I was thinking of the below points:

- Although this might need some thinking, but could we think of having interactive plots in some way to display area and phase on a sequence plot? As of now, I am not sure if this would solve the problem, but could be worth looking into?
- We could use annotations with arrows to prevent overlaps but it needs detecting overlaps beforehand. So this might still be difficult.

Could you let me know your thoughts on the above?

HGSilveri:
> Additionally, I was thinking of the below points:
> 
> * Although this might need some thinking, but could we think of having interactive plots in some way to display area and phase on a sequence plot? As of now, I am not sure if this would solve the problem, but could be worth looking into?
> * We could use annotations with arrows to prevent overlaps but it needs detecting overlaps beforehand. So this might still be difficult.
> 
> Could you let me know your thoughts on the above?

Both of these ideas would be very nice indeed, but I'm afraid I can't accurately predict how feasible they are. If you have some experience or are just willing to try, that would be great! Here are my thoughts:

- With an interactive plot, if you could hover over a pulse and it would tell it it's area and phase, that would solve this issue quite elegantly. In principle, we could even leave this feature as it stands as a backup, since the interactive plot would become the go-to.
- This one is probably less ambitious, though it also has it's challenges. However, if you can keep a box vertically aligned with the center of a pulse, I guess you could do without it. So the juice on this one might not be worth the squeeze.

I must say I have never tried to create interactive plots with matplotlib, so my assistance on this will probably be limited.

Yash-10:
Hello @HGSilveri, thanks for your guidance.

I tried to look into the idea of interactive plots. Although they might be visually appealing, it seems that it could complicate the code a bit more and additionally, even though not a major point, would require additional configuration to set up interactive figures.

I went back to the idea of simple text to display area and phase values. With some tweaking, I realized the plotting could get considerably robust to pulse shapes, by adding a check which ensures that text boxes for close pulses are never overlapped.

The advantages of this, I think are:

- The code doesn't get complicated since now we have a simple check to place the text box at the top and the center.
- It would be robust to various shapes of pulses since the boxes have negligible chances of overlap due to the way they have been placed.
- Micro improvement (with respect to the case where the text box is at the top or center only when there is an overlap): Takes slightly lesser time to render.

The plot now looks like this:

ss2

Possible disadvantages:

- The boxes may not follow a pattern (for eg: all boxes in the middle or all boxes on the top), 


Considering that the aim of writing phase and area values on the plot was just to be able to quickly identify phase and area for a particular pulse, this disadvantage might not be big.

Thanks!

HGSilveri:
Hi @Yash-10 , very nice job, I like how this is looking!

>The boxes may not follow a pattern (for eg: all boxes in the middle or all boxes on the top)

This is quite alright. As you said, this is just to quickly identify the phase and areas, so it's not a problem. 

I have a couple remarks:

- On the first row of your example, you see the box overlapping with the `t` orange box. To avoid this, perhaps the box should be in the center after a `target` command
- The legend as a title looks a bit out of place. If you can't bring it down, perhaps it's best to remove it. I think the symbols pretty much speak for themselves anyway
- When the phase is always zero, it seems a bit redundant to print it. Is it easy to omit it in those cases?

Yash-10:
Thanks again for your suggestions. I have added conditions such that all text boxes after the `target` command are placed in the center.

> The legend as a title looks a bit out of place. If you can't bring it down, perhaps it's best to remove it. I think the symbols pretty much speak for themselves anyway


Oh yes, it was a bit off since the title was for the whole figure and it seems matplotlib keeps some space if we have to set title for the subplots. Here, I have removed it as per your suggestions.


> When the phase is always zero, it seems a bit redundant to print it. Is it easy to omit it in those cases?

I had a question here. Do you mean dropping the text for phase when the phase is zero for each pulse in the channel or to ignore writing phase value whenever a particular pulse has phase 0 ? In the latest commit, I have added a check for the latter, as seen below:

ss3

If you meant the former, could you let me know?

Thanks.

HGSilveri:
@Yash-10 Looks very nice, I will review the code today. 

Regarding the phase, what I meant was what you did, to not write the it when it is 0. On second thought, I think the best way to go about it would be to omit it if it is 0 *and* it doesn't change along that channel (eg. the second channel in your example). If you have a channel where some pulses have phase 0 and others don't, I think it's good to explicitly state it.

pulser/_seq_drawer.py
```diff
@@ -199,6 +201,31 @@ def phase_str(phi):
...
+                    phase_val = seq_.type.phase / np.pi
+                    area_val = seq_.type.amplitude.integral / np.pi
+                    x_plot = (seq_.ti + seq_.tf) / 2
+                    if (
+                        seq._schedule[ch][pulse_num-1].type == "target"
+                        or pulse_num % 2 == 0
+                    ):
+                        y_plot = np.max(seq_.type.amplitude.samples) / 2
+                    else:
+                        y_plot = np.max(seq_.type.amplitude.samples)
```
HGSilveri, line 217: This will work most of the time, but what if you have two very close pulses after a `target`, and the first one has `pulse_num` odd? This will make them both place the box in the middle, which will make it overlap, right?

A solution to this is to have a variable storing whether the last box was placed on the top or the center, and to use its value to decide where the current box should go. Of course, after a `target` you would overwrite it.

HGSilveri (changes requested):
Alright, this is looking good! 

I realized there is one other thing to take into account, though: the `Simulation` class has a `draw()` method too, which uses the `sampling_rate` argument. This is supposed to display the actual shape being used in the simulation vs. the ideal sequence, which can differ when the `sampling_rate` is low. In these cases, I think it would make more sense to print the area of the actual pulse, rather than the ideal one. It won't be so simple like calling the `integral` method of the pulse, but in principle you can extract the information you need from `yaeff` using the `ti` and `tf` of the pulse.

If you have any questions, I'm happy to help out.

pulser/_seq_drawer.py
```diff
@@ -199,6 +202,34 @@ def phase_str(phi):
...
         b.set_ylabel(r'$\delta$ (rad/µs)', fontsize=14)
 
+        if draw_phase_area:
+            top = False  # Variable to track position of box, top or center.
+            for pulse_num, seq_ in enumerate(seq._schedule[ch]):
+                # Select only `Pulse` objects
+                if isinstance(seq_.type, Pulse):
+                    phase_val = seq_.type.phase / np.pi
+                    area_val = seq_.type.amplitude.integral / np.pi
+                    x_plot = (seq_.ti + seq_.tf) / 2
```
HGSilveri, line 212: Be careful, this needs to be adjusted for the `time_scale` (it changes to µs when the total duration reaches 10 µs). Try to make a long enough sequence and check if the boxes are still placed correctly.

pulser/_seq_drawer.py
```diff
@@ -199,6 +202,34 @@ def phase_str(phi):
...
+                    if (
+                        seq._schedule[ch][pulse_num-1].type == "target"
+                        or not top
+                    ):
+                        y_plot = np.max(seq_.type.amplitude.samples) / 2
+                        top = True  # Next box at the top.
+                    elif top:
+                        y_plot = np.max(seq_.type.amplitude.samples)
+                        top = False  # Next box at the center.
```
HGSilveri, line 221: Yup, that's it!

pulser/sequence.py
```diff
@@ -643,9 +643,11 @@ def deserialize(obj, **kwargs):
         return json.loads(obj, cls=PulserDecoder, **kwargs)
 
     @_screen
-    def draw(self):
+    def draw(self, draw_phase_area=False):
```
HGSilveri, line 646: Can you add info on the `draw_phase_area` to this docstring?

HGSilveri (changes requested):
Alright, thanks for addressing my previous comments! Here are a couple more.

After this, all that is left is to:
- Change the criterium for omitting the phase 
- Tackle the `sampling_rate` case (don't worry, it shouldn't be too bad). 

I'm happy to help out, let me know if you need it.

pulser/_seq_drawer.py
```diff
@@ -104,8 +106,8 @@ def phase_str(phi):
     time_scale = 1e3 if seq._total_duration > 1e4 else 1
 
     # Boxes for qubit and phase text
-    q_box = dict(boxstyle="round", facecolor='orange')
-    ph_box = dict(boxstyle="round", facecolor='ghostwhite')
+    q_box = dict(boxstyle="round", facecolor='orange', alpha=0.7)
+    ph_box = dict(boxstyle="round", facecolor='ghostwhite', alpha=0.7)
```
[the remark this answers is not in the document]
Yash-10: Thanks for this. I had tried it to check for visibility, just for checking purposes. Now it's removed.

pulser/_seq_drawer.py
Yash-10: I think it is addressed now..Thanks

pulser/sequence.py
Yash-10: Added it now, thanks, I forgot to do it earlier.

pulser/_seq_drawer.py
Yash-10: Thanks, will check this..

pulser/_seq_drawer.py
```diff
@@ -204,6 +204,12 @@ def phase_str(phi):
 
         if draw_phase_area:
             top = False  # Variable to track position of box, top or center.
+            draw_phase = True
+            if all(
+                seq_.type.phase == 0 for seq_ in seq._schedule[ch]
+                if isinstance(seq_.type, Pulse)
+            ):
+                draw_phase = False
```
HGSilveri, line 212: This is perfectly fine, but I guess you can be a bit more economical:
```suggestion
            draw_phase = any(
                seq_.type.phase != 0 for seq_ in seq._schedule[ch]
                if isinstance(seq_.type, Pulse)
            )
```
What do you think?

Yash-10: Thanks! I changed the condition..

HGSilveri:
> * I tried to add a condition where if all the phase values are zero, then omit writing phase values for each pulse. But if any pulse in a channel has a non-zero phase, it would print phase values for each pulse. I think I may have addressed it now.

Looks good, take a look at my (equivalent) suggestion.

> To summarize, pending things:
> 
> * [ ]  Tackling the `sampling_rate` case - I am currently trying to find a way to solve this. Will surely get back to you in case of any questions.

I should give you a hint here: The pulses are all encoded in `yaeff`, so from `ti` and `tf` you should be able to retrieve the real values. `yaeff` is a spline (I believe the times here are in µs, but you should confirm this and take that into account), so it will give you a value for whatever `t` you give it. To mimick the integral function of `Pulse`, I suggest you sum the values from `ti` to `tf` in 1ns steps.  

> * [ ]  Adjust for `time_scale`.

This is just a matter of dividing `x_plot` by `time_scale`

> * [ ]  Address test failures.

What you have is a coverage fail. No tests are failing, you just have some lines that are uncovered, so you have to add tests that cover them.

Yash-10:
Thank you for the hints @HGSilveri ! Currently, I am not so confident with the way I have implemented the condition for `sampling_rate`. The `time_scale` has been adjusted and the coverage seems to be fine too.

For `sampling_rate` condition:

- I was probably confused in the units of time, `ns` and `µs` since the units seems to depend on the `time_scale`. Hence I have used the `time_scale` parameter for handling this.

> To mimick the integral function of Pulse, I suggest you sum the values from ti to tf in 1ns steps.

I was slightly confused in summing in 1ns steps. If we use the `time_scale` parameter, this is same as simply summing (as done in the current commit), isn't it?

Thank you very much for putting up with me!

pulser/_seq_drawer.py
```diff
@@ -199,6 +202,43 @@ def phase_str(phi):
...
+                seq_.type.phase != 0 for seq_ in seq._schedule[ch]
+                if isinstance(seq_.type, Pulse)
+            )
+            for pulse_num, seq_ in enumerate(seq._schedule[ch]):
+                # Select only `Pulse` objects
+                if isinstance(seq_.type, Pulse):
+                    if sampling_rate:
+                        area_val = np.sum(
+                            yaeff[seq_.ti*time_scale:seq_.tf*time_scale+1]
+                        ) / time_scale / np.pi
```
HGSilveri, line 217: Okay, I can understand your confusion. So, `yaeff` is not actually an array, so you can't slice it like this. Rather, it is a function to which you can give a time `t` and it will return the respective value. Also, it always expects `t` to be in µs.

On the other hand, `seq_.ti` and `seq_.tf` are always in ns, so you shouldn't multiply by `time_scale` but rather by 1e3 before you give it to `yaeff`. 

Other remarks:
- You should stop right before `seq_tf`, so you shouldn't add the `+1`. 
- After summing you should indeed adjust the value, but not by `time_scale`, by `1e3`(as is done in `Waveform.integral`)

HGSilveri:
> For `sampling_rate` condition:
> 
> * I was probably confused in the units of time, `ns` and `µs` since the units seems to depend on the `time_scale`. Hence I have used the `time_scale` parameter for handling this.

So, I understand your confusion since this is not explicitly stated, but `time_scale` is just used to determine the units of the x axis of the plots. For the other objects, the units of time don't change.

> I was slightly confused in summing in 1ns steps. If we use the `time_scale` parameter, this is same as simply summing (as done in the current commit), isn't it?

See my last two comments, hopefully they help.

pulser/_seq_drawer.py
Yash-10: Thanks, I interpreted `yaeff` as a numpy array due to these lines which were already defined earlier:

```py
cs_amp = CubicSpline(solver_time, ya2)
...
yaeff = cs_amp(teff)
```

As a result, the types I get are:

```
type(cs_amp)
>> scipy.interpolate._cubic.CubicSpline
```

and

```
type(yaeff)
>> numpy.ndarray
```

Have I missed something here? Since I do get `yaeff` as an array.

Yash-10:
The earlier change gave unrealistic area values on graph. Now I tried to change it and it seems to be plausible!

(The plots are generated from the sequence defined in `test_simulation.py`)

**Sequence**
R1

**Simulation**
R2

The slight differences in areas, I think might be because the area is calculated on an interpolated pulse in the case of a simulation...

pulser/_seq_drawer.py
Yash-10: > Also, you can give cs_amp a list or array of all the times you want to evaluate over, you don't have to give it one at a time.

@HGSilveri since `teff` was already calculated, I thought the calculation would come down to just indexing on the array of times for the whole sequence. So does this solution also look fine?

Thanks!

HGSilveri: `yaeff` is indeed an array, which corresponds to the amplitude values at the times `teff`. When the sampling rate is 1, `teff` just goes in 1ns time steps, so that's why you seem to be able to index it like this. However, when the sampling rate is lower, this will not hold. I will make a comment on the relevant lines.

pulser/_seq_drawer.py
```diff
@@ -199,6 +202,43 @@ def phase_str(phi):
...
+                seq_.type.phase != 0 for seq_ in seq._schedule[ch]
+                if isinstance(seq_.type, Pulse)
+            )
+            for pulse_num, seq_ in enumerate(seq._schedule[ch]):
+                # Select only `Pulse` objects
+                if isinstance(seq_.type, Pulse):
+                    if sampling_rate:
+                        area_val = np.sum(
+                            np.abs(yaeff[int(seq_.ti):int(seq_.tf)])
+                        ) / 1e3 / np.pi
```
HGSilveri, line 217: What you need here is this:
```suggestion
                    if sampling_rate:
                        area_val = np.sum(cs_amp(
                            np.arange(seq_.ti, seq_.tf) * 1e-3
                        )) * 1e-3 / np.pi
```

HGSilveri: Test it with a low sampling rate

HGSilveri (changes requested):
Hope this helps

pulser/_seq_drawer.py
```diff
@@ -106,6 +108,7 @@ def phase_str(phi):
     # Boxes for qubit and phase text
     q_box = dict(boxstyle="round", facecolor='orange')
     ph_box = dict(boxstyle="round", facecolor='ghostwhite')
+    area_ph_box = dict(boxstyle='round', facecolor='ghostwhite')
```
HGSilveri, line 111: Also, I noticed you removed the box transparency. I think you should keep it, it's good to see the pulse behind

pulser/_seq_drawer.py
```diff
@@ -199,6 +202,43 @@ def phase_str(phi):
...
+                    if (
+                        seq._schedule[ch][pulse_num-1].type == "target"
+                        or not top
+                    ):
+                        y_plot = np.max(seq_.type.amplitude.samples) / 2
+                        top = True  # Next box at the top.
+                    elif top:
+                        y_plot = np.max(seq_.type.amplitude.samples)
+                        top = False  # Next box at the center.
+                    area_fmt = fr"A: {area_val:.2g}$\pi$"
```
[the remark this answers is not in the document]
Yash-10: Thanks, I have added that condition..

pulser/_seq_drawer.py
Yash-10: Oh yes, it would be better to have pulse still shown behind the box. I have added  the `alpha` parameter.

pulser/_seq_drawer.py
Yash-10: Thanks again for this suggestion.

When I tried your suggestion, I get these values on the plot for a simulation:

img1

If I try to use:

```diff
area_val = np.sum(cs_amp(
-                      np.arange(seq_.ti, seq_.tf) * 1e-3
                 )) * 1e-3 / np.pi
area_val = np.sum(cs_amp(
+                        np.arange(seq_.ti, seq_.tf)
                    )) * 1e-3 / np.pi
```
I get this plot:

img2

Both the above plots are with a low sampling rate. Moreover, it gives such plausible values for different sampling rates (eg: 1, 0.1, 0.001).

It seems that the lower plot gives correct values? Could you let me know your thoughts?

(The above diff is not in the latest commit)

pulser/_seq_drawer.py
```diff
@@ -199,6 +202,44 @@ def phase_str(phi):
...
+                    if (
+                        seq._schedule[ch][pulse_num-1].type == "target"
+                        or not top
+                    ):
+                        y_plot = np.max(seq_.type.amplitude.samples) / 2
+                        top = True  # Next box at the top.
+                    elif top:
+                        y_plot = np.max(seq_.type.amplitude.samples)
+                        top = False  # Next box at the center.
+                    area_fmt = (r"A: $\pi$" if area_val == 1
```
HGSilveri, line 231: You should check for the rounded value, ie:
```suggestion
                    area_fmt = (r"A: $\pi$" if round(area_val, 2) == 1
```

pulser/_seq_drawer.py
HGSilveri: Yes, you're clearly right. It seems that `cs_amp` is in ns instead of µs as I thought. Well spotted

Yash-10: It seems to be fixed now..thanks for confirming!

pulser/_seq_drawer.py
Yash-10: Thanks, I used your suggestion.

pulser/_seq_drawer.py
```diff
@@ -199,6 +202,44 @@ def phase_str(phi):
...
+            draw_phase = any(
+                seq_.type.phase != 0 for seq_ in seq._schedule[ch]
+                if isinstance(seq_.type, Pulse)
+            )
+            for pulse_num, seq_ in enumerate(seq._schedule[ch]):
+                # Select only `Pulse` objects
+                if isinstance(seq_.type, Pulse):
+                    if sampling_rate:
+                        area_val = np.sum(cs_amp(
+                            np.arange(seq_.ti, seq_.tf)
```
HGSilveri, line 216: nit: Please check if you can avoid the line break (I can't tell from here)

HGSilveri (approved):
Alright, this looks great to me! Can you please post the plots of how it's looking right now?

pulser/_seq_drawer.py
HGSilveri: Actually, I took yet another look at the code, and I think you need to adjust the indices to `cs_amp` according to `time_scale`, like this:
```suggestion
                            np.arange(seq_.ti, seq_.tf) / time_scale
```
Can you please confirm this by doing an extra plot where the total duration of the sequence goes over 10 µs (this will change the `time_scale`). Just add a large delay to the end of your current test sequence.

Yash-10: ### Plots

(Below are two examples)

a)
1. **Sequence**

f_seq1

2. **Simulation**

f_sim1

b)
1. **Sequence**

f_seq2

2. **Simulation**

f_sim2

---
#### Example of the long duration *simulation* plot

long_dur_sim

The time scale on the x-axis does change from `ns` to `µs`

Yash-10: > nit: Please check if you can avoid the line break (I can't tell from here)

Does the current commit's line break look good? It needs a bit of adjustment for `flake8` to pass :)

> Alright, this all looks correct. This is with the time_scale adjustment, right?

Yes, the plots are with the `time_scale` adjustment. Thanks for that too!

HGSilveri (approved):
This all looks good, I guess you're done! Good job!

Yash-10:
Thanks @HGSilveri for all the help throughout !

PR #167, opened 2021-05-19
[unitaryHACK] Add type hints to `sequence.py` (#16)

LaurentAjdnik:
Not many clues to figure out the exact data structures... => That was endless! 😱

MyPy-checked but probably lots of things left to adjust.

### DIFF
--- a/pulser/register.py
+++ b/pulser/register.py
@@ -55,7 +55,6 @@ class Register:
         self._coords = coords
 
     @property
-
     def qubits(self) -> dict[QubitId, np.ndarray]:
         """Dictionary of the qubit names and their position coordinates."""
         return dict(zip(self._ids, self._coords))
--- a/pulser/register.py
+++ b/pulser/register.py
@@ -12,15 +12,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-
 from __future__ import annotations
 
+from collections.abc import Mapping
 import matplotlib.pyplot as plt
 from matplotlib import collections as mc
 import numpy as np
 from numpy.typing import ArrayLike
 from scipy.spatial import KDTree
-from typing import Any, cast, Dict, Iterable, Optional, Mapping, Union
+from typing import Any, cast, Iterable, Optional, Union
 
 import pulser
 from pulser.json.utils import obj_to_dict
@@ -55,7 +55,8 @@ class Register:
         self._coords = coords
 
     @property
-    def qubits(self) -> Dict[QubitId, ArrayLike]:
+
+    def qubits(self) -> dict[QubitId, np.ndarray]:
         """Dictionary of the qubit names and their position coordinates."""
         return dict(zip(self._ids, self._coords))
 
@@ -364,7 +365,8 @@ class Register:
         self._coords = [rot @ v for v in self._coords]
 
     def draw(self, with_labels: bool = True,
-             blockade_radius: Optional[float] = None, draw_graph: bool = True,
+             blockade_radius: Optional[float] = None,
+             draw_graph: bool = True,
              draw_half_radius: bool = False) -> None:
         """Draws the entire register.
 
@@ -444,6 +446,6 @@ class Register:
 
         plt.show()
 
-    def _to_dict(self) -> dict:
+    def _to_dict(self) -> dict[str, Any]:
         qs = dict(zip(self._ids, map(np.ndarray.tolist, self._coords)))
         return obj_to_dict(self, qs)
--- a/pulser/register.py
+++ b/pulser/register.py
@@ -336,9 +336,12 @@ class Register:
             raise ValueError(f"Spacing ({spacing}) for this device must be"
                              f" {device.min_atom_distance} or above.")
 
-        if n_qubits == 1:
-            return cls.from_coordinates(np.array([(0.0, 0.0)], dtype=float),
-                                        center=False, prefix=prefix)
+        if n_qubits < 7:
+            hex_coords = np.array([(0.0, 0.0), (1.0, 0.0), (0.5, np.sqrt(3/4)),
+                                   (1.5, np.sqrt(3/4)), (2.0, 0.0),
+                                   (0.5, -np.sqrt(3/4))])
+            return cls.from_coordinates(spacing * hex_coords[:n_qubits],
+                                        prefix=prefix)
 
         full_layers = int((-3.0 + np.sqrt(9 + 12 * (n_qubits - 1))) / 6.0)
         atoms_left = n_qubits - 1 - (full_layers**2 + full_layers) * 3
--- a/pulser/register.py
+++ b/pulser/register.py
@@ -426,13 +426,10 @@ class Register:
                 raise NotImplementedError("Needs more than one atom to draw "
                                           "the blockade radius.")
 
-            delta_um = np.linalg.norm(pos[1] - pos[0])
-            r_pts = np.linalg.norm(
-                np.subtract(*ax.transData.transform(pos[:2]).tolist())
-            ) / delta_um * blockade_radius / 2
-            # A 'scatter' marker of size s has area pi/4 * s
-            ax.scatter(pos[:, 0], pos[:, 1], s=4*r_pts**2, alpha=0.1,
-                       c='darkgreen')
+            for p in pos:
+                circle = plt.Circle(tuple(p), blockade_radius/2, alpha=0.1,
+                                    color='darkgreen')
+                ax.add_patch(circle)
         if draw_graph and blockade_radius is not None:
             epsilon = 1e-9      # Accounts for rounding errors
             edges = KDTree(pos).query_pairs(blockade_radius * (1 + epsilon))
--- a/pulser/register.py
+++ b/pulser/register.py
@@ -15,12 +15,12 @@
 
 from __future__ import annotations
 
-from typing import Optional
-
 import matplotlib.pyplot as plt
 from matplotlib import collections as mc
 import numpy as np
+from numpy.typing import ArrayLike
 from scipy.spatial import KDTree
+from typing import Any, Dict, Iterable, Optional, Mapping, cast
 
 import pulser
 from pulser.json.utils import obj_to_dict
@@ -35,7 +35,7 @@ class Register:
             (e.g. {'q0':(2, -1, 0), 'q1':(-5, 10, 0), ...}).
     """
 
-    def __init__(self, qubits: dict):
+    def __init__(self, qubits: Mapping[Any, ArrayLike]):
         """Initializes a custom Register."""
         if not isinstance(qubits, dict):
             raise TypeError("The qubits have to be stored in a dictionary "
@@ -53,7 +53,7 @@ class Register:
         self._coords = coords
 
     @property
-    def qubits(self) -> dict:
+    def qubits(self) -> Dict[Any, np.ndarray]:
         """Dictionary of the qubit names and their position coordinates."""
         return dict(zip(self._ids, self._coords))
 
@@ -82,7 +82,7 @@ class Register:
             pre = str(prefix)
             qubits = {pre+str(i): pos for i, pos in enumerate(coords)}
         else:
-            qubits = dict(enumerate(coords))
+            qubits = dict(cast(Iterable, enumerate(coords)))
         return cls(qubits)
 
     @classmethod
--- a/pulser/register.py
+++ b/pulser/register.py
@@ -12,11 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+
+from __future__ import annotations
+
+from typing import Optional
+
 import matplotlib.pyplot as plt
 from matplotlib import collections as mc
 import numpy as np
 from scipy.spatial import KDTree
 
+import pulser
 from pulser.json.utils import obj_to_dict
 
 
@@ -29,7 +35,7 @@ class Register:
             (e.g. {'q0':(2, -1, 0), 'q1':(-5, 10, 0), ...}).
     """
 
-    def __init__(self, qubits):
+    def __init__(self, qubits: dict):
         """Initializes a custom Register."""
         if not isinstance(qubits, dict):
             raise TypeError("The qubits have to be stored in a dictionary "
@@ -47,12 +53,13 @@ class Register:
         self._coords = coords
 
     @property
-    def qubits(self):
+    def qubits(self) -> dict:
         """Dictionary of the qubit names and their position coordinates."""
         return dict(zip(self._ids, self._coords))
 
     @classmethod
-    def from_coordinates(cls, coords, center=True, prefix=None):
+    def from_coordinates(cls, coords: np.ndarray, center: bool = True,
+                         prefix: Optional[str] = None) -> Register:
         """Creates the register from an array of coordinates.
 
         Args:
@@ -65,6 +72,9 @@ class Register:
             prefix (str): The prefix for the qubit ids. If defined, each qubit
                 id starts with the prefix, followed by an int from 0 to N-1
                 (e.g. prefix='q' -> IDs: 'q0', 'q1', 'q2', ...).
+
+        Returns:
+            Register: A register with qubits placed on the given coordinates.
         """
         if center:
             coords = coords - np.mean(coords, axis=0)      # Centers the array
@@ -76,7 +86,8 @@ class Register:
         return cls(qubits)
 
     @classmethod
-    def rectangle(cls, rows, columns, spacing=4, prefix=None):
+    def rectangle(cls, rows: int, columns: int, spacing: float = 4.0,
+                  prefix: Optional[str] = None) -> Register:
         """Initializes the register with the qubits in a rectangular array.
 
         Args:
@@ -88,14 +99,32 @@ class Register:
             prefix (str): The prefix for the qubit ids. If defined, each qubit
                 id starts with the prefix, followed by an int from 0 to N-1
                 (e.g. prefix='q' -> IDs: 'q0', 'q1', 'q2', ...)
+
+        Returns:
+            Register: A register with qubits placed in a rectangular array.
         """
+        # Check rows
+        if rows < 1:
+            raise ValueError(
+                f"The number of rows ({rows}) must be 1 or above.")
+
+        # Check columns
+        if columns < 1:
+            raise ValueError(
+                f"The number of columns ({columns}) must be 1 or above.")
+
+        # Check spacing
+        if spacing <= 0.0:
+            raise ValueError(f"Spacing ({spacing}) must be above 0.0.")
+
         coords = np.array([(x, y) for y in range(rows)
                            for x in range(columns)], dtype=float) * spacing
 
         return cls.from_coordinates(coords, center=True, prefix=prefix)
 
     @classmethod
-    def square(cls, side, spacing=4, prefix=None):
+    def square(cls, side: int, spacing: float = 4.0,
+               prefix: Optional[str] = None) -> Register:
         """Initializes the register with the qubits in a square array.
 
         Args:
@@ -106,11 +135,21 @@ class Register:
             prefix (str): The prefix for the qubit ids. If defined, each qubit
                 id starts with the prefix, followed by an int from 0 to N-1
                 (e.g. prefix='q' -> IDs: 'q0', 'q1', 'q2', ...).
+
+        Returns:
+            Register: A register with qubits placed in a square array.
         """
+        # Check side
+        if side < 1:
+            raise ValueError(
+                f"The number of atoms per side ({side}) must be 1 or above.")
+
         return cls.rectangle(side, side, spacing=spacing, prefix=prefix)
 
     @classmethod
-    def triangular_lattice(cls, rows, atoms_per_row, spacing=4, prefix=None):
+    def triangular_lattice(cls, rows: int, atoms_per_row: int,
+                           spacing: float = 4.0,
+                           prefix: Optional[str] = None) -> Register:
         """Initializes the register with the qubits in a triangular lattice.
 
         Initializes the qubits in a triangular lattice pattern, more
@@ -126,7 +165,26 @@ class Register:
             prefix (str): The prefix for the qubit ids. If defined, each qubit
                 id starts with the prefix, followed by an int from 0 to N-1
                 (e.g. prefix='q' -> IDs: 'q0', 'q1', 'q2', ...).
+
+        Returns:
+            Register: A register with qubits placed in a triangular lattice.
         """
+
+        # Check rows
+        if rows < 1:
+            raise ValueError(
+                f"The number of rows ({rows}) must be 1 or above.")
+
+        # Check atoms per row
+        if atoms_per_row < 1:
+            raise ValueError(
+                f"The number of atoms per row ({atoms_per_row})"
+                " must be 1 or above.")
+
+        # Check spacing
+        if spacing <= 0.0:
+            raise ValueError(f"Spacing ({spacing}) must be above 0.0.")
+
         coords = np.array([(x, y) for y in range(rows)
                            for x in range(atoms_per_row)], dtype=float)
         coords[:, 0] += 0.5 * np.mod(coords[:, 1], 2)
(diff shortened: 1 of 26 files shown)


pulser/sequence.py
```diff
@@ -543,7 +563,7 @@ def align(self, *channels):
             if delta > 0:
                 self._delay(delta, id)
 
-    def build(self, **vars):
+    def build(self, **vars: Any) -> Sequence:
```
HGSilveri, line 566: Actually for the `vars`, you can specify them to be: `Union[int, float, str, ArrayLike]`

pulser/sequence.py
```diff
@@ -640,14 +660,14 @@ def deserialize(obj, **kwargs):
...
 
-    @_screen
-    def draw(self):
+    @ _screen
+    def draw(self) -> None:
         """Draws the sequence in its current state."""
         draw_sequence(self)
 
-    def _target(self, qubits, channel):
+    def _target(self, qubits: Union[Iterable, str], channel: str) -> None:
```
HGSilveri, line 670: Type of qubit ID

pulser/sequence.py
```diff
@@ -715,7 +735,8 @@ def _delay(self, duration, channel):
         self._add_to_schedule(channel,
                               _TimeSlot('delay', ti, tf, last.targets))
 
-    def _phase_shift(self, phi, *targets, basis='digital'):
+    def _phase_shift(self, phi: float, *targets: str,
```
HGSilveri, line 738: Type of qubit ID

pulser/sequence.py
```diff
@@ -734,20 +755,20 @@ def _phase_shift(self, phi, *targets, basis='digital'):
...
             return
 
-        for q in targets:
-            t = self._last_used[basis][q]
-            new_phase = self._phase_ref[basis][q].last_phase + phi
-            self._phase_ref[basis][q][t] = new_phase
+        for qubit in targets:
+            last_used = self._last_used[basis][qubit]
+            new_phase = self._phase_ref[basis][qubit].last_phase + phi
+            self._phase_ref[basis][qubit][last_used] = new_phase
```
HGSilveri, line 761: Thanks, it was indeed confusing

pulser/sequence.py
```diff
@@ -640,14 +660,14 @@ def deserialize(obj, **kwargs):
             warnings.warn("The given JSON formatted string does not encode a "
                           "Sequence.")
 
-        return json.loads(obj, cls=PulserDecoder, **kwargs)
+        return cast(Sequence, json.loads(obj, cls=PulserDecoder, **kwargs))
 
-    @_screen
-    def draw(self):
+    @ _screen
```
HGSilveri, line 665: nit: The space after the `@`. Is it your IDE that does this?

pulser/sequence.py
```diff
@@ -824,39 +845,41 @@ def _reset_parametrized(self):
...
-    def __init__(self, initial_phase):
-        self._times = [0]
-        self._phases = [self._format(initial_phase)]
+    def __init__(self, initial_phase: float):
+        self._times: List[int] = [0]
+        self._phases: List[float] = [self._format(initial_phase)]
 
-    @property
-    def last_time(self):
+    @ property
```
HGSilveri, line 852: nit: Space after the `@`

pulser/sequence.py
```diff
@@ -824,39 +845,41 @@ def _reset_parametrized(self):
...
+    def last_time(self) -> int:
         return self._times[-1]
 
-    @property
-    def last_phase(self):
+    @ property
```
HGSilveri, line 856: nit: Space after the `@`

pulser/sequence.py
```diff
@@ -824,39 +845,41 @@ def _reset_parametrized(self):
...
+    def last_phase(self) -> float:
         return self._phases[-1]
 
-    def changes(self, ti, tf, time_scale=1):
+    def changes(self, ti: float, tf: float,
```
HGSilveri, line 860: I guess `ti` and `tf` can be `int` too.

HGSilveri (changes requested):
Here is the my first review. I know these are a lot of comments, but trust me, this is a really nice job, especially considering the complexity of the task at hand.

pulser/sequence.py
```diff
@@ -133,39 +140,46 @@ def __init__(self, register, device):
...
+        self._device: Device = device
+        self._in_xy: bool = False
         self._calls = [_Call("__init__", (register, device), {})]
-        self._channels = {}
-        self._schedule = {}
-        self._phase_ref = {}  # The phase reference of each channel
+        self._channels: Dict[str, Channel] = {}
+        self._schedule: Dict[str, List[_TimeSlot]] = {}
+        # The phase reference of each channel
+        self._phase_ref: Dict[str, Dict[str, _PhaseTracker]] = {}
```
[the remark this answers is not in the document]
HGSilveri: On second thought, it would be best to use `TypeVar` to define a type for qubit ID's and use it throughout

LaurentAjdnik:
> this is a really nice job, especially considering the complexity of the task at hand.

Mostly considering this is my very first contact ever with Python. :blush:

I can't figure out how to solve this:

```
pulser/sequence.py:304: error: Item "float" of "Union[Set[Union[float, int, str]], float, str]" has no attribute "__iter__" (not iterable)
pulser/sequence.py:697: error: Argument 1 to "set" has incompatible type "Union[Iterable[Any], float, str]"; expected "Iterable[Any]"
```

Any advice ?

pulser/register.py
```diff
@@ -20,11 +20,13 @@
 import numpy as np
 from numpy.typing import ArrayLike
 from scipy.spatial import KDTree
-from typing import Any, Dict, Iterable, Optional, Mapping, cast
+from typing import Any, cast, Dict, Iterable, Optional, Mapping, Union
 
 import pulser
 from pulser.json.utils import obj_to_dict
 
+QubitId = Union[float, int, str]
```
HGSilveri, line 28: Instead of doing this, you should do:
```suggestion
QubitId = TypeVar('QubitId', float, int, str)
```

Change this on both files (or maybe define it here and import it in `sequence.py` from `pulser.register`) and see if it solves some of your issues. I will review the whole thing again after that

HGSilveri:
> I can't figure out how to solve this:
> 
> ```
> pulser/sequence.py:304: error: Item "float" of "Union[Set[Union[float, int, str]], float, str]" has no attribute "__iter__" (not iterable)
> pulser/sequence.py:697: error: Argument 1 to "set" has incompatible type "Union[Iterable[Any], float, str]"; expected "Iterable[Any]"
> ```
> 
> Any advice ?

These seem to be related to using `Set` to type-hint arguments. According to [this](https://docs.python.org/3/library/typing.html#typing.Set), you should use `AbstractSet` instead.

pulser/register.py
LaurentAjdnik: It did not solve the problem and, in fact, it made things much worse. ☹️

I think using a simple alias is fine in this case : https://mypy.readthedocs.io/en/stable/kinds_of_types.html#type-aliases.

```
pulser\sequence.py:154: error: Type variable "pulser.sequence.QubitId" is unbound
pulser\sequence.py:154: note: (Hint: Use "Generic[QubitId]" or "Protocol[QubitId]" base class to bind "QubitId" inside a class)
pulser\sequence.py:154: note: (Hint: Use "QubitId" in function signature to bind "QubitId" inside a function)
pulser\sequence.py:158: error: Argument 1 to "set" has incompatible type "KeysView[QubitId]"; expected "Iterable[QubitId]"
pulser\sequence.py:160: error: Type variable "pulser.sequence.QubitId" is unbound
pulser\sequence.py:160: note: (Hint: Use "Generic[QubitId]" or "Protocol[QubitId]" base class to bind "QubitId" inside a class)
pulser\sequence.py:160: note: (Hint: Use "QubitId" in function signature to bind "QubitId" inside a function)
pulser\sequence.py:173: error: Incompatible return value type (got "Dict[QubitId, Union[Union[int, float, complex, str, bytes, generic], Sequence[Union[int, float, complex, str, bytes, generic]], Sequence[Sequence[Any]], _SupportsArray]]", expected "Dict[float, Union[Union[int, float, complex, str, bytes, generic], Sequence[Union[int, float, complex, str, bytes, generic]], Sequence[Sequence[Any]], _SupportsArray]]")
pulser\sequence.py:173: error: Incompatible return value type (got "Dict[QubitId, Union[Union[int, float, complex, str, bytes, generic], Sequence[Union[int, float, complex, str, bytes, generic]], Sequence[Sequence[Any]], _SupportsArray]]", expected "Dict[int, Union[Union[int, float, complex, str, bytes, generic], Sequence[Union[int, float, complex, str, bytes, generic]], Sequence[Sequence[Any]], _SupportsArray]]")
pulser\sequence.py:173: error: Incompatible return value type (got "Dict[QubitId, Union[Union[int, float, complex, str, bytes, generic], Sequence[Union[int, float, complex, str, bytes, generic]], Sequence[Sequence[Any]], _SupportsArray]]", expected "Dict[str, Union[Union[int, float, complex, str, bytes, generic], Sequence[Union[int, float, complex, str, bytes, generic]], Sequence[Sequence[Any]], _SupportsArray]]")
pulser\sequence.py:304: error: Item "float" of "Union[Set[float], float]" has no attribute "__iter__" (not iterable)
pulser\sequence.py:304: error: Item "int" of "Union[Set[int], int]" has no attribute "__iter__" (not iterable)
```

pulser/sequence.py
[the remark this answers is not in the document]
LaurentAjdnik: > There might be a minor issue here with the type of the qubits, as it doesn't strictly have to be a `str`, they can be `int` or `float` too (as long as they serve as keys for `dict`).

I think `float` should not be allowed as keys of `dict` to identify qubits. It does not make much sense. And it seems to result in a lot of troubles.

pulser/register.py
HGSilveri: The problem with using just an alias for an `Union` is that it's inappropriate to be used as an argument to container types like `Dict` and `Set`, for example. It's only fine as an alias for the QubitId on its own

pulser/sequence.py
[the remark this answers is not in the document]
HGSilveri: I agree, I just mentioned it because it can indeed be used as a dictionary key, but I'm in favour of not allowing it explicitly, just `int` and `str`. Sorry for the mix-up

pulser/register.py
HGSilveri: But I believe you're right in saying that `TypeVar` is not appropriate either, now that I'm digging deeper into the docs (I'm sorry, I'm new to all this stuff too). I thought it was a solution to the problem I was encountering before of using `Union` to specify multiple possible argument types for `Dict`

LaurentAjdnik: I found a way to put it that passed all the tests... I'll push it now. I've been using `QubitId = Union[int, str]` and it seems to work fine with `dict`.

HGSilveri: Alright, let's go with it then, perhaps the error I encountered was particular to the case I was dealing with.

LaurentAjdnik:
OK... Tests all succeed locally.

There has been a merge conflict last time I commited. I tried to solve it. Now, it keeps on failing CI.

And it's getting on my nerves... Help!

HGSilveri:
@LaurentAjdnik which Python version do you use locally?

LaurentAjdnik:
> @LaurentAjdnik which Python version do you use locally? I'm asking out of suspicion that it is 3.9. Going back to PEP 585, here's what it says about earlier version (which are used in CI):

`Python 3.9.5` indeed.

> As far as I can tell, the CI tests are failing when the built-ins are used outside of type annotations (e.g. when using the `cast`).

That solved it all! Nice!

HGSilveri:
@LaurentAjdnik sorry about the style error, that was probably me while solving the merge conflicts from my last PR. Just delete the blank line I forgot at register.py: line 58.

As for the type error, I believe it is coming from me typing the output of `qubits` to `dict[QubitId, np.ndarray]` (you had `dict[QubitId, ArrayLike]`, which I don't think is quite right). I'll look further into it

pulser/sequence.py
```diff
@@ -133,39 +144,46 @@ def __init__(self, register, device):
...
+        self._variables: dict[str, Variable] = {}
+        self._to_build_calls: list[_Call] = []
+        self._building: bool = True
 
         # Initializes all parametrized Sequence related attributes
         self._reset_parametrized()
 
     @property
-    def qubit_info(self):
+    def qubit_info(self) -> dict[QubitId, ArrayLike]:
```
HGSilveri, line 171: If you change this to:
```suggestion
    def qubit_info(self) -> dict[QubitId, np.ndarray]:
```
it should fix the mypy error.

pulser/sequence.py
```diff
@@ -217,7 +236,10 @@ def current_phase_ref(self, qubit, basis='digital'):
 
         return self._phase_ref[basis][qubit].last_phase
 
-    def declare_channel(self, name, channel_id, initial_target=None):
+    def declare_channel(self, name: str, channel_id: str,
+                        initial_target: Optional[Union[set[QubitId],
+                                                       QubitId]] = None
```
HGSilveri, line 241: I know it says so in the docstring (you should change it too btw), but the `initial_target` doesn't actually have to be a `set[QubitId]`, but rather `Iterable[QubitId]` (same as the `target` method).

Also, each element can be a variable instead of `QubitId`, so the complete form would be something like:
```suggestion
                        initial_target: Optional[Union[Iterable[Union[QubitId, Parametrized]],
                                                       Union[QubitId, Parametrized]]] = None
```
(perhaps it's a good idea to use an alias, as this seems quite verbose)

pulser/sequence.py
```diff
@@ -296,11 +319,12 @@ def declare_channel(self, name, channel_id, initial_target=None):
...
-                            "declare_channel",
-                            (name, channel_id),
-                            {"initial_target": initial_target}))
+            "declare_channel",
+            (name, channel_id),
+            {"initial_target": initial_target}))
 
-    def declare_variable(self, name, size=1, dtype=float):
+    def declare_variable(self, name: str, size: int = 1,
+                         dtype: type = float) -> Variable:
```
HGSilveri, line 327: Let's specify the possible types here: `dtype` can be `Union[type[int], type[float], type[str]]`

pulser/sequence.py
```diff
@@ -332,7 +356,9 @@ def declare_variable(self, name, size=1, dtype=float):
         return var
 
     @_store
-    def add(self, pulse, channel, protocol='min-delay'):
+    def add(self, pulse: Union[Pulse, Parametrized],
+            channel: Union[str, Parametrized],
+            protocol: Union[str, Parametrized] = 'min-delay') -> None:
```
HGSilveri, line 361: Since when `protocol` is a `str`, it can only be `'min-delay', 'no-delay', 'wait-for-all'`, I think it would be nice to use a [`Literal`](https://docs.python.org/3/library/typing.html#typing.Literal) instead of `str`.

I would do an alias for valid protocols, like is done with `MODE` in the link above.

pulser/sequence.py
```diff
@@ -425,16 +454,17 @@ def add(self, pulse, channel, protocol='min-delay'):
...
+            if self._last_used[basis][qubit] < tf:
+                self._last_used[basis][qubit] = tf
 
         if pulse.post_phase_shift:
             self._phase_shift(pulse.post_phase_shift, *last.targets,
                               basis=basis)
 
     @_store
-    def target(self, qubits, channel):
+    def target(self, qubits: Union[QubitId, Parametrized],
```
HGSilveri, line 466: What about:
```suggestion
    def target(self, qubits: Union[QubitId, Iterable[QubitId], Parametrized],
```

pulser/sequence.py
```diff
@@ -640,25 +684,24 @@ def deserialize(obj, **kwargs):
...
             draw_phase_area (bool): Whether phase and area values need
                 to be shown as text on the plot, defaults to False.
         """
-        draw_sequence(
-            self, draw_phase_area=draw_phase_area,
-        )
+        draw_sequence(self, draw_phase_area=draw_phase_area)
 
-    def _target(self, qubits, channel):
+    def _target(self, qubits: Union[Iterable, QubitId], channel: str) -> None:
```
HGSilveri, line 699: Let's put `Iterable[QubitId]` instead.

pulser/sequence.py
```diff
@@ -640,25 +684,24 @@ def deserialize(obj, **kwargs):
...
         self._validate_channel(channel)
 
         try:
-            qs = set(qubits) if not isinstance(qubits, str) else {qubits}
+            qs = set(cast(AbstractSet, qubits)) if not isinstance(
```
HGSilveri, line 703: Cast to `Iterable` instead

HGSilveri (changes requested):
We're nearly there!

pulser/sequence.py
```diff
@@ -315,16 +316,17 @@ def declare_channel(self, name: str, channel_id: str,
                 initial_target = None
             else:
                 # "_target" call is not saved
-                self._target(initial_target, name)
+                self._target(cast(Union[Iterable[QubitId],
```
HGSilveri, line 319: Hum... This casting doesn't work with a parametrized `Iterable` imported from `collections.abc`. I you remove the `[QubitId]`, does mypy complain?

HGSilveri (commented):
See below the line that's failing the CI tests

HGSilveri:
@LaurentAjdnik The failed CI test is because `typing.get_args` was only added in Python 3.8. It's a shame, but I'm afraid we'll have to do without it, we can't give up on Python 3.7 just yet.

pulser/sequence.py
```diff
@@ -12,32 +12,42 @@
...
 from pulser.json.utils import obj_to_dict
 from pulser.parametrized import Parametrized, Variable
-from pulser._seq_drawer import draw_sequence
+from pulser.register import Register
+
+QubitId = Union[int, str]
+PROTOCOLS = Literal['min-delay', 'no-delay', 'wait-for-all']
 
 # Auxiliary class to store the information in the schedule
 _TimeSlot = namedtuple('_TimeSlot', ['type', 'ti', 'tf', 'targets'])
```
HGSilveri, line 45: I think it would be nice to have type annotations for `_TimeSlot` (no need to do it for `_Call`). Here's how that's done: https://docs.python.org/3/library/typing.html#typing.NamedTuple

LaurentAjdnik:
> @LaurentAjdnik The failed CI test is because `typing.get_args` was only added in Python 3.8. It's a shame, but I'm afraid we'll have to do without it, we can't give up on Python 3.7 just yet.

Yup. But removing it would fail:

```
        valid_protocols = PROTOCOLS
        if protocol not in valid_protocols:
```
Mypy:
```
pulser\sequence.py:393: error: Unsupported right operand type for in ("object")
pulser\sequence.py:395: error: Argument 1 to "join" of "str" has incompatible type "object"; expected "Iterable[str]"
Found 2 errors in 1 file (checked 34 source files)
```
Pytest:
```
TypeError: Parameters to generic types must be types. Got 0.
```

How can wecheck that some value belongs to a `Union` of `Literal`s in Python 3.7 ?

HGSilveri:
How about doing:
```
valid_protocols = ['min-delay', 'no-delay', 'wait-for-all`]
PROTOCOLS = Literal[*valid_protocols]
```

LaurentAjdnik:
> How about doing:
> 
> ```
> valid_protocols = ['min-delay', 'no-delay', 'wait-for-all`]
> PROTOCOLS = Literal[*valid_protocols]
> ```
> 
> In this way you can use the `valid_protocols` variable.

Yes, it's we had before. But I wanted to reuse the `PROTOCOLS` alias in order to avoid duplicates.

HGSilveri:
> But I wanted to reuse the `PROTOCOLS` alias in order to avoid duplicates.

Yes, I get that, it just might not be worth the trouble here. This way it's not a pure duplicate, as `PROTOCOLS` will be defined from `valid_protocols`

LaurentAjdnik:
> How about doing:
> 
> ```
> valid_protocols = ['min-delay', 'no-delay', 'wait-for-all`]
> PROTOCOLS = Literal[*valid_protocols]
> ```
> 
> In this way you can use the `valid_protocols` variable.

```
SyntaxError: invalid syntax
Unpack operator in subscript is not supported
Unpacked arguments cannot be used in type argument lists
```

HGSilveri:
Damn... I guess there's not much we can do, is there?

LaurentAjdnik:
> Edit: Other than `PROTOCOLS.__args__`, which is ugly but works

```
pulser\sequence.py:397: error: "object" has no attribute "__args__"
```

HGSilveri (approved):
Alright, this looks good to me! I only have a couple aesthetic comments

pulser/sequence.py
```diff
@@ -12,32 +12,47 @@
...
+
+QubitId = Union[int, str]
+PROTOCOLS = Literal['min-delay', 'no-delay', 'wait-for-all']
 
 # Auxiliary class to store the information in the schedule
-_TimeSlot = namedtuple('_TimeSlot', ['type', 'ti', 'tf', 'targets'])
+_TimeSlot: NamedTuple('_TimeSlot', [('type', str), ('ti', int), ('tf', int), (
+    'targets', Iterable[QubitId])]) = namedtuple(
+        '_TimeSlot', ['type', 'ti',
+                      'tf', 'targets'])
```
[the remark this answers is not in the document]
LaurentAjdnik: New errors appear:

```
pulser\sequence.py:444: error: Unsupported left operand type for & ("Iterable[Union[int, str]]")
pulser\sequence.py:465: error: Argument 1 to "_TimeSlot" has incompatible type "Pulse"; expected "str"
pulser\sequence.py:752: error: Unsupported operand types for - ("None" and "int")
pulser\sequence.py:752: note: Left operand is of type "Optional[int]"
pulser\sequence.py:756: error: Incompatible types in assignment (expression has type "int", variable has type "Union[number[Any], ndarray]")
pulser\sequence.py:757: error: Argument 1 to "validate_duration" of "Channel" has incompatible type "Union[number[Any], ndarray]"; expected "int"
pulser\sequence.py:765: error: Incompatible types in assignment (expression has type "Union[number[Any], Any]", target has type "int")
pulser\sequence.py:766: error: Argument 3 to "_TimeSlot" has incompatible type "Union[number[Any], Any]"; expected "int"
```

But I feel they make sense and have to be adressed.

For instance, for the error on line 752:

```
delta = np.clip(retarget - elapsed, 0, retarget)
```

We type hinted `Channel.retarget_time` as `Optional[int] = None`. Shouldn't we check if it is `None` and, if it is, force 0 in the calculations ?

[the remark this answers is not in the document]
HGSilveri: Ok, a few things:
- For `'type'`, it can indeed be `Union[Pulse, str]`
- `targets` is always `Set[QubitId]` (this is `collections.abc.Set`, which is substitutes `typing.AbstractSet`)
- For line 752, the channel has to be "Local", which forces a `retarget_time` to be defined. The parameter is `Optional` because it shouldn't be defined when the channel is "Global". So, in the case you mentioned, it's safe to cast to `int`

Try to address these and we'll see what remains

pulser/sequence.py
```diff
@@ -321,7 +324,7 @@ def declare_channel(self, name: str, channel_id: str,
                 initial_target = None
             else:
                 # "_target" call is not saved
-                self._target(cast(Union[Iterable,
+                self._target(cast(Union[Set[QubitId],
```
HGSilveri, line 327: Don't forget this casting doesn't work before Python 3.9... Just leave `Iterable` as before, it's fine as an argument to `_target`

pulser/sequence.py
```diff
@@ -243,7 +246,7 @@ def current_phase_ref(self, qubit: QubitId,
 
     def declare_channel(self, name: str, channel_id: str,
                         initial_target: Optional[
-                            Union[Iterable[Union[QubitId, Parametrized]],
+                            Union[Set[Union[QubitId, Parametrized]],
```
HGSilveri, line 249: Even though what's stored in `_TimeSlot.targets` is a `set`, we can (and want to) accept an `Iterable` here

HGSilveri (changes requested):
Almost there!

LaurentAjdnik:
The whole error message is `ERROR pulser/tests/test_simulation.py - TypeError: 'ABCMeta' object is not subscriptable when trying to annotate a hash variable`, I guess.

?!?

pulser/sequence.py
LaurentAjdnik: `pulser\sequence.py:327: error: Argument 1 to "_target" of "Sequence" has incompatible type "Union[Iterable[Union[int, str]], Union[int, str]]"; expected "Union[AbstractSet[Union[int, str]], Union[int, str]]"`

HGSilveri: Yes, you have to change back to `Iterable` on the `_target` type annotation too

pulser/sequence.py
```diff
@@ -324,7 +324,7 @@ def declare_channel(self, name: str, channel_id: str,
                 initial_target = None
             else:
                 # "_target" call is not saved
-                self._target(cast(Union[Set[QubitId],
+                self._target(cast(Union[Iterable[QubitId],
```
HGSilveri, line 327: Cast to `Iterable` instead of `Iterable[QubitId]` to fix the pytest error

pulser/sequence.py
```diff
@@ -12,32 +12,50 @@
...
 from pulser.parametrized import Parametrized, Variable
-from pulser._seq_drawer import draw_sequence
+from pulser.register import Register
+
+QubitId = Union[int, str]
+PROTOCOLS = Literal['min-delay', 'no-delay', 'wait-for-all']
+
+
+class _TimeSlot(NamedTuple):
+    # Auxiliary class to store the information in the schedule
```
HGSilveri, line 47: You can change this to a docstring: 
```suggestion
    """Auxiliary class to store the information in the schedule."""
```

pulser/sequence.py
```diff
@@ -12,32 +12,50 @@
...
+
 from collections import namedtuple
-from collections.abc import Iterable
+from collections.abc import Callable, Generator, Iterable, Set
 import copy
 from functools import wraps
 from itertools import chain
 import json
+from typing import Any, cast, NamedTuple, Optional, Tuple, Union
+from typing_extensions import Literal
```
HGSilveri, line 24: I understand you must be saturated at this point, but I found a solution where we can use `get_args` as before. If you want, I can also merge this and do it myself. What do you prefer?

LaurentAjdnik: Let me know and I'll try...

HGSilveri: Alright. In fact, you can import `get_args` from `typing_extensions`, so your first solution would work. However, I would rather we import from `typing_extensions` when necessary (i.e. when the python version being used is 3.7.). So, here's what I would suggest:
```suggestion
import platform
if platform.python_version()[:3] == '3.7':
    try:
        from typing_extensions import Literal, get_args
    except ImportError:
        raise ImportError("Using pulser with Python version 3.7 requires the `typing_extensions` module. Install it by running `pip install typing-extensions`.")
else:
    from typing import Literal, get_args
```
(you'll have to fix the style, it's hard to do it here)

Then you should be able to use `get_args` as before and avoid duplicating `valid_protocols`.

HGSilveri: Also, perhaps using `sys.version_info` might be preferable over `platform.python_version()`, which I suggested above. Just don't use `sys.version`

pulser/sequence.py
```diff
@@ -13,6 +13,18 @@
...
+from pulser.json.utils import obj_to_dict
+from pulser.json.coders import PulserEncoder, PulserDecoder
+from pulser.devices._device_datacls import Device
+from pulser.devices import MockDevice
+from pulser.pulse import Pulse
+from pulser.channels import Channel
+from pulser._seq_drawer import draw_sequence
+import pulser
+from numpy.typing import ArrayLike
+import numpy as np
```
HGSilveri, line 27: Please don't move these imports here. If you want, you can do the conditional imports after these.

LaurentAjdnik:
`FAIL Required test coverage of 100% not reached. Total coverage: 99.94%`

100 % locally. CI ignores `# pragma: no cover` ?

pulser/sequence.py
```diff
@@ -21,23 +33,20 @@
...
+if version_info[:2] == (3, 7):  # pragma: no cover
+    try:
+        from typing_extensions import Literal, get_args
+    except ImportError:
+        raise ImportError(
+            "Using pulser with Python version 3.7 requires the"
+            " `typing_extensions` module. Install it by running"
+            " `pip install typing-extensions`.")
+else:
+    from typing import Literal, get_args  # type: ignore
```
HGSilveri, line 48: Did you mean # pragma: no cover?

HGSilveri (commented):
Looks good, just address these issues

HGSilveri:
> `FAIL Required test coverage of 100% not reached. Total coverage: 99.94%`
> 
> 100 % locally. CI ignores `# pragma: no cover` ?

See my second comment on the last review.

pulser/sequence.py
LaurentAjdnik: ?

HGSilveri: You wrote `# type: ignore`, was `mypy` giving you an error? This line needs to be ignored in the `coverage`, it's what's giving you the failed CI test. So it seems you wrote `# type: ignore` instead of `# pragma: no cover`

LaurentAjdnik: `mypy` does complain without the `# type: ignore`: `pulser\sequence.py:48: error: Incompatible import of "Literal" (imported name has type "typing._SpecialForm", local name has type "typing_extensions._SpecialForm")`.

`# pragma: no cover` is on the same line as the `if` as advised in docs. It works fine when i `pytest` locally but not in CI.

Those are separate issues.

LaurentAjdnik: Well... The `# pragma: no cover` should be on the `else` too...

pulser/sequence.py
```diff
@@ -12,32 +12,60 @@
...
+from pulser.json.utils import obj_to_dict
+from pulser.json.coders import PulserEncoder, PulserDecoder
+from pulser.devices._device_datacls import Device
+from pulser.devices import MockDevice
+from pulser.pulse import Pulse
+from pulser.channels import Channel
 from pulser._seq_drawer import draw_sequence
+import pulser
+from numpy.typing import ArrayLike
+import numpy as np
```
HGSilveri, line 37: Please put these back on the second block of imports. For future reference, the PEP 8 style convention is to divide the imports into three blocks: 1. built-ins, 2. third-party libraries, 3. Local application/library specific imports.

Also, I think `import pulser` looks better at the top of the third block, but this is just a nit.

pulser/sequence.py
```diff
@@ -12,32 +12,60 @@
...
+
+from sys import version_info
```
HGSilveri, line 39: I guess you can put this on the first block too please

pulser/sequence.py
```diff
@@ -12,32 +12,60 @@
...
 from functools import wraps
 from itertools import chain
 import json
-import warnings
-
+from numpy.typing import ArrayLike
 import numpy as np
+from sys import version_info
+from typing import Any, cast, NamedTuple, Optional, Tuple, Union
+import warnings
```
HGSilveri, line 27: Just move these to the bottom of the first block and bring `numpy.typing` here, the rest is okay.

HGSilveri: Also, put `numpy` imports one the second block, as it is a third-party library

pulser/sequence.py
```diff
@@ -12,32 +12,60 @@
...
 from functools import wraps
 from itertools import chain
 import json
+from sys import version_info
+from typing import Any, cast, NamedTuple, Optional, Tuple, Union
 import warnings
 
 import numpy as np
-
+from numpy.typing import ArrayLike
```
HGSilveri, line 28: Ahhh, so close! Only needs a blank line after this one (I'm really sorry for being a pain with these things)

HGSilveri (approved):
Awesome, thank you so much for all your hard work and patience!

PR #168, opened 2021-05-20
Setting up mypy

HGSilveri:
- Add mypy config file
- Adds mypy to CI workflow
- Cleans up errors in mypy checks

PR #169, opened 2021-05-20
Add type hints in `_device_datacls.py` and `coders.py` [unitaryHACK] (#16)

TripleR47:
Type checked on `_device_datacls.py` and `coders.py`. 

Was getting `Skipping analyzing module_name: found module but no type hints or library stubs` error on line 19 of `_device_datacls.py`, wasn't able to correct it, used `#type: ignore` to ignore the same.

### DIFF
--- a/pulser/devices/_device_datacls.py
+++ b/pulser/devices/_device_datacls.py
@@ -12,8 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from __future__ import annotations
 from dataclasses import dataclass
-from typing import Tuple, Dict, Set
+from typing import Tuple, Dict, Set, Any
 
 import numpy as np
 from scipy.spatial.distance import pdist
@@ -52,12 +53,12 @@ class Device:
         self.__dict__["__doc__"] = self._specs(for_docs=True)
 
     @property
-    def channels(self) -> Dict:
+    def channels(self) -> Dict[str, Channel]:
         """Dictionary of available channels on this device."""
         return dict(self._channels)
 
     @property
-    def supported_bases(self) -> Set:
+    def supported_bases(self) -> Set[str]:
         """Available electronic transitions for control and measurement."""
         return {ch.basis for ch in self.channels.values()}
 
@@ -172,6 +173,6 @@ class Device:
 
         return "\n".join(lines + ch_lines)
 
-    def _to_dict(self) -> Dict:
+    def _to_dict(self) -> Dict[str, Any]:
         return obj_to_dict(self, _build=False, _module="pulser.devices",
                            _name=self.name)
--- a/pulser/devices/_device_datacls.py
+++ b/pulser/devices/_device_datacls.py
@@ -13,12 +13,12 @@
 # limitations under the License.
 
 from dataclasses import dataclass
-from typing import Tuple
+from typing import Tuple, Dict, Set
 
 import numpy as np
-from scipy.spatial.distance import pdist
+from scipy.spatial.distance import pdist #type: ignore
 
-from pulser import Register
+from pulser import Register, Pulse
 from pulser.channels import Channel
 from pulser.json.utils import obj_to_dict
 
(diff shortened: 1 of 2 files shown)


LaurentAjdnik:
> Was getting `Skipping analyzing module_name: found module but no type hints or library stubs` error on line 19 of `_device_datacls.py`, wasn't able to correct it, used `#type: ignore` to ignore the same.

Hi @TripleR47! You can use the `--ignore-missing-imports` in the command line and remove the `#type: ignore`.

This option is planned to be set as such in CI by @HGSilveri (see #168).

TripleR47:
> > Was getting `Skipping analyzing module_name: found module but no type hints or library stubs` error on line 19 of `_device_datacls.py`, wasn't able to correct it, used `#type: ignore` to ignore the same.
> 
> Hi @TripleR47! You can use the `--ignore-missing-imports` in the command line and remove the `#type: ignore`.
> 
> This option is planned to be set as such in CI by @HGSilveri (see #168).

Ya, I overlooked that, thanks for letting me know! :smiley:

pulser/devices/_device_datacls.py
```diff
@@ -47,31 +47,31 @@ class Device:
...
     interaction_coeff: float = 5008713.
 
-    def __post_init__(self):
+    def __post_init__(self) -> None:
         # Hack to override the docstring of an instance
         self.__dict__["__doc__"] = self._specs(for_docs=True)
 
     @property
-    def channels(self):
+    def channels(self) -> Dict:
```
HGSilveri, line 55: You can be a bit more specific here:
```suggestion
    def channels(self) -> Dict[str, Channel]:
```

pulser/devices/_device_datacls.py
```diff
@@ -13,12 +13,12 @@
 # limitations under the License.
```
HGSilveri, line 14: Add `from __future__ import annotations`

pulser/devices/_device_datacls.py
```diff
@@ -172,6 +172,6 @@ def _specs(self, for_docs=False):
 
         return "\n".join(lines + ch_lines)
 
-    def _to_dict(self):
+    def _to_dict(self) -> Dict:
```
HGSilveri, line 175: Specify that dict is `Dict[str, Any]`

pulser/json/coders.py
```diff
@@ -21,9 +21,11 @@
 from pulser.json.utils import obj_to_dict
 from pulser.parametrized import Variable
 
+from typing import Dict, Any
```
HGSilveri, line 24: This should be next to the built-in libraries imports (first block)

pulser/json/coders.py
```diff
@@ -21,9 +21,11 @@
...
+
 
 class PulserEncoder(JSONEncoder):
-    def default(self, o):
+    def default(self, o: Any) -> Dict:
```
HGSilveri, line 28: This one is also `Dict[str, Any]`

pulser/json/coders.py
```diff
@@ -37,12 +39,12 @@ def default(self, o):
 
 
 class PulserDecoder(JSONDecoder):
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
         # TODO: Check version compatibility (stored at the Sequence level)
-        self.vars = {}
+        self.vars: Dict = {}
```
HGSilveri, line 44:
```suggestion
        self.vars: Dict[str, Variable] = {}
```

pulser/json/coders.py
```diff
@@ -37,12 +39,12 @@ def default(self, o):
...
         super().__init__(object_hook=self.object_hook, *args, **kwargs)
 
-    def object_hook(self, obj):
+    def object_hook(self, obj: Dict) -> Dict:
```
HGSilveri, line 47:
```suggestion
    def object_hook(self, obj: Dict[str, Any]) -> Any:
```

HGSilveri (changes requested):
Same comment as before: sometimes, you can be more specific, especially in the types of the dictionaries keys and values.

TripleR47:
Updated the suggested changes.

pulser/json/coders.py
```diff
@@ -12,20 +12,20 @@
...
 
 from pulser.json.utils import obj_to_dict
 from pulser.parametrized import Variable
 
-from typing import Dict, Any
-
 
 class PulserEncoder(JSONEncoder):
-    def default(self, o: Any) -> Dict:
+    def default(self, o: Any) -> Union[Dict[str, Any], Any]:
```
HGSilveri, line 28: Doesn't just `Dict[str, Any]` work? A union with `Any` is a bit redundant, is the same as being `Any`

HGSilveri (commented):
Alright, this one seems done, there's just one question to address.

pulser/json/coders.py
TripleR47: Without `Any` it's showing `error: Returning Any from function declared to return "Dict[str, Any]"` in following lines of code - 
* https://github.com/pasqal-io/Pulser/blob/954fef70885a498df8d747c430b4c7d528e93615/pulser/json/coders.py#L28
* https://github.com/pasqal-io/Pulser/blob/954fef70885a498df8d747c430b4c7d528e93615/pulser/json/coders.py#L36

The return types are similar to a dictionary. I think casting will be a better option.

TripleR47: Done

HGSilveri (approved):
LGTM, nice job!

PR #172, opened 2021-05-21
[unitaryHACK] Add type hints to `pulse.py` (#16)

LaurentAjdnik:


pulser/pulse.py
```diff
@@ -60,7 +64,9 @@ def __new__(cls, *args, **kwargs):
         else:
             return object.__new__(cls)
 
-    def __init__(self, amplitude, detuning, phase, post_phase_shift=0):
+    def __init__(self, amplitude: Waveform, detuning: Waveform,
+                 phase: Union[float, Parametrized],
+                 post_phase_shift: Union[float, Parametrized] = 0):
```
HGSilveri, line 69: `amplitude` and `detuning` can actually be `Parametrized` too

pulser/pulse.py
```diff
@@ -136,17 +156,17 @@ def draw(self):
         fig.tight_layout()
         plt.show()
 
-    def _to_dict(self):
+    def _to_dict(self) -> Dict[str, Union[float, Waveform]]:
```
HGSilveri, line 159: I think the `Union` inside the `Dict` args don't really work as expected. Just put `Dict[str, Any]`

HGSilveri (changes requested):
I'll try to wait for `parametrized` and `waveforms` to be merged before this one, but here's a first review.

pulser/pulse.py
```diff
@@ -64,9 +64,10 @@ def __new__(cls, *args, **kwargs):
         else:
             return object.__new__(cls)
 
-    def __init__(self, amplitude: Waveform, detuning: Waveform,
+    def __init__(self, amplitude: Union[Parametrized, Waveform],
+                 detuning: Union[Parametrized, Waveform],
```
HGSilveri, line 68: Nit: Let's keep `Parametrized` as the second parameter, just for consistency's sake.

pulser/pulse.py
```diff
@@ -76,12 +83,18 @@ def __init__(self, amplitude, detuning, phase, post_phase_shift=0):
...
         self.detuning = detuning
+        phase = cast(float, phase)
         self.phase = float(phase) % (2 * np.pi)
+        post_phase_shift = cast(float, post_phase_shift)
         self.post_phase_shift = float(post_phase_shift) % (2 * np.pi)
 
     @classmethod
     @parametrize
-    def ConstantDetuning(cls, amplitude, detuning, phase, post_phase_shift=0):
+    def ConstantDetuning(cls, amplitude: Waveform,
```
HGSilveri, line 93: Don't forget the `Parametrized` option for the `Waveforms` on the classmethods.

pulser/pulse.py
```diff
@@ -95,7 +108,11 @@ def ConstantDetuning(cls, amplitude, detuning, phase, post_phase_shift=0):
 
     @classmethod
     @parametrize
-    def ConstantAmplitude(cls, amplitude, detuning, phase, post_phase_shift=0):
+    def ConstantAmplitude(cls, amplitude: Union[float, Parametrized],
+                          detuning: Waveform,
```
HGSilveri, line 112: Same here

HGSilveri (changes requested):
Alright, these are probably the final comments on this one.

pulser/pulse.py
LaurentAjdnik: > Don't forget the `Parametrized` option for the `Waveforms` on the classmethods.

How will it work with this check in `__init__` ?

```
        if not (isinstance(amplitude, Waveform) and
                isinstance(detuning, Waveform)):
            raise TypeError("'amplitude' and 'detuning' have to be waveforms.")
```

HGSilveri: It's the same as with the `Waveform`s, when it's parametrized it doesn't run (in this case, because of the `@parametrize` decorator). You should cast to `Waveform` inside the function block so that `mypy` doesn't complain, though.

HGSilveri (approved):
LGTM, one more down!

Issue #157, opened 2021-05-14
Consistency for `Error` messages


LaurentAjdnik:
There are sooooooo many ways to write an Error message.

For instance when starting the message:
* "The number of qubits..."
* "'n_qubits'..."
* "The number of qubits 'n_qubits'..."

Then we have to decide whether to include the parameter value that was provided and the allowed values (see #148). And how.

Even the phrasing for allowed values can vary a lot:
* "...must be below..."
* "...must be smaller than..."
* "...must be less than..."
* "...can not be above..."
* "...must not be above..."
* "...can not exceed..."
* "...must not exceed..."
 
Right now, along the code, we have a mix of different approaches.

Maybe we should enact rules to ensure consistency, especially since we intend to add lots of checks and hints.

As an option, we could even create an ErrorMessageFactory class.

One among very few functions:
* Signature: `ErrorMessageFactory.getMessage(subject, parameter, value, constraint, allowed)`
* Example: `ErrorMessageFactory.getMessage("The number of qubits", "n_qubits", -1, ErrorMessageFactory.EQUAL_OR_ABOVE, 1)`
* Output: `"The number of qubits ('n_qubits' = -1) must be 1 or above"`

Of course, we would have to identify the common types of constraints and allowed values (single or interval) but it seems feasable.

At the same time, this would allow us, in test functions, to check for full Error messages with exact matches rather than partial matches or regular expressions.

Maybe I'm splitting hairs here, but I would LOVE to discuss this issue. :grin:

HGSilveri:
Hi @LaurentAjdnik ! Sorry for not getting back to you on this sooner. I'm always in favour of having some systematic way of doing things, so I am on board with this idea. 
I do think it needs a bit more fleshing out though, so I propose we stand by until you're done with your current issue and then we can discuss it further, in case you're still up for it. Sounds good?

LaurentAjdnik:
Hi @HGSilveri, here are some more thoughts about it.

# TL;DR

* 92 errors raised in the `develop` branch
* 8 different error types, but mostly `TypeError` and `ValueError`
* These two can be further refined, depending on why they are raised
* That leads to error messages that could be made more consistent through some kind or "Error Factory"
* Anyhow, we really need to enact guidelines on how to write error messages because they're currently EXTREMELY diverse!
* Especially since we will certainly add more and more checks everywhere

# Diversity

Sometimes this information is given (when applicable), sometimes not, in error messages:

* The name of the parameter
[...]

Issue #148, opened 2021-05-11
Enhanced `ValueError` messages


romainfd:
When using values out of range (examples below for qubit positions and pulse detuning), the `ValueError` exception could indicate the minimum/maximum authorized value to make fixing the issue faster. The value is used in the test just above so printing it looks easy at first glance.

- Illustration for min qubits distance
![image](https://user-images.githubusercontent.com/32517197/117842634-6aa8c180-b27e-11eb-907a-dc4842225388.png)

- Illustration for max pulse detuning
![image](https://user-images.githubusercontent.com/32517197/117842658-6f6d7580-b27e-11eb-886f-74bf4007e33b.png)

LaurentAjdnik:
Hi! I've been doing so for the new functions `Register.hexagon()` and `Register.max_connectivity()` (soon to be PR-ed), so I might just as well continue for others.

Fortunately, some of them already do, like `Channel.validate_duration()` for instance.

Anyhow, this will be a good opportunity to:
* Make sure allowed values are always mentioned as part of Error messages
* Harmonize the way they are included in strings (throughout the code, you'll find F-strings, `.format()` and perhaps others)

By the way, I found almost no checks for `None` on parameters. Testing against allowed values will result in a TypeError (see [Ordering comparisons](https://docs.python.org/3/whatsnew/3.0.html#ordering-comparisons)).

Do we want to add checks and throw an explicit ValueError with a message rather than this default behavior ? This is already the case in `Register.draw()`, `SimulationResults.sample_final_state()` and `Variable.build()` but nowhere else.

HGSilveri:
@LaurentAjdnik thanks! I agree with your suggestions, although I think it would be better to do them in a separate PR if possible, otherwise the one you're currently working on might get too big. 

Regarding checks for `None` on parameters, those do sound a bit of overkill to me, but I will accept them if you feel they are necessary.

LaurentAjdnik:
@HGSilveri: You can assign this one to me, as it is related to #157.

LaurentAjdnik:
> @LaurentAjdnik If you're satisfied with the changes from your last PR, I'll close this, just let me know.

Yup, let's close this one. We might PR a few adjustments in the future but no need to keep this issue open.

PR #185, opened 2021-06-01
Enhance ValueError messages (#148 and #157)

LaurentAjdnik:
For feedback: First attempt to enhance ValueError messages (see #148 and #157), starting with `register.py`.

Guidelines:
* Functional explanation of parameter
* Technical name of parameter
* Provided value of parameter
* Using "must" rather than any other verb
* No negation
* Using "less than / equal to / greater than" rather than any other expression
* Recall allowed values
* All sentences follow the same structure

### DIFF
--- a/pulser/tests/test_register.py
+++ b/pulser/tests/test_register.py
@@ -147,11 +147,13 @@ def test_max_connectivity():
         reg = Register.max_connectivity(2, None)
 
     # Check min number of atoms
-    with pytest.raises(ValueError, match=r"The number of qubits(.+)or above"):
+    with pytest.raises(ValueError,
+                       match=r"The number of qubits(.+)greater than"):
         reg = Register.max_connectivity(0, device)
 
     # Check max number of atoms
-    with pytest.raises(ValueError, match=r"The number of qubits(.+)exceed"):
+    with pytest.raises(ValueError,
+                       match=r"The number of qubits(.+)less than"):
         reg = Register.max_connectivity(max_atom_num + 1, device)
 
     # Check spacing
--- a/pulser/register.py
+++ b/pulser/register.py
@@ -107,24 +107,25 @@ class Register:
         """
         # Check rows
         if rows < 1:
-            raise ValueError(
-                f"The number of rows ({rows}) must be 1 or above.")
+            raise ValueError(f"The number of rows (`rows` = {rows})"
+                             " must be greater than or equal to 1.")
 
         # Check columns
         if columns < 1:
-            raise ValueError(
-                f"The number of columns ({columns}) must be 1 or above.")
+            raise ValueError(f"The number of columns (`columns` = {columns})"
+                             " must be greater than or equal to 1.")
 
         # Check spacing
         if spacing <= 0.0:
-            raise ValueError(f"Spacing ({spacing}) must be above 0.0.")
+            raise ValueError(f"Spacing between atoms (`spacing` = {spacing})"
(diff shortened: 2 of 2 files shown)


pulser/tests/test_register.py
```diff
@@ -147,11 +147,13 @@ def test_max_connectivity():
         reg = Register.max_connectivity(2, None)
 
     # Check min number of atoms
-    with pytest.raises(ValueError, match=r"The number of qubits(.+)or above"):
+    with pytest.raises(ValueError,
+                       match=r"The number of qubits(.+)greater than"):
```
HGSilveri, line 151: Cool, I didn't know about this syntax

HGSilveri (commented):
No complaints, I agree that this is much cleaner

HGSilveri:
@LaurentAjdnik So, how do you want to go about this? Doing this for the whole code seems like a big task. But also, I was wondering about how we can enforce it automatically, for anyone contributing to the code base.

I had an idea to have a test in CI that would search the code for specific errors and check if their message is valid. For example, it would identify "threshold" value errors and check if the message follows the specific guidelines. What do you think about this?

LaurentAjdnik:
> @LaurentAjdnik So, how do you want to go about this? Doing this for the whole code seems like a big task.

Still, I had planned to do it. 😇

> But also, I was wondering about how we can enforce it automatically, for anyone contributing to the code base.

Automatically? I have not clue... I can only think of writing guidelines in `CONTRIBUTING.md` and checking if it's done when a PR arrives...

> I had an idea to have a test in CI that would search the code for specific errors and check if their message is valid. For example, it would identify "threshold" value errors and check if the message follows the specific guidelines. What do you think about this?

Cool! How would you do that?

HGSilveri:
> > @LaurentAjdnik So, how do you want to go about this? Doing this for the whole code seems like a big task.
> 
> Still, I had planned to do it. 😇

Thanks a lot for the initiative!

> > I had an idea to have a test in CI that would search the code for specific errors and check if their message is valid. For example, it would identify "threshold" value errors and check if the message follows the specific guidelines. What do you think about this?
> 
> Cool! How would you do that?

First, we have to layout all the guidelines for the error messages. At first, we could try to check for specific patterns that all errors of a given type should follow, based on these guidelines. This will be easier said than done, considering the variability that can occur within a single error type (eg `ValueError`). Depending on how far we want to take this, it might be necessary to create custom subclasses to further specify the error involved (e.g. `ValueThresholdError`, etc...), though this might be overkill for now.

LaurentAjdnik:
> @LaurentAjdnik do you have anything to add to this PR before I merge it?

Nope.

HGSilveri (approved):
LGTM

LaurentAjdnik:
> First, we have to layout all the guidelines for the error messages. At first, we could try to check for specific patterns that all errors of a given type should follow, based on these guidelines. This will be easier said than done, considering the variability that can occur within a single error type (eg `ValueError`). Depending on how far we want to take this, it might be necessary to create custom subclasses to further specify the error involved (e.g. `ValueThresholdError`, etc...), though this might be overkill for now.

Indeed, it does sound like a lot of hassle for something not so worth it.

I guess we should keep it simple for now: Let's officialize some guidelines, make sure all existing ValueErrors are updated accordingly (*) and check any PR coming in.

(*) Most of the time, when in doubt, people check existing parts of code to make sure they comply to guidelines. That's what I did anyway. :wink:

PR #186, opened 2021-06-02
Enhanced max_connectivity() under 7 atoms

LaurentAjdnik:
Extends the range of the test to N = 6 atoms.

Disables centering in order to leave the innermost atom on (0,0).

Changes the order of atoms for consistency with N >= 7 atoms (spiral pattern).

Uses `device.min_atom_distance` instead of constant `4` in tests.


sebgrijalva (approved):
LGTM

HGSilveri (approved):
LGTM too

PR #197, opened 2021-06-10
Add simconfig

julien-bremont:
Adds a `SimConfig` class to pulser to be used by a `Simulation` object. Contains noise settings, number of runs, system temperature... Part of the noisy simulations PR.

### DIFF
--- a/pulser/simulation/simconfig.py
+++ b/pulser/simulation/simconfig.py
@@ -77,9 +77,10 @@ class SimConfig:
     epsilon: float = 0.01
     epsilon_prime: float = 0.05
     solver_options: qutip.Options = qutip.Options(max_step=5)
-    spam_dict: dict[str, float] = field(init=False, default_factory=dict)
-    doppler_sigma: float = field(
-        init=False, default=KEFF * np.sqrt(KB * temperature / MASS))
+    spam_dict: dict[str, float] = field(init=False, default_factory=dict,
+                                        repr=False)
+    doppler_sigma: float = field(init=False,
+                                 default=KEFF * np.sqrt(KB * 50.e-6 / MASS))
 
     def __post_init__(self) -> None:
         self._process_temperature()
--- a/pulser/simulation/simconfig.py
+++ b/pulser/simulation/simconfig.py
@@ -77,13 +77,15 @@ class SimConfig:
     epsilon: float = 0.01
     epsilon_prime: float = 0.05
     solver_options: qutip.Options = qutip.Options(max_step=5)
-    spam_dict: dict[str, float] = field(default_factory=dict)
+    spam_dict: dict[str, float] = field(init=False, default_factory=dict)
+    doppler_sigma: float = field(
+        init=False, default=KEFF * np.sqrt(KB * temperature / MASS))
 
     def __post_init__(self) -> None:
         self._process_temperature()
-        self._check_noise_types()
         self.__dict__["spam_dict"] = {'eta': self.eta, 'epsilon': self.epsilon,
                                       'epsilon_prime': self.epsilon_prime}
+        self._check_noise_types()
         self._check_spam_dict()
         self._calc_sigma_doppler()
 
--- a/pulser/simulation/simconfig.py
+++ b/pulser/simulation/simconfig.py
@@ -15,7 +15,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass, field
-from typing import cast
+from typing import Union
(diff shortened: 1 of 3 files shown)


pulser/simulation/simconfig.py
```diff
@@ -0,0 +1,120 @@
...
+from typing_extensions import Literal, get_args
+
+import numpy as np
+import qutip
+
+
+NOISE_TYPES = Literal['doppler', 'amplitude', 'SPAM', 'dephasing']
+MASS = 1.45e-25
+KB = 1.38e-23
+KEFF = 8.7
```
HGSilveri, line 28: Maybe add the units here as comments

pulser/simulation/simconfig.py
```diff
@@ -0,0 +1,120 @@
...
+                    --  eta: Probability of each atom to be badly prepared
+                    --  epsilon: Probability of false positives
+                    --  epsilon_prime: Probability of false negatives.
+            runs (int): Number of runs needed : each run draws a new random
+                noise.
+            samples_per_run (int): Number of samples per noisy run.
+                Useful for cutting down on computing time, but unrealistic.
+            temperature (float): Temperature, set in µK, of the Rydberg array.
+                Also sets the standard deviation of the speed of the atoms.
+            laser_waist (float): Waist of the gaussian laser in global pulses.
```
HGSilveri, line 54: What are the units?

pulser/simulation/simconfig.py
```diff
@@ -0,0 +1,120 @@
...
+        # only one noise was given as argument : convert it to a tuple
+        if isinstance(self.noise, str):
+            self.__dict__["noise"] = (self.noise, )
+        for noise_type in self.noise:
+            if noise_type not in get_args(NOISE_TYPES):
+                raise ValueError(cast(str, noise_type) +
+                                 " is not a valid noise type."
+                                 "Valid noise types : " +
+                                 str(get_args(NOISE_TYPES))
+                                 )
```
HGSilveri, line 115: Use f-strings instead of casting to str and summing.

pulser/tests/test_simconfig.py
```diff
@@ -0,0 +1,28 @@
...
+# limitations under the License.
+
+import pytest
+
+from pulser.simulation import SimConfig
+
+
+def test_init():
+    config = SimConfig()
+    str(config)
```
HGSilveri, line 22: You should initialize a `SimConfig` with custom values and assert that `str(config)` matches what you would expect

HGSilveri (commented):
Alright, it looks like it's almost there!

pulser/simulation/simconfig.py
```diff
@@ -62,8 +68,9 @@ class SimConfig:
     epsilon: float = 0.01
     epsilon_prime: float = 0.05
     solver_options: qutip.Options = qutip.Options(max_step=5)
+    spam_dict: dict[str, float] = field(default_factory=dict)
```
HGSilveri, line 71: Why are you adding this one here? Is it the typing?

pulser/simulation/simconfig.py
```diff
@@ -0,0 +1,124 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from typing import Union
+from typing_extensions import Literal, get_args
```
HGSilveri, line 19: Can you please do the conditional import that is in `sequence.py`? I would rather have users install `typing_extensions` only if they're on Python 3.7, it doesn't make sense to demand that from a user working on Python 3.8.

HGSilveri (commented):
Alright, here are a couple more

pulser/simulation/simconfig.py
julien-bremont: Yeah, `mypy` complains that I use `self.spam_dict` otherwise, since I add it in `__post_init__`.

HGSilveri: Ok, let's hide it from init by setting the `field(init=False, ...)` option. And you can do the same with `doppler_sigma` while you're at it (make it's default be the value corresponding to the default temperature)

HGSilveri (approved):
Awesome, great job! Congrats on your first PR!

PR #183, opened 2021-05-31
XY simulation

Slimane33:
Here is the implementation of the XY simulation on the Mock Device for the `mw_global`channel. There is a joined notebook for the implementation of the examples.

A few remarks:
- the dependency of C_3 with the angle is not implemented yet.
- the specific tests are not implemented yet.
- only the global pulses are available (no local adressing of the atoms)

[test_pulser.pdf](https://github.com/pasqal-io/Pulser/files/6571168/test_pulser.pdf)

PR #198, opened 2021-06-11
[unitaryHACK] Xy simulation

Slimane33:
Implementation of the XY simulation on the Mock Device for the mw_global channel.
See #183


HGSilveri:
@Slimane33 In case you're struggling with the CI tests, you can install and run them locally. To install, just do `pip install -r requirements.txt`. Then you can run `flake8`, `mypy` and `pytest --cov`.

pulser/simulation/simulation.py
```diff
@@ -205,45 +213,87 @@ def _build_basis_and_op_matrices(self) -> None:
...
-            return sum(self._build_operator(op_id, q_id)
-                       for q_id in self._qdict)
-        if len(set(qubit_ids)) < len(qubit_ids):
-            raise ValueError("Duplicate atom ids in argument list.")
-        # List of identity operators, except for op_id where requested:
-        op_list = [self.op_matrix[op_id]
-                   if j in map(self._qid_index.get, qubit_ids)
-                   else self.op_matrix['I'] for j in range(self._size)]
+        op_list = [self.op_matrix['I'] for j in range(self._size)]
+        for operator, qubits in operations.items():
```
sebgrijalva, line 218: I put this comment on the previous PR: Please generalize to keys being `Qobj`s not just id strings and values being either ids (`'atomN'`) or directly indexes (`3, 5`).

pulser/register.py
```diff
@@ -53,6 +55,7 @@ def __init__(self, qubits: Mapping[Any, ArrayLike]):
             raise ValueError("All coordinates must be specified as vectors of"
                              " size 2 or 3.")
         self._coords = coords
+        self._mag_field = mag_field
```
sebgrijalva, line 58: Let's make the magnetic field a "property".  So it will have a `@property` decorator and a setter function (that corresponds in line 459)

pulser/register.py
```diff
@@ -452,3 +455,11 @@ def draw(self, with_labels: bool = True,
     def _to_dict(self) -> dict[str, Any]:
         qs = dict(zip(self._ids, map(np.ndarray.tolist, self._coords)))
         return obj_to_dict(self, qs)
+
+    def set_magnetic_field(self, mag_field: ArrayLike) -> None:
```
sebgrijalva, line 459: Add setter function here instead (Cf. comment above)

sebgrijalva (changes requested):
This looks better now, I think we are close to the desired functionality. Here are some changes to consider:

pulser/simulation/simulation.py
Slimane33: I am not sure that you can make a key of a Qobj

pulser/simulation/simulation.py
```diff
@@ -205,45 +213,87 @@ def _build_basis_and_op_matrices(self) -> None:
...
+            if qubits == 'global':
+                return sum(self._build_operator({operator: [q_id]})
+                           for q_id in self._qdict)
+            else:
+                if len(set(qubits)) < len(qubits):
+                    raise ValueError("Duplicate atom ids in argument list.")
+                for j, qubit in enumerate(qubits):
```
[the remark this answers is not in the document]
Slimane33: In this version it is indeed useless

pulser/simulation/simresults.py
```diff
@@ -104,8 +106,9 @@ def get_final_state(self, reduce_to_basis: Optional[str] = None,
             elif reduce_to_basis == "digital":
                 ex_state = "0"
             else:
-                raise ValueError("'reduce_to_basis' must be 'ground-rydberg' "
-                                 + f"or 'digital', not '{reduce_to_basis}'.")
+                raise ValueError(
+                    "'reduce_to_basis' must be 'ground-rydberg' "
+                    + f"'digital' or 'XY', not '{reduce_to_basis}'.")
```
HGSilveri, line 111: Actually, `reduce_to_basis` is not supported for `XY`, because it's meaningless (you will always have just a two-level system, so there's no reduction to be made). You should state this in the docstring of this method and change this message back.

pulser/simulation/simresults.py
```diff
@@ -180,9 +183,9 @@ def sample_final_state(self, meas_basis: Optional[str] = None,
                     )
             meas_basis = self._meas_basis
 
-        if meas_basis not in {'ground-rydberg', 'digital'}:
+        if meas_basis not in {'ground-rydberg', 'digital', 'XY'}:
             raise ValueError(
-                "'meas_basis' can only be 'ground-rydberg' or 'digital'."
+                "'meas_basis' can only be 'ground-rydberg', digital' or 'XY'."
```
HGSilveri, line 188: Same comment made in `__init__`

HGSilveri (changes requested):
Here are a few things to change on the "peripheral" classes

pulser/devices/_device_datacls.py
```diff
@@ -47,7 +51,8 @@ class Device:
     max_radial_distance: int
     min_atom_distance: int
     _channels: tuple[tuple[str, Channel], ...]
-    interaction_coeff: float = 5008713.
+    interaction_coeff_ising: float = 5008713.
```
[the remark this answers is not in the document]
Slimane33: Why ? What does that change for the rest of the codebase ?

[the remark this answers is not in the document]
HGSilveri: Nothing on this codebase, but there are custom devices outside of it who specify a different value for this coefficient, so this would be a breaking change.

HGSilveri:
@Slimane33 , please take note of the changes on PR #200.

Slimane33:
Thanks for implementing the feature @HGSilveri. I have one question: why enabling the magnetic field only in XY mode? It adds lots of checks...

HGSilveri:
> I have one question: why enabling the magnetic field only in XY mode? It adds lots of checks...

We don't want to have it available in the Ising mode, so it would be misleading if the user could define it. What checks are you worried about?

Slimane33:
> What checks are you worried about?
This part for example
```python
if not self._in_xy:
  if self._channels:
      raise ValueError("The magnetic field can only be set in 'XY "
                       "Mode'.")
  # No channels declared yet
  self._in_xy = True
  elif not self._empty_sequence:
      # Not all channels are empty
      raise ValueError("The magnetic field can only be set on an empty "
                       "sequence.")
```
For me it adds lots of maintenance and it doesn't have consequences on the workflow if it is set up in Ising mode

HGSilveri:
> For me it adds lots of maintenance and it doesn't have consequences on the workflow if it is set up in Ising mode

If the user is in Ising mode, these checks will never be touched unless they are trying to set a magnetic field, which they shouldn't be, since they are in the Ising mode. These checks prevent the user from doing something wrong and go unnoticed when things are done right.

pulser/simulation/simulation.py
sebgrijalva: The function that would work in the general case would be one that takes a list of `(local_operator, position)` tuples and assigns the corresponding local operators from it, for example:

```python
my_sim.build_operator([(qutip.basis(2,1).proj(), 'atom1'), ('sigma_gr', 3), (my_custom_qobj, 2)])
```

Then the code reads the couples one by one and assigns the nontrivial local operators to the initial list of identity matrics.

tutorials/applications/Spin chain of 3 atoms in XY mode.ipynb
```diff
@@ -0,0 +1,364 @@
+{
```
sebgrijalva, line 1: <p>Importing <code>Simulation</code>directly  from pulser will be deprecated soon. Add it as </p><p> </p><pre class="ql-syntax" spellcheck="false">from pulser.simulation import Simulation
</pre><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pasqal-io/Pulser/pull/198/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

sebgrijalva, line 1: <p>Please include some Introductory text so that people know the goals of the tutorial.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pasqal-io/Pulser/pull/198/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

sebgrijalva, line 1: <p>It would be nice to see the drawing of the sequence, to observe the MW channels acting.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pasqal-io/Pulser/pull/198/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

sebgrijalva, line 1: <p>Add labels, the oscillation is between what levels? Explain these are two Rydberg levels which are separated thanks to an applied magnetic field.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pasqal-io/Pulser/pull/198/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

sebgrijalva, line 1: <p>Why is this angle supressing the interaction? Explain that it depends on the square of the cosine and that this is the orientation that cancels the term.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pasqal-io/Pulser/pull/198/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

sebgrijalva, line 1: <p>*Chef's kiss*</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pasqal-io/Pulser/pull/198/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

sebgrijalva, line 1: <p>Explain that this is the principle used to program a topological phase in a longer chain.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pasqal-io/Pulser/pull/198/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

pulser/simulation/simulation.py
```diff
@@ -211,23 +211,44 @@ def _build_basis_and_op_matrices(self) -> None:
...
             )
 
-    def _build_operator(self, operations: dict,
-                        ) -> qutip.Qobj:
-        """Create qutip.Qobj with nontrivial action at *qubit_ids."""
+    def build_operator(self, operations: list) -> qutip.Qobj:
+        """Creates an operator with non trivial actions on some qubits.
+        Takes as argument a list of tuples [(operator_1, qubits_1),
+        (operator_2, qubits_2)...]. Returns the operator given by
+        sum_i {operator_i applied on qubits_i and Id on the rest}.
```
sebgrijalva, line 218: This is actually the tensor product, not a sum.

pulser/simulation/simulation.py
```diff
@@ -211,23 +211,44 @@ def _build_basis_and_op_matrices(self) -> None:
...
+        for operator, qubits in operations:
             if qubits == 'global':
-                return sum(self._build_operator({operator: [q_id]})
+                return sum(self.build_operator([(operator, [q_id])])
                            for q_id in self._qdict)
             else:
                 if len(set(qubits)) < len(qubits):
                     raise ValueError("Duplicate atom ids in argument list.")
+                if isinstance(operator, str):
+                    operator = self.op_matrix[operator]
```
sebgrijalva, line 245: Check if `'operator'` is a valid `op_id`(since now this can be used externally)

pulser/simulation/simulation.py
```diff
@@ -211,23 +211,44 @@ def _build_basis_and_op_matrices(self) -> None:
...
                 for qubit in qubits:
```
sebgrijalva, line 246: Check if `qubit` is either a valid `q_id` or a valid index (i.e. less than the number of atoms)

sebgrijalva (changes requested):
Nice job! It remains to add a few checks and correct a typo in the docstring

tutorials/applications/Spin chain of 3 atoms in XY mode.ipynb
Slimane33: <p>I would like to add a figure. May I create a folder <code>figures</code> in  the folder <code>tutorials</code>?</p><p> <span class="ql-cursor">﻿</span></p>

sebgrijalva: We are including figures in the `docs/source/files` folder, so that the online documentation is also rendered. Include it in the notebooks like this example:

```html
<img src="../../docs/source/files/charles_bronson.png" alt="The Stone Killer" style="width: 500px"/>
```

Slimane33: <p>Thanks</p>

pulser/simulation/simulation.py
```diff
@@ -205,21 +211,62 @@ def _build_basis_and_op_matrices(self) -> None:
...
+
+        for operator, qubits in operations:
+            if qubits == 'global':
+                return sum(self.build_operator([(operator, [q_id])])
+                           for q_id in self._qdict)
+            else:
+                if len(set(qubits)) < len(qubits):
+                    raise ValueError("Duplicate atom ids in argument list.")
+                if isinstance(operator, str):
+                    if operator not in self.op_matrix.keys():
```
HGSilveri, line 249: You don't need to check for it in `keys()` explicitly, it does it by default

pulser/simulation/simulation.py
```diff
@@ -205,21 +211,62 @@ def _build_basis_and_op_matrices(self) -> None:
...
+                        if qubit >= self._size:
+                            raise ValueError(
+                                "The index value is greater than the "
+                                "size of the system")
+                        k = qubit
+                    else:
+                        if qubit not in self._qid_index.keys():
+                            raise ValueError(
+                                f"{qubit} is not a valid qubit")
+                        k = self._qid_index[qubit]
```
HGSilveri, line 264: Maybe I'm missing something here, but can't you simplify this by doing:
```suggestion
                    try:
                        k = self._qid_index[qubit]
                    except KeyError:
                        raise ValueError(f"{qubit} is not a valid qubit")
```

pulser/simulation/simulation.py
```diff
@@ -230,19 +277,59 @@ def adapt(full_array: np.ndarray) -> np.ndarray:
...
+
+        def make_interaction_term() -> float:
+            """Construct interaction term.
+
+            Depending on the type of interaction.
+            """
+            if self._interaction == 'ising':
+                return make_vdw_term()
+            else:
+                return make_xy_term()
```
HGSilveri, line 331: You can actually do this synthetically with:
```suggestion
        make_interaction_term = make_xy_term if self._interaction == 'XY' else make_vdw_term
```

pulser/simulation/simulation.py
```diff
@@ -262,8 +351,8 @@ def build_coeffs_ops(basis: str, addr: str) -> list[list]:
                         # Build once global operators as they are needed
                         if op_id not in operators:
                             operators[op_id] =\
-                                self._build_operator(op_id, global_op=True)
-                        terms.append([operators[op_id], adapt(coeff)])
+                                self.build_operator([(op_id, 'global')])
+                            terms.append([operators[op_id], adapt(coeff)])
```
HGSilveri, line 355: Why should the append be inside the `if` now?

pulser/simulation/simulation.py
```diff
@@ -286,8 +375,8 @@ def build_coeffs_ops(basis: str, addr: str) -> list[list]:
         if self.basis_name == 'digital':
             qobj_list = []
         else:
-            # Van der Waals Interaction Terms
-            qobj_list = [make_vdw_term()] if self._size > 1 else []
+            # XY Interaction Terms
```
HGSilveri, line 378: I guess it's `Van der Walls or XY` right? You can also just delete the comment

HGSilveri (commented):
The implementation looks solid, very nice job! I have a few questions and remarks, most of them style related. I'll review the tests and the notebook on a later PR.

pulser/simulation/simulation.py
```diff
@@ -205,21 +211,62 @@ def _build_basis_and_op_matrices(self) -> None:
...
-        if global_op:
-            return sum(self._build_operator(op_id, q_id)
-                       for q_id in self._qdict)
-        if len(set(qubit_ids)) < len(qubit_ids):
-            raise ValueError("Duplicate atom ids in argument list.")
-        # List of identity operators, except for op_id where requested:
-        op_list = [self.op_matrix[op_id]
-                   if j in map(self._qid_index.get, qubit_ids)
-                   else self.op_matrix['I'] for j in range(self._size)]
+    def build_operator(self, operations: Union[list, tuple]) -> qutip.Qobj:
```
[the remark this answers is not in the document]
Slimane33: That was a suggestion of @sebgrijalva maybe it can be useful to the user

pulser/simulation/simulation.py
Slimane33: Not exactly because qubit can either be an index, or a key for the dictionnary. We can however put your suggestion in the `else`

pulser/simulation/simulation.py
Slimane33: I don't know I don't remember touching this part

HGSilveri: I think it was correct before. This way it seems like it will only work when you are building the operator, which happens only the first time.

pulser/simulation/simulation.py
HGSilveri: The thing is that it can be a key for the dictionary while also being an `int`, one does not exclude the other. You can have the indices of your qubits be `0`, `1`, ...

HGSilveri: If you create a Register with these qubits: `{1: (0,0), 2: (1, 1)}` what you have here will actually be wrong, because `k` for `qubit = 1` is `0` and `k` for `qubit=2` is `1`. This information will still be correct in your `_qid_index` however, which in this case will be `{1: 0, 2: 1}`. You should also use the `_qid_index` to get the value of `k`.

HGSilveri: What I'm getting at is that it is ambiguous to accept both qubit ids and indices, because of this situation where an `int` can be a `qid`, and you have no way of knowing. I would rather avoid this ambiguity altogether by accepting only one type.

Slimane33: True. I guess that would be the responsibility of the user not to mess up. What do you think @sebgrijalva ?

sebgrijalva: I guess keeping the qubit ids is more important since this is what the user chose and can see on the register. In a complicated graph, she will have to check `register.draw()` carefully to assign local operators.

HGSilveri: My vote goes to having just the qubit ids, they're the only ones being used here anyway right?

pulser/simulation/simulation.py
```diff
@@ -320,7 +321,8 @@ def make_xy_term() -> float:
                 )
             return xy
 
-        make_interaction_term = make_xy_term if self._interaction == 'XY' else make_vdw_term
+        make_interaction_term = make_xy_term\
+            if self._interaction == 'XY' else make_vdw_term
```
HGSilveri, line 325: Sorry for being a pain with these nits, but can you use `()` instead of `\` please?

pulser/simulation/simulation.py
```diff
@@ -252,17 +252,22 @@ def build_operator(self, operations: Union[list, tuple]) -> qutip.Qobj:
...
+                    #     if qubit >= self._size:
+                    #         raise ValueError(
+                    #             "The index value is greater than the "
+                    #             "size of the system")
+                    #     k = qubit
+                    # else:
+                    #     if qubit not in self._qid_index.keys():
+                    #         raise ValueError(
+                    #             f"{qubit} is not a valid qubit")
+                    #     k = self._qid_index[qubit]
```
HGSilveri, line 265: You can just delete these instead. If you ever need to retrieve this, you can just go back to a commit where you had it

pulser/simulation/simulation.py
HGSilveri: Did you figure out what's going on here?

pulser/tests/test_simulation.py
```diff
@@ -307,3 +334,57 @@ def test_run():
...
+        [[0, 10], [10, 0], [0, 0]], prefix='atom')
+    detun = 1.
+    amp = 3.
+    rise = Pulse.ConstantPulse(1500, amp, detun, 0.)
+    simple_seq = Sequence(simple_reg, MockDevice)
+    simple_seq.declare_channel('ch0', 'mw_global')
+    simple_seq.set_magnetic_field(0, 1., 0.)
+    simple_seq.add(rise, 'ch0')
+
+    assert np.abs(np.linalg.norm(simple_seq.magnetic_field[0:2]) - 1) < 1e-10
```
HGSilveri, line 350: Just a tip: You can use `np.isclose()` for these comparisons

pulser/tests/test_simulation.py
```diff
@@ -307,3 +334,57 @@ def test_run():
...
+    good_initial_qobj = qutip.tensor([qutip.basis(sim.dim, 0)
+                                      for _ in range(sim._size)])
+
+    sim.run(initial_state=good_initial_array)
+    sim.run(initial_state=good_initial_qobj)
+
+    assert not hasattr(sim._seq, '_measurement')
+    simple_seq.measure(basis='XY')
+    sim.run()
+    assert sim._seq._measurement == 'XY'
```
HGSilveri, line 390: Why are you testing this here?

HGSilveri (commented):
Very nice! I'll move on to the notebook

tutorials/applications/Spin chain of 3 atoms in XY mode.ipynb
HGSilveri: <p>Did you miss this comment from Seb?</p>

tutorials/applications/Spin chain of 3 atoms in XY mode.ipynb
```diff
@@ -0,0 +1,406 @@
...
```
HGSilveri, line 1: <ul><li>You should mention that the value of C3 is depends on the device.</li><li>Both your basis states are named |0&gt;. Also, weren't we using "up" and "down" here?</li></ul>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pasqal-io/Pulser/pull/198/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

HGSilveri, line 1: <p>As far as I can tell, you don't need:</p><ul><li><code>pprint</code></li><li><code>Chadoq2</code></li><li><code>Image</code></li></ul>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pasqal-io/Pulser/pull/198/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

HGSilveri, line 1: <p>This error comes from using <code>pulser</code> installed from PyPi instead of installed from source, but I'm sure you've solved it at this point. It would be good not to have it here though</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pasqal-io/Pulser/pull/198/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

pulser/tests/test_simulation.py
Slimane33: How do you think it should be ?

pulser/tests/test_simulation.py
Slimane33: Where should it be tested ?

tutorials/applications/Spin chain of 3 atoms in XY mode.ipynb
Slimane33: Yes we use 'up' and 'down' but isn't it equivalent to use |0> and |1>?

pulser/tests/test_simulation.py
HGSilveri: It would be 
```suggestion
    assert np.isclose(np.linalg.norm(simple_seq.magnetic_field[0:2]), 1)
```
You can optionally define the tolerance if you need to.

pulser/tests/test_simulation.py
HGSilveri: This should be tested in `test_sequence.py`, as it pertains to behaviour of `Sequence`, not `Simumation`. In fact, it is already tested there. I was just wondering why you felt the need to do it here too

tutorials/applications/Spin chain of 3 atoms in XY mode.ipynb
HGSilveri: It is equivalent (it's just a matter of naming after all) but "up" and "down" is probably clearer. |0> and |1> can be ambiguous

Slimane33: |0> and |1> are standard in quantum computing whereas 'up' and 'down' are completely made by us

pulser/tests/test_simulation.py
Slimane33: I was asking about your previous comment above actually

tutorials/applications/Spin chain of 3 atoms in XY mode.ipynb
HGSilveri: Sure, though you're not using this for digital quantum computation and what you decide to call |0> and |1> is also made up by you. My usual issue with |0> and |1> is that the convention in the Bloch sphere is to have |0> on the north pole and |1> on the south. Notice how here you are actually defining |down> = |0> and |up> = |1>,  so "down" would be on the north pole and "up" on the south (which is completely counter-intuitive, in my opinion). 

The reason behind my suggestion here was to have an unequivocal definition, which would also distinguish these states from the ones used in the Ising mode. If you insist on using these, I won't force you to change it, we can always do it later if necessary.

pulser/tests/test_simulation.py
HGSilveri: About the previous comment, I don't see a reason for it to change, but I might be missing something. I would revert to the previous version and check if everything still runs well.

tutorials/applications/Spin chain of 3 atoms in XY mode.ipynb
HGSilveri: I can see you are not keen to change it, so I'll let the nomenclature stay like this

pulser/simulation/simulation.py
HGSilveri: As far as I can tell, this is the only thing that needs to be changed before merging.

PR #200, opened 2021-06-11
Adding method to set the magnetic field on a Sequence

HGSilveri:
- Magnetic field can be set on a Sequence if working on the XY mode
- On an empty sequence, setting the magnetic field puts it into the XY mode
- When not specified, assumes a default of (0, 0, 1) G when entering the XY mode

### DIFF
--- a/pulser/sequence.py
+++ b/pulser/sequence.py
@@ -184,6 +184,8 @@ class Sequence:
         self._variables: dict[str, Variable] = {}
         self._to_build_calls: list[_Call] = []
         self._building: bool = True
+        # Marks the sequence as empty until the first pulse is added
+        self._empty_sequence: bool = True
 
         # Initializes all parametrized Sequence related attributes
         self._reset_parametrized()
@@ -217,6 +219,22 @@ class Sequence:
                     or self._device == MockDevice)
                     and (ch.basis == 'XY' if self._in_xy else ch.basis != 'XY')
                     }
+
+    @property
+    def magnetic_field(self) -> tuple[float, float, float]:
+        """The magnetic field acting on the array of atoms.
+
+        The magnetic field vector is defined on the reference frame of the
+        atoms in the Register (with the z-axis coming outside of the plane).
+
+
+        Note:
+            Only defined in "XY Mode", the default value being (0, 0, 1) G.
+        """
+        if not self._in_xy:
+            raise AttributeError("The magnetic field is only defined when the "
+                                 "sequence is in 'XY Mode'.")
+        return self._mag_field
 
     def is_parametrized(self) -> bool:
         """States whether the sequence is parametrized.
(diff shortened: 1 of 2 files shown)


pulser/tests/test_sequence.py
```diff
@@ -81,6 +81,39 @@ def test_channel_declaration():
...
+        seq.magnetic_field
+    seq.declare_channel('ch0', 'mw_global')    # seq in XY mode
+    assert seq.magnetic_field == (0., 0., 1.)    # mag field is the default
+    seq.set_magnetic_field(bx=1., by=-1., bz=0.5)
+    assert seq.magnetic_field == (1., -1., 0.5)
+    assert seq._empty_sequence
+    seq.add(Pulse.ConstantPulse(100, 1, 1, 0), 'ch0')
+    assert not seq._empty_sequence
+    with pytest.raises(ValueError, match="can only be set on an empty seq"):
+        seq.set_magnetic_field(0., 0., 0.)
```
sebgrijalva, line 97: Null vectors should raise an error, otherwise there would be no splitting of the Rydberg levels.

sebgrijalva (changes requested):
Looks good, but I just noticed that null vectors are an issue (see below)

pulser/sequence.py
```diff
@@ -256,6 +275,36 @@ def current_phase_ref(self, qubit: QubitId,
 
         return self._phase_ref[basis][qubit].last_phase
 
+    def set_magnetic_field(self, bx: float = 0., by: float = 0.,
```
sebgrijalva, line 278: I think now that since this is given in Gauss, perhaps 1 is too little, this is my bad. Published work shows values from 10 to 50 G for the magnetic field. I suggest we set 30 G for the moment.

sebgrijalva (changes requested):

pulser/tests/test_sequence.py
HGSilveri: As long as the there is a non-zero component, it's fine?

pulser/sequence.py
```diff
@@ -218,6 +221,22 @@ def available_channels(self) -> dict[str, Channel]:
...
+    @property
+    def magnetic_field(self) -> np.ndarray:
+        """The magnetic field acting on the array of atoms.
+
+        The magnetic field vector is defined on the reference frame of the
+        atoms in the Register (with the z-axis coming outside of the plane).
+
+
+        Note:
+            Only defined in "XY Mode", the default value being (0, 0, 1) G.
```
sebgrijalva, line 233:
```suggestion
            Only defined in "XY Mode", the default value being (0, 0, 30) G.
```

sebgrijalva (approved):
Typo in the docstring, but otherwise LGTM!

pulser/sequence.py
HGSilveri: Thanks for spotting it, I knew writing it in the docstring would backfire, I just didn't expect it to be this early haha

Issue #204, opened 2021-06-14
[Bug] 'Pulse too short' on some sequences of pulses


Slimane33:
The following code 
```python
def f(x): 
    reg = Register.square(3, 6)

    seq = Sequence(reg, Chadoq2)
    seq.declare_channel('ch0', 'rydberg_global')
    seq.declare_channel('ch1', 'rydberg_local')


    for i in range(2):
        pulse = Pulse.ConstantPulse(x * 4, 10, 0, 0)
        seq.add(pulse, 'ch0')
        for j in range(9):
            seq.target(j, 'ch1')
            pulse = Pulse.ConstantPulse(52, 0, (j+1)*10, 0)
            seq.add(pulse, 'ch1')

    seq.measure(basis='ground-rydberg')

    sim = Simulation(seq, sampling_rate=1)
    results = sim.run(nsteps=1000)
```

raises for `f(43), f(44), f(45)` 

```python
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-73-28c03cf16d82> in <module>
----> 1 f(45)

<ipython-input-72-15951c01fd5c> in f(x)
     13             seq.target(j, 'ch1')
     14             pulse = Pulse.ConstantPulse(52, 0, (j+1)*10, 0)
---> 15             seq.add(pulse, 'ch1')
     16 
     17     seq.measure(basis='ground-rydberg')

~/Desktop/unitaryhack/Pulser/pulser/sequence.py in wrapper(self, *args, **kwargs)
    109             verify_variable(x)
    110         storage = self._calls if self._building else self._to_build_calls
--> 111         func(self, *args, **kwargs)
    112         storage.append(_Call(func.__name__, args, kwargs))
    113     return wrapper

~/Desktop/unitaryhack/Pulser/pulser/sequence.py in add(self, pulse, channel, protocol)
    512         tf = ti + pulse.duration
    513         if ti > t0:
--> 514             self._delay(ti-t0, channel)
    515 
    516         prs = {self._phase_ref[basis][q].last_phase for q in last.targets}

~/Desktop/unitaryhack/Pulser/pulser/sequence.py in _delay(self, duration, channel)
    838         last = self._last(channel)
    839         ti = last.tf
--> 840         tf = ti + self._channels[channel].validate_duration(duration)
    841         self._add_to_schedule(channel,
    842                               _TimeSlot('delay', ti, tf, last.targets))

[...]

PR #205, opened 2021-06-15
Enforce minimal duration for delay (#204)

LaurentAjdnik:
Enforces a minimal duration for delays, equal to the minimal duration allowed by the channel (see #204).

Adds a corresponding test for 100% coverage.

### DIFF
--- a/pulser/sequence.py
+++ b/pulser/sequence.py
@@ -452,9 +452,19 @@ class Sequence:
                         current_max_t = op.tf
                         break
         ti = current_max_t
-        tf = ti + pulse.duration
         if ti > t0:
-            self._delay(ti-t0, channel)
+            # Insert a delay
+            delay_duration = ti - t0
+
+            # Delay must not be shorter than the min duration for this channel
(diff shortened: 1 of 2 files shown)


pulser/sequence.py
```diff
@@ -774,6 +774,11 @@ def _delay(self, duration: int, channel: str) -> None:
         if self.is_parametrized():
             return
 
+        # Delay must not be shorter than the minimal duration for this channel
+        min_duration = self._channels[channel].min_duration
+        if duration < min_duration:
+            duration = min_duration
+
```
HGSilveri, line 781: I would rather this didn't happen when the user calls `delay`, as I don't want to change an explicit input. Could you do this before `_delay` is called within `add` instead (outside of `_delay`, on line 514)?

HGSilveri (changes requested):
The solution looks good, it's just a matter of changing when it is enforced

HGSilveri (approved):
Perfect, thanks a lot!

Issue #203, opened 2021-06-14
Add __getitem__ to Waveform


HGSilveri:
The idea would be to have an easy access to the samples of a Waveform, preferentially without computing the full array of samples. Currently, the properties `first_value` and `last_value` do this for the first and last values, but a `__getitem__` would replace them and extend the functionality to the entire waveform. 

The keys would be the time in `ns`, also allowing specification from the end (e.g. `t=-1`) and slicing.

LaurentAjdnik:
You can assign me.

HGSilveri:
@LaurentAjdnik Great! There's a bit of nuance here, but I'm sure you're up to the task. A few pointers:
- We want just `__getitem__`, not `__setitem__`. By design, a waveform should not be changed in-place after initialization
- A big part of the motivation to do this is to save in computation time and/or memory by avoiding computing `samples` or, if it makes sense to compute it, do it just once. If we didn't care about this, we would just get items from `Waveform.samples`. You will see in the `first_value` and `last_value` examples on how this can be done. In the cases where it makes sense to compute the `samples`, it's worth considering some sort of caching mechanism.

Thanks again, and feel free to reach out with any questions!

LaurentAjdnik:
Lots of questions raised indeed. I will draft-PR something soon and we can discuss based on that.

PR #206, opened 2021-06-16
Add __getitem__ to waveforms (#203)

LaurentAjdnik:
Adds `__getitem__` to waveform.

Adds corresponding tests.

First PR : ConstantWaveform only, for review.

### DIFF
--- a/pulser/sequence.py
+++ b/pulser/sequence.py
@@ -39,16 +39,16 @@ from pulser.pulse import Pulse
 from pulser.register import Register
 from pulser._seq_drawer import draw_sequence
 
-if version_info[:2] == (3, 7):  # pragma: no cover
+if version_info[:2] >= (3, 8):  # pragma: no cover
+    from typing import Literal, get_args
+else:  # pragma: no cover
     try:
-        from typing_extensions import Literal, get_args
+        from typing_extensions import Literal, get_args  # type: ignore
     except ImportError:
         raise ImportError(
             "Using pulser with Python version 3.7 requires the"
             " `typing_extensions` module. Install it by running"
             " `pip install typing-extensions`.")
-else:  # pragma: no cover
-    from typing import Literal, get_args  # type: ignore
 
 
 QubitId = Union[int, str]
--- a/pulser/waveforms.py
+++ b/pulser/waveforms.py
@@ -17,6 +17,7 @@ from __future__ import annotations
 
 from abc import ABC, abstractmethod
 import functools
+from functools import cached_property
 import inspect
 import itertools
 import sys
@@ -73,15 +74,19 @@ class Waveform(ABC):
         """The duration of the pulse (in ns)."""
         pass
 
-    @property
+    @cached_property
     @abstractmethod
+    def _samples(self) -> np.ndarray:
+        pass
+
+    @property
     def samples(self) -> np.ndarray:
         """The value at each time step that describes the waveform.
 
         Returns:
             np.ndarray: A numpy array with a value for each time step.
         """
-        pass
+        return self._samples.copy()
 
     @property
     def first_value(self) -> float:
@@ -126,10 +131,14 @@ class Waveform(ABC):
     def __repr__(self) -> str:
         pass
 
-    @abstractmethod
     def __getitem__(self, index_or_slice:
                     Union[int, slice]) -> Union[float, np.ndarray]:
-        pass
+        if isinstance(index_or_slice, slice):
+            s: slice = self._check_slice(index_or_slice)
+            return cast(np.ndarray, self._samples[s])
+        else:
+            index: int = self._check_index(index_or_slice)
+            return cast(float, self._samples[index])
 
     def _check_index(self, i: int) -> int:
         if (i < -self.duration
@@ -230,8 +239,8 @@ class CompositeWaveform(Waveform):
             duration += wf.duration
         return duration
 
-    @property
-    def samples(self) -> np.ndarray:
+    @cached_property
+    def _samples(self) -> np.ndarray:
         """The value at each time step that describes the waveform.
 
         Returns:
@@ -262,56 +271,6 @@ class CompositeWaveform(Waveform):
     def __repr__(self) -> str:
         return f'CompositeWaveform({self.duration} ns, {self._waveforms!r})'
 
-    def __getitem__(self, index_or_slice:
-                    Union[int, slice]) -> Union[float, np.ndarray]:
-        wf_start: int
-        if isinstance(index_or_slice, slice):
-            s: slice = self._check_slice(index_or_slice)
-            start: int = s.start
-            stop: int = s.stop
-            array: np.ndarray = np.array([])
-            wf_start = 0
-            for wf in self._waveforms:
-                duration: int = wf.duration
-                if start > wf_start + duration:
-                    # Skip this waveform, which is before the start index
-                    wf_start += duration
-                    continue
-                elif wf_start <= start:
-                    # The start index belongs to this waveform
-                    if wf_start <= stop <= wf_start+duration:
-                        # The stop index also belongs to this waveform
-                        # Hence, we can return a subset of its samples
-                        return cast(np.ndarray,
-                                    wf[start-wf_start:stop-wf_start])
-                    else:
-                        # Copy the part of the waveform after the start index
-                        array = wf.samples[start-wf_start:]
-                else:
-                    if stop < wf_start:
-                        # No need to loop on the next waveforms
-                        break
-                    elif stop <= wf_start+wf.duration:
-                        # The stop index belongs to this waveform
-                        # Hence, we can concatenate a subset of it
-                        array = np.concatenate(
-                            (array, wf.samples[:stop-wf_start]))
-                    else:
-                        # This waveform fits entirely in the start~stop range
-                        array = np.concatenate((array, wf.samples))
-                wf_start += duration
-            return array
-        else:
-            index: int = self._check_index(index_or_slice)
-            wf_start = 0
-            value: float
-            for wf in self._waveforms:
-                if wf_start <= index < wf_start+wf.duration:
-                    value = cast(float, wf[index-wf_start])
-                    break
-                wf_start += wf.duration
-            return value
-
     def __mul__(self, other: float) -> CompositeWaveform:
         return CompositeWaveform(*(wf * other for wf in self._waveforms))
 
@@ -335,14 +294,15 @@ class CustomWaveform(Waveform):
         """The duration of the pulse (in ns)."""
         return self._duration
 
-    @property
-    def samples(self) -> np.ndarray:
+    @cached_property
+    def _samples(self) -> np.ndarray:
         """The value at each time step that describes the waveform.
 
         Returns:
             numpy.ndarray: A numpy array with a value for each time step.
         """
-        return self._samples.copy()
+        # self._samples is already cached when initialized in __init__
+        pass
 
     def _to_dict(self) -> dict[str, Any]:
         return obj_to_dict(self, self._samples)
@@ -352,15 +312,6 @@ class CustomWaveform(Waveform):
 
     def __repr__(self) -> str:
         return f'CustomWaveform({self.duration} ns, {self.samples!r})'
-
-    def __getitem__(self, index_or_slice:
-                    Union[int, slice]) -> Union[float, np.ndarray]:
-        if isinstance(index_or_slice, slice):
-            self._check_slice(index_or_slice)
-            return cast(np.ndarray, self._samples[index_or_slice])
-        else:
-            self._check_index(index_or_slice)
-            return cast(float, self._samples[index_or_slice])
 
     def __mul__(self, other: float) -> CustomWaveform:
         return CustomWaveform(self._samples * float(other))
@@ -386,8 +337,8 @@ class ConstantWaveform(Waveform):
         """The duration of the pulse (in ns)."""
         return self._duration
 
-    @property
-    def samples(self) -> np.ndarray:
+    @cached_property
+    def _samples(self) -> np.ndarray:
         """The value at each time step that describes the waveform.
 
         Returns:
@@ -419,9 +370,7 @@ class ConstantWaveform(Waveform):
     def __getitem__(self, index_or_slice:
                     Union[int, slice]) -> Union[float, np.ndarray]:
         if isinstance(index_or_slice, slice):
-            index_or_slice = self._check_slice(index_or_slice)
-            duration = index_or_slice.stop - index_or_slice.start
-            return np.full(duration, self._value)
+            return super().__getitem__(index_or_slice)
         else:
             self._check_index(index_or_slice)
             return self._value
@@ -448,22 +397,20 @@ class RampWaveform(Waveform):
         self._start: float = float(start)
         stop = cast(float, stop)
         self._stop: float = float(stop)
-        self._samples: np.ndarray = np.linspace(
-            self._start, self._stop, num=self._duration)
 
     @property
     def duration(self) -> int:
         """The duration of the pulse (in ns)."""
         return self._duration
 
-    @property
-    def samples(self) -> np.ndarray:
+    @cached_property
+    def _samples(self) -> np.ndarray:
         """The value at each time step that describes the waveform.
 
         Returns:
             numpy.ndarray: A numpy array with a value for each time step.
         """
-        return self._samples.copy()
+        return np.linspace(self._start, self._stop, num=self._duration)
 
     @property
     def slope(self) -> float:
@@ -494,15 +441,14 @@ class RampWaveform(Waveform):
     def __getitem__(self, index_or_slice:
                     Union[int, slice]) -> Union[float, np.ndarray]:
         if isinstance(index_or_slice, slice):
-            self._check_slice(index_or_slice)
-            return cast(np.ndarray, self._samples[index_or_slice])
+            return super().__getitem__(index_or_slice)
         else:
             index: int = self._check_index(index_or_slice)
             if index == 0:
                 return self._start
             if index == self._duration - 1:
                 return self._stop
-            return cast(float, self._samples[index])
+            return super().__getitem__(index)
 
     def __mul__(self, other: float) -> RampWaveform:
         k = float(other)
@@ -570,8 +516,8 @@ class BlackmanWaveform(Waveform):
         """The duration of the pulse (in ns)."""
         return self._duration
 
-    @property
-    def samples(self) -> np.ndarray:
+    @cached_property
+    def _samples(self) -> np.ndarray:
         """The value at each time step that describes the waveform.
 
         Returns:
@@ -603,13 +549,12 @@ class BlackmanWaveform(Waveform):
     def __getitem__(self, index_or_slice:
                     Union[int, slice]) -> Union[float, np.ndarray]:
         if isinstance(index_or_slice, slice):
-            self._check_slice(index_or_slice)
-            return cast(np.ndarray, self.samples[index_or_slice])
+            return super().__getitem__(index_or_slice)
         else:
             index: int = self._check_index(index_or_slice)
             if index == 0 or index == self._duration - 1:
                 return 0.
-            return cast(float, self.samples[index])
+            return super().__getitem__(index)
 
     def __mul__(self, other: float) -> BlackmanWaveform:
         return BlackmanWaveform(self._duration, self._area * float(other))
--- a/pulser/waveforms.py
+++ b/pulser/waveforms.py
@@ -34,7 +34,7 @@ from pulser.parametrized import Parametrized, ParamObj
 from pulser.parametrized.decorators import parametrize
 from pulser.json.utils import obj_to_dict
 
-if version_info[:2] >= (3, 7):  # pragma: no cover
+if version_info[:2] >= (3, 8):  # pragma: no cover
     from functools import cached_property
 else:  # pragma: no cover
     try:
--- a/pulser/waveforms.py
+++ b/pulser/waveforms.py
@@ -34,16 +34,16 @@ from pulser.parametrized import Parametrized, ParamObj
 from pulser.parametrized.decorators import parametrize
 from pulser.json.utils import obj_to_dict
 
-if version_info[:2] == (3, 7):  # pragma: no cover
+if version_info[:2] >= (3, 7):  # pragma: no cover
+    from functools import cached_property
+else:  # pragma: no cover
     try:
-        from backports.cached_property import cached_property
+        from backports.cached_property import cached_property  # type: ignore
     except ImportError:
         raise ImportError(
             "Using pulser with Python version 3.7 requires the"
             " `backports.cached-property` module. Install it by running"
             " `pip install backports.cached-property`.")
-else:  # pragma: no cover
-    from functools import cached_property  # type: ignore
 
 
 class Waveform(ABC):
--- a/pulser/waveforms.py
+++ b/pulser/waveforms.py
@@ -297,12 +297,12 @@ class CompositeWaveform(Waveform):
                     else:
                         # Copy the part of the waveform after the start index
                         array = wf.samples[start-wf_start:]
-                elif start < wf_start:
+                else:
                     if stop < wf_start:
                         # No need to loop on the next waveforms
                         break
                     elif stop <= wf_start+wf.duration:
-                        # This stop index belongs to this waveform
+                        # The stop index belongs to this waveform
                         # Hence, we can concatenate a subset of it
                         array = np.concatenate(
                             (array, wf.samples[:stop-wf_start]))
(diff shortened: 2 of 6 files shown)


LaurentAjdnik:
@HGSilveri: All done, all tested. I will update according to your feedback.

pulser/waveforms.py
```diff
@@ -233,6 +269,39 @@ def __str__(self) -> str:
...
+                if wf_start <= start and start <= wf_start+wf.duration:
+                    if wf_start <= stop and stop <= wf_start+wf.duration:
+                        return cast(np.ndarray,
+                                    wf.samples[start-wf_start:stop-wf_start])
+                    else:
+                        array = wf.samples[start-wf_start:]
+                if start < wf_start and stop > wf_start+wf.duration:
+                    array = np.concatenate((array, wf.samples))
+                if wf_start <= stop and stop <= wf_start+wf.duration:
+                    array = np.concatenate((array, wf.samples[:stop-wf_start]))
```
HGSilveri, line 291: I think it would be clearer to have the checks on `start` on the first stage and on `stop` on a second, like:
```suggestion
                elif start < wf_start:
                    if stop < wf_start:
                        break
                    elif stop <= wf_start+wf.duration:
                        array = np.concatenate((array, wf.samples[:stop-wf_start]))
                    else:
                        array = np.concatenate((array, wf.samples))
```
I think this even makes the if-clause above redundant.

pulser/waveforms.py
```diff
@@ -233,6 +269,39 @@ def __str__(self) -> str:
...
+        wf_start: int
+        if isinstance(index_or_slice, slice):
+            s: slice = self._check_slice(index_or_slice)
+            start: int = s.start
+            stop: int = s.stop
+            array: np.ndarray = np.array([])
+            wf_start = 0
+            for wf in self._waveforms:
+                if wf_start <= start and start <= wf_start+wf.duration:
+                    if wf_start <= stop and stop <= wf_start+wf.duration:
```
HGSilveri, line 283: You can write this as: 
```suggestion
                    if wf_start <= stop <= wf_start+wf.duration:
```
I feel like it's easier to read, don't you?

pulser/waveforms.py
```diff
@@ -233,6 +269,39 @@ def __str__(self) -> str:
...
+                if wf_start <= stop and stop <= wf_start+wf.duration:
+                    array = np.concatenate((array, wf.samples[:stop-wf_start]))
+                wf_start += wf.duration
+            return array
+        else:
+            index: int = self._check_index(index_or_slice)
+            wf_start = 0
+            value: float
+            for wf in self._waveforms:
+                if wf_start <= index and index < wf_start+wf.duration:
```
HGSilveri, line 299: Same suggestion here:
```suggestion
                if wf_start <= index < wf_start+wf.duration:
```

pulser/waveforms.py
```diff
@@ -373,7 +463,7 @@ def samples(self) -> np.ndarray:
         Returns:
             numpy.ndarray: A numpy array with a value for each time step.
         """
-        return np.linspace(self._start, self._stop, num=self._duration)
+        return self._samples
```
HGSilveri, line 466: The problem with doing this is (and the reason it wasn't done) that it give direct access to the samples, so the user could change it in-place:
```
samps = wf.samples
samps = np.array([1, 2 , 3])
```
After this block, `wf.samples` will be `np.array([1, 2, 3])`
I can't figure out a way out of it without doing the array from scratch or returning a `copy`. A compromise is to have `self.samples` return `self._samples.copy()` but then use `self._samples` internally.

pulser/waveforms.py
```diff
@@ -517,6 +616,15 @@ def __str__(self) -> str:
...
         return f"BlackmanWaveform({self._duration} ns, Area: {self._area:.3g})"
 
+    def __getitem__(self, index_or_slice:
+                    Union[int, slice]) -> Union[float, np.ndarray]:
+        if isinstance(index_or_slice, slice):
+            self._check_slice(index_or_slice)
+            return cast(np.ndarray, self.samples[index_or_slice])
+        else:
+            self._check_index(index_or_slice)
+            return cast(float, self.samples[index_or_slice])
```
HGSilveri, line 626: This is minutia, but if the `index` is `0` or `duration-1` you can return `0` instead.

HGSilveri (commented):
Alright, it's looking good! I just have a few questions and suggestions

pulser/waveforms.py
```diff
@@ -126,6 +126,42 @@ def __str__(self) -> str:
...
+                or i >= self.duration):
+            raise IndexError("Index ('index_or_slice' = "
+                             f"{i}) must be in the range "
+                             f"0~{self.duration-1}, or "
+                             f"{-self.duration}~-1 from the end.")
+        return i if i >= 0 else self.duration + i
+
+    def _check_slice(self, s: slice) -> slice:
+        if s.step is not None and s.step != 1:
+            raise IndexError("The step of the slice must be None or 1.")
```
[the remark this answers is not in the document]
LaurentAjdnik: Handling steps bigger than 1 would make things more complex in some cases (and MUCH more complex for composite waveforms in particular).

Considering it probably doesn't make much sense functionnaly, I prefered making it unallowed.

pulser/waveforms.py
```diff
@@ -126,6 +126,42 @@ def __str__(self) -> str:
...
+            s.stop if s.stop >= 0 else self.duration + s.stop)
+
+        if start >= stop:
+            raise IndexError(f"The start of the slice ({start}) "
+                             f"must be less than the stop ({stop}).")
+
+        if (start < 0 or stop < 0 or start > self.duration - 1
+                or stop > self.duration):
+            raise IndexError(f"The range of the slice ({start}~{stop}) "
+                             "must be included in the range of the waveform.")
```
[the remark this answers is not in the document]
LaurentAjdnik: > This is usually not enforced with slices. You might get an empty array or the full one, but it still works. I would say that we shouldn't deviate from the expected behaviour if not necessary, but I'm interested in hearing your motivation behind these choices.

I just wasn't aware of this default behavior in Python. I will stick to it. When completely out of bounds, I guess we would return an empty array?

pulser/waveforms.py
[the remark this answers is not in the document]
HGSilveri: Alright, that's good enough for me.

pulser/waveforms.py
```diff
@@ -233,6 +269,39 @@ def __str__(self) -> str:
...
+            s: slice = self._check_slice(index_or_slice)
+            start: int = s.start
+            stop: int = s.stop
+            array: np.ndarray = np.array([])
+            wf_start = 0
+            for wf in self._waveforms:
+                if wf_start <= start and start <= wf_start+wf.duration:
+                    if wf_start <= stop and stop <= wf_start+wf.duration:
+                        return cast(np.ndarray,
+                                    wf.samples[start-wf_start:stop-wf_start])
```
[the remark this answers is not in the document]
LaurentAjdnik: > Shouldn't you take advantage of the `__getitem__` of this waveform?

Indeed! Writing code at 11pm isn't always a good idea... :grin:

pulser/waveforms.py
LaurentAjdnik: > You can write this as:
> 
> I feel like it's easier to read, don't you?

Indeed. I didn't know this syntax.

pulser/waveforms.py
LaurentAjdnik: > Same suggestion here:

👍

pulser/waveforms.py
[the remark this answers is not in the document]
HGSilveri: Yup, that's it. And if you go over the maximum length (ie `stop > length`) it will give you the same array as if no `stop` was given.

pulser/waveforms.py
LaurentAjdnik: > This is minutia, but if the `index` is `0` or `duration-1` you can return `0` instead.

👍

pulser/waveforms.py
LaurentAjdnik: > The problem with doing this is (and the reason it wasn't done) that it give direct access to the samples, so the user could change it in-place:
> 
> ```
> samps = wf.samples
> samps = np.array([1, 2 , 3])
> ```
> 
> After this block, `wf.samples` will be `np.array([1, 2, 3])`
> I can't figure out a way out of it without doing the array from scratch or returning a `copy`. A compromise is to have `self.samples` return `self._samples.copy()` but then use `self._samples` internally.

Just to make sure: that applies to CustomWaveform and Rampwaveform that both return `self._samples`?

HGSilveri: You're right, `CustomWaveform` is returning `self._samples` although perhaps it shouldn't be. But `RampWaveform` is currently not returning `self._samples`, it's this change that would make it do so.

pulser/waveforms.py
LaurentAjdnik: Note sure, here...

I built my code around three cases, that happen sequentially:
* The start index belongs to the current waveform
* The current waveform is completely included in the start-stop range
* The end index belongs to the current waveform

I'll add comments to make this appear more clearly.

HGSilveri: What I'm suggesting here is to take advantage of the previous check(s) on `start`, such that at this point (if using an if-elif construction) you know the start index can only belong to a previous waveform. As such, you only need to check:
1. If the stop index also belongs to a previous waveform, you just quit the loop.
2. If you got here, you know `stop > wf_start`, so you check if `stop < wf_start+wf.duration` (ie is within the current waveform) (and you can `break` here too btw)
3. If you got here, none of the above are true, so you know `stop > wf_start+wf.duration`

Note how the check 3. covers the case you are isolating above (`start < wf_start and stop > wf_start+wf.duration`)

pulser/waveforms.py
```diff
@@ -233,6 +269,39 @@ def __str__(self) -> str:
...
+                    Union[int, slice]) -> Union[float, np.ndarray]:
+        wf_start: int
+        if isinstance(index_or_slice, slice):
+            s: slice = self._check_slice(index_or_slice)
+            start: int = s.start
+            stop: int = s.stop
+            array: np.ndarray = np.array([])
+            wf_start = 0
+            for wf in self._waveforms:
+                if wf_start <= start and start <= wf_start+wf.duration:
```
HGSilveri, line 282: Btw, I think you should start this block by checking if the start index is on a subsequent waveform, that way you can just skip to the next iteration of the loop:
```suggestion
                if start > wf_start + duration:
                    wf_start += duration
                    continue
                elif wf_start <= start:
```
Note how you can skip checking `start <= wf_start+wf.duration` too with the `elif`

pulser/waveforms.py
```diff
@@ -279,24 +282,41 @@ def __getitem__(self, index_or_slice:
...
+                                    wf[start-wf_start:stop-wf_start])
                     else:
+                        # Copy the part of the waveform after the start index
                         array = wf.samples[start-wf_start:]
-                if start < wf_start and stop > wf_start+wf.duration:
-                    array = np.concatenate((array, wf.samples))
-                if wf_start <= stop and stop <= wf_start+wf.duration:
-                    array = np.concatenate((array, wf.samples[:stop-wf_start]))
-                wf_start += wf.duration
+                elif start < wf_start:
```
HGSilveri, line 300: You can omit the comparison, it's the only option at this point

pulser/waveforms.py
```diff
@@ -463,7 +483,7 @@ def samples(self) -> np.ndarray:
         Returns:
             numpy.ndarray: A numpy array with a value for each time step.
         """
-        return self._samples
+        return self._samples.copy()
```
HGSilveri, line 486: I'm thinking it might make sense to have all Waveforms defining a `self._samples` property instead of `self.samples`, and then have `self.samples` defined to return a copy of `self._samples` on the parent class. This way we can call `wf._samples` instead of `wf.samples` in `__getitem__` and avoid copying repeatedly. What you think?

pulser/waveforms.py
```diff
@@ -622,8 +642,10 @@ def __getitem__(self, index_or_slice:
             self._check_slice(index_or_slice)
             return cast(np.ndarray, self.samples[index_or_slice])
         else:
-            self._check_index(index_or_slice)
-            return cast(float, self.samples[index_or_slice])
+            index: int = self._check_index(index_or_slice)
+            if index == 0 or index == self._duration - 1:
+                return 0.
+            return cast(float, self.samples[index])
```
HGSilveri, line 648: You can do something similar on `RampWaveform` to return `start` and `stop`.

HGSilveri (commented):
Alright, looking good

pulser/waveforms.py
LaurentAjdnik: :+1:

pulser/waveforms.py
LaurentAjdnik: 👍

pulser/waveforms.py
LaurentAjdnik: Working on it.

LaurentAjdnik:
BTW, shouldn't we replace this...

```
    @property
    def first_value(self) -> float:
        """The first value in the waveform."""
        return float(self.samples[0])

    @property
    def last_value(self) -> float:
        """The last value in the waveform."""
        return float(self.samples[-1])
```

...by that in the `Waveform` parent class?

```
    @property
    def first_value(self) -> float:
        """The first value in the waveform."""
        return float(self[0])

    @property
    def last_value(self) -> float:
        """The last value in the waveform."""
        return float(self[-1])
```

pulser/waveforms.py
LaurentAjdnik: I have doubts about the consequences and benefits. I must be missing something.

Currently, only `CustomWaveform` (before my commits) and `RampWaveform` (since my commits) are storing their samples, hence requiring a `copy()` before returning them.

Others (`BlackmanWaveform`, `CompositeWaveform` and `ConstantWaveform`) return arrays of samples computed on the fly, without a need for copy. Would you still add a `copy()` here?

HGSilveri:
> BTW, shouldn't we replace this...
> 
> ```
>     @property
>     def first_value(self) -> float:
>         """The first value in the waveform."""
>         return float(self.samples[0])
> 
>     @property
>     def last_value(self) -> float:
>         """The last value in the waveform."""
>         return float(self.samples[-1])
> ```
> 
> ...by that in the `Waveform` parent class?
> 
> ```
>     @property
>     def first_value(self) -> float:
>         """The first value in the waveform."""
>         return float(self[0])
> 
>     @property
>     def last_value(self) -> float:
>         """The last value in the waveform."""
>         return float(self[-1])
> ```

Yup, and also remove the `first_value` and `last_value` definitions on all child classes.

pulser/waveforms.py
HGSilveri: My idea would be for the ones that are computed on the fly, to only be once. Might be a good idea to use a [cached_property](https://docs.python.org/3/library/functools.html#functools.cached_property) on `_samples`.

LaurentAjdnik: I wonder if we're taking the right direction here.

I assume that, in the initial code, not saving samples was mostly a matter of saving memory.

I also assume that a call to the property `samples` will ALWAYS occur, sooner or later. Otherwise, why would we create a waveform anyway? :grin:

Hence, if we were to cache samples, we would ALWAYS end up with no memory gain.

Maybe we should just store samples in `__init__` most of the time (`ConstantWaveform` and `CompositeWaveform` possibly being exceptions). We would gain a lot in terms of code readability and speed of execution.

But I might be completely wrong along the way. Maybe I'm not taking parametrization into account.

HGSilveri: My concern is not really the memory gain, but rather the time wasted on multiple calls to `samples`. It's true that it always gets called once, in fact it gets called much more than once on average. The caching would simply be a way of only wasting time computing it once. That being said, it's very much the same thing as storing self._samples in __init__. The only difference is that we can mark `self._samples` with `@abstractmethod` on the parent class, so it enforces that all child classes have `self._samples` defined.

To be clear, `@cached_property` will run the first time it's called and store the result. All ensuing calls will essentially lookup the stored value. I feel like it is in line with what you're suggesting, with some added benefits.

LaurentAjdnik: OK, I'll be adding `@cached_property` then.

BTW, since it always gets called once (at least), maybe we can improve/simplify some of the code I wrote for `__getitem__`.

For instance, inside `CompositeWaveform`: what about computing/caching samples once and for all, and then simply index them, rather than using the loops we've been working on (so hard)?

HGSilveri: Yes, I think that's sensible. I'm sorry that it will make you get rid of the loops, though...

LaurentAjdnik: We'll miss the loops but we'll forever remember and cherish the nice work we achieved on them. 😁

If we are to create `samples` in the parent class as `return self._samples.copy()`:
* That would also apply to `ConstantWaveform`, in which case we should cache the constant samples instead of always returning a `np.full()` built on the fly?
* All `__getitem__` could be simplified as mere `return self._samples[index_or_slice]` in the parent class?
* Except maybe with a few optimizations, like:
  * `ConstantWaveform`: Simply returning the constant value when `index_or_slice` is an `int` instead of indexing `self._samples`?
  * `BlackmanWaveform` and `RampWaveform`: Simply returning the corresponding stored values when we index the first or last sample?

HGSilveri: Yes to all!

LaurentAjdnik:
Why "cancelled"? 😢

LaurentAjdnik:
Well, I was expecting the fail on v3.7 since `@cached_property` was introduced in v3.8...

Maybe we can find a workaround, just like we did with `get_args` in `sequence.py`...

Perhaps by using https://pypi.org/project/cached-property/ or https://pypi.org/project/backports.cached-property/?

HGSilveri:
> Well, I was expecting the fail on v3.7 since `@cached_property` was introduced in v3.8...
> 
> Maybe we can find a workaround, just like we did with `get_args` in `sequence.py`...
> 
> Perhaps by using https://pypi.org/project/cached-property/ or https://pypi.org/project/backports.cached-property/?

Yes, that's the way to go. I'm adding the `typing_extensions` to the dependencies for Python 3.7 in `setup.py` and `requirements.txt`. Take not of how that's done and do the same for `cached-property`.

HGSilveri:
@LaurentAjdnik Take note of PR #209

pulser/waveforms.py
```diff
@@ -17,6 +17,7 @@
 
 from abc import ABC, abstractmethod
 import functools
+from functools import cached_property
```
HGSilveri, line 20: Let's make this import conditional on the version, like was done with `typing-extensions`. If in 3.7,   do `from cached_property import cached_property` instead.

pulser/waveforms.py
```diff
@@ -524,11 +441,14 @@ def __repr__(self) -> str:
...
+            return super().__getitem__(index_or_slice)
         else:
-            self._check_index(index_or_slice)
-            return cast(float, self._samples[index_or_slice])
+            index: int = self._check_index(index_or_slice)
+            if index == 0:
+                return self._start
+            if index == self._duration - 1:
+                return self._stop
+            return super().__getitem__(index)
```
HGSilveri, line 451: I know I agreed to leaving these cases in, but could please you check if there's actually any significant performance gain vs. just getting the value from `self._samples`?

HGSilveri (approved):
This looks really neat, nice job!

pulser/waveforms.py
HGSilveri: On second-thought, use `backports.cached_property` instead.

LaurentAjdnik: :+1:

LaurentAjdnik:
@HGSilveri: What should I do with this fail? Is this something I can handle personally or do you have to change something on your side ?

```
Run mypy
  mypy
  shell: /usr/bin/bash -e {0}
  env:
    pythonLocation: /opt/hostedtoolcache/Python/3.8.10/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.8.10/x64/lib
pulser/waveforms.py:39: error: Library stubs not installed for "backports.cached_property" (or incompatible with Python 3.8)
pulser/waveforms.py:39: note: Hint: "python3 -m pip install types-backports"
pulser/waveforms.py:39: note: (or run "mypy --install-types" to install all missing stub packages)
pulser/waveforms.py:39: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
pulser/waveforms.py:46: error: unused "type: ignore" comment
pulser/waveforms.py:100: error: Returning Any from function declared to return "ndarray"
Found 3 errors in 1 file (checked 37 source files)
Error: Process completed with exit code 1.
```

pulser/waveforms.py
LaurentAjdnik: Checking indexes first is indeed faster, but not so much (less than twice):

```
Benchmarking access to the first value (1 million times):
* Direct access    : 0.09557080000000018
* Check index first: 0.05514600000000014

Benchmarking access to the last value (1 million times):
* Direct access    : 0.09351679999999996
* Check index first: 0.06454629999999995
```

But the absolute difference in duration is extremely tiny (the difference between measurements above, in seconds, divided by 1 million...).

I think we should remove the checks, considering:
* This does not really affect performance
* It makes code more verbose
* The calls on the first and alst value probably don't occur really often

pulser/waveforms.py
```diff
@@ -33,6 +34,17 @@
...
+if version_info[:2] == (3, 7):  # pragma: no cover
+    try:
+        from backports.cached_property import cached_property
+    except ImportError:
+        raise ImportError(
+            "Using pulser with Python version 3.7 requires the"
+            " `backports.cached-property` module. Install it by running"
+            " `pip install backports.cached-property`.")
+else:  # pragma: no cover
+    from functools import cached_property  # type: ignore
```
HGSilveri, line 46: `mypy` CI tests are made with python 3.8, so you don't need the `# type: ignore`, but rather need it on line 39.

pulser/waveforms.py
```diff
@@ -73,25 +85,29 @@ def duration(self) -> int:
...
+
+    @property
     def samples(self) -> np.ndarray:
         """The value at each time step that describes the waveform.
 
         Returns:
             np.ndarray: A numpy array with a value for each time step.
         """
-        pass
+        return self._samples.copy()
```
HGSilveri, line 100: Cast this to `np.ndarray`.

HGSilveri (commented):
See if these fix the errors you are getting

pulser/waveforms.py
LaurentAjdnik: > Cast this to `np.ndarray`.

`pulser\waveforms.py:100: error: Redundant cast to "ndarray"`

It is not necessary since the return type is already included in the function def.

pulser/waveforms.py
LaurentAjdnik: > `mypy` CI tests are made with python 3.8, so you don't need the `# type: ignore`, but rather need it on line 39.

```
pulser\waveforms.py:39: error: unused 'type: ignore' comment
pulser\waveforms.py:46: error: Incompatible import of "cached_property" (imported name has type "Type[functools.cached_property[Any]]", local name has type "Type[backports.cached_property.cached_property[Any]]")
```

The `# type: ignore` seems correctly placed to me. It is exactly what I did with `sequence.py`, for the same reasons.

I think the `Library stubs not installed for "backports.cached_property" (or incompatible with Python 3.8)` error comes from elsewhere.

There is a reason why this works with `sequence.py` and not `waveforms.py` but I can't figure it out.

pulser/waveforms.py
HGSilveri: But without it you have this error: 
```
pulser/waveforms.py:100: error: Returning Any from function declared to return "ndarray"
```

How come?

pulser/waveforms.py
HGSilveri: I agree. Isn't this also the case for the other special cases, though?

pulser/waveforms.py
HGSilveri: The reason why we put the `# type: ignore` on the second import is because it complains if you import an object with the same name twice. Since we are testing with mypy on Python 3.8, you want `from functools import chached_property` to come first, and then have the `# type: ignore` on the `from backports.cached_property ...`. Can you try this please?

pulser/waveforms.py
HGSilveri: The reason is because it is importing the `cached_property` with stubs, so `cached_property` is returning `Any`. My suggestion above might fix this.

pulser/waveforms.py
LaurentAjdnik: Yes, yes, I meant removing the checks for all the special cases. Let's avoid unnecessary complexity whenever performance gains are not that important.

LaurentAjdnik:
Here we go! All checks succesful ! Yeeeeeeh hah!

pulser/sequence.py
```diff
@@ -39,16 +39,16 @@
 from pulser.register import Register
 from pulser._seq_drawer import draw_sequence
 
-if version_info[:2] == (3, 7):  # pragma: no cover
+if version_info[:2] >= (3, 8):  # pragma: no cover
+    from typing import Literal, get_args
+else:  # pragma: no cover
     try:
-        from typing_extensions import Literal, get_args
+        from typing_extensions import Literal, get_args  # type: ignore
```
HGSilveri, line 46: If you're changing this here, can you change it in `simconfig.py` too please?

HGSilveri (approved):
Great!

pulser/sequence.py
LaurentAjdnik: I wasn't aware of that one. Done.

PR #222, opened 2021-06-28
`BlackmanWaveform.from_max_val()`: Move back to previous even duration when necessary (#218)

### LINKED ISSUE #218: BlackmanWaveform.from_max_val(): Side-effect of adjusting for rounding errors
I have been wondering about this in `BlackmanWaveform.from_max_val()`:

```
        # Adjust for rounding errors to make sure max_val is not surpassed
        while np.abs(wf._scaling) > np.abs(max_val):
            duration += 1
            wf = cls(duration, area)
```

We check `_scaling` against `max_val`, based on the hypothesis that the max value of the samples is always 1.

But the `numpy.blackman()` [documentation](https://numpy.org/doc/stable/reference/generated/numpy.blackman.html) states that "**the value one appears only if the number of samples is odd**".

Hence, the current code always returns a correct waveform but, once in a while, when its duration is odd, the previous solution was better (the max value is closer to `max_val`).

I tested with an area of `2 * np.pi` for 1000 different max values (100~1100), using the existing method and then checking with a slightly shorter duration:

`75 fails out of 503 odd durations (14.91%) and 1000 total checks (7.50%).`

For instance:
```
Failed for max_val=230: 229.933 (66 ns) is  3.267 above 226.666 (67 ns)
Failed for max_val=245: 244.979 (62 ns) is  3.689 above 241.290 (63 ns)
Failed for max_val=282: 281.857 (54 ns) is  4.821 above 277.036 (55 ns)
Failed for max_val=293: 292.877 (52 ns) is  5.185 above 287.692 (53 ns)
[...]
Failed for max_val=997: 979.536 (16 ns) is 44.538 above 934.998 (17 ns)
```

Do we want to correct this behavior or should we stick to assuming 1 as the max values of samples (interpolated somehow between the real max values of even durations)?

LaurentAjdnik:
Fixes #218.

### DIFF
--- a/pulser/waveforms.py
+++ b/pulser/waveforms.py
@@ -538,13 +538,24 @@ class BlackmanWaveform(Waveform):
             raise ValueError(
                 "The maximum value and the area must have " "matching signs."
             )
+
         # A normalized Blackman waveform has an area of 0.42 * duration
         duration = np.ceil(area / (0.42 * max_val) * 1e3)  # in ns
         wf = cls(duration, area)
+
         # Adjust for rounding errors to make sure max_val is not surpassed
         while np.abs(wf._scaling) > np.abs(max_val):
             duration += 1
+            previous_wf = wf
             wf = cls(duration, area)
+
+        # According to the documentation for numpy.blackman(), "the value one
+        # appears only if the number of samples is odd". Hence, the previous
+        # even duration can reach a maximal value closer to max_val.
+        if (previous_wf is not None and duration % 2 == 1 and
+                np.max(wf.samples) < np.max(previous_wf.samples) <= max_val):
+            wf = previous_wf
+
         return wf
 
     @property
--- a/pulser/waveforms.py
+++ b/pulser/waveforms.py
@@ -542,6 +542,7 @@ class BlackmanWaveform(Waveform):
         # A normalized Blackman waveform has an area of 0.42 * duration
         duration = np.ceil(area / (0.42 * max_val) * 1e3)  # in ns
         wf = cls(duration, area)
+        previous_wf = None
 
         # Adjust for rounding errors to make sure max_val is not surpassed
         while np.abs(wf._scaling) > np.abs(max_val):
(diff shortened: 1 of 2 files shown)


pulser/waveforms.py
```diff
@@ -538,13 +538,27 @@ def from_max_val(
...
         while np.abs(wf._scaling) > np.abs(max_val):
             duration += 1
+            previous_wf = wf
             wf = cls(duration, area)
+
+        # According to the documentation for numpy.blackman(), "the value one
+        # appears only if the number of samples is odd". Hence, the previous
+        # even duration can reach a maximal value closer to max_val.
+        if (
+            previous_wf is not None
```
HGSilveri, line 556: I think you have to initialize `previous_wf = None` to make this work. Also, maybe add a test that covers the case where the `while` loop is not even entered.

pulser/waveforms.py
```diff
@@ -538,13 +538,27 @@ def from_max_val(
...
+            and duration % 2 == 1
+            and np.max(wf.samples) < np.max(previous_wf.samples) <= max_val
```
HGSilveri, line 558: This check might be the most costly and it will fail when `previous_wf = None`, so you should nest it as a second `if`-statement.

HGSilveri (changes requested):
Looks good overall, just a couple things to take care of

pulser/waveforms.py
LaurentAjdnik: According to https://docs.python.org/3.8/reference/expressions.html#boolean-operations:

`The expression 'x and y' first evaluates x; if x is false, its value is returned; otherwise, y is evaluated and the resulting value is returned.`

Hence, no further condition is tested as soon as one fails. That's why I ordered them in increasing computation costs.

pulser/waveforms.py
LaurentAjdnik: > I think you have to initialize `previous_wf = None` to make this work.

It worked though. The variable is declared above but never initialized, so it's set to None by default. But I will add it for clarity/safety.

> Also, maybe add a test that covers the case where the `while` loop is not even entered.

I assume this is already achieved by the existing tests. Plus, this part of code is not impacted by/related to this PR.

HGSilveri: Thanks for adding it explicitly

HGSilveri (approved):
LGTM, thanks!

PR #223, opened 2021-06-29
Measurement errors incoporated into `CoherentResults`

### LINKED ISSUE #220: Have a simulation with only measurement errors return CoherentResults instead of NoisyResults
Currently, when a simulation includes `SPAM` noise, it returns a `NoisyResults` instance. However, when there are only measurement errors, the results could actually be given as `CoherentResults`. As such, we are currently throwing away information that could be given to the user.

Naturally, every time `doppler`, `amplitude` and/or "preparation" errors are present, the result has to be given as a `NoisyResults` instance (unless, perhaps, when `N_samples=1`). On the rest of the cases, the result should rather be `CoherentResults`.

Incorporating measurement errors into `CoherentResults` is not very hard: 
- The class already features the `_sampling_with_detection_errors()` hidden method, whose outcome can bypass those of `sample_state()` and `sample_final_state()`.  
- The most uncertain part is the behaviour of `expect()`, which should use the results post measurement. Although it would work, it would be messy to sample every state in `CoherentResults` and feed them to `NoisyResults`, which can then calculate the expectation values from diagonal operators. Instead, the most straightforward approach would rather be to turn the state vectors pre-measurement into density matrices post-measurement and use them to calculate the expectation values.

HGSilveri:
These changes incorporate measurement errors into `CoherentResults`, such that:
- When calling `sample_state()`, the result automatically includes the measurement errors (i.e. `_sampling_with_detection_errors()` was incorporated in `sample_state()`)
- When calling `expect()`, it calculates the "pseudo-density matrices", corresponding to the state of the system post-measurement. When measurement errors are present, these are used instead of the state vectors

A few extra features were added in the process:
- The ability to calculate the pseudo-density matrix of the system at a given time is part of the parent class, as it is needed in both child classes now. The calculation might be expensive for large systems, so I'm caching them so that they're only computed once.
- The measurement errors are built into these density matrices by changing the projectors into the basis states when measurement errors are present
- `NoisyResults._calc_weights()` was refactored to avoid the exponential complexity with the system size.
- `expect()` is now defined solely in `SimulationResults`

Fixes #220 .

### DIFF
--- a/pulser/simulation/simresults.py
+++ b/pulser/simulation/simresults.py
@@ -18,6 +18,7 @@ from __future__ import annotations
 from collections import Counter
 import collections.abc
 from abc import ABC, abstractmethod
+from functools import lru_cache
 from typing import Optional, Union, cast, Tuple
 
 import matplotlib.pyplot as plt
@@ -56,6 +57,8 @@ class SimulationResults(ABC):
         self._basis_name = basis_name
         self._sim_times = sim_times
         self._results: Union[list[Counter], list[qutip.Qobj]]
+        # Use the pseudo-density matrix when calculating expectation values
+        self._use_pseudo_dens: bool = False
 
     @property
     @abstractmethod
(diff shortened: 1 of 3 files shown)


julien-bremont:
This both agrees with former results and allows for a significant speed-up in targeted cases. Nice !
Former method used on Rabi oscillations with measurement errors only (there are error bars since results come from multiple measurements) :
Unknown-2
New method :
Unknown

PR #229, opened 2021-07-01
Add KaiserWaveform class (#228)

### LINKED ISSUE #228: Implement Kaiser waveform
### Discussed in https://github.com/pasqal-io/Pulser/discussions/227

<div type='discussions-op-text'>

<sup>Originally posted by **LaurentAjdnik** July  1, 2021</sup>
Hi all!

From https://arxiv.org/abs/2104.15044:

`In practice, a properly shaped waveform—like the Blackman waveform, which we will use here — can mitigate unwanted modifications to the modulated signal, stemming from spectral leakage or noise sensitivity.`

From https://numpy.org/doc/stable/reference/generated/numpy.blackman.html:

`The Blackman window is a taper formed by using the first three terms of a summation of cosines. It was designed to have close to the minimal leakage possible. It is close to optimal, only slightly worse than a Kaiser window.`

Hence, should we implement a `KaiserWaveform` based on https://numpy.org/doc/stable/reference/generated/numpy.kaiser.html (probably with `beta = 8.6`)?
</div>

To implement, with `beta=14` by default, as suggested in numpy's documentation.

LaurentAjdnik:
Adds KaiserWaveform class, based on `numpy.kaiser()`.

Adds the corresponding tests.

Fixes #228.


pulser/waveforms.py
```diff
@@ -739,6 +739,94 @@ def __mul__(self, other: float) -> InterpolatedWaveform:
         )
 
 
+class KaiserWaveform(Waveform):
+    """A Kaiser window of a specified duration and beta parameter.
+
```
HGSilveri, line 744: Can you write a bit about this window? Maybe a `See Also` for the numpy docs? I feel like it's necessary to explain the role of the `beta` parameter, somehow.

pulser/waveforms.py
```diff
@@ -739,6 +739,94 @@ def __mul__(self, other: float) -> InterpolatedWaveform:
...
+    def change_duration(self, new_duration: int) -> KaiserWaveform:
+        """Returns a new waveform with modified duration.
+
+        Args:
+            new_duration(int): The duration of the new waveform.
+
+        Returns:
+            KaiserWaveform: The new waveform with the same scaling and beta
+            but a new duration.
```
HGSilveri, line 805: nit: Indentation
```suggestion
                but a new duration.
```

HGSilveri (commented):
This looks very nice, thanks for the enhancement! I just have a couple thinks to discuss about what the signature of this waveform should be

pulser/waveforms.py
```diff
@@ -739,6 +739,94 @@ def __mul__(self, other: float) -> InterpolatedWaveform:
...
+            The default value is 1. Can be negative, in which case it takes
+            the positive waveform and changes the sign of all its values.
+        beta (Optional[float]): The beta parameter of the Kaiser window.
+            The default value is 14.
+    """
+
+    def __init__(
+        self,
+        duration: Union[int, Parametrized],
+        scaling: Optional[Union[float, Parametrized]] = 1.0,
```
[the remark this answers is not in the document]
LaurentAjdnik: I will go for the `area` approach.

BTW, why the `1e-3` in `self._area / float(np.sum(self._norm_samples)) / 1e-3` for `BlackmanWaveform`?

It must come from some unit conversion but it is not very clear to me.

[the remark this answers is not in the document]
HGSilveri: Exactly, it's a unit conversion. You have `ns = 1e-3 µs` on the x-axis and `rad/µs `on the y-axis. The product of the two gives you the factor of `1e-3` that you have to correct for.

pulser/waveforms.py
```diff
@@ -739,6 +739,108 @@ def __mul__(self, other: float) -> InterpolatedWaveform:
...
+        beta (Optional[float]): The beta parameter of the Kaiser window.
+            The default value is 14.
+    """
+
+    def __init__(
+        self,
+        duration: Union[int, Parametrized],
+        area: Union[float, Parametrized],
+        beta: Optional[Union[float, Parametrized]] = 14.0,
+    ):
```
HGSilveri, line 763: Cool! How do you feel about having `KaiserWaveform.from_max_val()` too?

LaurentAjdnik: `BlackmanWaveform.from_max_val()` relies on this to compute a value of `duration` to start with:

```
        # A normalized Blackman waveform has an area of 0.42 * duration
        duration = np.ceil(area / (0.42 * max_val) * 1e3)  # in ns
```

I could not find anything similar for a Kaiser window, especially since `beta` has a huge impact.

The only way I can think of is to start with `duration = 1` and then increment it... Not very "elegant"...

Do you think of any other approach?

HGSilveri: Indeed, it's not feasible to integrate the Kaiser window. What I have in mind is not much better than what you suggest, but it can take advantage of a couple facts:
- For a fixed value of ß, the Kaiser window's integral is proportional to the duration of the waveform, which means there is some proportionality factor `k(ß)` such that the area `A` is `A = k * duration`. Conversely, `duration = A / k`. 
- We know this factor decreases monotonically with ß. It starts as `k(0) = 1`, then `k(5) = 0.54`, `k(6) = 0.5`, `k(8.3) = 0.42`, and so on... Essentially, the window slims down with increasing ß, and so the area will decrease for a fixed duration.

My proposal is to take the given `beta` and get the nearest know `k(ß)` factor obeying `ß < beta`. This will give an upper-bound on the area for a fixed duration and, conversely, a lower-bound for the `duration` for the given area, which we can use as a starting point. From there, it's a matter of gradually increasing the duration, like we do for the Blackman. We can either increase in steps of 1 to make sure we don't miss the sweet spot, or try a bigger step and then correct in case of an overshoot.

Also, the value of `k` decreases progressively slower as ß increases, so for large values of ß we won't be that far off if we start off with the value of `k(14) = 0.321`, which can be the `k` factor corresponding to the largest ß we care to store.

LaurentAjdnik: Working (hard) on it. But quite busy lately 😓. I will write more and PR something ASAP.

HGSilveri: No rush at all, take your time. Thank you for your efforts! 😄

LaurentAjdnik: Here we go again!

Out of curiosity, I've been working on some kind of dichotomic approach but yours if much wiser and more efficient.

I spent a lot of time exploring this method and here are a few conclusions.

The ratio `area / duration` fluctuates a lot on small durations (mostly because of the even/odd durations artifact already mentioned in #218) but it tends to stabilize  after a while.

Hence, we can compute a satisfying value, without having to store any of them, by first creating a disposable waveform of `duration = 100` with the given `beta`.

Using this ratio, the computed duration guess is perfect most of the time, with a few exceptions by +/- 2. For 100% success, we can start from `duration_guess - 2` and increment step by step, as we do for `BlackmanWaveform`.

But this is true only when the duration guess is above 10. Below, the seesawing effect is very strong and I recommend testing all waveforms for duration from 1 to 10 and selecting the best one.

I wrote the corresponding code. It works perfectly. But there is still room for some improvement.

What do you think?

HGSilveri: This sounds like a great solution, I'll take a look right now.

LaurentAjdnik: When the duration guess is less than or equal to 10, all durations from 1 to 15 are tested (stopping at 10 proved to be a bad idea since a guess on 9 might yield an optimal value on 11).

Otherwise: I enhanced the path to the best duration. Starting from the guessed duration minus 2 was not always a good pick. I tried with 99 % of the guessed duration but it is not very efficient. So, I simply start from the guessed duration and decide if it needs to be incremented or decremented towards the optimal duration. This leads to as few loops as possible.

Note: Not tested yet for negative areas and I think it would fail...

HGSilveri: I didn't think of negative areas for the `BlackmanWaveform` either. The best solution is probably to use the absolute value and then set the appropriate sign on the final waveform.

pulser/waveforms.py
```diff
@@ -794,6 +794,74 @@ def __init__(
...
+                sign of `area`.
+            area (float): The area under the waveform.
+            beta (Optional[float]): The beta parameter of the Kaiser window.
+                The default value is 14.
+        """
+        # Compute the ratio area / duration for a long duration
+        # and use this value for a first guess of the best duration
+
+        max_val = cast(float, max_val)
+        area = cast(float, area)
```
HGSilveri, line 824: To solve the case where `max_val` and `area` are negative, I think the way to go is to check that's the case here (raise an error if not), then store the sign in a variable and use the absolute values for `max_val` and `area`. Then when you create the final class that you return, you use the stored sign on the `area`.

pulser/waveforms.py
```diff
@@ -794,6 +794,74 @@ def __init__(
...
+                    duration_best = duration
+
+        else:
+            # Increase or decrease duration depending on
+            # the max value for the guessed duration
+
+            kaiser_guess = np.kaiser(duration_guess, beta)
+            scaling_guess = 1000 * area / np.sum(kaiser_guess)
+
+            step = 1 if np.max(kaiser_guess) * scaling_guess > max_val else -1
```
HGSilveri, line 851: Just so I'm sure I have this right: you have that `step = np.sign(np.max(kaiser_guess) * scaling_guess - max_val)` when they are not equal, right?

pulser/waveforms.py
```diff
@@ -794,6 +794,74 @@ def __init__(
...
+            duration = duration_guess
+
+            max_val_temp = max_val + step
```
HGSilveri, line 854: Hum, wasn't `step` supposed to be for the `duration`?  Shouldn't `max_val_temp` be `np.max(kaiser_guess) * scaling_guess` instead?  I get that this works because it makes you go inside the loop below, but it doesn't seem necessary and is potentially misleading.

pulser/waveforms.py
```diff
@@ -794,6 +794,74 @@ def __init__(
...
+            while np.sign(max_val_temp - max_val) == np.sign(step):
+                duration += step
+                kaiser_temp = np.kaiser(duration, beta)
+                scaling = 1000 * area / np.sum(kaiser_temp)
+                max_val_temp = np.max(kaiser_temp) * scaling
+
+            duration_best = duration if step == 1 else duration + 1
+
+        return cls(duration_best, area, beta)
```
HGSilveri, line 863: This looks neat, very nice job!

HGSilveri (commented):
I like the approach you took! I just have a few remarks

pulser/waveforms.py
LaurentAjdnik: I agree.

pulser/waveforms.py
LaurentAjdnik: `step` tells us if the duration must be incremented (the guess is greater than `max_val`) or decremented (the guess is less than `max_val`).

So, yes, both lines of code are more or less equivalent.

The only difference is that I never have `step == 0` that could end up in a infinite loop.

But something might be wrong here... I didn't take into account a perfect match before looping...

pulser/waveforms.py
LaurentAjdnik: This is a trick to initialize `max_val_temp` before entering the loop and to make sure the `while` will happen at least once.

But it is indeed misleading. I should force to `0` or `np.inf` (which I wanted to avoid at first).

HGSilveri: But do you always want it to enter at least once? If you set  `max_val_temp = np.max(kaiser_guess) * scaling_guess`, it will always enter except when `max_val_temp == max_val`, in which case you are already at the right result and you don't want to enter. Am I missing something here?

pulser/waveforms.py
HGSilveri: I think you should have this check for `np.max(kaiser_guess) * scaling_guess >= max_val` instead. That way, when they match perfectly, it will set `step == 1` which means it won't make the adjustment of `duration_best` at the end. Combine this with it not entering the while loop if you set the `max_val_temp` as I suggested before, and you go straight to the right result.

HGSilveri (approved):
Great! LGTM, thanks a lot

PR #236, opened 2021-07-13
Fixes bug with SimResults.expect()

julien-bremont:
Fixes a bug regarding the computation of expectation values in the all basis via SimResults.expect().


HGSilveri (approved):
LGTM, thanks!

PR #237, opened 2021-07-13
`BlackmanWaveform.from_last_val()` fully compatible with negative inputs

HGSilveri:
Although the original implementation of `BlackmanWaveform.from_last_val()` had negative values in mind, the changes introduced in #222 were designed with only positive values in mind. These changes serve to make those changes work for negative waveforms as well.

### DIFF
--- a/pulser/tests/test_waveforms.py
+++ b/pulser/tests/test_waveforms.py
@@ -194,6 +194,13 @@ def test_blackman():
     wf2 = BlackmanWaveform(duration + 1, area)
     assert np.max(wf2.samples) < np.max(wf.samples) <= max_val
 
+    # Same but with a negativa max value
+    wf: BlackmanWaveform = BlackmanWaveform.from_max_val(-max_val, -area)
+    duration = wf.duration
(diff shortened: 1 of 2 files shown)


pulser/tests/test_waveforms.py
```diff
@@ -194,6 +194,13 @@ def test_blackman():
     wf2 = BlackmanWaveform(duration + 1, area)
     assert np.max(wf2.samples) < np.max(wf.samples) <= max_val
 
+    # Same but with a negativa max value
```
LaurentAjdnik, line 197: Tiny typo on `negativa` :wink:

```suggestion
    # Same but with a negative max value
```

LaurentAjdnik (changes requested):
Nice job!

pulser/tests/test_waveforms.py
HGSilveri: Nice catch, cheers!

LaurentAjdnik (approved):
LGTM!

PR #239, opened 2021-07-15
Fixes dimension bug in `Simulation.initial_state` setter

julien-bremont:
Same bug as in #236 but for `initial_state`.


HGSilveri (approved):
LGTM, thanks for the fix!

PR #242, opened 2021-07-26
Make Pulse a frozen dataclass

HGSilveri:
By making `Pulse` a frozen dataclass, we ensure that it does not change after initialization, which prevents a user from changing a `Pulse` that has been added to a Sequence (which would change the already added `Pulse` too). 

Additionally, it automatically adds an `__eq__` magic method, which can be useful at times.

Unfortunately, the type hints of the arguments to `__init__` had to be changed, but I feel that's a small price to pay for the added benefits.


darcangelomauro (approved):
LGTM

PR #245, opened 2021-08-05
Forbid combination of custom initial_state with state preparation errors

### LINKED ISSUE #244: Incompatibility between setting of an initial state and having state preparation errors on a simulation
The current release allows the user to set a custom initial state and a configuration that includes state preparation errors. However, the way a "badly prepared" state is based on pretending that it is left on the ground-state the entire time, with no terms in the Hamiltonian acting on it. This means that a custom initial state preparing some atoms on a state other than the ground is in a potential conflict with a state preparation error, which for our intentions effectively forces a state to stay in the ground the entire time.

The most complete approach here would be to check, on each run, if the atoms that have been randomly selected to be on "badly prepared" are on the ground-state and if not, either: 1. Force them to be on the ground-state, or 2. Interrupt the run.

Option 1. is non-trivial, computationally heavy and perhaps even non-sensical when the total state is entangled. We could consider using iff the state is a product state, but that also can be resource-intensive to check from the whole state vector.

Option 2. is counting on luck, essentially hoping that only states starting from the "ground-state" are picked to be "badly prepared". If this does not happen, the run would be interrupted, potentially wasting all the elapsed time. 

As both options above seem unsatisfactory, I would say that the solution, at least for the time being, is to forbid the setting of a custom initial state when state preparation errors are involved in the simulation.

As this is technically a bug on the current release, it should be treated as an hotfix to be merged directly into `master`.

HGSilveri:
This PR prevents the existing conflict between the setting of a custom initial state and the way we simulate state preparation errors, as described in #244.

Closes #244 .

### DIFF
--- a/pulser/simulation/simulation.py
+++ b/pulser/simulation/simulation.py
@@ -238,7 +238,7 @@ class Simulation:
     def initial_state(self, state: Union[str, np.ndarray, qutip.Qobj]) -> None:
         """Sets the initial state of the simulation."""
         self._initial_state: qutip.Qobj
-        if state == "all-ground":
+        if isinstance(state, str) and state == "all-ground":
             self._initial_state = qutip.tensor(
                 [self.basis["g"] for _ in range(self._size)]
             )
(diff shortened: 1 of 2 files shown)


Louis-PaulHenry (approved):
I agree this version is the more satisfiable hotfix for the issue.

PR #247, opened 2021-08-09
Protected deserialization

HGSilveri:
Restricts objects that support serialization and deserialization to guarantee it is safe (ie it either works or fails without harm) to deserialize any JSON formatted string on a clean install of Pulser.


bejito:
I tried to hack this but didn't succeed so far.
I'm still a bit scared of this pattern, especially with the getattr

HGSilveri:
> I tried to hack this but didn't succeed so far. I'm still a bit scared of this pattern, especially with the getattr

I think we can make do without the `getattr`, would you be in favour of preventively removing it?

bejito:
> I think we can make do without the `getattr`, would you be in favour of preventively removing it?

If that's not too complicated to do, I think we should. It will remove the biggest potential vector of attack.

HGSilveri:
> If that's not too complicated to do, I think we should. It will remove the biggest potential vector of attack.

Ok, it's done. The loss of functionality was marginal, in my opinion.

pulser/json/supported.py
```diff
@@ -0,0 +1,82 @@
...
+        raise TypeError("Invalid 'obj_dict'.")
+
+    if module_str not in SUPPORTED_MODULES:
+        raise SystemError(
+            f"No serialization support for module '{module_str}'."
+        )
+
+    if "__submodule__" in obj_dict:
+        submodule_str = obj_dict["__submodule__"]
+        if submodule_str not in SUPPORTS_SUBMODULE:
```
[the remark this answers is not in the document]
HGSilveri: I think the whole list of supported modules and submodules being extracted from the code (i.e. no having to be explicitly written in `supported.py`) would be a nice feature to have, indeed, but it doesn't seem trivial. 
For now, I'll leave it as it is and we can consider improving it later on

pulser/tests/test_json.py
```diff
@@ -57,6 +61,23 @@ def test_register_3d():
...
+def test_layout():
+    custom_layout = RegisterLayout([[0, 0], [1, 1], [1, 0], [0, 1]])
+    new_custom_layout = encode_decode(custom_layout)
+    assert new_custom_layout == custom_layout
+    assert type(new_custom_layout) is RegisterLayout
+
+    tri_layout = TriangularLatticeLayout(100, 10)
+    new_tri_layout = encode_decode(tri_layout)
+    assert new_tri_layout == tri_layout
+    assert isinstance(new_tri_layout, TriangularLatticeLayout)
```
Codoscope, line 73: Why not also using`type(…) is …` here?

pulser/json/supported.py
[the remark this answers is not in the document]
Codoscope: Sure, no problem, and maybe we can create an issue for it (I may do it). It would also be great to automatize run-time type checking for the constructors exposed to outside, to make sure not to forget any.

pulser/register/special_layouts.py
```diff
@@ -166,3 +170,6 @@ def __str__(self) -> str:
             f"TriangularLatticeLayout({self.number_of_traps}, "
             f"{self._spacing}µm)"
         )
+
+    def _to_dict(self) -> dict[str, Any]:
+        return obj_to_dict(self, self.number_of_traps, self._spacing)
```
Codoscope, line 175: Warning: it seems the only way the parameter `n_traps`'s type is "checked" is inside `triangular_hex` with `if n_points < 7`

Codoscope (approved):

pulser/tests/test_json.py
HGSilveri: No good reason, I'll change it for consistency's sake.

pulser/register/special_layouts.py
HGSilveri: Do you think this is an issue? Do you want me to make sure that it's an `int` with `int(n_traps)`?

Codoscope: I think it's an issue because `TriangularLatticeLayout` is included in `SUPPORTED_MODULES` and thus is exposed to the outside. We probably want to check that users won't inject something else than an `int`.

HGSilveri: Ok, but note that what they inject there still needs to be in the supported modules, so I don't think they can put anything harmful. 

Still, how do you propose we make this check?

PR #254, opened 2021-08-25
Fixed global-local overwriting bug and added test

darcangelomauro:
The error was caused by amp, det and phase samples replacing the old values instead of adding. That would be fine it two pulses with different addressing and basis could never write on the same samples, but with the introduction of noise sometimes a global pulse becomes local and a conflict might happen.

Since the amp, det and phase samples are always initialized to 0, it's fine to add instead of replacing.

I also added a unit test in test_simulation.py that the old code does not pass regarding this issue.

### DIFF
--- a/pulser/_version.py
+++ b/pulser/_version.py
@@ -12,4 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = "0.3.1.dev"
+__version__ = "0.3.1"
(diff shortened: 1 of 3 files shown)


darcangelomauro:
There are three commits in the branch by @HGSilveri that I didn't see. I'm not sure how that happened since I was the only one working on this branch. One of the commits changed the version to `0.3.1.dev`, I believe this should not happen in `master`, right?

sebgrijalva:
Yes, these were adaptations to the `develop` branch to keep track of the version.

sebgrijalva (approved):
LGTM, Thanks a lot for noticing this bug!

Issue #12, opened 2020-11-10
Add a LICENSE


karalekas:
Before open-sourcing, you'll need to pick what type of license you want to use (e.g. `Apache-2.0`). After that, do the following:
1. Put a copy of it in the repository.
2. Update the `setup` function in `setup.py` to tell it what license you chose.
3. Add `include LICENSE` to `MANIFEST.in`.
4. Take the shortened license and apply it to all the source code files in the repository (see top of [this file](https://github.com/unitaryfund/mitiq/blob/master/mitiq/conversions.py)).

PR #258, opened 2021-08-31
SLM mask

darcangelomauro:
Implemented the SLM mask for initial state preparation.

`Sequence` has now a new attribute, `Sequence._slm_mask_targets`.
`Sequence._slm_mask_targets` is a set of qubits that specifies which qubits will be masked.

`Sequence` has now a new method, `Sequence.config_slm_mask(qubits: Set[QubitId])`.
`Sequence.config_slm_mask` is the way users can set the `Sequence._slm_mask_targets` attribute.

`Simulation.initial_state` now accepts a new string: `"slm-mask"`.
When `Simulation.initial_state` is set to be `"slm-mask"`, the initial state is prepared so that all masked qubits are in 0, and all unmasked qubits are in 1.
What 0 and 1 is, depends on the basis that the simulation uses.
XY: 0 is "d", 1 is "u"
ground-rydberg: 0 is "g", 1 is "r"
digital: 0 is "g", 1 is "h"
all: 0 is "g", 1 is "h"


darcangelomauro:
Sequence has now two new attributes, Sequence._slm_mask_targets and Sequence._slm_mask_time.
Sequence._slm_mask_targets is a set of qubits that specifies which qubits will be masked.
Sequence._slm_mask_time is a list of the type [t_i, t_f] that represent the time when the mask was turned on and off respectively.

Sequence has now a new method, Sequence.config_slm_mask(qubits: Set[QubitId]).
Sequence.config_slm_mask(qubits: Set[QubitId]) tells the sequence which qubits are to be masked, and can be called only one time.

The mask is automatically turned on during the first pulse and off during the rest of the simulation. This ensure that its effect is to effectively remove the masked qubits from the register. This means that amp, det and phase samples of the first pulse are put to zero if targeting the masked qubits, as well as the XY interaction terms.
Two interaction terms are calculated in Simulation now: one with all the qubits (call it V), and one without the masked qubits (call it V_masked). The interaction terms are now time-dependent (of type qutip.QobjEvo) with a coefficient that is either 0 or 1. When the mask is on, V has coefficient 0 and V_mask has coefficient 1, and vice versa when the mask is off.

When visualizing the sequence, the mask appears as a darker shade in the background. At the bottom left of the shaded region, the masked qubits are specified in a grey box with bars.

darcangelomauro:
This is the way the mask looks when the sequence is drawn
draw_mask

pulser/sequence.py
```diff
@@ -1111,6 +1128,45 @@ def _reset_parametrized(self) -> None:
...
+                tf = first.tf
+            # If try fails, it's because the variable 'first' is not defined,
+            # meaning there are no pulses in this channel
+            except NameError:
+                continue
+            try:
+                if ti < self._slm_mask_time[0]:
+                    self._slm_mask_time = [ti, tf]
+            except IndexError:
+                self._slm_mask_time = [ti, tf]
```
HGSilveri, line 1168: Instead of breaking the for-loop right away, why don't you make these checks once you find a first pulse? Also, it's often considered a good practice to avoid using `try-except` statements and to "move the code to the left" as much as possible. Combining all this together would look something like this:
```suggestion
            for slot in self._schedule[channel]:
                if not isinstance(slot.type, Pulse):
                    continue
                ti = slot.ti
                tf = slot.tf
                if self._slm_mask_time:
                    if ti < self._slm_mask_time[0]:
                        self._slm_mask_time = [ti, tf]
                else:
                    self._slm_mask_time = [ti, tf]
                break
```

pulser/simulation/simulation.py
```diff
@@ -419,13 +419,26 @@ def write_samples(
...
-            samples_dict["amp"][slot.ti : slot.tf] = (
+            samples_dict["amp"][slot.ti : slot.tf] += (
                 _pulse.amplitude.samples * noise_amp
             )
-            samples_dict["det"][slot.ti : slot.tf] = (
+            samples_dict["det"][slot.ti : slot.tf] += (
                 _pulse.detuning.samples + noise_det
             )
-            samples_dict["phase"][slot.ti : slot.tf] = _pulse.phase
+            samples_dict["phase"][slot.ti : slot.tf] += _pulse.phase
```
HGSilveri, line 428: It would be nice that these changes were not repeated here. They are already in `develop`, so if you want them to be here the way to go would be to get them from there (whether you merge `develop` to `noisy-xy` and then from `noisy-xy` to here or straight from `develop` will depend on whether we will keep this PR targeting `noisy-xy` or change to `develop`)

pulser/sequence.py
```diff
@@ -1111,6 +1128,45 @@ def _reset_parametrized(self) -> None:
         self._variables = {}
         self._to_build_calls = []
 
+    def config_slm_mask(self, qubits: Set[QubitId]) -> None:
```
HGSilveri, line 1131: Don't forget you need to store calls to this function for serialization purposes

HGSilveri (commented):
Ok, here's a review of half of the files. From looking at `Simulation`, there seems to be a lot of code replication that we should probably try to avoid. I'll keep looking into it

pulser/simulation/simulation.py
```diff
@@ -700,11 +781,56 @@ def build_coeffs_ops(basis: str, addr: str) -> list[list]:
...
+
+        def build_unmasked_coeff() -> np.ndarray:
+            coeff = np.ones(self._tot_duration)
+            try:
+                ti = self._seq._slm_mask_time[0]
+                tf = self._seq._slm_mask_time[1]
+            except IndexError:
+                return coeff
+            coeff[ti:tf] = 0
+            return coeff
```
HGSilveri, line 802: Two things:

1. One seems to be just the logical NOT of the other
2. Why do you need these in functions?

pulser/simulation/simulation.py
```diff
@@ -700,11 +781,56 @@ def build_coeffs_ops(basis: str, addr: str) -> list[list]:
...
+                qobj_list += (
+                    [
+                        [
+                            make_masked_interaction_term(),
+                            self._adapt_to_sampling_rate(m_coeff),
+                        ]
+                    ]
+                    if self._size > 1
+                    else []
+                )
```
HGSilveri, line 831: Bear in mind that `self._size` will be `>1` because of the first `if`.

pulser/simulation/simulation.py
```diff
@@ -640,9 +677,53 @@ def make_xy_term() -> qutip.Qobj:
                 )
             return xy
 
-        make_interaction_term = (
-            make_xy_term if self._interaction == "XY" else make_vdw_term
-        )
+        def make_masked_xy_term() -> qutip.Qobj:
```
HGSilveri, line 680: There's an awful lot of code replication from this function. Why not just modifiy the `make_xy_term` function to accept a `masked=True/False` keyword argument instead?

HGSilveri (commented):
Here go a few suggestions for the part in `Simulation`. Although they will be reduced, you are right about not merging this straight into `develop`. I did not think of the modifications that were needed for the interactions

pulser/devices/_mock_device.py
```diff
@@ -47,5 +47,11 @@
             "mw_global",
             Microwave.Global(1000, 200, clock_period=1, min_duration=1),
         ),
+        (
+            "mw_local",
+            Microwave.Local(
+                1000, 200, 0, max_targets=2000, clock_period=1, min_duration=1
+            ),
+        ),
```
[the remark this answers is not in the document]
darcangelomauro: At first I did it because I thought that the samples dictionary in `Simulation` was based on the existing channels, but that's not the case so you're right

pulser/sequence.py
```diff
@@ -635,6 +644,14 @@ def add(
         if self._empty_sequence:
             self._empty_sequence = False
 
+        # Update the SLM mask initial and final time if necessary
+        if self._slm_mask_targets:
+            try:
+                if self._slm_mask_time[0] > ti:
+                    self._slm_mask_time = [ti, tf]
+            except IndexError:
+                self._slm_mask_time = [ti, tf]
```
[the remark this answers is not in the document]
darcangelomauro: Oh I thought that it was supposed to be on only during the pulse.
So if the mask is set, it will be on from time 0 no matter what?

pulser/sequence.py
darcangelomauro: You mean decorating with `@_store`?
I'm still not sure when to do that. Is it because some of its values depend on pulses?

pulser/sequence.py
darcangelomauro: looks good

pulser/simulation/simulation.py
```diff
@@ -419,13 +419,26 @@ def write_samples(
...
+        def affected_by_slm(slot: _TimeSlot) -> bool:
+            """Check if the SLM is on during a slot."""
+            if not self._seq._slm_mask_targets:
+                return False
+            if self._seq._slm_mask_time:
+                ti_slot = slot.ti
+                tf_slot = slot.tf
+                ti_mask = self._seq._slm_mask_time[0]
+                tf_mask = self._seq._slm_mask_time[1]
+                if ti_mask < tf_slot and tf_mask > ti_slot:
```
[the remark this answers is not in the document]
darcangelomauro: I think it's clearer with an example.

pulse: (1, 3)
mask: (0, 2)
0<3 and 2>1: true

pulser/simulation/simulation.py
```diff
@@ -419,13 +419,26 @@ def write_samples(
...
+                    return True
```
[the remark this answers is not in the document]
darcangelomauro: right

pulser/simulation/simulation.py
```diff
@@ -419,13 +419,26 @@ def write_samples(
...
+
+        def affected_by_slm(slot: _TimeSlot) -> bool:
+            """Check if the SLM is on during a slot."""
+            if not self._seq._slm_mask_targets:
+                return False
+            if self._seq._slm_mask_time:
+                ti_slot = slot.ti
+                tf_slot = slot.tf
+                ti_mask = self._seq._slm_mask_time[0]
+                tf_mask = self._seq._slm_mask_time[1]
```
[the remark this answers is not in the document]
darcangelomauro: I thought it made the code more readable and the condition more compact, but if you think the overhead is not worth it I'll call the variables directly

pulser/simulation/simulation.py
```diff
@@ -435,21 +448,33 @@ def write_samples(
...
+                        # Otherwise, populate corresponding local
+                        else:
+                            samples_dict = self.samples["Local"][basis]
+                            for qubit in slot.targets:
+                                if qubit not in samples_dict:
+                                    samples_dict[qubit] = prepare_dict()
+                                write_samples(
+                                    slot, samples_dict[qubit], True, qubit
+                                )
+                            self.samples["Local"][basis] = samples_dict
```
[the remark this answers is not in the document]
darcangelomauro: Yes I thought about it, this is what happened.
At first I implemented the mask as if you could turn it on and off many times mid simulation, and in that case any slot could be affected.
But then we decided to leave it just for the first pulse because the effect on superposition states was not trivial. The code was basically the same as before, but instead of having a list of times like [[t1_i, t1_f], [t2_i, t2_f], ...] now we just have one [t_i, t_f].
So I ended up using the code I had already written, where checking every slot made sense. I thought it could be useful to leave the code like this in case we eventually decide to implement turning the mask on and off many times.

Your way seems a good idea though, and it should be possible to expand beyond state preparation with minimal modifications

pulser/simulation/simulation.py
```diff
@@ -458,7 +483,19 @@ def write_samples(
...
+            if self._seq._slm_mask_targets:
+                try:
+                    ti = self._seq._slm_mask_time[0]
+                    tf = self._seq._slm_mask_time[1]
+                    for qubit in self._seq._slm_mask_targets:
+                        self.samples["Local"][basis][qubit]["amp"][ti:tf] = 0
+                        self.samples["Local"][basis][qubit]["det"][ti:tf] = 0
+                        self.samples["Local"][basis][qubit]["phase"][ti:tf] = 0
+                except IndexError:
+                    pass
```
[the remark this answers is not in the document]
darcangelomauro: looks good

pulser/simulation/simulation.py
darcangelomauro: 1. True, I could just leave one
2. I thought it looked more neat and clearer. You would prefer it without its own function?

pulser/simulation/simulation.py
darcangelomauro: It will? I don't think I ever check if there is more than one qubit when I set the mask (which might actually be a good idea)

pulser/simulation/simulation.py
darcangelomauro: You're right, I'll do that

pulser/sequence.py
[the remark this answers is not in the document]
HGSilveri: Yes, in practice that's what will happen. If configured, it's turned on until the end of the first pulse.

pulser/sequence.py
HGSilveri: Yeah, it will depend on whether we want to allow parametrized objects to serve as qubit ID's. I would say no for now, so we should talk about how to do it.

pulser/simulation/simulation.py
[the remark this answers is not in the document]
HGSilveri: You're right. All I would say is that the first check is probably redundant, seeing as ti_mask will always be 0.

pulser/simulation/simulation.py
```diff
@@ -606,43 +627,59 @@ def make_vdw_term() -> qutip.Qobj:
                     vdw += U * self.build_operator([("sigma_rr", [q1, q2])])
             return vdw
 
-        def make_xy_term() -> qutip.Qobj:
+        def make_xy_term(masked: bool) -> qutip.Qobj:
```
HGSilveri, line 630: This would work better with a keyword arg (`masked: bool = False`), imo

pulser/simulation/simulation.py
```diff
@@ -606,43 +627,59 @@ def make_vdw_term() -> qutip.Qobj:
...
-                )
-                xy += U * self.build_operator(
-                    [("sigma_du", [q1]), ("sigma_ud", [q2])]
-                )
             return xy
 
-        make_interaction_term = (
-            make_xy_term if self._interaction == "XY" else make_vdw_term
-        )
+        def make_interaction_term(masked: bool) -> qutip.Qobj:
```
HGSilveri, line 678: Same here

HGSilveri (approved):
LGTM, I only have this one suggestion

tutorials/applications/Spin chain of 3 atoms in XY mode.ipynb
```diff
@@ -34,7 +34,7 @@
   },
```
HGSilveri, line 34: <pre class="ql-syntax" spellcheck="false">Line #12.    masked_pulse = Pulse.ConstantPulse(110, 2*np.pi*4.6, 0, 0)
</pre><p>This is quite obscure to be honest, why not use a BlackmanWaveform with an area of π instead?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pasqal-io/Pulser/pull/258/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

darcangelomauro:
A check on the effective size of the register was added inside `make_xy_term`. This checks whether the total number of qubits minus the number of bad and masked qubits is less than 2. In that case, it returns no interactions.
This check cannot be put in `build_coeffs_ops`, which already checks for bad qubits, because bad qubits are bad from start to finish, while whether a qubit is masked or not is time-dependent. 
But perhaps there's a better way of doing this which does not calculate an effective size in two different places.

pulser/simulation/simulation.py
```diff
@@ -640,44 +640,51 @@ def make_xy_term(masked: bool = False) -> qutip.Qobj:
             The units are given so that the coefficient
             includes a 1/hbar factor.
             """
-            # Check if the total number of unmasked qubits is less than 2
+            # Calculate the total number of good, unmasked qubits
             effective_size = (
-                self._size - len(self._seq._slm_mask_targets)*int(masked)
+                self._size
+                - len(self._seq._slm_mask_targets) * int(masked)
+                - sum(self._bad_atoms.values())
```
HGSilveri, line 647: What about when a bad atom is also a masked one? Aren't you subtracting it twice?

HGSilveri (changes requested):
I think what you mention is fine, but there is a different small problem though

pulser/simulation/simulation.py
darcangelomauro: Correct, I'll fix it

darcangelomauro: Now it should be fixed. I've added a couple of unit test to check if it's working properly, just to be sure.
The check inside `make_xy_term` happens only if the masked term is being created, because for the unmasked case the check was already done outside, right?

pulser/simulation/simulation.py
```diff
@@ -641,26 +641,24 @@ def make_xy_term(masked: bool = False) -> qutip.Qobj:
...
-                    or (
-                        masked
-                        and (
-                            q1 in self._seq._slm_mask_targets
-                            or q2 in self._seq._slm_mask_targets
-                        )
+                    or masked
+                    and (
+                        q1 in self._seq._slm_mask_targets
+                        or q2 in self._seq._slm_mask_targets
```
HGSilveri, line 661: Why did you remove the parameters? Now it will do this every time it is `masked`

pulser/simulation/simulation.py
HGSilveri: You mean the check for `effective_size`?

pulser/simulation/simulation.py
darcangelomauro: sorry I'm not sure what happened there. It should be fixed now

pulser/simulation/simulation.py
darcangelomauro: Yes that one

HGSilveri: That is correct, this is enough

HGSilveri (approved):
LGTM

Issue #248, opened 2021-08-10
SLM mask for partial register addressing with a Global channel


HGSilveri:
When working in the XY mode, all the atoms start off in the "down" state by default. Due to the lack of a `Microwave.Local` channel, they currently cannot be prepared into another initial configuration through Pulser.

However, experimentally it is possible to prepare the system in specifc initial configurations by overlaying an SLM pattern that brings certain atoms out-of-resonance with the microwave, effectively allowing the driving of the transition on only the atoms that are still in resonance. 

This SLM pattern cannot be modified in the middle of a run, so the chosen atoms are fixed for the entirety of the sequence. This is why it is most useful as a way of preparing the initial state of the system.

The interface I have in mind is as follows:

- A method (e.g. `Sequence.config_slm_mask()`) can be defined, where one determines which qubits will be masked.
- A new instruction `Sequence.set_slm_mask(on_state: bool, align_with_channel: str)` allows for turning on pattern on and off as the sequence is built. The time at which it changes state should be aligned with a given channel.

Important remarks: 

- The effect of the SLM mask will affect all channels. It might make sense to forbid (or at least warn of) targetting of masked qubits with local channels when the SLM is on, for example.
- The representation of this in the Sequence drawing can be a slight darker shade of the background colour on all channels when the pattern is on.
- Attention needs to be had regarding all the times where a Global channel is impliciltly assumed to be acting on all qubits (e.g. in the `Simulation` module)
- The option to use this feature should depend on whether the chosen device supports it

HGSilveri:
> Hi, I've got a question on the issue.
> Unless I'm misinterpreting the code, it seems to me that the SLM activation and deactivation should be linked to a `Pulse` (or a `_TimeSlot` perhaps?) rather than a `Channel`. Otherwise you can't deactivate it.
> What do you think?
> 
> edit: Or do you mean that `Sequence.set_slm_mask` should work similarly to `Sequence.add`?

@darcangelomauro indeed, that's what the I was trying to convey with `align_with_channel` argument: that you would "align" the moment when you turn on the SLM with the time the chosen channel is free (i.e. it's last instruction ends). Does it make sense?

darcangelomauro:
Ok I see, thanks.
Could that bring to a situation where the slm turns on and off in the middle of a different pulse than the one it's aligned with?
And should I be mindful in general of a user potentially mixing together microwave and rydberg/raman channels in the same sequence?

darcangelomauro:
[...]

PR #264, opened 2021-09-15
Adding support for XY Simulation

### LINKED ISSUE #147: Add Support for Simulation in XY Mode
With the recently added support for `Microwave` channels and a corresponding "XY Mode" in a Pulser sequence (Refer to [this reference](https://arxiv.org/pdf/1408.1055.pdf) for some background on the physics), it remains to add the corresponding simulation support. This implies adding an XY Hamiltonian which takes as time-dependent coefficients the pulses of the sequence, and will imply several changes to the `Simulation` class:

- A  `self._interaction` or `self._sequence_mode` private attribute will allow to classify the incoming sequence depending on the chosen mode (should be related to `Sequence._in_xy`).
- The `self.samples` dictionary has to correspond to the case of 'Ising' or 'XY' interaction: In the case of the latter, since there is only one basis (`'xy'`), we only need slots for `'Global'` and `'Local'`.
- Since the sequence has been verified to have consistent basis of pulses throughout, the `write_samples()` function inside the `self._extract_samples()`method would only be slightly modified. On the other hand, the `self._build_basis_and_op_matrices()` method should now reflect the two-Rydberg level nature of the interaction , while following the existing scheme of the (actually more complex) Ising interaction. Perhaps just call these levels 'up' and 'down'?
- In the `make_interaction_term()` function of the `self._construct_hamiltonian()` method, one needs to add the proper term `0.5 * C_3 / R_ij**3`, which depends on the cube of the interatomic distance and has now a different coupling constant `C_3` (which should be taken as  `C_3 = 3.7 10**3 Mhz * µm**3`). 
- In fact, the coefficient `C_3` depends on the relative position vector of the atoms (Cf. Supplementary Material of [the Reference](https://arxiv.org/pdf/1408.1055.pdf)) as `C_3(theta) = C_3 * (1 - 3*cos(theta)**2)`. This dependance should be included by reading from the `Register` the position of the corresponding atoms.
- The `build_coeffs_ops()`function inside `self._construct_hamiltonian()`should also take into account the differently built operators.
- Finally, one must edit the `SimulationResults` class: In this case, both the `basis_name` and `meas_basis` correspond to the "XY mode" which implies `SimulationResults._dim = 2`. Thus, many of the checks and basis reduction methods are simpler, although they have to be clearly assessed.

To check things are working correctly, one could try to add a test displaying Rabi Oscillations with similar properties as that of Fig. 1(c) of the [Reference](https://arxiv.org/pdf/1408.1055.pdf), with one and then several atoms in an array and global/local pulses.

HGSilveri:
- Adds simulation capabilities in the XY mode, as described in detail #147 and originally executed in #198 
- Combines the changes of #198 with the latest version of `pulser.simulation`, which supports noisy simulation. In this PR, only support for simulation of `"SPAM"` noise in the XY mode is added.
- Adds an SLM Mask feature for initial state preparation (as described in #248 and executed in #258)

Closes #147.


pulser/sequence.py
```diff
@@ -479,6 +543,9 @@ def add(
         if self.is_parametrized():
             if not isinstance(pulse, Parametrized):
                 self._validate_pulse(pulse, channel)
+            # Sequence is marked as non-empty on the first added pulse
+            if self._empty_sequence:
+                self._empty_sequence = False
```
darcangelomauro, line 548: If I understand correctly, this is needed  because the magnetic field can be set only on an empty sequence. But I don't understand why the empty sequence constraint is necessary

pulser/sequence.py
```diff
@@ -299,6 +320,48 @@ def current_phase_ref(
...
+            # Not all channels are empty
+            raise ValueError(
+                "The magnetic field can only be set on an empty " "sequence."
+            )
+
+        mag_vector = (bx, by, bz)
+        if np.linalg.norm(mag_vector) == 0.0:
+            raise ValueError(
+                "The magnetic field must have a magnitude greater" " than 0."
+            )
```
darcangelomauro, line 359: Is it worth considering imposing a magnitude greater than a tolerance instead of different from zero? Something like `1e-8`?

pulser/simulation/simulation.py
```diff
@@ -82,17 +85,10 @@ def __init__(
...
-                "No instructions given for the channels in the " "sequence."
-            )
-        not_supported = (
-            set(ch.basis for ch in sequence._channels.values())
-            - SUPPORTED_BASES
-        )
-        if not_supported:
-            raise NotImplementedError(
-                "Sequence with unsupported bases: " + "".join(not_supported)
+                "No instructions given for the channels in the sequence."
```
darcangelomauro, line 88: This still allows a sequence with only a delay in it and no pulses. Could we use `_empty_sequence` instead since we have it?

pulser/simulation/simulation.py
```diff
@@ -139,6 +136,12 @@ def set_config(self, cfg: SimConfig) -> None:
         """
         if not isinstance(cfg, SimConfig):
             raise ValueError(f"Object {cfg} is not a valid `SimConfig`.")
+        not_supported = set(cfg.noise) - SUPPORTED_NOISE[self._interaction]
```
darcangelomauro, line 139: What if instead of creating the new variable `not_supported` every time, we check `set(cfg.noise).issubset(SUPPORTED_NOISE[self._interaction])`, and compute the difference only when the error is raised?

pulser/simulation/simulation.py
```diff
@@ -192,6 +194,13 @@ def add_config(self, config: SimConfig) -> None:
         if not isinstance(config, SimConfig):
             raise ValueError(f"Object {config} is not a valid `SimConfig`")
 
+        not_supported = set(config.noise) - SUPPORTED_NOISE[self._interaction]
```
darcangelomauro, line 197: Same as before

pulser/simulation/simulation.py
```diff
@@ -442,23 +459,54 @@ def write_samples(
...
+
+        if not isinstance(operations, list):
+            operations = [operations]
+
+        for operator, qubits in operations:
+            if qubits == "global":
+                return sum(
+                    self.build_operator([(operator, [q_id])])
+                    for q_id in self._qdict
+                )
```
darcangelomauro, line 494: Right now it's safe because the second argument in the tuple `(operator, [q_id])` is a list. If in the future `build_operator` is changed to accept the name of a single qubit as the second argument, there could be a potential infinite recursion if for some reason the qubit is called "global". Just a note

pulser/simulation/simulation.py
```diff
@@ -442,23 +459,54 @@ def write_samples(
...
+                if isinstance(operator, str):
+                    try:
+                        operator = self.op_matrix[operator]
+                    except KeyError:
+                        raise ValueError(f"{operator} is not a valid operator")
+                for qubit in qubits:
+                    try:
+                        k = self._qid_index[qubit]
+                    except KeyError:
+                        raise ValueError(f"{qubit} is not a valid qubit")
```
darcangelomauro, line 507: Perhaps it's worth checking whether `set(qubits).issubset(set(self._qdict.keys()))` at the beginning, to avoid ever entering in the loop in the first place if there's a bad qubit

pulser/simulation/simulation.py
```diff
@@ -604,19 +702,21 @@ def build_coeffs_ops(basis: str, addr: str) -> list[list]:
...
+            qobj_list.append(make_interaction_term())
 
         # Time dependent terms:
         for addr in self.samples:
             for basis in self.samples[addr]:
                 if self.samples[addr][basis]:
                     qobj_list += cast(list, build_coeffs_ops(basis, addr))
 
+        if not qobj_list:  # If qobj_list ends up empty
+            qobj_list = [0 * self.build_operator([("I", "global")])]
```
darcangelomauro, line 718: I think this might be an unnecessary overhead, especially because it cycles over every qubit. Can something simpler like `qutip.Qobj(shape=(n, n))` be used instead?

darcangelomauro (changes requested):
Just some minor comments, I couldn't see anything major

pulser/devices/_device_datacls.py
```diff
@@ -47,7 +51,9 @@ class Device:
     max_radial_distance: int
     min_atom_distance: int
     _channels: tuple[tuple[str, Channel], ...]
+    # Ising interaction coeff
     interaction_coeff: float = 5008713.0
+    interaction_coeff_xy: float = 3700.0
```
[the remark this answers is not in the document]
HGSilveri: Sure, it would just need changes that are not relevant here.

pulser/sequence.py
```diff
@@ -299,6 +320,48 @@ def current_phase_ref(
 
         return self._phase_ref[basis][qubit].last_phase
 
+    def set_magnetic_field(
```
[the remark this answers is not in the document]
HGSilveri: Yes, but I usually don't like setter that much. It either makes the user think that they are changing an attribute directly or has them guess that they can do it. I find this formulation clearer

pulser/sequence.py
HGSilveri: You mean the `if`, right? It's not necessary, actually. I just thought it would be a bit cleaner and faster, which is true but completely negligible. Would you advocate for changing it to just setting `self._empty_sequence = False` every time?

pulser/sequence.py
HGSilveri: Sure, but I'm not sure what that tolerance will be (or even if it will depend on the device) and I didn't want to make up a value. It's a good idea though, it would make more sense that way

pulser/simulation/simulation.py
HGSilveri: I think t's not an issue for the simulation to have a sequence without pulses, it just needs something in the channels. I agree it's not very useful, but does it warrant forbiding it?

pulser/simulation/simulation.py
HGSilveri: I think it's a trade-off here: when you fail, you already have the noises you don't support, so you save on computation time there. In the other case, indeed it is a bit more expensive, though I would venture guessing it's negligible. Considering that this is not called often, is it worth the potential optimization?

pulser/simulation/simulation.py
HGSilveri: True, it is a risk, but it will exist for whatever name we choose here, so it would have to be done differently. I don't see a reason why we would change the signature, so I think this will be fine

pulser/simulation/simulation.py
HGSilveri: I agree, but I'm not sure that exists. Plus, bear in mind this will only be run for an empty sequence, which is the same as saying "almost never"

pulser/tests/test_simulation.py
```diff
@@ -474,7 +501,7 @@ def test_config():
 
 def test_noise():
     sim2 = Simulation(
-        seq, sampling_rate=0.01, config=SimConfig(noise=("doppler"))
+        seq, sampling_rate=0.01, config=SimConfig(noise=("SPAM"), eta=0.4)
     )
     sim2.run()
     with pytest.raises(NotImplementedError, match="Cannot include"):
```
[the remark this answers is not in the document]
HGSilveri: The reason why it's done that way is because `self.basis_name` is defined on the first time `self._construct_hamiltonian()` is called, so that check can only happen afterwards. I don't like it either, but it is not as simple as moving the check upwards, unfortunately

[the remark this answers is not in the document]
HGSilveri: Ok, I patched this one

pulser/simulation/simulation.py
HGSilveri: I agree

Slimane33 (commented):
Not much comment from my side either since I made a large part of the implementation. I still have to get more familiar to the part related to the noise.

pulser/sequence.py
```diff
@@ -299,6 +320,48 @@ def current_phase_ref(
...
+                raise ValueError(
+                    "The magnetic field can only be set in 'XY " "Mode'."
+                )
+            # No channels declared yet
+            self._in_xy = True
+        elif not self._empty_sequence:
+            # Not all channels are empty
+            raise ValueError(
+                "The magnetic field can only be set on an empty " "sequence."
+            )
```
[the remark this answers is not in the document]
HGSilveri: It's just because we wanted it to be clear that it is set before the sequence begins and cannot be changed. If you could do it at the middle of a sequence, it might give the impression that it only turns on then

pulser/sequence.py
darcangelomauro: Ah sorry, I didn't explain myself very well. I agree with the `if`, I wouldn't change it. I was just wondering why we can only set the magnetic field on an empty sequence

HGSilveri: Slimane had the same question just now so I'll quote my reply:

> It's just because we wanted it to be clear that it is set before the sequence begins and cannot be changed. If you could do it at the middle of a sequence, it might give the impression that it only turns on then

pulser/sequence.py
HGSilveri: I would say we keep it like this for now and then define more specific values once we have them

darcangelomauro (approved):
For me it's good to go

PR #271, opened 2021-10-05
2D registers

Louis-PaulHenry:
Created a 3D device, and split the Register class into to child classes, with their respective draw() :
For the 3D register, the default is to output 2 perspective views of the register, but it can be changed to three projection views on the xy, yz and xz planes

### DIFF
--- a/pulser/__init__.py
+++ b/pulser/__init__.py
@@ -18,7 +18,7 @@ from pulser._version import __version__
 
 from pulser.pulse import Pulse
 
-from pulser.register import Register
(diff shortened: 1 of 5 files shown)


Louis-PaulHenry:
replacing it by `ax.autoscale()` does the trick

Louis-PaulHenry:
Capture d’écran 2021-10-11 à 16 18 41

HGSilveri:
@Louis-PaulHenry Did you make changes after my last review? You might have forgot to push them

PR #272, opened 2021-10-06
Added register drawing capabilities to Sequence.draw()

darcangelomauro:
`Sequence.draw()` now accepts a new boolean argument `draw_register` that defaults to `False`.
When `True`, the full register is shown after the pulse sequence. If an SLM mask is set, the masked qubits are shown as red crosses.

There are some points to discuss: 

1. The code is a bit redundant, as the register drawing part is copied from `Register.draw()`. One could think for example to split `Register.draw()` in two methods, the first returning an `ax` that could be used directly into `Sequence.draw()`. On the other hand, `Sequence` has the extra information on the SLM mask which is missing in `Register`, so I couldn't see a way to do that. 
2. Should the register drawing options (like blockade radius) be passed in `Sequence.draw()` as well? Or do we want to fall back on a set of default ones?
3. Should we introduce a flag to choose whether to highlight the masked qubits or not?
4. Related to the previous question: do we want to change the way masked qubits are shown? A sequence can be very long and complicated, with state preparation being just a small part of it. So perhaps showing a red cross is too strong as a visual cue?
5. When the register is drawn, an extra vertical padding between the different `axes` was added to avoid overlaps.

See attached image for the drawing result, where qubit `"q0"` and `"q10"` are masked.

reg_draw

### DIFF
--- a/pulser/_seq_drawer.py
+++ b/pulser/_seq_drawer.py
@@ -20,11 +20,13 @@ from typing import Any, cast, Optional, Union
 import matplotlib.pyplot as plt
 import numpy as np
 from scipy.interpolate import CubicSpline
+from itertools import combinations
 
 import pulser
 from pulser.waveforms import ConstantWaveform, InterpolatedWaveform
 from pulser.pulse import Pulse
 from pulser.channels import Channel
+from pulser import Register, Register3D
 
 
 def gather_data(seq: pulser.sequence.Sequence) -> dict:
@@ -152,45 +154,53 @@ def draw_sequence(
     area_ph_box = dict(boxstyle="round", facecolor="ghostwhite", alpha=0.7)
     slm_box = dict(boxstyle="round", alpha=0.4, facecolor="grey", hatch="//")
 
+    # Register related
+    pos = np.array(seq._register._coords)
+
+    if draw_register:
+        if isinstance(seq._register, Register3D):
+            labels = "xyz"
+            fig_reg, axes_reg = seq._register._initialize_fig_axes_projection(
+                pos,
+            )
+
+            for ax_reg, (ix, iy) in zip(axes_reg, combinations(np.arange(3), 2)):
+                seq._register._draw_2D(
+                    ax=ax_reg,
+                    pos=pos,
+                    ids=seq._register._ids,
+                    plane=(ix,iy),
+                    masked_qubits=seq._slm_mask_targets,
+                )
+                ax_reg.set_title(
+                    "Masked register projected onto\n the "
+                    + labels[ix]
+                    + labels[iy]
+                    + "-plane"
+                )
+
+        elif isinstance(seq._register, Register):
+            fig_reg, ax_reg = seq._register._initialize_fig_axes(pos)
+            seq._register._draw_2D(
+                ax=ax_reg,
+                pos=pos,
+                ids=seq._register._ids,
+                masked_qubits=seq._slm_mask_targets,
+            )
+            ax_reg.set_title("Masked register", pad=10)
+
     fig = plt.figure(
         constrained_layout=False,
-        figsize=(20, 4.5 * n_channels + 10 * int(draw_register)),
+        figsize=(20, 4.5 * n_channels),
     )
     gs = fig.add_gridspec(
-        n_channels + int(draw_register),
+        n_channels,
         1,
-        hspace=0.075 + 0.027 * int(draw_register),
+        hspace=0.075,
     )
-
-    # Draw the register
-    if draw_register:
-        big_ax = fig.add_subplot(gs[0])
-        big_ax.set_ylabel("Masked register", labelpad=40, fontsize=18)
-        big_ax.spines["top"].set_color("none")
-        big_ax.spines["bottom"].set_color("none")
-        big_ax.spines["left"].set_color("none")
-        big_ax.spines["right"].set_color("none")
-        big_ax.tick_params(
-            labelcolor="w", top=False, bottom=False, left=False, right=False
-        )
-
-        subgs = gs[0].subgridspec(1, 2)
-        ax = fig.add_subplot(subgs[0, 0])
-        seq._register._draw_2D(
-            ax=ax,
-            pos=np.array(seq._register._coords),
-            ids=seq._register._ids,
-            masked_qubits=seq._slm_mask_targets,
-        )
 
     ch_axes = {}
-    channel_plots = (
-        {} if not draw_register else {"dummy": cast(Channel, "dummy")}
-    )
-    channel_plots.update(seq._channels)
-    for i, (ch, gs_) in enumerate(zip(channel_plots, gs)):
-        if ch == "dummy":
-            continue
+    for i, (ch, gs_) in enumerate(zip(seq._channels, gs)):
         ax = fig.add_subplot(gs_)
         ax.spines["top"].set_color("none")
         ax.spines["bottom"].set_color("none")
--- a/pulser/_seq_drawer.py
+++ b/pulser/_seq_drawer.py
@@ -25,7 +25,6 @@ from itertools import combinations
 import pulser
 from pulser.waveforms import ConstantWaveform, InterpolatedWaveform
 from pulser.pulse import Pulse
-from pulser.channels import Channel
 from pulser import Register, Register3D
 
 
@@ -154,9 +153,9 @@ def draw_sequence(
     area_ph_box = dict(boxstyle="round", facecolor="ghostwhite", alpha=0.7)
     slm_box = dict(boxstyle="round", alpha=0.4, facecolor="grey", hatch="//")
 
-    # Register related
     pos = np.array(seq._register._coords)
 
+    # Draw masked register
     if draw_register:
         if isinstance(seq._register, Register3D):
             labels = "xyz"
@@ -164,12 +163,14 @@ def draw_sequence(
                 pos,
             )
 
-            for ax_reg, (ix, iy) in zip(axes_reg, combinations(np.arange(3), 2)):
+            for ax_reg, (ix, iy) in zip(
+                axes_reg, combinations(np.arange(3), 2)
+            ):
                 seq._register._draw_2D(
                     ax=ax_reg,
                     pos=pos,
                     ids=seq._register._ids,
-                    plane=(ix,iy),
+                    plane=(ix, iy),
                     masked_qubits=seq._slm_mask_targets,
                 )
                 ax_reg.set_title(
--- a/pulser/_seq_drawer.py
+++ b/pulser/_seq_drawer.py
@@ -111,7 +111,7 @@ def draw_sequence(
     draw_interp_pts: bool = True,
     draw_phase_shifts: bool = False,
     draw_register: bool = False,
-) -> Figure:
+) -> tuple[Figure, Figure]:
     """Draws the entire sequence.
 
     Args:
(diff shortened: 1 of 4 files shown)


HGSilveri:
You raise very valid points. Here are my answers:

> 1. The code is a bit redundant, as the register drawing part is copied from `Register.draw()`. One could think for example to split `Register.draw()` in two methods, the first returning an `ax` that could be used directly into `Sequence.draw()`. On the other hand, `Sequence` has the extra information on the SLM mask which is missing in `Register`, so I couldn't see a way to do that.

I agree, and I think it's avoidable. In the same way that is done in `Waveform` for example, we can split the `draw` in an private method `_plot` which creates the figure object and returns it, but does not show it. Since this is a private method, we can also add an optional argument like `masked` that takes the masked qubits and draws the differently already.

> 2. Should the register drawing options (like blockade radius) be passed in `Sequence.draw()` as well? Or do we want to fall back on a set of default ones?

I would go with the default here.

> 3. Should we introduce a flag to choose whether to highlight the masked qubits or not?

I don't think so, this feature is pretty redundant if you're not masking the qubits.

> 4. Related to the previous question: do we want to change the way masked qubits are shown? A sequence can be very long and complicated, with state preparation being just a small part of it. So perhaps showing a red cross is too strong as a visual cue?

I agree, I would maybe just change their color to grey or something along those lines.


Also, I would maybe draw the Register first, and put a title on it or legend for the masked qubits.

darcangelomauro:
I've overwritten this branch because the code was rewritten using the recent changes introduced by #271.

A couple of comments:

- The mask is now shown as a square halo around the qubits
- The masked register is drawn before the pulse sequence. This change in particular required using some trickery (lines 187-193 in `_seq_drawer.py`). Perhaps there's an easier solutions which I'm not seeing.

See attached for an example

reg_draw_new_overlap
reg_draw_new

darcangelomauro:
Right now `Sequence` requires a `Register`, not a `BaseRegister`. This behaviour will be changed in another pull request. After that is done, support for drawing the mask on 3D registers will be added here

pulser/_seq_drawer.py
```diff
@@ -122,6 +125,9 @@ def draw_sequence(
             top of the respective waveforms (defaults to True).
         draw_phase_shifts (bool): Whether phase shift and reference information
             should be added to the plot, defaults to False.
+        draw_register (bool): Whether to draw the register after the pulse
```
HGSilveri, line 128: I guess now it's `before the pulse sequence`?

pulser/register.py
```diff
@@ -612,12 +629,34 @@ def rotate(self, degrees: float) -> None:
         )
         self._coords = [rot @ v for v in self._coords]
 
+    def _initialize_fig_axes(
+        self,
+        pos: np.ndarray,
+        blockade_radius: Optional[float] = None,
+        draw_half_radius: bool = False,
+    ) -> Tuple[plt.figure.Figure, plt.axes.Axes]:
```
HGSilveri, line 637: You don't need this `Tuple`,  you can use `tuple`

pulser/register.py
```diff
@@ -115,7 +132,7 @@ def _draw_2D(
         if with_labels:
             # Determine which labels would overlap and merge those
             plot_pos = list(pos[:, (ix, iy)])
-            plot_ids = [f"{i}" for i in ids]
+            plot_ids = [f"{i}" + "(m)" * (i in masked_qubits) for i in ids]
```
HGSilveri, line 135: Very compact, cool!

HGSilveri (commented):
This is looking good, how's the issue with the padding coming along?

darcangelomauro:
I've implemented some changes that should fix the overlapping issues.
I've reintroduced `tight_layout`, which #271 had removed because it was creating some discrepancy in the visualization with and without labels.

The reason for the discrepancy was that the bounding box for the qubit labels was set to 50 by overloading `_get_wrap_line_width` (which seems to be the only way to do it, see [this](https://stackoverflow.com/questions/48079364/wrapping-text-not-working-in-matplotlib) discussion). But `tight_layout` apparently didn't know that, and instead calculated the padding by thinking that the text would never wrap.
The solution was to set `tight_layout` before the qubit labels are generated. It seems to be working (see attached).
2d_proj
2d_proj_wl


As for the SLM mask, I've made sure the square halo doesn't go outside the axes by faking a `blockade_radius`, see attached for the resulting plot.
reg_draw

HGSilveri:
Judging from the pictures, I'd say it's looking good! 
The latest PR merged into develop has some conflicts that need to be resolved, please take care of that and then ping me for review.

darcangelomauro:
The last merge to develop allows to save the sequence as an image.

I extended that to the case where the masked register is drawn from a sequence.
In `Sequence.draw` you can specify separately an image name for the sequence and for the register. 

I had to switch from `matplotlib.pyplot.savefig` to `matplotlib.figure.Figure.savefig` because otherwise it would not be possible to save both images.

darcangelomauro:
We decided to change the label of masked qubits from `qubit_name(m)` to `[qubit_name]`.
If the labels of two masked qubits are merged, and the qubits are written consecutively, a single square bracket enclosing both qubits will be displayed (i.e., the label will read `[qubit1, qubit2]`).

In the attached example, qubits 1, 2 and 15 are masked.
sq_brack

HGSilveri:
> Of course this will cause undefined behaviour if the user specifies custom names for the qubits that include square brackets.

Won't they just be brackets inside square brackets? Might look odd, but that's pretty much it, right?

darcangelomauro:
> Won't they just be brackets inside square brackets? Might look odd, but that's pretty much it, right?

I was referring more specifically to the merging of square brackets, when for example `[qubit_1], [qubit_2]` becomes `[qubit_1, qubit_2]`.

The way it is done is by checking that the last character of one label is `]` and the first of the other is `[`, and if that's the case, they are deleted.

HGSilveri:
> > Won't they just be brackets inside square brackets? Might look odd, but that's pretty much it, right?
> 
> I was referring more specifically to the merging of square brackets, when for example `[qubit_1], [qubit_2]` becomes `[qubit_1, qubit_2]`.
> 
> The way it is done is by checking that the last character of one label is `]` and the first of the other is `[`, and if that's the case, ~they are deleted~.
> 
> edit: rather, they are not displayed

Oh I see. Sorry, I didn't look at the code yet. There might be a more robust way of doing it though

pulser/register.py
```diff
@@ -133,7 +150,9 @@ def _draw_2D(
         if with_labels:
             # Determine which labels would overlap and merge those
             plot_pos = list(pos[:, (ix, iy)])
-            plot_ids = [f"{i}" for i in ids]
+            plot_ids = [
+                f"[{i}]" if i in masked_qubits else f"{i}" for i in ids
+            ]
```
HGSilveri, line 155: What if instead you join all masked qubits under the same brackets, irrespectively of the order? E.g; if '0', '1' and '4' are masked the label would be: '2, 3, [0, 1, 4]'

Then you can just group all the masked qubits together.

pulser/sequence.py
```diff
@@ -913,8 +913,11 @@ def draw(
         draw_phase_area: bool = False,
         draw_interp_pts: bool = True,
         draw_phase_shifts: bool = False,
-        fig_name: str = None,
-        kwargs_savefig: dict = {},
+        draw_register: bool = False,
+        fig_name_seq: str = None,
+        fig_name_reg: str = None,
+        kwargs_savefig_seq: dict = {},
+        kwargs_savefig_reg: dict = {},
```
HGSilveri, line 920: This seems a bit excessive... But ok, let's leave it this way.

pulser/tests/test_sequence.py
```diff
@@ -327,8 +327,9 @@ def test_sequence():
     seq.phase_shift(np.pi, "q0", basis="ground-rydberg")
 
     with patch("matplotlib.pyplot.show"):
-        with patch("matplotlib.pyplot.savefig"):
-            seq.draw(fig_name="my_sequence.pdf")
+        with patch("matplotlib.figure.Figure.savefig"):
+            seq.draw(fig_name_seq="my_sequence.pdf")
+            seq.draw(draw_register=True, fig_name_reg="my_register.pdf")
```
HGSilveri, line 332: Did you check this locally? Is it not saving these files?

HGSilveri (approved):
This looks good to me, I guess what I have are suggestions/questions at this points. Let me know what you think

pulser/register.py
darcangelomauro: I guess the problem here is that at that line you still don't know if their label will be merged, because that only happens if they are drawn very close together (like in the case of 2d projections of 3d registers). 
For example if you look at the picture I posted in the previous comment, `[1,2]` are merged in the x-y view but they are not in the x-z and y-z.

pulser/sequence.py
darcangelomauro: In a previous version I left just one `fig_name`, and I saved two images with an added prefix `pulses_` and `register_`. 

If you think it's better I could revert to that.

pulser/tests/test_sequence.py
darcangelomauro: Yes it's not saving them, that's the reason why I changed it from `patch("matplotlib.pyplot.savefig")` to `patch("matplotlib.figure.Figure.savefig")`

pulser/sequence.py
darcangelomauro: Ah right, now that I think about it the reason I separated them is because this way you can choose whether you want to save just the pulses, just the register, or both

pulser/tests/test_sequence.py
HGSilveri: Sure, I just wanted to make sure because `patch` is a bit finicky to work with sometimes

pulser/sequence.py
HGSilveri: Yeah, I would go for that, just detail it in the docstring. Also share the `kwargs`, I guess

pulser/register.py
HGSilveri: Well yeah, you wouldn't be doing it here exactly. When you're deciding which qubits go in a label, you also check whether they are masked. By the end you have a group of masked and unmasked and you can compose your label. Doesn't this work?

pulser/sequence.py
HGSilveri: > Ah right, now that I think about it the reason I separated them is because this way you can choose whether you want to save just the pulses, just the register, or both

Sorry, I didn't see this before replying. Since they are saved separately, that doesn't sound so bad to me.

darcangelomauro:
I've implemented the requested changes. The register drawing now looks like the attached image, and there should be no conflict with qubit names.

sq_brack

pulser/register.py
```diff
@@ -150,9 +150,7 @@ def _draw_2D(
         if with_labels:
             # Determine which labels would overlap and merge those
             plot_pos = list(pos[:, (ix, iy)])
-            plot_ids = [
-                f"[{i}]" if i in masked_qubits else f"{i}" for i in ids
-            ]
+            plot_ids: list[Union[list, str]] = [[f"{i}"] for i in ids]
```
HGSilveri, line 153: The type here is now `list[list[str]]`, right?

HGSilveri (commented):
Cool! Final comment

pulser/register.py
darcangelomauro: Yes, but later on the sublists become strings. If I don't force the `list[Union[list, str]]` type then mypy complains

HGSilveri: Ok, roger that

PR #273, opened 2021-10-08
Minimizing number of evolution simulations for specific noise configurations

### LINKED ISSUE #221: Make noisy simulation more efficient under specific circumstances
PR #199 added the option of including several types of noise to the simulation. 

When noise is specified, the current implementation generates a new Hamiltonian before each run, which allows us to introduce random noise into it. However, such random noise is only strictly necessary when either `doppler` and/or `amplitude` noise are involved, as those are the two types of noise that are simulated through small, random changes to the Hamiltonian. 

If neither `amplitude` nor `doppler` noise are involved, then there are a few modifications that can be done to speed up the current implementation. Separating by noise type:
- `dephasing`: The case where this is the only noise is treated separately already.
- `SPAM`:
    - Preparation errors: These do actually change the Hamiltonian on each run, but the different initial configurations can be computed beforehand. For small values, most runs will have a perfectly prepared register, which means they will have the same hamiltonian. The idea here would be to make one Hamiltonian per initial configuration, simulate the unitary evolution of each Hamiltonian, sample each one according to the number of times they occur and sum the samples.
    - Measurement errors: These do not affect the Hamiltonian and can instead be added in post-processing. All samples can be extracted from a single `CoherentResults` instance (see #220).

HGSilveri:
Closes #221 .


HGSilveri:
Ok, onto this one @sebgrijalva :pray:

sebgrijalva:
> Ok, onto this one @sebgrijalva 🙏

Already reading the code!

sebgrijalva (approved):
Nice! This makes a lot of sense, thank you.

PR #276, opened 2021-10-15
Add the possibility to save figures of register and sequences

Slimane33:
This PR aims at adding the possibility for the user to save figures from the drawing of sequences and register. I tried 2 ways, but at the end we should only use one in order to be consistent:
- for the sequence, creating a new function `savefig`
- for the register, adding the option in `Register.draw`


HGSilveri:
Take note of the changes freshly merged from #271.

pulser/_seq_drawer.py
```diff
@@ -446,4 +447,4 @@ def phase_str(phi: float) -> str:
                 pts = np.array(all_points["detuning"])
                 b.scatter(pts[:, 0], pts[:, 1], color="indigo")
 
-    plt.show()
+    return fig
```
HGSilveri, line 450: Is this still necessary, in the end?

pulser/register.py
```diff
@@ -18,11 +18,12 @@
 from abc import ABC, abstractmethod
 from collections.abc import Mapping, Iterable
 from collections.abc import Sequence as abcSequence
-import matplotlib.pyplot as plt
+
 from matplotlib import collections as mc
 import numpy as np
 from numpy.typing import ArrayLike
 from scipy.spatial import KDTree
+import matplotlib.pyplot as plt
```
HGSilveri, line 26: nit: Could move this one next to the `matplotlib` import?

pulser/register.py
```diff
@@ -18,11 +18,12 @@
...
 from typing import Any, cast, Optional, Union, TypeVar, Type
 from itertools import combinations
```
HGSilveri, line 28: Hum, these shouldn't be here... I know it's not your fault, but could you please put them on the first block while you're at it?

pulser/sequence.py
```diff
@@ -934,6 +942,9 @@ def draw(
             draw_interp_pts=draw_interp_pts,
             draw_phase_shifts=draw_phase_shifts,
         )
+        if fig_name is not None:
+            plt.savefig(fig_name, **kwargs_savefig)
+        plt.show()
```
HGSilveri, line 947: You'll probably need to do this for `Simulation.draw()` too.

HGSilveri (commented):
Ok, there only a few minor things to take care of now

HGSilveri (approved):
Perfect, thanks a lot @Slimane33 !

PR #279, opened 2021-11-03
Custom labels

Louis-PaulHenry:
Minor fix : `Register.from_coordinates()`now has an additional optional argument to specify custom labels on each qubit. This change is applied to the `to_2D()`method

### DIFF
--- a/pulser/register.py
+++ b/pulser/register.py
@@ -17,6 +17,7 @@ from __future__ import annotations
 
 from abc import ABC, abstractmethod
 from collections.abc import Mapping, Iterable
+from collections.abc import Sequence as abcSequence
 import matplotlib.pyplot as plt
 from matplotlib import collections as mc
 import numpy as np
@@ -63,7 +64,7 @@ class BaseRegister(ABC):
         coords: np.ndarray,
         center: bool = True,
         prefix: Optional[str] = None,
-        labels: Optional[ArrayLike] = None,
+        labels: Optional[abcSequence[QubitId]] = None,
     ) -> T:
         """Creates the register from an array of coordinates.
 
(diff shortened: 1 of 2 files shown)


pulser/register.py
```diff
@@ -85,6 +88,14 @@ def from_coordinates(
         if prefix is not None:
             pre = str(prefix)
             qubits = {pre + str(i): pos for i, pos in enumerate(coords)}
+            if labels is not None:
+                raise NotImplementedError(
+                    "It is impossible to specify a prefix and "
+                    "a set of labels at the same time"
+                )
+
+        elif labels is not None:
```
HGSilveri, line 97: Maybe check that the lengths match?

pulser/register.py
```diff
@@ -826,7 +837,7 @@ def to_2D(self, tol_width: float = 0.0) -> Register:
             coords_2D = np.array(
                 [np.array([e_x.dot(r), e_y.dot(r)]) for r in coords]
             )
-            return Register.from_coordinates(coords_2D, prefix=prefix)
+            return Register.from_coordinates(coords_2D, labels=labels)
```
HGSilveri, line 840: I guess you can go directly with `labels=self._ids` then.

pulser/register.py
```diff
@@ -63,6 +63,7 @@ def from_coordinates(
         coords: np.ndarray,
         center: bool = True,
         prefix: Optional[str] = None,
+        labels: Optional[ArrayLike] = None,
```
HGSilveri, line 66: I think instead of `ArrayLike`, we can be a bit stricter and make it a `collections.abc.Sequence[QubitId]` (I know, it's ugly but it's the way we got not to confuse it with `pulser.Sequence`)

HGSilveri (commented):
Looks good! I just have a couple remarks

pulser/register.py
Louis-PaulHenry: That a very reasonable idea! ^^

pulser/register.py
Louis-PaulHenry: I'll try that

darcangelomauro (approved):
Looks good to me

Issue #249, opened 2021-08-16
Classical simulation does not allow simulation time equal to the duration of the pulse


Adrien-Kahn:
When simulating a 4ns sequence, `res.states` only contains 4 states, including the initial state. However, since 4 values for the Hamiltonian were provided, the initial state could be updated 4 times, so that we would have a total of 5 states in `res.states` with the initial state.

Manually setting the `evaluation_times` of the `simulation` does not help because the setter checks that the largest time is smaller than `self.times[-1]` which is equal to the duration of the sequence minus 1.

Adrien-Kahn:
Here is a minimal working example:
```
reg = Register.rectangle(1,2)
p = Pulse.ConstantPulse(4, 40, 80, 0)
seq = Sequence(reg, devices.MockDevice)
seq.declare_channel('ryd','rydberg_local', 0)
seq.add(p1, 'ryd')

simtest = Simulation(seqtest)
restest = simtest.run()

# Overlap with |gg>
print([abs(state.overlap(q.basis(4,3)))**2 for state in restest.states])
```

Output:
```
[1.0, 0.9996002665950946, 0.998404263138335, 0.9964215525372951]
```

The first value corresponds to the initial state, so we are only given 3 updated states, while 4 values for the Hamiltonian were provided.

sebgrijalva:
While implementing these changes, there appeared some issues related to the interpolation of the call method of the `QobjEvo`. I'm still working on this, but it will take longer than expected.

PR #280, opened 2021-11-03
Change evaluation times to include the final time.

sebgrijalva:
This addresses #249.


sebgrijalva:
For the moment I changed the way `Simulation.evaluation_times` is assigned. Notice also that `Simulation._times` is changed to `Simulation._sampling_indexes` (although it keeps the factor of 1/1000 so far), since these indexes are used to write the samples from `Sequence`

HGSilveri:
> For the moment I changed the way `Simulation.evaluation_times` is assigned. Notice also that `Simulation._times` is changed to `Simulation._sampling_indexes` (although it keeps the factor of 1/1000 so far), since these indexes are used to write the samples from `Sequence`

Wouldn't `sampling_times` be more appropriate than `sampling_indexes`?

sebgrijalva:
> > For the moment I changed the way `Simulation.evaluation_times` is assigned. Notice also that `Simulation._times` is changed to `Simulation._sampling_indexes` (although it keeps the factor of 1/1000 so far), since these indexes are used to write the samples from `Sequence`
> 
> Wouldn't `sampling_times` be more appropriate than `sampling_indexes`?

So far what's happening is that `Simulation._sampling_times` stores the indexes from the Sequence that are consistent with the sampling rate and that will be used to construct the `Qutip.QobjEvo` hamiltonian... and then divides by 1000, so it's Ok to call them "times" in microseconds.

Then, the solver itself is using a `tlist` consisting on an array from 0 to `_tot_duration` (now including it) mutliplied by the sampling rate. This seems more adapted to be called a "time", although it loses connection with the Sequence samples.

Finally, the evaluation times are being fed to the `SimResults` object to give the output that the user wanted (e.g. if "Minimal", gives two states, two expectation values, etc.)

sebgrijalva:
I see that some notebooks are being affected by these changes, most likely because they use `Simulation._times`at some point. I leave this note here for myself when I come back to this

pulser/simulation/simulation.py
```diff
@@ -886,15 +887,25 @@ def _run_solver() -> CoherentResults:
                 result = qutip.mesolve(
                     liouvillian,
                     self.initial_state,
-                    self._eval_times_array,
+                    np.linspace(
+                        0,
+                        self._tot_duration,
+                        int(self._sampling_rate * self._tot_duration) + 1,
+                    )
+                    / 1000,
```
HGSilveri, line 895: I don't get this one... Why not keep using `self._eval_times_array`?

pulser/tests/test_simresults.py
```diff
@@ -210,7 +219,9 @@ def test_expect():
     sim_single.set_config(config)
     sim_single.evaluation_times = "Minimal"
     results_single = sim_single.run()
+    print("states after run are:", results_single.states)
     exp = results_single.expect(op)[0]
+    print("expect results are", exp)
```
HGSilveri, line 224: I guess you can remove these

pulser/tests/test_simresults.py
```diff
@@ -284,22 +294,30 @@ def test_sample_final_state():
 
 
 def test_sample_final_state_noisy():
+    # Sample final state of noisy simulation and check resulting Counter:
+    print(results_noisy.states[-1])
+    print(results_noisy.get_final_state())
```
HGSilveri, line 299: And these... The `print` statements in general

HGSilveri (commented):
I guess I'm a bit puzzled about some of the changes...

sebgrijalva:
Closing this for a fresh start on #282

PR #281, opened 2021-11-16
Changing rydberg level

LucasGitQ:


HGSilveri (changes requested):
Some initial comments

pulser/devices/interaction_coefficients/__init__.py
```diff
@@ -0,0 +1,19 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""C6 coefficients for Rydberg levels between 50 and 100"""
+
+import json
+from pathlib import PurePath
+
+C6_dict = json.load(open(PurePath(__file__).parent / "C6_coeffs.json"))
```
HGSilveri, line 19: It sucks a bit that the keys are `str`, you could maybe make a new one where they are `int` instead. Something like this:
```suggestion
_json_dict = json.load(open(PurePath(__file__).parent / "C6_coeffs.json"))
C6_dict = {int(key): value for key, value in _json_dict.items()}
```

pulser/devices/_device_datacls.py
```diff
@@ -54,30 +53,14 @@ class Device:
     max_radial_distance: int
     min_atom_distance: int
     _channels: tuple[tuple[str, Channel], ...]
-    rydberg_level: int = 60
+    rydberg_level: int
```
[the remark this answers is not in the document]
HGSilveri: For example, right now there is no `rydberg_level` for `Chadoq2`, so there isn't an `interaction_coeff` either.

HGSilveri (changes requested):
It's quite important to make sure the units are matching. `interaction_coeff` is not `C6`, it's `C6/hbar`

pulser/devices/_device_datacls.py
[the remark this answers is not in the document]
HGSilveri: > For example, right now there is no `rydberg_level` for `Chadoq2`, so there isn't an `interaction_coeff` either.

Nevermind, I see that it was previously defined.

pulser/tests/test_devices.py
```diff
@@ -43,10 +45,12 @@ def test_init():
 
 def test_mock():
     dev = pulser.devices.MockDevice
-    assert dev.dimensions == 3
+    assert dev.dimensions == 2
```
HGSilveri, line 48: It's actually `dev.dimensions == 3` now, this is here because your branch way outdated with respect to `develop`.

LucasGitQ: Oups sorry

pulser/devices/interaction_coefficients/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""C_6/hbar (in  um^6 / us`), coeffs for Rydberg levels between 50 and 100."""
+
+import json
+from pathlib import PurePath
+
+_json_dict = json.load(open(PurePath(__file__).parent / "C6_coeffs.json"))
+C6_dict = {int(key): value for key, value in _json_dict.items()}
```
HGSilveri, line 20: nit: This is a style thing, we shouldn't name a variable with an upper-case letter (sorry, I just remembered this now). Let's go with `c6_dict` instead

pulser/devices/_device_datacls.py
```diff
@@ -37,23 +38,21 @@ class Device:
         max_radial_distance: The furthest away an atom can be from the center
             of the array (in μm).
         min_atom_distance: The closest together two atoms can be (in μm).
-        interaction_coeff: :math:`C_6/\hbar`
-            (in :math:`\mu m^6 / \mu s`),
-            which sets the van der Waals interaction strength between atoms in
-            the same Rydberg state.
+        rybderg level : The value of the principal quantum number n
```
HGSilveri, line 41:
```suggestion
        rybderg_level : The value of the principal quantum number n
```

pulser/devices/interaction_coefficients/__init__.py
HGSilveri: Reminder to do this too

HGSilveri (commented):
Awesome, all tests passed! Only tiny things to take care of now

HGSilveri (approved):
Great, thanks a lot!

Issue #30, opened 2020-11-30
Dataclasses vs. ABC + property


karalekas:
Reading through the source, I've noticed some places that I think would be better served by using [dataclasses](https://docs.python.org/3/library/dataclasses.html), which are part of the Python standard library as of py37. For example, see the following definition of a `PasqalDevice`:

```python
class PasqalDevice(ABC):
    def __init__(self, qubits):
        ...

    @property
    @abstractmethod
    def name(self):
        """The device name."""
        pass

    @property
    @abstractmethod
    def max_dimensionality(self):
        """Whether it works at most with a 2D or 3D array (returns 2 or 3)."""
        pass

    @property
    @abstractmethod
    def max_atom_num(self):
        """Maximum number of atoms that can be simultaneously trapped."""
        pass

    @property
    @abstractmethod
    def max_radial_distance(self):
        """Maximum allowed distance from the center of the array."""
        pass

    @property
    @abstractmethod
    def min_atom_distance(self):
        """Minimal allowed distance of atoms in the trap (in um)."""
        pass

    @property
    @abstractmethod
    def channels(self):
        """Channels available on the device."""
        pass
```

The properties can be much more succinctly defined using a dataclass, as follows:

```python
from dataclasses import dataclass
from typing import Dict

@dataclass
class PQDevice:
    name: str
    max_dimensionality: int
    max_atom_num: int
    max_radial_distance: int
    min_atom_distance: int
    channels: Dict[str, Channel]
```

There are a couple things that are missing from my dataclass implementation, namely the input parsing and validation that goes into `__init__.py`, and the definitions of the `supported_bases` and `qubits` properties, but these can all be managed by implementing a simple [`__post_init__`](https://www.geeksforgeeks.org/data-classes-in-python-set-5-post-init/) magic method.

As for the ABC part, now instead of defining each device as a subclass of `PQDevice`, each device would simply be an instantiation (object) of class `PQDevice`:

```python
Chadoq2 = PQDevice(
    name = "Chadoq2",
    max_dimensionality = 2,
    max_atom_num = 100,
    max_radial_distance = 50,
    min_atom_distance = 4,
    channels = {
            "rydberg_global": Rydberg.Global(50, 2.5),
            "rydberg_local": Rydberg.Local(50, 10, 100),
            "rydberg_local2": Rydberg.Local(50, 10, 100),
            "raman_local": Raman.Local(50, 10, 100),
        },
)
```

[...]

PR #282, opened 2021-11-24
Evaluation Times and Sampling Times

sebgrijalva:
I'm restarting this PR. Related to #280

I included a (for now private) `Simulation._eval_times_instruction` attribute in case it might be useful.


HGSilveri:
What is (or is going to be) different here with respect to #280?

pulser/simulation/simulation.py
```diff
@@ -106,7 +106,7 @@ def __init__(
         self._sampling_rate = sampling_rate
         self._qid_index = {qid: i for i, qid in enumerate(self._qdict)}
         self._collapse_ops: list[qutip.Qobj] = []
-        self._times = self._adapt_to_sampling_rate(
+        self.sampling_times = self._adapt_to_sampling_rate(
```
HGSilveri, line 109: So we're not going with `_sampling_times` so the user can have access to them, is that right?

HGSilveri (commented):
What's here looks good, but is it all there is?

sebgrijalva:
> What is (or is going to be) different here with respect to #280?

Mostly restarting the changes done in the SimResults class because I misunderstood the tlist from QuTip.

pulser/simulation/simulation.py
sebgrijalva: Yes, although it could be confusing. Notice this doesn't include the final instant, because it's just the indices from the pulse sequence divided by 1000 and adapted to sampling rate.

sebgrijalva:
I will continue with the CI requirements and try to catch any inconsistencies with the previous code (so far there's already a mistake when the QObjEvo is being constructed...)

sebgrijalva:
There's a typing error caused by calling the setter of `evaluation_times` during initialization with `Union[str, ArrayLike, float]` , since the variable was defined as `numpy.ndarray`. Maybe there's a better way to define things, but ignoring this typing (and the types are checked manually to create `eval_times_array` anyway), all tests run.

HGSilveri (approved):
This looks good, do you want me to take a look at the failing tutorials?

sebgrijalva:
> This looks good, do you want me to take a look at the failing tutorials?

Yes, thanks a lot!

tutorials/applications/Building 1D Rydberg Crystals.ipynb
```diff
@@ -74,7 +74,7 @@
      "name": "stdout",
```
HGSilveri, line 74: <pre class="ql-syntax" spellcheck="false">Line #30.    simul = Simulation(seq, sampling_rate=0.6)
</pre><p>I don't know why, but this was giving me an <code>ODEError</code> with the previous sampling_rate, so I had to increase it</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pasqal-io/Pulser/pull/282/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='38'/>

sebgrijalva: Yes, that's the tricky part about that tutorial, since the atoms are so close (in a `MockDevice`) , the terms in the hamiltonian can get really big, so one may have to play with the sampling rate and/or the nsteps

Issue #263, opened 2021-09-08
Change the way we modify attributes in frozen dataclasses


darcangelomauro:
Frozen dataclasses do not allow class attributes to be modified.
However, sometimes in the code we need to override that behaviour.

For example, the `Device` dataclass is frozen:
```python
@dataclass(frozen=True, repr=False)
class Device:
```

but `__post_init__` modifies an attribute:
```python
def __post_init__(self) -> None:
        # Hack to override the docstring of an instance
        self.__dict__["__doc__"] = self._specs(for_docs=True)
```

Following [this](https://stackoverflow.com/questions/60699202/python-frozen-dataclass-allow-changing-of-attribute-via-method) discussion of Stack Overflow, there could be an argument for changing the way this is done.

The first consideration is that it might be preferable to never modify an attribute in a frozen class, and instead replace the object with a new instance.

The second consideration is that, when necessary, modifying an attribute of a frozen class should be done with `object.__setattr__` instead of `self.__dict__`  (for technical reasons, see link above).

PR #283, opened 2021-11-24
Changed the way attributes of frozen classes are modified

darcangelomauro:
See issue #263 for details


pulser/parametrized/variable.py
```diff
@@ -63,8 +63,9 @@ def variables(self) -> dict[str, Variable]:
         return {self.name: self}
 
     def _clear(self) -> None:
-        self.__dict__["value"] = None
-        self.__dict__["_count"] += 1
+        object.__setattr__(self, "value", None)
+        old_count = object.__getattribute__(self, "_count")
+        object.__setattr__(self, "_count", old_count + 1)
```
HGSilveri, line 68: Did this not work?
```suggestion
        object.__setattr__(self, "_count", self._count + 1)
```

pulser/pulse.py
```diff
@@ -87,8 +87,10 @@ def __post_init__(self) -> None:
                 "greater than or equal to zero."
             )
 
-        self.__dict__["phase"] %= 2 * np.pi
-        self.__dict__["post_phase_shift"] %= 2 * np.pi
+        old_phase = object.__getattribute__(self, "phase")
+        old_pps = object.__getattribute__(self, "post_phase_shift")
```
HGSilveri, line 91: I think getting the attributes normally (e.g. `self.phase`) is not blocked, so I think this is not necessary.

pulser/simulation/simconfig.py
```diff
@@ -93,11 +93,15 @@ class SimConfig:
 
     def __post_init__(self) -> None:
         self._process_temperature()
-        self.__dict__["spam_dict"] = {
-            "eta": self.eta,
-            "epsilon": self.epsilon,
-            "epsilon_prime": self.epsilon_prime,
-        }
+        object.__setattr__(
```
HGSilveri, line 96: I'm thinking that on cases where you do this a lot, it might make sense to make a private method that "hides away" the ugliness, something like:
```
def _change_value(self, name, new_value):
    object.__setattr__(self, name, new_value)
```
What do you think?

HGSilveri (commented):
It works fine, let me know what you think of my suggestions

pulser/simulation/simconfig.py
darcangelomauro: I guess in `simconfig.py` and `variable.py` can make sense. Or would you add it globally?

darcangelomauro: Actually though I'm not sure..It's not like the syntax or the line length changes much

HGSilveri: It's true, but it prevents other users to have to know this weird thing. I would suggest doing it on the classes that have potential to grow, which would be just `SimConfig`.

darcangelomauro: sounds good, I've made the change let me know if it's okay

HGSilveri (approved):
LGTM, thank you!

PR #291, opened 2021-11-29
Generalizing XY mode simulation for 3D

HGSilveri:
Minor changes so that XY mode can be simulated on 3D Registers.


Slimane33:
I think it is enough.

Slimane33 (approved):

HGSilveri:
@Slimane33 Thanks!

PR #304, opened 2022-01-03
[Hotfix] Restricting numpy version to fix qutip incompatibility 

### LINKED ISSUE #303: `numpy` 1.22.0 is not compatible with `qutip`
It seems that the latest release of `numpy` (version 1.22.0, released on 31/12/2021) breaks the compatibility with `qutip`, right on import. Specifically, I'm seeing this error:

```bash
>>> import qutip
  File "/Users/hsilver/opt/miniconda3/envs/pulser-test/lib/python3.8/site-packages/qutip/__init__.py", line 73, in <module>
    (_blas_info() == "OPENBLAS" and platform.system() == 'Darwin')
  File "/Users/hsilver/opt/miniconda3/envs/pulser-test/lib/python3.8/site-packages/qutip/utilities.py", line 430, in _blas_info
    blas_info = config.blas_opt_info
AttributeError: module 'numpy.__config__' has no attribute 'blas_opt_info'
```

I've found that downgrading `numpy` to version 1.21.5 fixes this, so for now I will enforce this in `requirements.txt` and in `setup.py`. I will do this as an hotfix and make a new release, to make sure new `pulser` users are not affected.

HGSilveri:
As detailed in #303 , the latest version of `numpy` is not compatible with `qutip`, so I'm restricting the `numpy` version for now.

Additionally, the latest version of `mypy` also seems to have raised a couple `typing` errors, which i'm fixing here too.

Fixes #303.

### DIFF
--- a/pulser/devices/_device_datacls.py
+++ b/pulser/devices/_device_datacls.py
@@ -15,7 +15,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import Any
+from typing import Any, cast
 
 import numpy as np
 from scipy.spatial.distance import pdist, squareform
@@ -111,7 +111,9 @@ class Device:
         Returns:
             float: The rydberg blockade radius, in μm.
         """
-        return (self.interaction_coeff / rabi_frequency) ** (1 / 6)
+        return cast(
+            float, (self.interaction_coeff / rabi_frequency) ** (1 / 6)
+        )
 
     def rabi_from_blockade(self, blockade_radius: float) -> float:
         """The maximum Rabi frequency value to enforce a given blockade radius.
--- a/pulser/simulation/simulation.py
+++ b/pulser/simulation/simulation.py
@@ -881,7 +881,7 @@ class Simulation:
             if progress_bar is True:
                 p_bar = True
             elif (progress_bar is False) or (progress_bar is None):
-                p_bar = None  # type: ignore
+                p_bar = None
             else:
                 raise ValueError("`progress_bar` must be a bool.")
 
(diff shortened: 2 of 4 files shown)


Codoscope (approved):

pulser/devices/_device_datacls.py
```diff
@@ -111,7 +111,9 @@ def rydberg_blockade_radius(self, rabi_frequency: float) -> float:
         Returns:
             float: The rydberg blockade radius, in μm.
         """
-        return (self.interaction_coeff / rabi_frequency) ** (1 / 6)
+        return cast(
```
Codoscope, line 114: I tried this on my computer and mypy is not returning an error if I omit this cast. However, it's raising the error that there's a redundant cast to float now, on the most commit of develop (e0f7849a1747523e1503f008e5c2a050cbb22086). @HGSilveri

HGSilveri: It seems that there has been a yet another release (0.931) of mypy on the 7th of January. This change was made with 0.930, so that might be why. However, I just updated to 0.931 and didn't get this error you mention. What `mypy` version do you have locally?

In any case, I think it's clear that we have to pick a `mypy` version and stick with it.

Codoscope: I'm using mypy 0.920

HGSilveri: Ok, I confirm that I can replicate the behaviour with that version. The same goes for 0.921.

I didn't like that mypy had me do this cast, so perhaps I'll fix the mypy version to 0.921 (the next one is already 0.930) and revert this change. What do you thinK?

Codoscope: I think it's a bug, because storing the result in an intermediate float variable before returning it works with the 0.931 version. I agree that it would be easier to stick on 0.921 for now and consider later to update if we need new features and the bug has been fixed.

Issue #297, opened 2021-12-08
Create new Device instances


HGSilveri:
Following the changes in #134 and #296, we will add new, more realistic device instances. Specifically:
- `Fresnel`, which will feature only global channels
- `FresnelPlus`, which will also have local channels that will feature the newly added constraints

PR #305, opened 2022-01-06
Buffer times for retargeting and phase jumps

### LINKED ISSUE #296: Add Channel specific buffer times for retargeting and phase jumps
Realistically, hardware channels will require some buffer time to update their output. Two particular instances where that is relevant are:
- Retargeting: Right now, there is a variable retargeting time being enforced based on the time of the last retargeting. In addition to this, it is also possible that some channels take an intrinsic, fixed time to retarget. This time should be added in the presence of a target instruction.
- Phase jumps: The way phase is encoded in a pulse makes it seem like changing it comes for free, but this is not necessarily the case. In the presence of a phase shift, the channel might need some time to adjust. For two consequent pulses where the phase changes between them, this buffer time should be enforced whenever it is not already present.

HGSilveri:
Test coverage is at 100%, but more specific tests should be added once devices with channels using these quantities are defined (see #297).

Closes #296 .

### DIFF
--- a/pulser/channels.py
+++ b/pulser/channels.py
@@ -39,8 +39,8 @@ class Channel:
         max_amp: Maximum pulse amplitude (in rad/µs).
         phase_jump_time: Time taken to change the phase between consecutive
             pulses (in ns).
-        min_retarget_interval: Minimum time required between two target
-            instructions (in ns).
+        min_retarget_interval: Minimum time required between the ends of two
+            target instructions (in ns).
         fixed_retarget_t: Time taken to change the target (in ns).
         max_targets: How many qubits can be addressed at once by the same beam.
         clock_period: The duration of a clock cycle (in ns). The duration of a
--- a/pulser/channels.py
+++ b/pulser/channels.py
@@ -37,7 +37,11 @@ class Channel:
         max_abs_detuning: Maximum possible detuning (in rad/µs), in absolute
             value.
         max_amp: Maximum pulse amplitude (in rad/µs).
-        retarget_time: Maximum time to change the target (in ns).
+        phase_jump_time: Time taken to change the phase between consecutive
+            pulses (in ns).
+        retarget_time: Minimum time required between two target instructions
+            (in ns).
+        fixed_retarget_t: Time taken to change the target (in ns).
         max_targets: How many qubits can be addressed at once by the same beam.
         clock_period: The duration of a clock cycle (in ns). The duration of a
             pulse or delay instruction is enforced to be a multiple of the
@@ -55,7 +59,9 @@ class Channel:
     addressing: str
     max_abs_detuning: float
     max_amp: float
+    phase_jump_time: int = 0
     retarget_time: Optional[int] = None
+    fixed_retarget_t: Optional[int] = None
     max_targets: Optional[int] = None
     clock_period: int = 4  # ns
     min_duration: int = 16  # ns
@@ -66,7 +72,9 @@ class Channel:
         cls,
         max_abs_detuning: float,
         max_amp: float,
+        phase_jump_time: int = 0,
         retarget_time: int = 220,
+        fixed_retarget_t: int = 0,
         max_targets: int = 1,
         **kwargs: int,
     ) -> Channel:
@@ -76,7 +84,11 @@ class Channel:
             max_abs_detuning (float): Maximum possible detuning (in rad/µs), in
                 absolute value.
             max_amp(float): Maximum pulse amplitude (in rad/µs).
-            retarget_time (int): Maximum time to change the target (in ns).
+            phase_jump_time (int): Time taken to change the phase between
+                consecutive pulses (in ns).
+            retarget_time (int): Minimum time required between two target
+                instructions (in ns).
+            fixed_retarget_t (int): Time taken to change the target (in ns).
             max_targets (int): Maximum number of atoms the channel can target
                 simultaneously.
         """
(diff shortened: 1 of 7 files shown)


pulser/sequence.py
```diff
@@ -638,39 +657,30 @@ def add(
...
+                        self._channels[channel].phase_jump_time - (t0 - op.tf),
+                    )
+                break
+
+        if delay_duration > 0:
             # Delay must not be shorter than the min duration for this channel
             min_duration = self._channels[channel].min_duration
             if delay_duration < min_duration:
-                ti += min_duration - delay_duration
                 delay_duration = min_duration
```
Codoscope, line 677: Could be simplified to `delay_duration = max(delay_duration, min_duration)`

pulser/sequence.py
```diff
@@ -1051,7 +1061,9 @@ def _target(
                     delta = self._channels[channel].validate_duration(
                         16 if delta < 16 else delta
                     )
-            tf = ti + delta
+            tf = ti + max(
+                delta, cast(int, self._channels[channel].fixed_retarget_t)
+            )
```
Codoscope, line 1066: Shouldn't this also take into account `retarget_time`?
Do you confirm that retargetting and phase jumping can be performed in parallel?

pulser/sequence.py
```diff
@@ -622,6 +622,25 @@ def add(
...
+        if len(prs) != 1:
+            raise ValueError(
+                "Cannot do a multiple-target pulse on qubits with different "
+                "phase references for the same basis."
+            )
+        else:
+            phase_ref = prs.pop()
+
+        if phase_ref != 0:
+            # Has to recriate the original pulse with a new phase
```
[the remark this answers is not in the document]
HGSilveri: Yes, good catch!

pulser/sequence.py
```diff
@@ -622,6 +622,25 @@ def add(
         last = self._last(channel)
         t0 = last.tf  # Preliminary ti
         basis = channel_obj.basis
+
+        prs = {self._phase_ref[basis][q].last_phase for q in last.targets}
```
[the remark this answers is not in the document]
HGSilveri: Changed to `ph_refs`, `phase_refs` would require extra lines

pulser/sequence.py
```diff
@@ -638,39 +657,28 @@ def add(
...
+        # Find last pulse and compare phase
+        for op in self._schedule[channel][::-1]:
+            if isinstance(op.type, Pulse):
+                if op.type.phase != pulse.phase:
+                    delay_duration = max(
+                        delay_duration, self._channels[channel].phase_jump_time
+                    )
+                break
+
+        if delay_duration > 0:
```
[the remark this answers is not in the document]
HGSilveri: It can't be `< 0` because `delay_duration <= current_max_t - t0` and `current_max_t => t0` by definition. If `== 0` then you'll simply end up with `ti = t0`; `min_duration` only needs to be respected if there is a delay.

pulser/sequence.py
```diff
@@ -664,7 +664,9 @@ def add(
             if isinstance(op.type, Pulse):
                 if op.type.phase != pulse.phase:
                     delay_duration = max(
-                        delay_duration, self._channels[channel].phase_jump_time
+                        delay_duration,
+                        # Considers that the last pulse might not be at t0
+                        self._channels[channel].phase_jump_time - (t0 - op.tf),
```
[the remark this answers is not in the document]
HGSilveri: Good idea, I must have forgot I had defined that

pulser/sequence.py
HGSilveri: Indeed it can.

pulser/channels.py
```diff
@@ -37,7 +37,11 @@ class Channel:
         max_abs_detuning: Maximum possible detuning (in rad/µs), in absolute
             value.
         max_amp: Maximum pulse amplitude (in rad/µs).
-        retarget_time: Maximum time to change the target (in ns).
+        phase_jump_time: Time taken to change the phase between consecutive
+            pulses (in ns).
+        retarget_time: Minimum time required between two target instructions
```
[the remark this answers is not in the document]
HGSilveri: I agree it should be changed, but that's a huge name. Let's try to come up with something smaller if possible.
> By the way, I can't see where this value is used to calculate the delay?
It's used to calculate the `delta` on line 1055.

pulser/sequence.py
HGSilveri: `retarget_time` is already considered in the `delta` value. 

Yes, retargeting and phase jumping can be performed in parallel because they take place on different hardware components.

pulser/channels.py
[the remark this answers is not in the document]
HGSilveri: I changed it to `min_retarget_interval` instead.

pulser/sequence.py
HGSilveri: It's actually the delta between the two retarget ends that matters, note that you always keep `ti` the same and shift `tf` to satisfy your criteria. This also means that if `fixed_retarget_t > retarget_time`, then `retarget_time` is irrelevant because it will always be assured by definition. Does it make sense now?

Codoscope: Yes, makes perfect sense. It would probably be worth it to precise that the interval is between both ends on the variable's comment (‶interval between instructions″ might be confusing).

Codoscope (approved):
Maybe just a comment to improve, but I approve overall.

pulser/sequence.py
HGSilveri: Ok, I changed it too.

HGSilveri:
Thanks for the thorough review @Codoscope !

Issue #299, opened 2021-12-16
Enable serialization for Register3D


HGSilveri:
The `Register3D` class was added with a `_to_dict()` method, which prevents it from being serialized. The task here is simply to add this method, along with the relevant tests.

PR #306, opened 2022-01-19
Add Register3D _to_dict method and UTs

Codoscope:
#299

### DIFF
--- a/pulser/register.py
+++ b/pulser/register.py
@@ -288,20 +288,16 @@ class BaseRegister(ABC):
         return obj_to_dict(self, qs)
 
     def __eq__(self, other: Any) -> bool:
-        if not isinstance(other, BaseRegister):
+        if type(other) is not type(self):
             return False
 
-        return (
-            type(self) is type(other)
-            and set(self._ids) == set(other._ids)
-            and all(
-                (
-                    np.array_equal(
-                        self._coords[i],
-                        other._coords[other._ids.index(self._ids[i])],
-                    )
-                    for i in range(len(self._ids))
+        return set(self._ids) == set(other._ids) and all(
+            (
+                np.array_equal(
+                    self._coords[i],
+                    other._coords[other._ids.index(id)],
                 )
+                for i, id in enumerate(self._ids)
             )
         )
 
--- a/pulser/register.py
+++ b/pulser/register.py
@@ -282,6 +282,29 @@ class BaseRegister(ABC):
                     "Needs more than one atom to draw " "the blockade radius."
                 )
 
+    @abstractmethod
+    def _to_dict(self) -> dict[str, Any]:
+        qs = dict(zip(self._ids, map(np.ndarray.tolist, self._coords)))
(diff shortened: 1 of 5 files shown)


pulser/register.py
```diff
@@ -282,6 +282,29 @@ def _draw_checks(
                     "Needs more than one atom to draw " "the blockade radius."
                 )
 
+    @abstractmethod
```
HGSilveri, line 285: What's the motivation to define it as an `abstractmethod` that is the same on all subclasses? Is it to force the definition of `_to_dict()` on each subclass (even if redundant) so it requires coverage by the unit tests?

pulser/register.py
```diff
@@ -282,6 +282,29 @@ def _draw_checks(
...
+    def _to_dict(self) -> dict[str, Any]:
+        qs = dict(zip(self._ids, map(np.ndarray.tolist, self._coords)))
+        return obj_to_dict(self, qs)
+
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, BaseRegister):
```
HGSilveri, line 291: Isn't this unnecessary considering you then check for `type(self) is type(other)` ?

pulser/register.py
```diff
@@ -282,6 +282,29 @@ def _draw_checks(
...
+            type(self) is type(other)
+            and set(self._ids) == set(other._ids)
+            and all(
+                (
+                    np.array_equal(
+                        self._coords[i],
+                        other._coords[other._ids.index(self._ids[i])],
+                    )
+                    for i in range(len(self._ids))
+                )
```
HGSilveri, line 304: I would loop over the `_ids` instead:
```suggestion
                (
                    np.array_equal(
                        self._coords[i],
                        other._coords[other._ids.index(id)],
                    )
                    for i, id in enumerate(self._ids)
                )
```

pulser/sequence.py
```diff
@@ -215,6 +215,11 @@ def qubit_info(self) -> dict[QubitId, np.ndarray]:
         """Dictionary with the qubit's IDs and positions."""
         return self._register.qubits
 
+    @property
+    def register(self) -> BaseRegister:
+        """Register with the qubit's IDs and positions."""
+        return self._register
```
HGSilveri, line 221: I agree with this addition. I think that `Sequence._register` is being accessed in `_seq_drawer.py`, do you think you could also replace those calls with `Sequence.register`?

HGSilveri (commented):
Looking good! I only have a few remarks and requests

pulser/register.py
Codoscope: The motivation is to make sure that any future subclass won't be wrongly serializable by default, and will need to make it explicit by adding a new `_to_dict()` method. The developer will also be forced to think about whether they want the method to be identical, or whether they want to extend it.

HGSilveri: Yes, that's a fair point.

HGSilveri (approved):
Looks good to me, thanks!

PR #318, opened 2022-02-07
Ship typed hints to the library's users

Codoscope:
Enable to use typed hints with mypy when using the Pulser library

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,6 +33,7 @@ setup(
         ],
     },
     packages=find_packages(),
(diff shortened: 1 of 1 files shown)


HGSilveri:
Good idea! Did you check that it's working locally now?

Codoscope:
Yes, this fixes the `missing library` error that I had on the MIS project — although I have other type errors to fix now that mypy is running correctly.

HGSilveri (approved):
LGTM

Issue #319, opened 2022-02-07
Various formatting issues


lvignoli:
There are various formatting issues:

- the code in notebooks is not formatted by black
- imports are sorted in a very conservative way by black, which let them messy
- some files could be formatted systematically, like markdown ones

I propose adding the following requirements to the project:

- `black[jupyter]`
- `isort`

and format some other files.

lvignoli:
It seems there is [this extension](https://jupyterlab-code-formatter.readthedocs.io/en/latest/) for jupyterlab. I will try to test it tomorrow.

I actually have another branch with some enhancements for the development environment, one of them offering a git pre-push hook that check for black, flake8, mypy, tests, etc before actually allowing a push. I'll make a distinct PR, as it is not exactly the same scope.

PR #320, opened 2022-02-07
Consistent formatting and style

lvignoli:
Fixing #319


.flake8
```diff
@@ -1,15 +1,17 @@
...
-exclude = ./build, ./docs
 docstring-convention = google
+exclude = ./build, ./docs
+extend-ignore =
+    # D105 Missing docstring in magic method
+    D105,
+    # E203 whitespace before ':' (for compliance with black)
+    E203,
+# for compliance with black
+max-line-length = 88
```
HGSilveri, line 10: When we adopted `black`, we decided to stick with the `flake8` 80 line limit instead

HGSilveri (commented):
LGTM in general, just note the detail about the line length

.flake8
lvignoli: Indeed, it's fixed!

.flake8
```diff
@@ -7,7 +7,7 @@ extend-ignore =
     # E203 whitespace before ':' (for compliance with black)
     E203,
 # for compliance with black
-max-line-length = 88
+max-line-length = 80
```
HGSilveri, line 10: Eish, I'm sorry, I just checked and it's actually 79... Maybe just remove this line instead, since we're using the `flake8` default

HGSilveri (approved):

PR #324, opened 2022-02-08
Changing the line length for the `isort` config

HGSilveri:
Since we don't use the default `black` line length, only setting the `isort` profile to `black` will make `isort` go up to the 88 character limit, which causes an incompatibility with our custom 79 character limit. Adding an explicit `line_lenght=79` parameter to the `isort` config fixes this.


lvignoli (approved):
Yes good point!

PR #326, opened 2022-02-10
Removing restriction on numpy version following qutip's update

HGSilveri:
`qutip` 4.6.3 now supports the latest version of `numpy`, so I'm removing the restriction that `numpy < 1.22` and rather place it on qutip to be `>=4.6.3`(ie it's latest version).

This effectively reverses the hotfix introduced in #304 , which was needed due to an update of `numpy` that `qutip` did not support.


Codoscope (approved):

lvignoli:
🎉🎉🎉

M1 users thank you greatly 🥰

Issue #325, opened 2022-02-09
Give consistent behavior to variables


Codoscope:
Currently, an instance of the class `Variable` can have two different behaviors, depending on whether its size is 1 or greater. This dynamic polymorphism can be annoying when trying to treat the instance only relying on its type. For example, from qaoa_mis, we have to write a specific condition to make sure `t_list` and `s_list` will be iterable. It's cumbersome, and makes it impossible to correctly type-hint them:

```python
t_list = seq.declare_variable('t_list', size=LAYERS)
s_list = seq.declare_variable('s_list', size=LAYERS)

if LAYERS == 1:
    t_list = [t_list]
    s_list = [s_list]

for t, s in zip(t_list, s_list):
    pulse_1 = Pulse.ConstantPulse(1000*t, 1., 0., 0)
    pulse_2 = Pulse.ConstantPulse(1000*s, 1., 1., 0)

    seq.add(pulse_1, 'ch0')
    seq.add(pulse_2, 'ch0')
```

In usual programming languages, arrays having only one element are allowed and have the same behavior than if they were bigger.

I think it would be great that the `Variable` keeps the same behavior regardless of the size (currently, `__getitem__` will raise an error when it's equal to 1). It could also be renamed `VariableArray`, because it will be an array in all cases, and when we want to really create a variable that is not an array, we could benefit from `VariableItem` that could be renamed to just `Variable`.

The above code would be simplified to:

```python
t_list = seq.declare_variable('t_list', size=LAYERS)
s_list = seq.declare_variable('s_list', size=LAYERS)

for t, s in zip(t_list, s_list):
    pulse_1 = Pulse.ConstantPulse(1000*t, 1., 0., 0)
    pulse_2 = Pulse.ConstantPulse(1000*s, 1., 1., 0)

    seq.add(pulse_1, 'ch0')
    seq.add(pulse_2, 'ch0')
```

The function `declare_variable` could additionally be split into `declare_variable` (without size) and `declare_variable_array` (with size equal to or greater than 1).

PR #327, opened 2022-02-10
Generalize variable

Codoscope:
PR corresponding to the issue [https://github.com/pasqal-io/Pulser/issues/325](https://github.com/pasqal-io/Pulser/issues/325)

I still have to fix a problem when trying to rename the classes, falling on a `TypeError: 'ABCMeta' object is not subscriptable` error.

### DIFF
--- a/pulser/parametrized/paramobj.py
+++ b/pulser/parametrized/paramobj.py
@@ -19,7 +19,6 @@ import inspect
 import operator
 import warnings
 from collections.abc import Callable
-from functools import partialmethod
 from itertools import chain
 from typing import TYPE_CHECKING, Any, Union
 
--- a/pulser/parametrized/paramobj.py
+++ b/pulser/parametrized/paramobj.py
@@ -56,6 +56,48 @@ class OpSupport:
     def __abs__(self) -> ParamObj:
         return ParamObj(operator.abs, self)
 
+    def __add__(self, other: Union[int, float]) -> ParamObj:
+        pass
+
+    def __radd__(self, other: Union[int, float]) -> ParamObj:
+        pass
+
+    def __sub__(self, other: Union[int, float]) -> ParamObj:
+        pass
+
+    def __rsub__(self, other: Union[int, float]) -> ParamObj:
+        pass
+
+    def __mul__(self, other: Union[int, float]) -> ParamObj:
+        pass
+
+    def __rmul__(self, other: Union[int, float]) -> ParamObj:
+        pass
+
+    def __truediv__(self, other: Union[int, float]) -> ParamObj:
+        pass
+
+    def __rtruediv__(self, other: Union[int, float]) -> ParamObj:
+        pass
+
+    def __floordiv__(self, other: Union[int, float]) -> ParamObj:
+        pass
+
+    def __rfloordiv__(self, other: Union[int, float]) -> ParamObj:
+        pass
+
+    def __pow__(self, other: Union[int, float]) -> ParamObj:
+        pass
+
+    def __rpow__(self, other: Union[int, float]) -> ParamObj:
+        pass
+
+    def __mod__(self, other: Union[int, float]) -> ParamObj:
+        pass
+
+    def __rmod__(self, other: Union[int, float]) -> ParamObj:
+        pass
+
 
 # Inject operator magic methods into OpSupport
 for method in reversible_ops:
--- a/pulser/parametrized/paramobj.py
+++ b/pulser/parametrized/paramobj.py
@@ -29,26 +29,9 @@ from pulser.parametrized import Parametrized
 if TYPE_CHECKING:
     from pulser.parametrized import Variable  # pragma: no cover
 
-# Available operations on parametrized objects with OpSupport
-reversible_ops = [
-    "__add__",
-    "__sub__",
-    "__mul__",
-    "__truediv__",
-    "__floordiv__",
-    "__pow__",
-    "__mod__",
-]
-
 
 class OpSupport:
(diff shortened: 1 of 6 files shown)


HGSilveri:
> I still have to fix a problem when trying to rename the classes, falling on a `TypeError: 'ABCMeta' object is not subscriptable` error.

Ok, I'll mark it as a Draft for now then

HGSilveri:
Is this ready for review or are you still going to rename the `Variable` and `_VariableItem` classes?

pulser/sequence.py
```diff
@@ -538,12 +548,31 @@ def declare_channel(
...
         dtype: Union[type[int], type[float], type[str]] = float,
     ) -> Variable:
+        pass
+
+    def declare_variable(
+        self,
+        name: str,
+        size: Optional[int] = None,
+        dtype: Union[type[int], type[float], type[str]] = float,
+    ) -> Union[Variable, _VariableItem]:
```
HGSilveri, line 575: I like this solution, I think it's better than having two separate methods.

pulser/parametrized/paramobj.py
```diff
@@ -56,6 +56,48 @@ def __neg__(self) -> ParamObj:
...
+        pass
+
+    def __rpow__(self, other: Union[int, float]) -> ParamObj:
+        pass
+
+    def __mod__(self, other: Union[int, float]) -> ParamObj:
+        pass
+
+    def __rmod__(self, other: Union[int, float]) -> ParamObj:
+        pass
```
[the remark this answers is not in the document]
Codoscope: The benefit of doing this is that the implementation is still dynamically factorized. Only the redundancy of declarations necessary to mypy is added. But I can also include the implementation if you think it's better.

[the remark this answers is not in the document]
HGSilveri: In your opinion, what's the benefit of having it dynamically factorized?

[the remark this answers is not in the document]
Codoscope: It avoids the burden of repeating `ParamObj(operator.__add__, self, other)`, `ParamObj(operator.__radd__, other, self)` and so on in all functions. The repetition is pretty small however, so it's true that the dynamic code might not be worth it anymore.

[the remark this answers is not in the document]
HGSilveri: To me, it won't add any lines and it will make the code more explicit, so I say go for it.

pulser/tests/test_paramseq.py
```diff
@@ -29,7 +29,7 @@
 def test_var_declarations():
     sb = Sequence(reg, device)
     assert sb.declared_variables == {}
-    var = sb.declare_variable("var")
+    var = sb.declare_variable("var", size=1)
```
HGSilveri, line 32: Maybe you could add a test where the variable is declared with `size=None` too

pulser/tests/test_paramseq.py
```diff
@@ -175,7 +175,8 @@ def test_str():
     sb.add(pls, "ch1")
     s = (
         f"Prelude\n-------\n{str(seq)}Stored calls\n------------\n\n"
-        + "1. add(Pulse.ConstantPulse(mul(var, 100), var, -1, var), ch1)"
+        + "1. add(Pulse.ConstantPulse(mul(var[0], 100), var[0],"
```
HGSilveri, line 178: I guess that it would be nice that a variable declared with `size=None` would keep being described without the `[0]`, but that's a small detail.

HGSilveri (approved):
LGTM, I only have a couple of minor remarks

pulser/tests/test_paramseq.py
Codoscope: This corresponds to the line `var = sb.declare_variable("var")` since `size` is defaulted to `None`. Do you think I should add a test where it is explicitely `None` too?

pulser/tests/test_paramseq.py
Codoscope: Do you think we could leave it for another PR and keep the Issue open? I even wonder if the `_VariableItem` class shouldn't be removed in favor of an instanciation of `ParamObj` doing the indexation. I have some draft about it but it's having UT issues.

pulser/tests/test_paramseq.py
HGSilveri: Yes, but you changed it to have `size=1` so now I don't know if there is one where size was not specified (ie defaulted to `None`). No need for a test where `None` is explicitly stated

pulser/tests/test_paramseq.py
HGSilveri: Sure, that's okay with me

pulser/tests/test_paramseq.py
Codoscope: I meant that the line `var = sb.declare_variable("var")` is already present in other parts of the test file. For example, on line 134 in `test_build()`

HGSilveri: Sure, of course. But I think it would be worth it to have the checks that are done after this declaration for this case too. For example:
```python
var_ = seq.declare_variable("var_")
assert sb.declared_variables["var_"] == var_
assert isinstance(var_, _VariableItem)
```
You see what I mean?

Codoscope: Yes, thanks for your remark, I have applied it, although a downside of this approach in my sense is that it seems to be testing the implementation instead of the behavior.

HGSilveri (approved):
Thanks a lot @Codoscope !

PR #330, opened 2022-02-14
Make evaluation times uniform including final time

sebgrijalva:
This PR addresses some instabilities on the way the `evaluation_times` attribute of `Simulation` is generated. 

For very small sequences and full sampling rates it made sense to include the final time and be able to look at every instant (in ns) of evolution. But for larger sequences and smaller sampling rates this results in a 1 ns step added to `evaluation_times`. This makes the convergence of the ODE integration more difficult.

An additional time step has been added in order to append a sample of `0.0` to the used `Simulation` samples dict.

I also updated the case in which the `evaluation_times` value is set to a `float`, by including always the initial (`0`) and final times to the resulting `_eval_times_array` attribute. This avoids the bug of having a very small float that makes the evaluation array empty !


pulser/simulation/simulation.py
```diff
@@ -460,6 +462,12 @@ def write_samples(
             )
             samples_dict["phase"][slot.ti : slot.tf] += _pulse.phase
 
+            # Final time-step: repeat last sample
+            if slot.tf == self._tot_duration - 1:
+                samples_dict["amp"][-1] = _pulse.amplitude.samples[-1]
```
HGSilveri, line 467: I think you should put 0 in the amplitude (so actually, not touch it)

pulser/simulation/simulation.py
```diff
@@ -308,14 +309,13 @@ def evaluation_times(self) -> np.ndarray:
...
         if isinstance(value, str):
             if value == "Full":
-                self._eval_times_array = np.append(
-                    self.sampling_times, self._tot_duration / 1000
-                )
+                self._eval_times_array = np.copy(self.sampling_times)
             elif value == "Minimal":
                 self._eval_times_array = np.array(
-                    [self.sampling_times[0], self._tot_duration / 1000]
+                    [self.sampling_times[0], ]
```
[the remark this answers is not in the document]
sebgrijalva: These are added in lines 356-359:

```python
if t_min > 0:
    eval_times = np.insert(eval_times, 0, 0.0)
if t_max < final_seq_time:
    eval_times = np.append(eval_times, final_seq_time)
```

I wonder about `self.sampling_times[0]` . When is it not zero?

pulser/simulation/simulation.py
sebgrijalva: Removed copying of last value so it's now by default 0.

pulser/simulation/simulation.py
```diff
@@ -317,14 +318,15 @@ def evaluation_times(self) -> np.ndarray:
...
             if value == "Full":
-                self._eval_times_array = np.append(
-                    self.sampling_times, self._tot_duration / 1000
-                )
+                self._eval_times_array = np.copy(self.sampling_times)
             elif value == "Minimal":
                 self._eval_times_array = np.array(
-                    [self.sampling_times[0], self._tot_duration / 1000]
+                    [
+                        self.sampling_times[0],
```
HGSilveri, line 328: I'm sorry @sebgrijalva , I still don't get it. Can you point me to the line where the `final_seq_time` and the `0` are being injected into `_eval_times_array` for this case?

sebgrijalva:
I'm going to need some input @lvignoli  on how to adapt the tests for the sampler. Note that this PR will make the simulation sequences one step longer, with a sample equal to `0.0`.

pulser/simulation/simulation.py
```diff
@@ -362,15 +368,11 @@ def evaluation_times(self, value: Union[str, ArrayLike, float]) -> None:
...
-            if t_max < final_seq_time:
-                eval_times = np.append(eval_times, final_seq_time)
-
-            self._eval_times_array = eval_times
-            # always include initial and final times
+            # Ensure 0 and final time are included:
+            self._eval_times_array = np.union1d(
+                value, 
+                [0.0, self._tot_duration / 1000]
+                )
```
HGSilveri, line 375: Isn't this something that's always required? And if so, shouldn't we just do it at the end of the function for all cases?

sebgrijalva: Yup, good idea.

pulser/tests/test_simulation.py
```diff
@@ -659,8 +660,8 @@ def test_get_xy_hamiltonian():
         )
         < 1e-10
     )
-    assert simple_ham[0, 1] == 0.5 * amp
-    assert simple_ham[3, 3] == -2 * detun
+    assert np.isclose(simple_ham[0, 1], 0.5 * amp)
+    assert np.isclose(simple_ham[3, 3], -2 * detun)
```
HGSilveri, line 664: Do you know why it's no longer strictly equal?

HGSilveri (commented):
It's much cleaner this way, good job!

pulser/tests/test_sequence_sampler.py
```diff
@@ -31,14 +31,16 @@ def assert_nested_dict_equality(got, want: dict) -> None:
         for qty in want["Global"][basis]:
             np.testing.assert_array_equal(
                 got["Global"][basis][qty],
-                want["Global"][basis][qty],
+                # skip final 0 sample from comparison
+                want["Global"][basis][qty][:-1],
```
lvignoli, line 35: This helper function is used elsewhere (in the SLM tests for example), and by the name it is expected to check the dicts without any logic assumption on the form.

I would rather keep it untouched, and copy patch it in the `assert_same_samples_as_sim()` function.

Or, add a bool kwarg `skip_last` defaulting to `False` and do

```suggestion
                (
                        want["Global"][basis][qty][:-1]
                        if skip_last
                        else want["Global"][basis][qty]
                ),
```

and call `assert_nested_dict_equality(got, want, skip_last=True)` in `assert_same_samples_as_sim()`.

pulser/tests/test_sequence_sampler.py
```diff
@@ -31,14 +31,16 @@ def assert_nested_dict_equality(got, want: dict) -> None:
...
             )
     for basis in want["Local"]:
         for qubit in want["Local"][basis]:
             for qty in want["Local"][basis][qubit]:
                 np.testing.assert_array_equal(
                     got["Local"][basis][qubit][qty],
-                    want["Local"][basis][qubit][qty],
+                    # skip final 0 sample from comparison
+                    want["Local"][basis][qubit][qty][:-1],
                 )
```
lvignoli, line 44: Same thing here

pulser/tests/test_sequence_sampler.py
```diff
@@ -128,7 +130,7 @@ def test_SLM_samples(seq_with_SLM):
     a_samples = pulse.amplitude.samples
 
     def z() -> np.ndarray:
-        return np.zeros(seq_with_SLM.get_duration())
+        return np.zeros(seq_with_SLM.get_duration() + 1)
```
lvignoli, line 134: Then this can be reverted

pulser/tests/test_sequence_sampler.py
```diff
@@ -139,9 +141,9 @@ def z() -> np.ndarray:
             }
         },
     }
-    want["Local"]["ground-rydberg"]["batman"]["amp"][200:401] = a_samples
+    want["Local"]["ground-rydberg"]["batman"]["amp"][200:400] = a_samples
     want["Local"]["ground-rydberg"]["superman"]["amp"][0:200] = a_samples
-    want["Local"]["ground-rydberg"]["superman"]["amp"][200:401] = a_samples
+    want["Local"]["ground-rydberg"]["superman"]["amp"][200:400] = a_samples
```
lvignoli, line 147: Thanks for catching this inconsequential index mistake 😃.

lvignoli (commented):
`assert_same_samples_as_sim()` should bear responsibility for the simulation details.
Apart that it is good to me!

pulser/tests/test_sequence_sampler.py
lvignoli: I think it is much better now! Thanks 😃

pulser/tests/test_simulation.py
sebgrijalva: `simple_ham` was returning a complex number (with zero imaginary part)

HGSilveri (approved):
Great, very glad to see this one go through!

PR #331, opened 2022-02-16
Adding CI checks with git hooks and pre-commit

lvignoli:
- add pre-commit hooks
- update contributing guidelines with pre-commit usage

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,27 @@
+repos:
+  - repo: https://github.com/psf/black
+    rev: 22.1.0
+    hooks:
+      - id: black-jupyter
+
+  - repo: https://github.com/PyCQA/flake8
+    rev: 4.0.1
+    hooks:
+      - id: flake8
+
+  - repo: https://github.com/pycqa/isort
+    rev: 5.10.1
+    hooks:
+      - id: isort
+        name: isort (python)
+        args: [--profile black]
+
+  # Calling mypy from a bash script, as I cannot figure out how to pass the
+  # .mypy.ini config file when using the hook at
+  # https://github.com/pre-commit/mirrors-mypy
+  - repo: local
+    hooks:
+      - id: mypy
+        name: mypy
+        entry: ./.mypy_script
+        language: script
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,7 +14,6 @@ repos:
     hooks:
       - id: isort
         name: isort (python)
-        args: [--profile black]
 
   # Calling mypy from a bash script, as I cannot figure out how to pass the
   # .mypy.ini config file when using the hook at
--- a/.mypy_script
+++ b/.mypy_script
@@ -1,0 +1,2 @@
+#!/usr/bin/env bash
+mypy --config-file .mypy.ini
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -115,3 +115,23 @@ annotations and pass the type checks from running:
 
     **Note**: Type hints for `numpy` have only been added in version 1.20. Make sure you have `numpy >= 1.20`
     installed before running the type checks.
+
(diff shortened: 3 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,27 @@
...
+    rev: 4.0.1
+    hooks:
+      - id: flake8
+
+  - repo: https://github.com/pycqa/isort
+    rev: 5.10.1
+    hooks:
+      - id: isort
+        name: isort (python)
+        args: [--profile black]
```
HGSilveri, line 17: So, `black-jupyter` and `isort` have access to their config files but `isort` doesn't? Or is the `--profile black` option redundant here? Because if `isort` doesn't go looking for what's in the `pyproject.toml` file, then it won't know about the custom line length.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,27 @@
...
+
+  # Calling mypy from a bash script, as I cannot figure out how to pass the
+  # .mypy.ini config file when using the hook at
+  # https://github.com/pre-commit/mirrors-mypy
+  - repo: local
+    hooks:
+      - id: mypy
+        name: mypy
+        entry: ./.mypy_script
+        language: script
```
HGSilveri, line 27: Anecdotally, this part initially failed when I first tried the hook. I eventually tried to run the script from the terminal (which worked) and afterwards the hook was working too. I did some random stuff before that though, so I'm not sure that's what made it work... (I know, very useful feedback)

HGSilveri (commented):
This is cool, thank you! I only have a couple remarks

.pre-commit-config.yaml
lvignoli: Did the hook fail, or did `mypy` complained?
At the moment (7976013add4bb642d9b5e6e4bc78586f522e728f) `mypy` raises two issues in `waveforms.py`.

I have check that the hook does discover `.mypy.ini` in fact.

I think this hook works fine, but I have been suspicious about it because `pre-commit run mypy --all-files` gives me a quite different output than the one given by my local `mypy` installation my dev venv, with have the same version.
That's why I am using the script workaround.

.pre-commit-config.yaml
lvignoli: `isort` hook actually read the `pyproject.toml` configuration, so I removed the args line, see fe2e49f4a9a4420dd20f518af5df6a18753a4a39.

.pre-commit-config.yaml
lvignoli: I think it's because the `--all-file` flags pass really all the files to `mypy`, so it also checks the `.py` files in the docs.
But it's not the whole story, since it raises me errors my local `mypy` don't, and vice versa.

They are supposed to run the same version with the same config file... I don't know who to trust!

HGSilveri: > Did the hook fail, or did `mypy` complained?
It failed to run the script, it was not `mypy` itself, it didn't even reach that point. Once the script started running, it passed.



> But it's not the whole story, since it raises me errors my local `mypy` don't, and vice versa.

This could be an issue with the `mypy` version. `mypy` is too volatile, so we fixed the version in the requirements. Perhaps the hook is using the latest release instead

HGSilveri:
@lvignoli Should we go with this? I'm ok with that, but it's up to you

.pre-commit-config.yaml
lvignoli: > This could be an issue with the mypy version. mypy is too volatile, so we fixed the version in the requirements. Perhaps the hook is using the latest release instead

I checked multiple times, passing the version number v0.921 to the hook make it use this exact version of `mypy`.

lvignoli:
Yes we can merge.
I am a bit puzzled and unsatisfied by this mypy mystery, but it's actually working the way it is now so I guess it's fine.

HGSilveri (approved):
Thanks for sharing this!

PR #333, opened 2022-02-17
Adding Register Layouts

### LINKED ISSUE #278: Creating a Register from a specific layout
For multiple reasons, it would be valuable to have register created from an underlying, predefined ~~pattern~~ layout, rather than with arbitrary coordinates. 

There are a lot of different things to do here, but the first would be to create the foundational class that stores the layout. We could call it `Layout` or `RegisterLayout` and it would store all the coordinates of a layout. 

Then, the layout class would enable the creation of a `Register` from it by specification of the trap indices with which to form the `Register` itself, with specific constraints like the limitation of <50% of the sites being chosen, so that rearrangement is still possible afterwards.

A method to check whether a layout is compatible with a device should be added too. 

An open question is how we will differentiate layouts. We want to have layouts like square and triangular lattices, with different spacings between the atoms, but should these be instances of the same layout class, or child classes that specify the overall layout? To me, the child class approach seems cleaner and it would mimic what is done with the `Waveform`s.

HGSilveri:
This is a substantial change that does a few things:
- Adds the `RegisterLayout` class, in `register/register_layout.py`
- Adds a few special subclasses, in `register/special_layout.py`
- Moves the functionality to draw 2D patterns, previously in the `Register`class, to a separate file `register/_patterns.py` so that they can be used by the special layouts
- Moves the register drawing functionality, which is also used by the `RegisterLayout`, to a `RegDrawer` class stored in `register/_reg_drawer.py`
- Adds support to associated a `RegisterLayout` to a `BaseRegister` instance
- Adds functionality to the `Device` class to store pre-calibrated register layouts and to validate layouts

Closes #278 .


pulser/tests/test_register_layout.py
```diff
@@ -111,3 +118,64 @@ def test_eq():
...
+        layout.get_traps_from_coordinates((0.9999994, 1))
+
+
+def test_square_lattice_layout():
+    square = SquareLatticeLayout(9, 7, 5)
+    assert str(square) == "SquareLatticeLayout(9x7, 5µm)"
+    assert square.square_register(3) == Register.square(
+        3, spacing=5, prefix="q"
+    )
+    # An even number of atoms on the side won't align the center with an atom
```
Codoscope, line 143: I'm not sure to understand the comment: is its intent to explain why registers are different on lines 144 and 145? If so, the used sizes `3` and `4` are different, which is already an explanation.

pulser/register/base_register.py
```diff
@@ -52,7 +52,7 @@ class BaseRegister(ABC):
     """The abstract class for a register."""
 
     @abstractmethod
-    def __init__(self, qubits: Mapping[Any, ArrayLike]):
+    def __init__(self, qubits: Mapping[Any, ArrayLike], **kwargs: Any):
```
Codoscope, line 55: How about taking `Optional[_LayoutInfo]` as input instead of `**kwargs: Any`, for better type safety and documentation?

pulser/tests/test_devices.py
```diff
@@ -139,3 +146,31 @@ def test_validate_layout():
...
+
+    TestDevice = Device(
+        name="TestDevice",
+        dimensions=2,
+        rydberg_level=70,
+        max_atom_num=100,
+        max_radial_distance=50,
+        min_atom_distance=4,
+        _channels=(),
+        pre_calibrated_layouts=(TriangularLatticeLayout(100, 5),),
```
Codoscope, line 172: Maybe also try with `200`

pulser/register/register_layout.py
```diff
@@ -146,24 +178,42 @@ def draw(
...
+        hash.update(self.coords.tobytes())
+        return hash.digest()
+
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, RegisterLayout):
+            return False
+        return self._safe_hash() == other._safe_hash()
+
+    def __hash__(self) -> int:
+        return hash(self._safe_hash())
```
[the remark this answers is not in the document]
HGSilveri: Unfortunately no, because `__hash__` returns an `int` and `self._safe_hash()` is in `bytes`. I considered just turning the `bytes` to `int` but since the `hash` in Python has the property of not being the same on different runs, I though it would be best to keep it that way.

pulser/register/base_register.py
```diff
@@ -124,7 +124,19 @@ def from_coordinates(
...
         self, register_layout: RegisterLayout, *trap_ids: int
     ) -> None:
+        """Sets the RegisterLayout that originated this register."""
         trap_coords = register_layout.coords
+        if register_layout.dimensionality != self._dim:
+            raise ValueError(
+                "The RegisterLayout dimensionality is not the same as this "
+                "register's."
+            )
+        if len(trap_ids) != len(self._ids):
```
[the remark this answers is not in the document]
HGSilveri: You're right, I just thought it would be redundant considering that's already done in `RegisterLayout.define_register()`, but I'll add it

pulser/register/register.py
```diff
@@ -366,60 +275,26 @@ def max_connectivity(
...
-        return cls._hexagon_helper(full_layers, atoms_left, spacing, prefix)
+        return cls.from_coordinates(coords, center=False, prefix=prefix)
 
     def rotate(self, degrees: float) -> None:
         """Rotates the array around the origin by the given angle.
 
         Args:
             degrees (float): The angle of rotation in degrees.
         """
+        if self._layout_info is not None:
```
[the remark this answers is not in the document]
HGSilveri: I think that will only serve to confuse things. The idea of the RegisterLayout is that it is this fixed pattern that cannot be changed, so allowing any type of rotations is contradictory. And when they don't do anything, then why do them at all?

pulser/tests/test_register_layout.py
HGSilveri: You're right, this is a typo, it should be `4` on both

pulser/register/base_register.py
HGSilveri: I though about that, but my intention is to hide this from the users. I want the way of associating a layout with a register to be only through the methods of `RegisterLayout` and its subclasses. To have a `layout_info` keyword argument in `Register` will give the impression that they should provide a layout when creating a `Register`, which is not the intended way.

pulser/register/register.py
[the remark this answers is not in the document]
Codoscope: The rotation would be performed on the register (not the layout), which would enable tuning to try different configurations and check results are the same on the machine for example. That's not a priority though, but just a small comment.

Codoscope (approved):

PR #336, opened 2022-02-24
Accounting for rounding errors in Register and RegisterLayout validation

HGSilveri:


Louis-PaulHenry (approved):
LGTM! Quick and clean fix!

Register.max_connectivity now works without specifying `spacing`
(Just note that the `COORD_PRECISION`is not used in `max_connectivity` that is strictly enforcing a minimal distance of 4.0)

PR #337, opened 2022-02-24
WIP: add modulation to the simuation module

lvignoli:
**This is work in progress.**

This PR is for discussion on how to handle modulation in the simulation module.

Here is a minimum example for the modulation of global channel.
It discards any noises or SLM settings.

### DIFF
--- a/pulser/simulation/simulation.py
+++ b/pulser/simulation/simulation.py
@@ -67,12 +67,15 @@ class Simulation:
               those specific times.
 
             - A float to act as a sampling rate for the resulting state.
+        mod_output (bool, default False): toggles the use of modulated samples
+            in the simulation.
     """
 
     def __init__(
         self,
         sequence: Sequence,
         sampling_rate: float = 1.0,
+        mod_output: bool = False,
         config: Optional[SimConfig] = None,
         evaluation_times: Union[float, str, ArrayLike] = "Full",
     ) -> None:
@@ -98,6 +101,13 @@ class Simulation:
         self._qdict = self._seq.qubit_info
         self._size = len(self._qdict)
         self._tot_duration = self._seq.get_duration()
+
+        self.mod_output = mod_output
+        if self.mod_output:
+            max_rise_time = max(
+                [ch.rise_time for ch in self._seq.declared_channels.values()]
+            )
+            self._tot_duration += 2 * max_rise_time
 
         # Type hints for attributes defined outside of __init__
         self.basis_name: str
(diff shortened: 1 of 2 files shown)


lvignoli:
@HGSilveri could you share your thoughts on how to test the extraction and modulation of samples?
For now I naively compare the final states.

lvignoli:
The idea is to merge the `_extract_and_modulate_samples` method inside the `_extract_samples` one, but it necessitate to first refactor the latter in my opinion, to avoid many if branch by checking each time for noises sources, slm, modulation etc.
Hence the separate method for now.

@HGSilveri could you share your thoughts on how to properly test the extraction and modulation of samples?
For now I naively compare the final states.

HGSilveri:
@lvignoli Thank you for your work on this. I haven't been following the progress, so I have a bit of catching up to do.

Actually, I was thinking about this the other day and, borrowing from your idea of a "FeltSequence", I think it could make sense to encapsulate the sequence samples related logic in a separate, auxiliary class. This class would receive the sequence object and be capable of returning the samples for the multiple scenarios: input or output, noise, masked qubits, ... I think this could lighten up the bloated `Simulation` class and at the same time encapsulate what you've introduced here. What are your thoughts? I didn't look at the code in detail when thinking about this, so it's possible I'm missing some interconnections that will make this an awkward split.

Regarding the tests, maybe you can check for some extra things (like the streching of the total duration), but it's true that these are particularly tricky to test. I'll try to make some suggestions when reviewing.

By the way, is this ready for review now?

lvignoli:
I am very fond of your suggestion of a sample extractor class/function. I think it will be a substantial enhancement to the code. I am willing to implement it, I am opening an issue to track this!

lvignoli:
> By the way, is this ready for review now?

No, it was a PoC for feedback on how to do it the right way. It is not compatible with noises or SLM right now.
Unless this feature is very desirable for someone, I'll say it can wait for the sample extractor component implementation.

Still I am leaving the PR open for discussion about the right way to test the modulation.

HGSilveri:
> I am very fond of your suggestion of a sample extractor class/function. I think it will be a substantial enhancement to the code. I am willing to implement it, I am opening an issue to track this!

Great! In that case, it might make sense to tackle that first (without these changes), and then adapt these changes to the new format. This would mean, of course, that this PR would be put on stand by. Is that ok with you?

lvignoli:
> Great! In that case, it might make sense to tackle that first (without these changes), and then adapt these changes to the new format. This would mean, of course, that this PR would be put on stand by. Is that ok with you?

Exactly!

HGSilveri:
> > By the way, is this ready for review now?
> 
> No, it was a PoC for feedback on how to do it the right way. It is not compatible with noises or SLM right now. Unless this feature is very desirable for someone, I'll say it can wait for the sample extractor component implementation.
> 
> Still I am leaving the PR open for discussion about the right way to test the modulation.

I can see we are on the same page 😄

PR #346, opened 2022-03-18
Initial version of abstract sequence format

stared:
An initial version of an abstract serialization of the Pulser sequences. 
As of now, while it is not general, it supports https://github.com/pasqal-io/Pulser/blob/develop/tutorials/advanced_features/Serialization.ipynb.

It might serve as a starting point and comparison for more robust export approaches.

### DIFF
--- a/pulser/sequence.py
+++ b/pulser/sequence.py
@@ -1071,6 +1071,70 @@ class Sequence:
         """
         return json.dumps(self, cls=PulserEncoder, **kwargs)
 
+    def abstract_repr(self) -> dict[str, Any]:
+        """Serializes the Sequence into an abstract JSON object.
+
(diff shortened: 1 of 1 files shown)


HGSilveri (approved):
Cheers, thanks!

stared:
Thanks, @HGSilveri.  I wanted to add one more commit with tests. It was a quick merge, so there is a new PR #347.

As a small note - this format may differ a bit from the others we discussed. It is not necessarily indented, but we will need to make updates anyway. The test function (which will be updated as well) is important as with each new change we want to make sure that the output is as desired.

HGSilveri:
Sorry @stared , I guess I jumped the gun 😅

PR #347, opened 2022-03-18
Abstract represenation test

stared:
Vide #346


HGSilveri:
@stared To be clear, for the differences between this one and the schema, the schema format is the correct one, right?

stared:
@HGSilveri `vulcan-schema` is the reference. I didn't update all as first I want to see if you have requirements or remarks to `vulcan-schema`.

PR #348, opened 2022-03-21
Feat/ Define custom serialization errors

MatthieuMoreau0:
- Create custom exception `SerializationError` when sequence serialization fails instead of generic `SystemError`.
- Raise it when appropriate
- Update tests

Not sure where you wanted me to add the definition of the exception, if this doesn't fit let me know!


HGSilveri (approved):
LGTM, thanks for the addition!

PR #350, opened 2022-03-23
Abstract representation schema check

stared:
- [X] Added a schema example
- [X] Automatic schema check with [jsonschema](https://python-jsonschema.readthedocs.io)
- [X] Modified proof-of-concept to fit the schema
- [X] Moved abstract repr test to a separate file and split into parts

The schema file `abstract_repr_schema.json` will be modified according to needs. It should be generated rather than typed by hand. 

The same schema verification can be performed for other export formats and options.


pulser/sequence.py
```diff
@@ -1081,29 +1081,49 @@ def abstract_repr(self) -> dict[str, Any]:
...
-        res = {"register": [], "channels": [], "operations": []}
+        mapping = {
+            "CompositeWaveform": "composite",
+            "CustomWaveform": "custom",
+            "ConstantWaveform": "constant",
+            "RampWaveform": "ramp",
+            "BlackmanWaveform": "blackman",
+            "InterpolatedWaveform": "interpolated",
+            "KaiserWavewform": "kaiser",
+        }
```
HGSilveri, line 1092: Actually, I have this being taken care of in a more general way in some local commits. It's still WIP, but I'll push them so that you can see how it's looking. Beware you will get some merge conflicts because I changed this file.

pulser/sequence.py
```diff
@@ -1081,29 +1081,49 @@ def abstract_repr(self) -> dict[str, Any]:
...
-                    res["register"].append({"name": name, "x": x, "y": y})
+                    res["register"][name] = {"x": x, "y": y}
             elif call.name == "declare_channel":
                 ch_name, ch_kind = call.args
-                res["channels"].append({"name": ch_name, "kind": ch_kind})
+                res["channels"][ch_name] = {"hardware_channel": ch_kind}
+                if "initial_target" in call.kwargs:
+                    res["channels"][ch_name]["initial_target"] = call.kwargs[
+                        "initial_target"
+                    ]
```
HGSilveri, line 1115: So we are storing the "initial_target" after all? Doesn't this make the following target operation redundant?

pulser/tests/test_abstract_repr.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+def test_schema():
+    with open("pulser/json/abstract_repr_schema.json") as f:
+        schema = json.load(f)
+
+    jsonschema.validate(instance=abstract, schema=schema)
+
+
+def test_values():
```
HGSilveri, line 74: Bear in mind the changes I made to the `test_abstract_repr()` function I made in `test_sequence.py`

pulser/tests/test_sequence.py
```diff
@@ -16,6 +16,7 @@
 
 import numpy as np
 import pytest
+import jsonschema
```
HGSilveri, line 19: I guess you don't need this import

pulser/tests/test_sequence.py
```diff
@@ -705,86 +706,3 @@ def test_mappable_register():
     )
     with pytest.raises(ValueError, match="already has a concrete register"):
         seq_.build(qubits={"q2": 20, "q0": 10})
-
-
-def test_abstract_repr():
```
HGSilveri, line 710: Bear in mind I changed some things in this function

HGSilveri (commented):
I'm fully on board with validating against the schema. Perhaps we should wait until it is a bit more stable, otherwise we'll be making PRs with every update.

I do wish you told me you were working on this though, as I have been making changes locally on this branch. I'll push them, so expect some merge conflicts coming your way. I'm sorry about that... 

Do you intend to keep making changes here? Should I start working on a branch of `schema-check`?

pulser/sequence.py
```diff
@@ -1081,29 +1081,49 @@ def abstract_repr(self) -> dict[str, Any]:
...
+        }
         operations = res["operations"]
         for call in self._calls:
             if call.name == "__init__":
                 register, device = call.args
                 # process register
                 res["device"] = device.name
                 for name, (x, y) in register.qubits.items():
-                    res["register"].append({"name": name, "x": x, "y": y})
+                    res["register"][name] = {"x": x, "y": y}
```
HGSilveri, line 1108: The qubit `name` can actually be an `int`, is this a problem if being used as a key in the JSON?

pulser/sequence.py
stared: This one is clearly "quick and dirty" and I don't intend to keep it for the final version. It was a minimal change so that I can generate JSON for the new schema.

stared:
@HGSilveri Thank you for your response! 

My main comment is that so far it is a minimal change so that it would work with the current schema. The schema will be changed, to fulfill any needs we have. This version is mostly to give an example for the verification process.

Not working on this branch, for now, so feel free to make any changes you need.

When it comes to the main export part, I am to assist you if needed, but otherwise, I leave it to you.

HGSilveri:
@stared Ok, great! In that case, can you please take care of the conflicts that came from the code I pushed and I'll take over from there?

An important remark: I'm using a custom JSONEncoder, so `Sequence.abstract_repr()` is returning a JSON-formatted string now. If you want to turn it into a `dict` like you had originally, you can just use `json.loads(seq.abstract_repr())`  (which is what I did for the tests).

pulser/sequence.py
HGSilveri: Ping in case you missed this question

PR #351, opened 2022-03-24
Update runs and samples per run.

sebgrijalva:
This PR adds the option for updating `runs` and `samples_per_run` when using the `Simulation.add_config()` method, which has been rising recently as something users want to include.

I also added some comments for future reading.

### DIFF
--- a/pulser/simulation/simulation.py
+++ b/pulser/simulation/simulation.py
@@ -223,9 +223,13 @@ class Simulation:
         old_noise_set = set(self.config.noise)
         new_noise_set = old_noise_set.union(config.noise)
         diff_noise_set = new_noise_set - old_noise_set
+        # Create temporary param_dict to add noise parameters:
         param_dict: dict[str, Any] = asdict(self._config)
+        # remove redundant `spam_dict`:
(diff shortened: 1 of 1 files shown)


sebgrijalva:
Not sure how to pass the docs check. Some help, @HGSilveri @lvignoli ?

lvignoli:
> Not sure how to pass the docs check. Some help, @HGSilveri @lvignoli ?

No idea, I also have a failing doc build on #345 with weird errors. I will push news commits there this morning, we will see if the problem remains.

HGSilveri:
> Not sure how to pass the docs check. Some help, @HGSilveri @lvignoli ?

It has to do with the latest update to `jinja2`, which has removed features being used by `nbconvert` (see [here](https://github.com/pallets/jinja/issues/1626)). A PR has been submitted on `nbconvert` to fix this ([here](https://github.com/jupyter/nbconvert/pull/1737)), which hopefully will be accepted soon. Until then, let's just ignore it.

HGSilveri (approved):
This should have been there from the start but I guess we missed it... Thanks for the PR!

PR #354, opened 2022-04-05
[Fix] Simulation with state preparation errors fixed

HGSilveri:
This fixes a bug on the simulation in the presence of state preparation errors - essentially, the `self._bad_atoms` were being changed but then the samples were not being extracted again, which meant the Hamiltonian stayed the same and didn't take the bad atoms into account.


pulser/simulation/simulation.py
```diff
@@ -642,19 +642,18 @@ def _build_basis_and_op_matrices(self) -> None:
...
         Args:
-            update_and_extract(bool=True): Whether to update the noise
-                parameters and extract the samples from the sequence.
+            update(bool=True): Whether to update the noise parameters.
         """
-        if update_and_extract:
+        if update:
             self._update_noise()
-            self._extract_samples()
+        self._extract_samples()
```
sebgrijalva, line 656: This is correct, the initial configs (from SP errors) may have changed and so the extraction needs to take place for every run.

sebgrijalva (approved):
Thanks a lot! This fixes the wrong plot in the tutorial.

PR #356, opened 2022-04-07
Address qubits with indices

CdeTerra:
Allow to address qubits with indices, allowing to use integer parameters to select them. Will cover `target` and `phase_shift` functions.


pulser/sequence.py
```diff
@@ -1214,11 +1268,38 @@ def _target(
...
+
+    def _target(
+        self, qubits: Union[Iterable[QubitId], QubitId], channel: str
+    ) -> None:
+        qubits_set = self._precheck_target_qubits_set(qubits, channel)
+        if not self.is_parametrized():
+            self._perform_target_non_parametrized(qubits_set, channel)
+
+    @_store
+    def _target_index(
```
HGSilveri, line 1282: So, the real reason we have `_target` and `_phase_shift` is that we need these to be called without being stored, not that they don't accept parametrized parameters. In this case, does it make sense to do the same or can you just store everything under `target_index()` and store that?

pulser/sequence.py
```diff
@@ -833,6 +781,33 @@ def target(
...
+                of qubit indices, when multi-qubit addressing is possible.
+            channel (str): The channel's name provided when declared. Must be
+                a channel with 'Local' addressing.
+        """
+        if not self.is_parametrized():
+            raise RuntimeError(
+                f"Sequence.{self.target_index.__name__} can't be called in"
+                " non parametrized sequences."
+            )
+
```
HGSilveri, line 806: Do we need to restrict this to when the sequence is parametrized? I agree that it makes more sense to use it when there are variables involved, but it doesn't have to be the case.

HGSilveri (commented):
A first batch of comment, but it's looking pretty good already!

pulser/sequence.py
CdeTerra: I think we agreed to allow the index only for parametrized sequences, because non-parametrized sequences can still use a mappable register, so it will fail to compute the indices.

pulser/sequence.py
CdeTerra: Actually, the reason I did this was to drop the parametrization check from the stored function, since the deferred call happens when it's not parametrized anymore.

pulser/register/mappable_reg.py
```diff
@@ -67,9 +68,50 @@ def build_register(self, qubits: Mapping[QubitId, int]) -> BaseRegister:
             raise ValueError(
                 "All qubits must be labeled with pre-declared qubit IDs."
             )
+        trap_ordered_qubits = {
```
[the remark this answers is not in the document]
CdeTerra: Nice catch, I forgot to rename the variable when I changed the code to calculate the id from the index.

pulser/register/mappable_reg.py
```diff
@@ -67,9 +68,50 @@ def build_register(self, qubits: Mapping[QubitId, int]) -> BaseRegister:
...
+            seq.build(qubits=qubit_map, qubit_indices=indices)
+            ``
+
+        Args:
+            chosen_ids: IDs of the qubits that are chosen to map the
+                MappableRegister
+            id_list: IDs of the qubits to denote
+
+        Returns:
+            Indices of the qubits to denote, only valid for the given mapping.
```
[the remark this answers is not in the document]
CdeTerra: Shouldn't it be automatized, to avoid to repeat what is already in the signature?

pulser/sequence.py
CdeTerra: I think we agreed on allowing the index only for parametrized sequences, but I can't remember why.

HGSilveri: Or rather, `self.is_register_mappable() and not self.is_parametrized()`

pulser/sequence.py
HGSilveri: I think what we want to avoid is that the register is mappable in a non-parametrized sequence, I think we should still allow it for non-parametrized sequences with normal registers (see my other comment)

pulser/sequence.py
CdeTerra: Do you mean, only if `self.is_register_mappable() and not self.is_parametrized()`? Since an index target makes little sense without parametrization, wouldn't it be simpler to keep the only parametrization condition? We may also arbitrarily mark the sequence as parametrized in such cases, even if there are no parameters, but this doesn't waive out the complexity.

CdeTerra: I think it is about the same complexity for the support, but I had in mind the complexity of the documentation for the users. Maybe the best would be to automatically switch the sequence to parametrized every time `target_index` is called with `self.is_register_mappable()`. Thus, it will always work, but will simply switch to parametrized in a small documented edge case.

pulser/register/mappable_reg.py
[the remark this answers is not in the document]
CdeTerra: We may try https://github.com/tox-dev/sphinx-autodoc-typehints in another PR. For now, I will simply rewrite the types as you propose.

pulser/sequence.py
```diff
@@ -1276,9 +1381,14 @@ def _phase_shift(
                     raise ValueError(
                         "All non-variable targets must belong to the register."
                     )
-            return
 
-        elif not set(targets) <= self._qids:
+    def _phase_shift_non_parametrized(
+        self,
+        phi: Union[float, Parametrized],
+        *targets: Union[QubitId, Parametrized],
```
HGSilveri, line 1388: I guess here the args should never be `Parametrized`

pulser/sequence.py
```diff
@@ -1170,11 +1234,23 @@ def draw(
...
+    ) -> Union[Set[int]]:
+        pass
+
+    @overload
+    def _precheck_target_qubits_set(
+        self, qubits: Union[Iterable[QubitId], QubitId], channel: str
+    ) -> Union[Set[QubitId], Set[int]]:
+        pass
+
+    def _precheck_target_qubits_set(
```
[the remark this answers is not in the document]
HGSilveri: You're absolutely right, I meant to delete this comment but forgot to

pulser/sequence.py
```diff
@@ -98,38 +98,53 @@ def wrapper(self: Sequence, *args: Any, **kwargs: Any) -> Any:
...
+                raise ValueError(
+                    f"{x} has variables that don't come from this "
+                    "Sequence. Use only what's returned by this"
+                    "Sequence's 'declare_variable' method as your"
+                    "variables."
+                )
+    elif isinstance(x, Iterable) and not isinstance(x, str):
+        # Recursively look for parametrized objs inside the arguments
+        for y in x:
+            _verify_variable(seq, y)
```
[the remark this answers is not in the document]
HGSilveri: It's a nit, if you prefer to keep it like this I'm okay with that. It's a matter of personal taste, to me it usually makes more sense to have functions that are only used within a certain context to be defined inside that context.

pulser/sequence.py
```diff
@@ -825,8 +840,38 @@ def target(
...
+
         self._target(qubits, channel)
 
+    @_verify_parametrization
+    def target_index(
+        self,
+        qubits: Union[int, Iterable[int], Parametrized],
+        channel: Union[str, Parametrized],
+    ) -> None:
+        """Acts as ``target`` using indices instead of Qubit IDs.
```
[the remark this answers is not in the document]
HGSilveri: Yes, I would prefer it, I think it's better that things are kept self-explanatory. In an IDE, I would prefer that the docstring gave me all the information, instead of me having to go look at the docstring of another function.

pulser/tests/test_sequence.py
```diff
@@ -698,10 +699,116 @@ def test_mappable_register():
         seq.build(qubits={"q0": 1, "q1": 10})
 
     seq_ = seq.build(qubits={"q2": 20, "q0": 10})
-    seq_._last("ryd").targets == {"q2", "q0"}
+    assert seq_._last("ryd").targets == {"q2", "q0"}
```
HGSilveri, line 702: Thanks for spotting this!

pulser/tests/test_sequence.py
```diff
@@ -698,10 +699,116 @@ def test_mappable_register():
...
+        IndexError, match="Indices must exist for the register"
+    ):
+        seq.target_index(20, channel="ch0")
+    with pytest.raises(
+        IndexError, match="Indices must exist for the register"
+    ):
+        seq.phase_shift_index(phi, 20)
+    seq.target_index(index, channel="ch0")
+    seq.phase_shift_index(phi, index)
+    built_seq = seq.build(**build_params)
```
HGSilveri, line 788: Why build in this case?

HGSilveri (approved):
Very nice! Only a couple things left to address, thanks a lot for your help!

pulser/tests/test_sequence.py
CdeTerra: You're welcome!

pulser/sequence.py
```diff
@@ -900,6 +945,25 @@ def phase_shift(
...
         self._phase_shift(phi, *targets, basis=basis)
 
+    @_verify_parametrization
+    def phase_shift_index(
+        self,
+        phi: Union[float, Parametrized],
+        *targets: Union[int, Parametrized],
+        basis: str = "digital",
+    ) -> None:
+        """Acts as ``phase_shift`` using indices instead of Qubit IDs.
```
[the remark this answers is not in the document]
CdeTerra: Oh, I forgot to do it.

pulser/sequence.py
```diff
@@ -958,12 +958,23 @@ def phase_shift_index(
...
-        It is then converted to a Qubit ID using the order in which they
-        were declared when instantiating the ``Register``
-        or ``MappableRegister``.
+        This is equivalent to an :math:`R_z(\phi)` gate (i.e. a rotation of the
+        target qubit's state by an angle :math:`\phi` around the z-axis of the
+        Bloch sphere).
+
+        Args:
+            phi (float): The intended phase shift (in rads).
+            targets (Union[int, str]): The indices of the qubits to apply the
```
HGSilveri, line 969: It cannot be `str`

CdeTerra: Oh, thanks, you're absolutely right.

PR #360, opened 2022-04-11
Updates to `pulser.sampler`

HGSilveri:
- Separates the specific noise definitions from general noise objects to isolate `pulser.sampler` and `pulser.simulation`
- Adds missing license headers

### DIFF
--- a/pulser/sampler/__init__.py
+++ b/pulser/sampler/__init__.py
@@ -1,4 +1,4 @@
-# Copyright 2020 Pulser Development Team
+# Copyright 2022 Pulser Development Team
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


lvignoli (commented):
When the simulation module will be an extra package, one would have to install it to sample a sequence with the amplitude or doppler noises.
It is what we want?

HGSilveri:
> When the simulation module will be an extra package, one would have to install it to sample a sequence with the amplitude or doppler noises. It is what we want?

Yes, I would say so, mainly because sampling with noises is something that's essentially used for simulation. 
Bear in mind, however, that I plan on keeping the default `pulser` installation include the simulation features.

lvignoli (approved):
Ok then!

If the number of noises grows too much, maybe they will deserve their own module.

lvignoli:
Oops sorry I missed it too :/

PR #362, opened 2022-04-12
Splitting `pulser` into `pulser-core` and `pulser-simulation`.

### LINKED ISSUE #358: Splitting the `pulser.simulation` module
## Objectives:
- Have all the features of the `pulser.simulation` module in a separate distribution, thus allowing for a leaner `pulser-core` distribution that's entirely focused on the sequence design aspect.
- Keep existing code working, at least until the next major release (which will be 1.0.0)
- Minimize the need for code adaptations on the user side as much as possible
- `pip install pulser` would still install everything, but now there is the option to `pip install pulser-core`(or similar) for situations where the simulation package is not needed.

### 1. Namespace packages

[Namespace packages](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/) allow for a single package to be split into multiple sub-packages, which can be distributed, installed and used independently.

The advantage of this approach is that we could make `pulser` a namespace package and `pulser-simulation` a subpackage, which would allow us to keep the `from pulser.simulation import ...` imports unchanged. It would also mean that all the other modules would be moved to a `pulser-core` subpackage. We would thus move from this:

```
pulser/
    __init__.py
    setup.py
    simulation/
        __init__.py
        ...
    devices/
        __init__.py
        ...
    sequence.py
    [all the other pulser modules]
```

to this:

```
pulser-simulation/
    setup.py
    pulser/
        simulation/
            __init__.py
            ...

pulser-core/
    setup.py
    pulser/
        devices/
            __init__.py
            ...
        sequence.py
        [all the other pulser modules]
```

When `pulser-simulation` and `pulser-core` are installed, their contents would then be located under the `pulser` namespace package.

As far as I could tell, not being able to have an `__init__.py` means that we have to drop the direct imports from `pulser` (like `from pulser import Sequence`), which would defeat the purpose of keeping existing code working. I haven't experiment with this yet, so perhaps there is a way, but if there is it is obscure. **If there is indeed no way to do this, I would say this makes this approach unviable.**

### 2. Make a separate `pulser_simulation` package

This is a more straightforward solution, which would essentially just extract the simulation module and keep the rest the same. In this case, the file structure would look something like this.

```
pulser-simulation/
    setup.py   # pulser_simulation package setup, with pulser as a dependency (distribution: 'pulser-simulation')
    pulser_simulation/
        __init__.py
        simulation.py
        simresults.py
        ...         

pulser-core/
    setup.py   # pulser package setup (distribution: 'pulser-core')
    pulser/
        devices/
            __init__.py
            ...
        sequence.py
        [all the other pulser modules]
setup.py  # 'pulser' distribution, just installs 'pulser-core' and 'pulser-simulation'
```

As it is, this approach will have the inconvenience of requiring that all imports of `pulser.simulation` be changed to `pulser_simulation`. I think a workaround here is to keep an empty `simulation` module inside `pulser` with only an `__init__.py` containing code that checks if `pulser_simulation` is installed and imports the relevant classes from it (i.e. `Simulation` and `SimConfig`). I think this would be fine for now, I'm not so sure it's something we would like to keep forever or deprecate at some point.

HGSilveri:
## Distributions

Following the motivations and objectives in #358, this PR splits the current `pulser` distribution into two:

- `pulser-simulation`, which installs the `pulser_simulation` package and contains all contents previously in `pulser.simulation`
- `pulser-core`, which installs the `pulser` package containing everything else.

Note that the `pulser` distribution is still kept, but it becames only a meta-package that installs `pulser-core` and `pulser-simulation`.

## Versions

In terms of versioning, I went with a global version for all distributions and packages, which is recorded in `VERSION.txt`.

## Backwards compatibility

Support for importing `Simulation` and `SimConfig` from `pulser.simulation` is kept (if `pulser-simulation` is installed), but a warning is raised recommending imports are switched to `pulser_simulation`. **I'm still considering if we should update all imports on the tutorials right away or not**.

- [ ] Consider updating the `pulser.simulation` imports on the tutorials

## Advice for reviewers

Almost all files were touched, though many of them were only renamed. The commits where files are moved purposefully do not introduce any changes, so it is advised that the changes are reviewed in a commit-by-commit basis, such that the file renames don't pollute the diff.

Closes #358 .


pulser-core/README.md
```diff
@@ -0,0 +1,21 @@
+# Pulser-Simulation
```
bejito, line 1: nit: pulser-core

pulser-core/rtd_requirements.txt
```diff
@@ -0,0 +1 @@
+pulser-core/.
```
bejito, line 1: What is this for ?

bejito (approved):
Minor comment, otherwise LGTM

pulser-core/rtd_requirements.txt
HGSilveri: This is needed for the Read The Docs builder to install `pulser-core`

pulser-core/README.md
HGSilveri: Nice catch, thanks!

HGSilveri:
Thanks @bejito ! Do you think this will have an impact on the work being done on serialization to the abstract representation, or is it okay to merge right away?

sebgrijalva (approved):
Looks good! Thanks for the work here.

bejito:
There shouldn't be any impact on serialization, I think it is safe to merge.

pulser-simulation/setup.py
```diff
@@ -0,0 +1,67 @@
...
+# Overwrites the _version.py for the source distribution (reverted at the end)
+with open(local_version_fpath, "w") as f:
+    f.write(f"__version__ = '{__version__}'\n")
+
+setup(
+    name=distribution_name,
+    version=__version__,
+    install_requires=[
+        f"pulser-core=={__version__}",
+        "qutip>=4.6.3",
```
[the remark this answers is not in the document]
HGSilveri: I would say we leave it to the `pulser-core` installation, unless we have version requirements particular to `pulser-simulation` on those packages, in which case we specify them here.

README.md
```diff
@@ -15,7 +15,7 @@ and also allows the creation of **analog** quantum simulations, outside of the
 scope of traditional quantum circuit approaches. Whatever the type of experiment
 or paradigm, if it can be done on the device, it can be done with Pulser.
 
-Additionally, Pulser features built-in tools for classical simulation (using [QuTiP][qutip] libraries) to aid in the development and testing of new pulse sequences.
+Additionally, the ``pulser_simulation`` extensions provides tools for classical simulation (using [QuTiP][qutip] libraries) to aid in the development and testing of new pulse sequences.
```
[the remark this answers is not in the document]
HGSilveri: Yes, thanks!

PR #365, opened 2022-04-13
Remove string variables

CdeTerra:
Remove support of string variables, and use index functions instead where relevant.


pulser-core/pulser/sequence.py
```diff
@@ -846,7 +846,7 @@ def target(
     def target_index(
         self,
         qubits: Union[int, Iterable[int], Parametrized],
-        channel: Union[str, Parametrized],
+        channel: str,
```
HGSilveri, line 849: Sorry I missed this... 
Also, I think you also need to remove the `Parametrized` option on the regular `target`

CdeTerra:
Thanks Henrique for fixing the tutorial!

HGSilveri (commented):
Looking good! Just needs to make sure all arguments accepting `Parametrized` that would rely on a `str` variable are adjusted

pulser-core/pulser/sequence.py
CdeTerra: Do you mean on the "channel" argument? I think it's already away.

HGSilveri: I meant the `qubits` on `target()`, isn't it still allowing `Parametrized`?

CdeTerra: Do you think we should forbid it? Couldn't it be useful when the QubitId is an integer?

HGSilveri: I think there should be only one way to use variables on a target operation, and that's `target_index`. I would prefer that the `target()` operation stops supporting parametrized objects as arguments now that `target_index` exists.

CdeTerra: Ok. Then, I guess I should also remove the `Parametrized` possibility of the `initial_target` of `declare_channel`, to keep consistency.

HGSilveri: Indeed, please do!

pulser-core/pulser/sequence.py
```diff
@@ -547,6 +545,17 @@ def declare_channel(
...
                 raise ValueError(f"Channel {channel_id} is not available.")
 
+        if initial_target is not None:
+            try:
+                cond = any(
+                    isinstance(t, Parametrized)
+                    for t in cast(Iterable, initial_target)
+                )
+            except TypeError:
+                cond = isinstance(initial_target, Parametrized)
```
HGSilveri, line 555: Shouldn't we raise this TypeError if `cond == False` ? I think something like this would work:
```suggestion
            except TypeError as e:
                cond = isinstance(initial_target, Parametrized)
                if not cond:
                    raise e
```

pulser-core/pulser/sequence.py
```diff
@@ -1290,31 +1292,47 @@ def _precheck_target_qubits_set(
...
         qubits_set = self._precheck_target_qubits_set(qubits, channel)
-        if not self.is_parametrized():
+
+        if self.is_parametrized():
+            for q in qubits_set:
+                if q not in self._qids and not isinstance(q, Parametrized):
+                    raise ValueError(
+                        f"All non-variable qubits must belong to the register."
+                        f" Wrong id: {q!r}"
+                    )
```
HGSilveri, line 1310: Isn't this still allowing the qubits in `qubits_set` to be parametrized? I think the check needs to be done regardless of whether the sequence is parametrized or not

pulser-core/pulser/sequence.py
```diff
@@ -1428,11 +1438,18 @@ def _phase_shift_non_parametrized(
...
-        self._precheck_phase_shift(*targets, basis=basis)
-        if not self.is_parametrized():
+        self._check_basis(basis)
+
+        if self.is_parametrized():
+            for t in targets:
+                if t not in self._qids and not isinstance(t, Parametrized):
+                    raise ValueError(
+                        "All non-variable targets must belong to the register."
+                    )
```
HGSilveri, line 1451: Same here, this is allowing variable targets

tests/test_paramseq.py
```diff
@@ -44,14 +44,30 @@ def test_var_declarations():
         sb.declare_variable("qubits", size=10, dtype=int)
 
 
+def test_parametrized_channel_initial_target():
```
HGSilveri, line 47: Do you mean...
```suggestion
def test_parametrized_sequence_initial_target():
```
?

tests/test_paramseq.py
```diff
@@ -120,8 +135,23 @@ def test_stored_calls():
     with pytest.raises(ValueError, match="targets the given 'basis'"):
         sb.phase_shift_index(var, *q_var)
 
-    with pytest.raises(ValueError, match="non-variable targets must belong"):
+    with pytest.raises(
+        ValueError,
+        match="All non-variable targets must" " belong to the register.",
```
HGSilveri, line 140: nit: You can remove the break in the string

tests/test_sequence.py
```diff
@@ -751,10 +751,10 @@ def test_mappable_register():
 def test_parametrized_index_functions(
     reg, build_params, index, expected_target
 ):
+    phi = np.pi / 4
     seq = Sequence(reg, Chadoq2)
     seq.declare_channel("ch0", "rydberg_local")
     seq.declare_channel("ch1", "raman_local")
-    phi = np.pi / 4
     index_var = seq.declare_variable("index", dtype=int)
     seq.target_index(index_var, channel="ch0")
```
HGSilveri, line 759: Can you add a test where you target an array instead of a single item? Plus, one where the array is subject to some operation like `var_array = var_array + 2` or something like that.

CdeTerra: Alright, I will do this; it will require changing the code to accept iterables of parametrized values.

CdeTerra: Actually, I think this should be in a separate PR. This is a new feature, and is not part of the string variables removal.

CdeTerra: Actually, there will be no need to change the code. I thought meant arrays like `[1, variable, 2]`, but in fact, only so-called variable arrays.

pulser-core/pulser/sequence.py
HGSilveri: I was thinking of the case where you catch a different `TypeError` that doesn't come from the initial target and just ignore it

pulser-core/pulser/sequence.py
HGSilveri: Yes, but what is the `not isinstance(q, Parametrized)` doing?

tests/test_sequence.py
```diff
@@ -855,7 +855,18 @@ def test_multiple_index_targets():
...
+    seq.target_index([0, 1], channel="ch0")
+    assert seq._last("ch0").targets == {"q0", "q1"}
+
+    seq.target_index(var_array, channel="ch0")
+    built_seq = seq.build(var_array=[1, 2])
+    assert built_seq._last("ch0").targets == {"q1", "q2"}
+
+    seq.target_index(var_array + 1, channel="ch0")
+    built_seq = seq.build(var_array=[1, 2])
+    assert built_seq._last("ch0").targets == {"q2", "q3"}
```
HGSilveri, line 872: Great, thanks!

tests/test_paramseq.py
HGSilveri: Ah ok, gotcha

pulser-core/pulser/sequence.py
CdeTerra: Discussed together.

pulser-core/pulser/sequence.py
CdeTerra: Got it, thanks for the remark!

pulser-core/pulser/sequence.py
CdeTerra: Got it, thanks for the remark!

HGSilveri (approved):
Awesome, thanks a lot for your work on this!

Issue #368, opened 2022-05-12
Split`sequence.py` in multiple files


HGSilveri:
The `sequence.py` function is by far the largest in the codebase and could use a refactoring into a module multiple files.

In the [`seq-module`](https://github.com/pasqal-io/Pulser/tree/seq-module) branch, I have already split what was straightforward to split. Still, the file remains quite large, so I propose the following tasks:

- [x] Assess what other parts of the code can naturally be moved out of the `Sequence` class
- [x] Move out the identified parts
- [x] Adapt the autodocs generation settings to the new file structure

PR #373, opened 2022-05-18
Feat: deserialize abstract repr for non param sequences (+refacto abstract serialization)

bejito:
## Deserializer

`Sequence.from_abstract_rep` allows users to deserialize strings in the abstract JSON format into Pulse sequences. For now, only non-parametrized sequences have been implemented. The deserializer starts by validating the input JSON string against the JSON Schema, so we need to make sure the version of the JSON Schema is up-to-date in Pulser. Later, we might put the schema in a separate github repo and import it in both Pulser and Vulcan.

note: `Sequence.abstract_repr` is now called `Sequence.to_abstract_repr`, and the deserializer is in `Sequence.from_abstract_repr`

## Refacto

In order not to clog the sequence.py file any more, some refactoring was done:
- All the logic related to abstract serialization and deserialization was moved to  `pulser.json.abstract_repr`. 
- In order to avoid nasty circular dependencies, the imports are **inline** in `Sequence.to_abstract_repr`. This means that the modules will be loaded only when the functions are called for the first time. The modules will then be in `sys.modules`, so any subsequent call to the function won't require an module reload. Once #368 is done, we should be able to have a cleaner split that doesn't raise circular import errors and we should be able to get rid of these inline imports.

Next steps: parametrized sequences


HGSilveri:
> * In order to avoid nasty circular dependencies, the imports are **inline** in `Sequence.to_abstract_repr`. This means that the modules will be loaded only when the functions are called for the first time. The modules will then be in `sys.modules`, so any subsequent call to the function won't require an module reload. Once [Split`sequence.py` in multiple files #368](https://github.com/pasqal-io/Pulser/issues/368) is done, we should be able to have a cleaner split that doesn't raise circular import errors and we should be able to get rid of these inline imports.

What circular dependencies were giving you problems? I tried to replicate the issue locally by placing the import at the top and it worked fine.

bejito:
Weird ! If you remove both inline imports, put them at the top and then run the tests you have no errors ?
One example: pulser.sequence -> pulser.json.abstract_repr.deserializer -> pulser.sequence

HGSilveri:
> Weird ! If you remove both inline imports, put them at the top and then run the tests you have no errors ?
> One example: pulser.sequence -> pulser.json.abstract_repr.deserializer -> pulser.sequence

I had tried with only the serializer and it worked, it seems that the problem is only the deserializer. I did some small changes to make it work as well, do you mind if I push them here?

bejito:
Yes, go ahead !

pulser-core/pulser/json/abstract_repr/deserializer.py
```diff
@@ -0,0 +1,157 @@
...
+    """Deserialize a sequence from an abstract JSON object.
+
+    Args:
+        obj_str (str): the JSON string representing the sequence encoded
+            in the abstract JSON format.
+
+    Returns:
+        Sequence: The Pulser sequence.
+    """
+    pass
```
HGSilveri, line 118: Did you perhaps forget this here?

pulser-core/pulser/json/abstract_repr/deserializer.py
```diff
@@ -0,0 +1,157 @@
...
+
+    # Variables
+    vars = {}
+    for name, desc in obj["variables"].items():
+        v = seq.declare_variable(
+            cast(str, name),
+            size=len(desc["value"]),
+            dtype=VARIABLE_TYPE_MAP[desc["type"]],
+        )
+        vars[name] = v
```
HGSilveri, line 147: What's this dict for?

HGSilveri (commented):
Sorry this took so long! Looks okay to proceed, I only a few minor remarks

pulser-core/pulser/json/abstract_repr/deserializer.py
```diff
@@ -0,0 +1,157 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Deserializer from JSON in the abstract representation."""
+from __future__ import annotations
+
+import json
+from typing import TYPE_CHECKING, cast
+
+import jsonschema
```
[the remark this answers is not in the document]
bejito: I thought it was already there ! I'll add it to pulser-core dependencies, it's pretty lightweight

pulser-core/pulser/json/abstract_repr/deserializer.py
bejito: indeed...

pulser-core/pulser/json/abstract_repr/deserializer.py
bejito: This dict will be used for keeping a ref of all variables in order to use them during the deserilization of param objects. Not used yet though

HGSilveri:
@bejito I enabled the CI tests on this branch. Let me know if you would like me to take over on the failing ones.

tests/test_abstract_repr.py
```diff
@@ -201,3 +216,614 @@ def test_exceptions():
...
+                    },
+                    "detuning": {
+                        "kind": "constant",
+                        "duration": 1000,
+                        "value": 42,
+                    },
+                }
+            ],
+            variables={
+                "var1": {"type": "int", "value": [1.5]},
```
HGSilveri, line 772: Same here

tests/test_abstract_repr.py
```diff
@@ -201,3 +216,614 @@ def test_exceptions():
...
+        seq_var1 = seq._variables["var1"]
+
+        # init + declare channels + 1 operation
+        offset = 1 + len(s["channels"])
+        assert len(seq._calls) == offset
+        # No parametrized call
+        assert len(seq._to_build_calls) == 1
+
+        c = seq._to_build_calls[0]
+        pulse: Pulse = c.kwargs["pulse"]
```
HGSilveri, line 786: Isn't this a `ParamObj`?

tests/test_abstract_repr.py
```diff
@@ -201,3 +216,614 @@ def test_exceptions():
...
+            {"expression": "sin", "lhs": {"variable": "var1"}},
+            {"expression": "cos", "lhs": {"variable": "var1"}},
+            {"expression": "tan", "lhs": {"variable": "var1"}},
+            {"expression": "index", "lhs": {"variable": "var1"}, "rhs": 0},
+            {"expression": "add", "lhs": {"variable": "var1"}, "rhs": 0.5},
+            {"expression": "sub", "lhs": {"variable": "var1"}, "rhs": 0.5},
+            {"expression": "mul", "lhs": {"variable": "var1"}, "rhs": 0.5},
+            {"expression": "div", "lhs": {"variable": "var1"}, "rhs": 0.5},
+            {"expression": "pow", "lhs": {"variable": "var1"}, "rhs": 0.5},
+            {"expression": "mod", "lhs": {"variable": "var1"}, "rhs": 2},
```
HGSilveri, line 746: I think it would be better to use the actual `var1` (ie the VariableItem rather than the array of size 1) in at least some of these.

tests/test_abstract_repr.py
```diff
@@ -201,3 +216,614 @@ def test_exceptions():
...
+        "register": [
+            {"name": "q0", "x": 0, "y": 2},
+            {"name": "q42", "x": -2, "y": 9},
+            {"name": "q666", "x": 12, "y": 0},
+        ],
+        "channels": {"digital": "raman_local", "global": "rydberg_global"},
+        "operations": operations or [],
+        "variables": variables or {},
+        "measurement": None,
+    }
```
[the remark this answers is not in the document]
HGSilveri: I pushed a function to check the serialization-deserialization roundtrip, WDYT?

pulser-core/pulser/parametrized/paramobj.py
```diff
@@ -316,3 +316,11 @@ def __str__(self) -> str:
         else:
             name = self.cls.__name__
         return f"{name}({', '.join(args+kwargs)})"
+
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, ParamObj):
+            return False
+        return self.args == other.args and self.kwargs == other.kwargs
+
+    def __hash__(self) -> int:
```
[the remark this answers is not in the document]
bejito: This is used to make an object hashable in order to be used as a key in a dict. 
The default for `__hash__` for an object is to return its `id()`, but that default gets overridden when `__eq__` is redefined, so this is just a matter of setting that up again.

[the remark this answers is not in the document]
HGSilveri: Ahh ok. I didn't find it being used as a dict key anywhere, so I wasn't sure why you had defined it. Good to know about the `__eq__` overwritting

tests/test_abstract_repr.py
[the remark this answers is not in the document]
bejito: Looks good IMO. We could also add the other way around: Seq -> JSON -> Seq and see if we have the same seq, but equality is harder to assess then ?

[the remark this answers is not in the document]
HGSilveri: Yes, much harder unfortunately

PR #376, opened 2022-06-05
Add sphinx typehints auto generation [unitaryhack]

### LINKED ISSUE #359: Automatize docstrings types
Currently, functions' parameters and return types are precised in the docstrings. It is redundant, because they are already given as typehints.

```python
    @_store
    def target(
        self,
        qubits: Union[QubitId, Iterable[QubitId], Parametrized],
        channel: str,
    ) -> None:
        """Changes the target qubit of a 'Local' channel.

        Args:
            qubits (Union[int, str, Iterable]): The new target for this
                channel. Must correspond to a qubit ID in device or an iterable
                of qubit IDs, when multi-qubit addressing is possible.
            channel (str): The channel's name provided when declared. Must be
                a channel with 'Local' addressing.
        """
```

It would be great to remove these repetitions, by automatically inferring the types in the docs from the typehints. We might for example try to use the library https://github.com/tox-dev/sphinx-autodoc-typehints for that purpose.

WingCode:
closes #359


HGSilveri (commented):
Looking great, but it does seems like not all type hints are appearing in the build (you can check the Readthedocs build on the CI tests to see the docs build). I'm not sure what's making some work and others not, perhaps changing some of the [options of `sphinx_autodoc_typehints`](https://pypi.org/project/sphinx-autodoc-typehints/)?

CdeTerra:
This looks good @WingCode, many typehints appear automatically. Unfortunately, there are still some of them that don't show up, for example: https://pulser--376.org.readthedocs.build/en/376/apidoc/creation.html#pulser.sequence.Sequence.declare_channel

It seems that autodoc typehints doesn't like where `Args` is not the first label.

pulser-core/pulser/sequence.py
```diff
@@ -514,12 +514,12 @@ def declare_channel(
...
         Args:
-            name (str): Unique name for the channel in the sequence.
-            channel_id (str): How the channel is identified in the device.
+            name: Unique name for the channel in the sequence.
+            channel_id: How the channel is identified in the device.
                 Consult ``Sequence.available_channels`` to see which channel
                 ID's are still available and the associated channel's
                 description.
-            initial_target (Optional[Union[int, str, Iterable]]): For 'Local'
+            initial_target: For 'Local'
```
HGSilveri, line 522: You should fill this line, this might be causing troubles too

pulser-core/pulser/sequence.py
```diff
@@ -627,14 +627,14 @@ def declare_variable(
...
 
         Args:
-            name (str): The name for the variable. Must be unique within a
+            name: The name for the variable. Must be unique within a
                 Sequence.
 
-        Keyword Args:
-            size (Optional[int]=None): The number of entries stored in the
+        Args:
+            size: The number of entries stored in the
```
HGSilveri, line 634: Here you ended up with two `Args` section and an half-filled line

pulser-core/pulser/sequence.py
```diff
@@ -457,7 +457,7 @@ def config_slm_mask(self, qubits: Iterable[QubitId]) -> None:
         """Setup an SLM mask by specifying the qubits it targets.
 
         Args:
-            qubits (Iterable[QubitId]): Iterable of qubit ID's to mask during
+            qubits: Iterable of qubit ID's to mask during
```
HGSilveri, line 460: Half-filled line here too

pulser-core/pulser/sequence.py
```diff
@@ -835,10 +835,10 @@ def target(
         """Changes the target qubit of a 'Local' channel.
 
         Args:
-            qubits (Union[int, str, Iterable]): The new target for this
+            qubits: The new target for this
```
HGSilveri, line 838: Here too

pulser-core/pulser/sequence.py
```diff
@@ -852,14 +852,14 @@ def target_index(
         """Changes the target qubit of a 'Local' channel.
 
         Args:
-            qubits (Union[int, Iterable, Parametrized]): The new target for
+            qubits: The new target for
```
HGSilveri, line 855: And here

HGSilveri (commented):
I have no idea if this will actually help, but there seems to be some correlation between these and the cases where the automatic typing fails, so it's worth the shot.

WingCode:
Nice catch @HGSilveri. Thank you! Let me check this out and get back

WingCode:
@HGSilveri I tried filling lines, changing docstring lines, remove variables from docstring, changing types of Pulser object to str but alas it doesn't fix the issues unfortunately. Do you have another suggestions that I can try?

pulser-core/pulser/sequence.py
HGSilveri: It seems you missed this

HGSilveri:
Hey @WingCode , thanks for your efforts! After comparing the successful with the unsuccessful cases, I think the problem is coming from the `annotations` which we typically import from `__future__` but can't do so here. I think that if we start building the docs on Python 3.9 most of the failed cases should go away.

Can you change [this line](https://github.com/pasqal-io/Pulser/blob/b551e406a0369daceb2bb2d082977d6b6eab36cf/.readthedocs.yml#L14) to `3.9` and try again?

WingCode:
@HGSilveri It works! Thank you for your patience and constant support.

HGSilveri (commented):
Looking much better! Other than the comments below, the thing that is missing is removing the return types from the docstrings too. Otherwise, what's written in the docstring will override the auto-generated return type

docs/source/conf.py
```diff
@@ -39,8 +39,12 @@
     "sphinx.ext.autodoc",
     "sphinx.ext.mathjax",
     "sphinx.ext.napoleon",
+    "sphinx_autodoc_typehints",
 ]
 
+# Show the defaults for the keyword args.
+typehints_default = "comma"
```
HGSilveri, line 46: This doesn't seem to be having the desired effect. It is my bad, I gave the wrong name for the variable, it's actually
```suggestion
typehints_defaults = "comma"
```

Also, can you please place this more to the end of the file, after the configuration parameters for the other extensions that come first?

docs/source/conf.py
```diff
@@ -73,3 +74,6 @@
 html_css_files = [
     "css/max_width.css",
 ]
+
+# Show the defaults for the keyword args.
+typehints_defaults = "comma"
```
HGSilveri, line 79: The lack of a newline at the end is throwing a formatting error with black. In any case, I was thinking this would fit better after line 60 instead

HGSilveri (approved):
Looks great! Only a couple of return type annotations got left behind, once you fix those it's ready to merge. Thanks, @WingCode !

Issue #315, opened 2022-02-02
Add option to simulate a Sequence with the modulated output


HGSilveri:
#312 introduces output modulation into Pulser. The next logical step is to use the expected output to get more reliable results out of a simulation. Important notes:

- The option to choose between using the input or the output samples should be given upon initialization of `Simulation`. For consistency with the current behaviour, the default should be to use the input.
- In principle, this change should only affect the samples that are extracted and used for simulation.
- Considering pulses in isolation can lead to mistakes, because `modulation(pulse1 + pulse2) != modulation(pulse1) + modulation(pulse2)`. 
- The correct approach is to consider the full channel output and modulate that, as is done when drawing the sequence. This is fine for global channels, but can be tricky to do for local channels because the samples are separated according to the target qubit. Note, however, that a sequence of pulses in between two target operations can be safely considered in isolation because the necessary fall time is included.

lvignoli:
Modulation is now handled at the sampler level with the merging of #345.
Issue #349 tracks the progress of using the sampler in simulations.

HGSilveri:
Something to take into account here: the behaviour of the SLM mask in the presence of modulation is still ill-defined, so it was decided that for now we will just not support the simulation of sequences with modulated output and an SLM mask. 

Practically, this implies that in `Simulation.__init__()`, if the user chooses the option to use the modulated output, we check whether the SLM mask is configured and raise a `NotImplementedError` if so.

PR #379, opened 2022-06-21
WIP: simulation with samples from pulser.sampler.sample()

lvignoli:
Draft PR for substituting the legacy sampling in the `Simulation` class with the dedicated `pulser.sampler` module.

It is in particular a prerequisite to simulation of modulated sequence (see #315).

Blindly substituting the core of `Simulation._extract_samples()` with `pulser.sampler.sample()` does not work: weirdly the final state of the two simulations differ by the number of levels considered, the new one outputting a 3-level state.
Also, restricted to the first 2 components, they are numerically not the same at 1e-3 accuracy.

I have been unable to pinpoint where the divergence occurs.

Run the notebook `tutorials/simulation_with_sampler.ipynb` to see the issue in action.

### DIFF
--- a/pulser-core/pulser/sampler/sampler.py
+++ b/pulser-core/pulser/sampler/sampler.py
@@ -107,6 +107,16 @@ def sample(
     # format the samples in the simulation dict form
     d = _write_dict(seq, samples, addrs)
 
+    # Replace empty defaultdict with empty dicts.
+    # NOTE(lvignoli): this is the format required by the simulation module. We
+    # keep the above code for flexibility in the construction of QubitSamples
+    # and for other samples formats.
+    return _default_to_regular(d)
+
+
+def _default_to_regular(d: dict | defaultdict) -> dict:
+    if isinstance(d, defaultdict):
+        d = {k: _default_to_regular(v) for k, v in d.items()}
     return d
 
 
--- a/pulser-core/pulser/sampler/sampler.py
+++ b/pulser-core/pulser/sampler/sampler.py
@@ -33,6 +33,7 @@ from pulser.sequence import Sequence, _TimeSlot
 def sample(
     seq: Sequence,
     modulation: bool = False,
+    additional_final_sample: bool = False,
     common_noises: Optional[list[NoiseModel]] = None,
     global_noises: Optional[list[NoiseModel]] = None,
 ) -> dict:
@@ -101,6 +102,14 @@ def sample(
                     ti, tf = seq._slm_mask_time[0], seq._slm_mask_time[1]
                     s[i].amp[ti:tf] = 0.0
                     # apply only on amp since it's just a shutter
+
+        # Add final 0 samples to comply with the current format of the
+        # simulation module.
+        if additional_final_sample:
+            for sample in s:
+                sample.amp = np.append(sample.amp, 0.0)
+                sample.det = np.append(sample.det, 0.0)
+                sample.phase = np.append(sample.phase, 0.0)
 
         samples[ch_name] = s
 
--- a/pulser-simulation/pulser_simulation/simconfig.py
+++ b/pulser-simulation/pulser_simulation/simconfig.py
@@ -90,6 +90,8 @@ class SimConfig:
     doppler_sigma: float = field(
         init=False, default=KEFF * np.sqrt(KB * 50.0e-6 / MASS)
     )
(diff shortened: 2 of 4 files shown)


HGSilveri:
Thanks! I'll take a look

HGSilveri:
@lvignoli I found the culprit: `Simulation._build_basis_and_op_matrices()` relies on the dictionary for a given basis being empty to determine whether to ignore that basis (e.g. if `samples["Global"]["digital"] == {}`, it will make a two-level system with just the `ground-rydberg` basis). 
The problem is that the sampler is using a `defaultdict`, so when you check for e.g. `samples["Global"]["digital”]` it populates it (with arrays of 0s, but still) and so assumes the `digital` basis is also being used.

HGSilveri:
> Actually this can't be the fix, since any key is valid in the `defaultdict`, and even read operation modify the `defaultdict` by adding a zero value for the read keys.
> 
> Why not simply getting rid of the `defaultdict`? I am unsure of the exact reason they are here in the first place, probably because it was much easier to test since it provides default zero values.

I think they make for a cleaner implementation by avoiding the recurrent check of whether a given key is already in the dictionary or not, but this shouldn't come at the expense of complicating things down the road. 

In this case, however, I would maybe add that relying on the samples to determine the basis being used is not the most robust approach here. I would maybe suggest we change `Simulation._build_basis_and_op_matrices()` to look at the bases of the declared channels instead.

lvignoli:
I back the change for `Simulation._build_basis_and_op_matrices()`.

But I am still a bit concern about the defaultdicts, since

```python
_ = samples["Global"]["whatever-channel-🦄"]["amp"]
```

is a valid call to the `samples` dictionary, returning an array of zeros. Worse, everytime one tries to read a non-existing value, it instantiates a numpy array of zeros for "amp", "det" and "phase" for this new key, and any future loop over the keys will have to deal with this.

It is tricky stuff…

HGSilveri:
> But I am still a bit concerned about the defaultdicts, since
> 
> ```python
> _ = samples["Global"]["whatever-channel-🦄"]["amp"]
> ```
> 
> is a valid call to the `samples` dictionary, returning an array of zeros. Worse, everytime one tries to read a non-existing key, it instantiates a numpy array of zeros for "amp", "det" and "phase" for this new key, and any future loop over the keys will have to deal with this.
> 
> It is tricky stuff…

Yes, I understand your concerns. What if you convert the `defaultdict`s to `dict`s right before you return the samples?

lvignoli:
We could do that, we have to introspect the samples nested dict for non-zero basis keys, discard them if the three amp, let and phase samples are all zeros, and then assign the regular dict with retained keys in place.

It is not that bad, but given the supplementary intricacy, I seriously reconsider the use of `defaultdict` in the first place, if, aside from testing, it is just to get rid of them with overcomplicated code in the end.

HGSilveri:
> We could do that, we have to introspect the samples nested dict for non-zero basis keys, discard them if the three amp, det and phase arrays are all zeros, and then assign the regular dict with retained keys in place.

I don't think it needs to be that complicated. Wouldn't [this simple solution](https://stackoverflow.com/questions/26496831/how-to-convert-defaultdict-of-defaultdicts-of-defaultdicts-to-dict-of-dicts-o) do the trick?

lvignoli:
Okay it sounded bad to my ears when describing it but yeah it is not that bad one you just _write_ it

lvignoli:
Still, it's weird to have them just to discard them

Do you find a place in the `pulser.sampler` code where we actually benefit from having a default zero value for non existing keys?

lvignoli:
Okay it makes the [`_write_dict()`](https://github.com/pasqal-io/Pulser/blob/2b49cf4b6156ed3724777fcc56b08bf07108edfe/pulser-core/pulser/sampler/sampler.py#L141) function straightforward, but that's it for me

lvignoli:
@HGSilveri all good on the `defaultdicts` now.

Still puzzled by the additional zero sample, should we revive the eternal discussion with @sebgrijalva?

lvignoli:
Added the additional 0 samples through the `additional_zero_sample` flag.

Still miss tests.

HGSilveri (commented):
Looks pretty solid!

HGSilveri:
A new PR will be opened incorporating the changes introduced in #388

PR #384, opened 2022-07-04
Bugfix in simulation with SLM mask

HGSilveri:
This fixes a bug that appeared when simulating a sequence where an SLM mask has a target that is never addressed by the local channel. The accompanying test is an example where a `KeyError` would be thrown without this change.


darcangelomauro (approved):
good spot, looks good to me

Issue #349, opened 2022-03-22
Sequence sampler feature extension


lvignoli:
The sequence sampler added by #345 deserves the following features

- [ ] a notebook to demonstrate its use and extend the User Guide
- [x] multiple output format: for now it give only the nested dict form required by the simulation module
- [x] a better way to specify the sampling-related noises
- [x] add the possibility to specify the sampling rate

It would then be possible to use it in the `simulation` module and the `_seq_drawer()` function.

arthurfaria:
I would like to work in this issue

lvignoli:
Hi @arthurfaria! Thank you for your interest. The sampler is currently being reworked in PR #388, so this issue is not up do date.
We'll update the sampler-related issue after #388, you might be interesting in working on it after that 😃.

HGSilveri:
Very cool to know you're interested, @arthurfaria ! It is as @lvignoli says, we're currently reworking the sampler. I can always ping you once that's done.

This issues actually encompasses multiple working points, are you interested in any one in particular?

arthurfaria:
Hi @HGSilveri and @lvignoli, thanks for the reply. Well, right now I am more interested in the simple ones. I have lately grown very interested in your open-source and working on an open `good first issue` would help me very much to grasp better the ideas behind it. So among those tasks, which one do you think is the best to start with?

lvignoli:
Sure, writing an example notebook is the obvious entry point to the code.
Probably #135 is a safer route, since it does not rely on current WIP.
In any case @arthurfaria your contributions are welcomed 🙂.

arthurfaria:
alright! thanks a lot

HGSilveri:
@arthurfaria Can you please leave a comment on PR #367 so that I can assign it to you?

arthurfaria:
Of course: Hi!

HGSilveri:
> Of course: Hi!

Thanks, but can you do it on that issue #367 instead? It is more focused on what you'll be doing

HGSilveri:
Update on this issue:
- The notebook will be addressed in issue #367 
- The new sampler introduced in #388 allows for the different output formats desired
- In #390, noise and sampling rate are handled externally by the `Simulation` class (because that's the only place they're needed), so they don't need to be incorporated in the sampler.

With that being said, I think we can consider this issue concluded, so I'll close it.

PR #385, opened 2022-07-07
Flexible Sequence drawing

### LINKED ISSUE #332: More flexible sequence drawing
Currently, sequence drawing always has a curve for amplitude and another for detuning, in spite of the fact that detuning is sometimes not used and thus could be omitted. 
On the other hand, jumps in phase could be sometime that would be useful to track, especially considering that it is also affected by channel modulation.

The idea here is to enable the drawing function to assess which quantities and adapt accordingly. It will require a bit of refactoring because the drawing function is currently built around the fact that the number of subplots per channel is fixed, which will no longer be the case.

HGSilveri:
- Makes the detuning curve appear only when the detuning is being used (i.e. is not zero throughout the channel)
- Adds the option to draw the phase in a curve (`draw_phase_curve`) in drawing methods
- Simplifies the color scheme for programmed vs modulated waveforms by using a hatch on output waveforms instead.

Examples:
- Bell state creation sequence, which uses only resonant pulses so the detuning curve is automatically omitted
image

- The phase curves show up when `draw_phase_curve=True. The detuning is still omitted if not used (as happens for  the`"raman_local"` channel)
image

**To Do:**
- The phase curves are for now being extracted by the `_seq_drawer.gather_data()` function, as is still the case for the amplitude and detuning. As outlined in #349, we would like these to come from the sampler instead. 
- There is some freedom in the moment the phase is changed that can be used to minimize the number of phase jumps and the effects of modulation. This should also be done centrally by the sampler.

Closes #332 .


pulser-core/pulser/_seq_drawer.py
```diff
@@ -154,6 +174,16 @@ def phase_str(phi: float) -> str:
     data = gather_data(seq)
     total_duration = data["total_duration"]
     time_scale = 1e3 if total_duration > 1e4 else 1
+    curves_per_ch = {}
+    for ch in seq._schedule:
+        curves_per_ch[ch] = [True, False, False]
+        if np.nonzero(data[ch]["detuning"])[0].size > 0:
+            curves_per_ch[ch][1] = True
+        if draw_phase_curve and np.nonzero(data[ch]["phase"])[0].size > 0:
+            curves_per_ch[ch][2] = True
```
CdeTerra, line 183: Here, the correspondence between the indices and "detuning/phase" seems magic. It might be good to have it at one common place.

pulser-core/pulser/_seq_drawer.py
```diff
@@ -273,77 +298,96 @@ def phase_str(phi: float) -> str:
...
 
-    for ch, (a, b) in ch_axes.items():
+    for ch, axes in ch_axes.items():
         ch_obj = seq._channels[ch]
         basis = ch_obj.basis
         times = np.array(data[ch]["time"])
         t = times / time_scale
-        ya = data[ch]["amp"]
-        yb = data[ch]["detuning"]
+        ys = [data[ch][qty] for qty in ("amp", "detuning", "phase")]
```
CdeTerra, line 306: "amp/detuning/phase" seem to be magic, it would be nice that they be clear from a structure returned by `gather_data`.

pulser-core/pulser/_seq_drawer.py
```diff
@@ -88,12 +103,14 @@ def gather_data(seq: pulser.sequence.Sequence) -> dict:
             time += [time[-1] + 1, total_duration - 1]
             amp += [0, 0]
             detuning += [0, 0]
+            phase += [phase[-1] if len(phase) else 0] * 2
         # Store everything
         time.pop(0)  # Removes the -1 in the beginning
         data[ch] = {
```
CdeTerra, line 109: Maybe a real data structure would improve the code readability. It could additionally integrate some logic. For example:

```python
class ChannelDrawContent:
    time: list[int]
    amp: list[float]
    detuning: list[float]
    phase: list[float]
    target: list[dict[Union[str, tuple[int, int]], Any]]

    def curves(self, draw_phase_curve: bool): ...
```

CdeTerra (approved):
Some general style remarks, to be discussed and that could be applied later.

HGSilveri:
@CdeTerra I agree with your remarks in general, I admit I tend to be more careless with functions for plotting because I don't expect people to read their contents often, but it's a bad habit. I'll give your suggestions a shot before merging, if I find they're not so hard to implement I'll still include them here.

pulser-core/pulser/_seq_drawer.py
```diff
@@ -594,11 +608,11 @@ def phase_str(phi: float) -> str:
             if ax_lims[i][0] < 0:
                 ax.axhline(0, **hline_kwargs)
 
-        if "interp_pts" in data[ch] and draw_interp_pts:
-            all_points = data[ch]["interp_pts"]
-            for i, qty in enumerate(("amplitude", "detuning")):
-                if qty in all_points and curves_per_ch[ch][i]:
-                    pts = np.array(all_points[qty])
-                    axes[i].scatter(pts[:, 0], pts[:, 1], color=COLORS[i])
+        if ch_data.interp_pts and draw_interp_pts:
```
CdeTerra, line 611: `if draw_interp_pts` would probably be enough, since `qty in ch_data.interp_pts` will always be `False` if `ch_data.interp_pts` is empty.

PR #386, opened 2022-07-11
Automatic packaging and CI test updates

### LINKED ISSUE #363: Automatic packaging
Once #362 goes through, it would be nice to have some automatic packaging routine and some CI test on it.

The packaging process is a bit unorthodox because we have three different PyPI distributions under the same repository:
- `pulser-core`, which contains the `pulser` package
- `pulser-simulation`, which contains the `pulser_simulation` package
- `pulser`, which has no package in itself but installs a specific version of `pulser-core` and `pulser-simulation`

AFAIK, [`cirq`](https://github.com/quantumlib/Cirq) has a similar setup, so it might be worth looking into for inspiration.

HGSilveri:
- Separates the `requirements` and `dev_requirements`
- Refactors the `build` workflow to install only the necessary packages for each job
- Creates a `test` workflow to run the unit tests across different OSs and python versions (triggered only on push events to `master` and `develop`)
- Adds a workflow to ensure the validity of the version numbers on `master` and `develop`
- Adds a workflow to build and publish and check stable releases to PyPI
- Documents the release process in detail

Closes #363 .


AdrienFontaine-PASQAL:
Since I'm not familiar with Github Actions I had a quick question about the following piece of code (used for instance in `.github/workflows/ci.yml`):
```
on:
  pull_request:
  push:
    branches:
      - master
      - develop
```
Do we allow pushing commits directly to those branches (if so why/when)? Or are they "push protected" in a way? Maybe is this CI rule used for tagging the release commit?

The `release.md` only address tagging on `master`, so maybe the push rule is for detecting a PR that got merged?

HGSilveri:
> Since I'm not familiar with Github Actions I had a quick question about the following piece of code (used for instance in `.github/workflows/ci.yml`):
> 
> ```
> on:
>   pull_request:
>   push:
>     branches:
>       - master
>       - develop
> ```
> 
> Do we allow pushing commits directly to those branches (if so why/when)? Or are they "push protected" in a way? Maybe is this CI rule used for tagging the release commit?
> 
> The `release.md` only address tagging on `master`, so maybe the push rule is for detecting a PR that got merged?

Hi @AdrienFontaine-PASQAL , thanks for taking the time to look into this! 
Those branches are "push protected" except for collaborators with Admin rights. The "push" case will be triggered every time a PR is merged into one of these branches (which is admittedly redundant because the PR needs to pass all these tests to be merged) or on the rare occasions when an Admin does a direct push (what we want to catch). 
There is no relation with tagging on this particular workflow.

AdrienFontaine-PASQAL (approved):
As far as my expertise go on Github Actions this sounds good to me.

release.md
```diff
@@ -0,0 +1,73 @@
...
+
+Currently (as of July 2022), Pulser is still in an early development phase and has a MAJOR of 0 - as such, both breaking and backwards compatible changes and additions to the API will be introduced through increments in the MINOR until version 1.0.0 is released. 
+
+During this phase, only two type of releases are envisioned:
+
+- A scheduled release, where the MINOR is bumped and the PATCH is reset (`0.{x}.{y} -> 0.{x+1}.0`)
+- A hotfix, where the PATCH is bumped (`0.{x}.{y} -> 0.{x}.{y+1}`)
+
+Only releases are tracked and tagged in the `master` branch, while development is done in the `develop` branch. To signal this, the version in the `develop` branch should always be one MINOR ahead of `master` and follow the `MAJOR.{MINOR+1}devPATCH` format (e.g. if the latest release tagged in `master` was `0.4.3`, then the version in `develop` should be `0.5dev3`). Through this format, we mark which release is under development and how many patches have occured since its development started (which tells us how many times we brought in changes done directly in `master` through an hotfix).
```
CdeTerra, line 15: Shouldn't it be `MAJOR.{MINOR+1}dev{PATCH}`?

release.md
```diff
@@ -0,0 +1,73 @@
...
+
+In the release branch, no other features can be added. Changes to the documentation and bug fixes are allowed, but should only be done when the development in the `develop` branch needs to continue while the release is being prepared; otherwise, do all the changes in `develop` before checking out the release branch. Note that the release branch will ultimately be *merged* to the `master` branch *without being squashed*, so keep the ammount of commits small and document them well to preserve the quality of the history.
+
+Crucially, the release branch must feature a commit changing the development version in `VERSION.txt` to the desired version of the release. For a minor release, this should be of the form `{x}.{y}dev{z} -> {x}.{y}.0`. 
+
+Finally, open a PR from the `release/v{x}.{y}.{z}` branch to `master`, have someone review and accept the changes introduced in the `release` branch (all the changes done in `develop` will be there as well, but those have already been reviewed) and merge the branch to `master` **without squashing the commits**.
```
CdeTerra, line 28: "In the `release` branch": the code quotes might be removed to keep consistency with former sentences that refer to it as "release branch" (not "`release` branch"). It also helps to understand that it is not about a branch named "release" (but actually "release/v{x}.{y}.{z}").

release.md
```diff
@@ -0,0 +1,73 @@
...
+
+
+## Preparing a hotfix
+
+Unlike with a scheduled release, a hotfix serves only to fix bugs found in the latest release. The hotfix branch must be branched out from `master` and feature only the changes required to fix any bugs.
+
+Along with the bug fixes, the hotfix branch must also have a commit updating the version with an increment of the PATCH, ie `{x}.{y}.{z} -> {x}.{y}.{z+1}`.
+
+When ready, open a PR to merge the hotfix branch to `master` and, once that is reviewed and accepted, **squash and merge the commits** (note the difference with respect to the scheduled release procedure).
```
CdeTerra, line 37: From my understanding, this squash rule is new (just like the naming convention), because v0.5.1 did not apply it (0731ae4, 9304b65 and 5bf2d57 all appear in the master's commit history).

CdeTerra:
Great work, and thanks for the nice and clear documentation! I only have a few tiny comments.

release.md
HGSilveri: Yes, you're right

release.md
HGSilveri: Good catch!

release.md
HGSilveri: There was no explicit rule before, so it is true you will find violations of these rules in the past history.
That being said, these commits don't all point to the squash rule you're referring to:
- [0731ae4](https://github.com/pasqal-io/Pulser/commit/0731ae486fe7056ce7be344fbb811d100b0e47f9) : This is an hotfix + dev version bump which was squashed and merged back to develop. Following these rules, it would still appear in the `master` history, with two differences: 1) the version would be `0.6dev1`; 2) it would not be squashed, so the hotfix changes would appear in a separate commit from the version bump
- [9304b65](https://github.com/pasqal-io/Pulser/commit/9304b651587568044076241948d17993e7364355) & [5bf2d57](https://github.com/pasqal-io/Pulser/commit/5bf2d57eb8692adc16a13a1c6f4b962705b6fb61) : These refer to what you mention, the version bump was done separately from the hotfix, while now it would be done together in the same squashed commit

.github/workflows/ci.yml
```diff
@@ -12,83 +12,60 @@ jobs:
...
+      - name: Pulser + mypy setup
+        uses: ./.github/workflows/pulser-setup
         with:
-          python-version: 3.8
-      - name: Install Python dependencies
-        run: |
-          python -m pip install --upgrade pip
-          pip install -e ./pulser-core -e ./pulser-simulation
-          pip install -r requirements.txt
+          extra-packages: '''mypy\|pytest'''
```
MatthieuMoreau0, line 52: nit: I find this syntax a big ugly but it's just a matter of taste.. I would prefer something a bit more classic (maybe this is very common) like `mypy,pytest` but it might be a pain to parse in the bash script then.

As a more general note (probably for another PR), It might be nice to add these dev dependencies in the `extras_require` section of the `setup.py` file instead of having an external `dev_requirements.txt`. You could separate optional dependencies in different categories, such as testing, linting and so on and instead of having this custom logic to identify packages to install (e.g. in this case pytest => find all packages which start with pytest in the `dev_requirements.txt`). Instead you could just do `pip install -e .[testing]` which would install all dependencies needed for testing.

The annoying thing with `extras_require` is that you have no way to install all `extras_require` categories except by listing them all, e.g. you can't `pip install -e .[all]` but instead `pip install -e .[linting, testing, blabla, blabla]`. To go the extra mile [this article](https://hanxiao.io/2019/11/07/A-Better-Practice-for-Managing-extras-require-Dependencies-in-Python/) suggests a cool idea to automatically generate the extras_require from a .txt file with a given format. The extras_require will then have an `all` category which lets you install everything with `pip install -e .[all]`

Anyway just a nice to have, nothing critical!

.github/workflows/ci.yml
```diff
@@ -12,83 +12,60 @@ jobs:
...
       - name: Type check with mypy
         run: mypy
   test:
+    if: github.event_name != 'push'
```
MatthieuMoreau0, line 56: Question: Why do we not run test on `push` events ? Not super familiar with all the event types in github actions, so maybe test are automatically triggered through a `merge_request` event or something

.github/workflows/publish.yml
```diff
@@ -0,0 +1,79 @@
...
+        shell: bash
+        run: |
+          python -m pip install --upgrade pip
+          pip install pulser
+      - name: Test the installation
+        shell: bash
+        run: |
+          version="$(head -1 VERSION.txt)"
+          python -c "import pulser; assert pulser.__version__ == '$version'"
+          grep -e pytest dev_requirements.txt | sed 's/ //g' | xargs pip install
```
MatthieuMoreau0, line 78: nit: As per my other comment, having specific `extras_require` would have the extra benefit to avoid duplicating this `grep` magic several times

anneclairelh (commented):
These are mostly questions about your overall flow (therefore hopefully complementary to Adrien and Matthieu's review ) :)

.github/workflows/pulser-setup/action.yml
```diff
@@ -0,0 +1,31 @@
...
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: ${{ inputs.python-version }}
+        cache: 'pip'
+    - name: Install Pulser
+      shell: bash
+      run: |
+        python -m pip install --upgrade pip
+        pip install -e ./pulser-core -e ./pulser-simulation
```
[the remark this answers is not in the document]
HGSilveri: I didn't think so either and I tried to do it without it, but then actually `pytest-cov` failed because the source code was located in `site-packages` rather than in the root of the repo.

.github/workflows/ci.yml
HGSilveri: I agree, for now it's just the regex that is given directly to grep, so there's no parsing required. I think that if this stops being an isolated case I would try to figure out how to parse it, but for now I think I'll stick with this little bit of ugliness. 

The `extras_require` is indeed a nice workaround, though admittedly out of scope for this PR perhaps. However, I will say I do like having the `dev_requirements` all neatly in a single file, rather than a bit more hidden away in the setup.py. Perhaps we can even find a way to combine the two, it would indeed be a nice to have!

.github/workflows/ci.yml
HGSilveri: Just because I wrote a more complete test workflow to run only on `push` to the protected branches, so these would be redundant.

.github/workflows/ci.yml
MatthieuMoreau0: Yes definitely not a priority. If you prefer to have the dependencies isolated in a txt file then the article I linked might be a good hint for the future

.github/workflows/ci.yml
```diff
@@ -12,83 +12,60 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Check out Pulser
-        uses: actions/checkout@v2
-      - name: Set up Python
-        uses: actions/setup-python@v2
+        uses: actions/checkout@v3
+      - name: Pulser + flake8 setup
```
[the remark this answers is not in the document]
HGSilveri: I agree, consider it done.

.github/workflows/ci.yml
```diff
@@ -12,83 +12,60 @@ jobs:
...
-        run: |
-          python -m pip install --upgrade pip
-          pip install -e ./pulser-core -e ./pulser-simulation
-          pip install -r requirements.txt
+          extra-packages: '''mypy\|pytest'''
       - name: Type check with mypy
         run: mypy
   test:
+    if: github.event_name != 'push'
     runs-on: ${{ matrix.os }}
```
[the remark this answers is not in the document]
HGSilveri: True, done.

.github/workflows/ci.yml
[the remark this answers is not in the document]
HGSilveri: The workflow will be triggered on every update to an open PR and on pushes to `master` and `develop`. It's a lighter version of the other, more complete test workflow that runs only on push to the protected branches (including when a PR is merged)
This line is just to avoid the same tests to run twice on a push.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,79 @@
...
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.9
+      - name: Install Python dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install setuptools wheel twine
+      - name: Build packages
+        shell: bash
+        run: ./.github/scripts/package.sh
```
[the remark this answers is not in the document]
HGSilveri: Good question! There are a few reasons:
1. I had the packaging script already written and I think it is pretty self-contained
2. For the install, I did that by removing steps of an existing action, so it seemed natural to use a Github action. Tbh, I didn't even consider turning it into a script.
3. Github actions is convenient to handle inputs (which I don't need to do on the packaging script)

.github/workflows/publish.yml
```diff
@@ -0,0 +1,79 @@
...
+          TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
+        run: twine upload dist/*
+
+  check-release:
+    needs: deploy
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        python-version: ['3.7', '3.8', '3.9', '3.10']
```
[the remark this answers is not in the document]
HGSilveri: For some reason it is needed once the minor has more than one digit (ie from 3.10), so I figured they should all have quotes for uniformity. I'll change the ones in `ci.yml` to have quotes too

.github/workflows/test.yml
```diff
@@ -0,0 +1,25 @@
+name: test
+
+on:
+  push:
+    branches:
+      - master
+      - develop
```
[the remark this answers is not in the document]
HGSilveri: Exactly. I figured they are a bit excessive to be ran on every update to a PR. There are still tests in the `ci.yml` that should catch everything, these tests are run just to really ensure we didn't miss anything related to a specifc OS or python version. 
On the rare case something fails, we can always follow up with a fix.

.github/workflows/publish.yml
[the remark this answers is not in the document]
HGSilveri: Good question! There are a few reasons:
1. I had the packaging script already written and I think it is pretty self-contained
2. For the install, I did that by removing steps of an existing action, so it seemed natural to use a Github action. Tbh, I didn't even consider turning it into a script.
3. Github actions is convenient to handle inputs (which I don't need to do on the packaging script)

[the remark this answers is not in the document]
anneclairelh: Ok, I see. Side note for the future, I feel like if we see the need for it (like using some options), we could at some point move the shell script to a Github actions file.

.github/workflows/ci.yml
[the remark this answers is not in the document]
anneclairelh: I see, so this line is meant to cover the "update to an open PR" case, while the other workflow is meant to cover the "push to `master` or `develop` " cases, is that right ?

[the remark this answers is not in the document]
HGSilveri: Exactly!

PR #388, opened 2022-07-19
New sampler implementation

lvignoli:
The following sampler implementation is channel-centric—by opposing to the former one which was qubit centric.

Samples dataclasses are defined in `pulser-core/samples.py`. Samples are extracted at the channel level from the `_ChannelSchedule.get_samples()` method. For the regular user, there is only one function to know, the `sample(seq)` one exposed by the `pulser-core/sampler/` module.

Improvements:
- [x] tests
- [x] creation of `ChannelDrawContent` from `ChannelSamples`.\
There is much more info needed to draw, its a wider scope than the time series, so I am unsure about the right design.
- [ ] method on the `Sequence` class to get samples?


lvignoli:
Last commits restore the tests and update them to the new architecture.
They actually allow to catch an issue with channel duration not being the same as the parent 
sequence hence messing up the samples arrays.

HGSilveri:
@lvignoli the changes are done. I'm noticing there's a tutorial that I'm guessing should be untracked?

lvignoli:
Thank you @HGSilveri!
I'll review them soon.

The tutorial was simply to check that it was working as intended, so you can delete safely.

arthurfaria:
> Thank you @HGSilveri! I'll review them soon.
> 
> The tutorial was simply to check that the sampler was working as intended, so you can delete it safely. Besides I think @arthurfaria would like to write a notebook tutorial once this is reviewed and merge, see #349.

Sounds good. Perhaps I will start with that one. Is there any other channel we can communicate like slack or something?

HGSilveri:
> Sounds good. Perhaps I will start with that one. Is there any other channel we can communicate like slack or something?

Indeed, we have a Slack channel that has been dormant but that we can use. I sent you an invite.

pulser-core/pulser/sampler/samples.py
```diff
@@ -46,21 +45,15 @@ def new_qdict() -> dict:
...
-def _pairwise(iterable: typing.Iterable) -> typing.Iterable:
-    a, b = itertools.tee(iterable)
-    next(b, None)
-    return zip(a, b)
-
-
 def _default_to_regular(d: dict | defaultdict) -> dict:
     """Helper function to convert defaultdicts to regular dicts."""
-    if isinstance(d, defaultdict):
+    if isinstance(d, dict):
```
lvignoli, line 50: So it now catches both defaultdicts and dicts?

pulser-core/pulser/sampler/samples.py
```diff
@@ -46,21 +45,15 @@ def new_qdict() -> dict:
...
         d = {k: _default_to_regular(v) for k, v in d.items()}
     return d
 
 
 @dataclass
-class _TimeSlot:
+class _TargetSlot:
```
lvignoli, line 56: 👍👍

pulser-core/pulser/sampler/samples.py
```diff
@@ -83,37 +76,91 @@ class ChannelSamples:
...
+        self.duration = len(self.amp)
 
         for t in self.slots:
             assert t.ti < t.tf  # well ordered slots
-        for t1, t2 in _pairwise(self.slots):
+        for t1, t2 in zip(self.slots, self.slots[1:]):
             assert t1.tf <= t2.ti  # no overlaps on a given channel
 
+    def extend_duration(self, new_duration: int) -> ChannelSamples:
+        """Extends the duration of the samples.
```
lvignoli, line 91:
```suggestion
        """Extends the duration of the samples by padding zero samples at the end.
```

pulser-core/pulser/sampler/samples.py
```diff
@@ -83,37 +76,91 @@ class ChannelSamples:
...
+        """Modulates the samples for a given channel.
+
+        Args:
+            channel_obj: The channel object for which to modulate the samples.
+
+        Returns:
+            The modulated channel samples.
+        """
+        if not isinstance(channel_obj, Channel):
+            raise TypeError("'channel_obj' must be a Channel instance.")
```
lvignoli, line 123: Why having a runtime exception here when we have the type hint to prevent passing something else?
Is this function high enough to be potentially misused by a regular user?

pulser-core/pulser/sampler/samples.py
```diff
@@ -83,37 +76,91 @@ class ChannelSamples:
...
+    samples_list: list[ChannelSamples]
+    _ch_objs: dict[str, Channel]
+
+    @property
+    def channel_samples(self) -> dict[str, ChannelSamples]:
+        """Mapping between the channel name and its samples."""
+        return dict(zip(self.channels, self.samples_list))
+
+    @property
+    def duration(self) -> int:
```
lvignoli, line 145:
```suggestion
    def maximum_duration(self) -> int:
```
otherwise I expect the sequence duration because of the name

pulser-core/pulser/sequence/_schedule.py
```diff
@@ -78,26 +77,38 @@ def adjust_duration(self, duration: int) -> int:
...
         """Returns the samples of the channel.
 
         Args:
-            N (int): the duration of the parent sequence. Necessary to give the
-            right length to the sample arrays.
+            modulated: Whether to return the modulated samples.
         """
         # Keep only pulse slots
+        dt = self.get_duration()
         channel_slots = [s for s in self.slots if isinstance(s.type, Pulse)]
```
lvignoli, line 88: Swapping lines for consistent comment

```suggestion
        channel_slots = [s for s in self.slots if isinstance(s.type, Pulse)]
        dt = self.get_duration()
```

pulser-core/pulser/sequence/_schedule.py
```diff
@@ -78,26 +77,38 @@ def adjust_duration(self, duration: int) -> int:
...
-            phase[s.ti : s.tf] += pulse.phase
-            slots.append(samples._TimeSlot(s.ti, s.tf, s.targets))
+            ph_jump_t = self.channel_obj.phase_jump_time
+            t_start = max(0, (s.ti - ph_jump_t))
+            t_end = (
+                channel_slots[ind + 1].ti - ph_jump_t
+                if ind < len(channel_slots) - 1
+                else dt
+            )
+            phase[t_start:t_end] += pulse.phase
```
lvignoli, line 103: I was not aware of this mechanics, glad you took over!

pulser-core/pulser/sampler/samples.py
```diff
@@ -83,37 +76,91 @@ class ChannelSamples:
...
+
+        Returns:
+            The modulated channel samples.
+        """
+        if not isinstance(channel_obj, Channel):
+            raise TypeError("'channel_obj' must be a Channel instance.")
+
+        new_amp = channel_obj.modulate(self.amp)
+        new_detuning = channel_obj.modulate(self.det)
+        new_phase = channel_obj.modulate(self.phase, keep_ends=True)
```
lvignoli, line 127: Why only keep_end only for phase?
I think a comment would make it clear

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -44,100 +46,107 @@
...
     target: dict[Union[str, tuple[int, int]], Any]
-    measurement: Optional[str] = None
     interp_pts: dict[str, list[list[float]]] = field(default_factory=dict)
 
     def __post_init__(self) -> None:
+        self.samples.phase = self.samples.phase / (2 * np.pi)
+        self._samples_from_curves = {
+            "amplitude": "amp",
+            "detuning": "det",
+            "phase": "phase",
```
lvignoli, line 60: Very sad to have different names across the code base 😢

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -44,100 +46,107 @@
...
         return [i for i, qty in enumerate(CURVES_ORDER) if self.curves_on[qty]]
 
+    def _give_curves_from_samples(
+        self, samples: ChannelSamples
+    ) -> list[np.ndarray]:
+        return [
+            getattr(samples, self._samples_from_curves[qty])
+            for qty in CURVES_ORDER
+        ]
+
```
lvignoli, line 103: Its there because of the naming conflict, right?

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -44,100 +46,107 @@
...
         # Store everything
-        time.pop(0)  # Removes the -1 in the beginning
-        data[ch] = ChannelDrawContent(time, amp, detuning, phase, target)
-        if hasattr(seq, "_measurement"):
-            data[ch].measurement = seq._measurement
+        samples = sch.get_samples()
+        data[ch] = ChannelDrawContent(
+            samples.extend_duration(total_duration), target
+        )
```
lvignoli, line 145: So it trims down to this thanks to the sampler, isn't it?

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -192,13 +202,13 @@ def phase_str(phi: float) -> str:
     n_channels = len(seq.declared_channels)
     if not n_channels:
         raise RuntimeError("Can't draw an empty sequence.")
-    data = gather_data(seq)
+    data = gather_data(seq, gather_output=draw_modulation)
     total_duration = data["total_duration"]
     time_scale = 1e3 if total_duration > 1e4 else 1
     for ch in seq._schedule:
-        if np.nonzero(data[ch].detuning)[0].size > 0:
+        if np.nonzero(data[ch].samples.det)[0].size > 0:
```
lvignoli, line 209: nit: use np.count_nonzero

tests/test_sequence_sampler.py
```diff
@@ -106,20 +106,41 @@ def test_inXY() -> None:
...
+
+    want_phase = np.ones(mod_seq.get_duration(include_fall_time=True))
+    got_phase = mod_samples.to_nested_dict()["Global"]["ground-rydberg"][
+        "phase"
+    ]
+    np.testing.assert_array_equal(got_phase, want_phase)
+
+    input_samples = sample(mod_seq)
+    with pytest.raises(TypeError, match="must be a Channel instance"):
+        input_samples.channel_samples["ch0"].modulate("ch0")  # type: ignore
```
lvignoli, line 133: I am still puzzled by this exception

tests/test_sequence_sampler.py
```diff
@@ -188,6 +209,41 @@ def test_SLM_against_simulation(seq_with_SLM):
     assert_same_samples_as_sim(seq_with_SLM)
 
 
+def test_samples_repr(seq_rydberg):
```
lvignoli, line 212: It feels really wrong we have to test for this.

lvignoli (commented):
Up to my few comments, everything looks good to me!
Thank you for taking over the development 💪

NB: I cannot approve since I am the author of the PR

pulser-core/pulser/sampler/sampler.py
```diff
@@ -1,299 +1,26 @@
...
+from typing import TYPE_CHECKING
 
-from pulser.channels import Channel
-from pulser.pulse import Pulse
-from pulser.sampler.noise_model import NoiseModel, apply_noises
-from pulser.sampler.samples import QubitSamples
-from pulser.sequence.sequence import Sequence, _ChannelSchedule, _TimeSlot
+from pulser.sampler.samples import SequenceSamples
 
+if TYPE_CHECKING:  # pragma: no cover
```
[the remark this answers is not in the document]
HGSilveri: No, it's just to avoid the circular imports, the `Sequence` is only used in type hints

pulser-core/pulser/sampler/samples.py
HGSilveri: Yes, it has to so that it can catch defaultdicts inside dicts.

pulser-core/pulser/sampler/samples.py
HGSilveri: It's not exactly this, as the `phase` is not padded with zeros, but I can make the docstring more informative

pulser-core/pulser/sampler/samples.py
HGSilveri: Hum, you're right, for now this won't be directly used by a user, I guess I can remove this check.

pulser-core/pulser/sequence/_seq_drawer.py
HGSilveri: I agree, but I think it's not worth changing now

pulser-core/pulser/sequence/_seq_drawer.py
HGSilveri: Yes, pretty much.

pulser-core/pulser/sequence/_seq_drawer.py
HGSilveri: Yes, exactly. In the end, it didn't make sense to have `ChannelSamples` make the `ChannelDrawContent`, but rather have give the `ChannelSamples` to `ChannelDrawContent`

tests/test_sequence_sampler.py
HGSilveri: Why? I think it makes sense it is touched and it's giving the output we expect.

pulser-core/pulser/sampler/samples.py
HGSilveri: Good point, I'll add it

pulser-core/pulser/sampler/samples.py
lvignoli: Good catch

pulser-core/pulser/sampler/samples.py
lvignoli: Thanks!

pulser-core/pulser/sequence/_seq_drawer.py
lvignoli: Definitely out of scope for this PR indeed

tests/test_sequence_sampler.py
lvignoli: I wrote this __repr__  to help me debug by visual inspection, not for actually printing this info somewhere since I doubt thousands of numerical samples are informative.
Now We have to check against an "expected" output format, while I just made it on the fly.
So my point was that's not at all a hard design constraint and with this test it now becomes one 😅

So I am unsure now, maybe we could just remove the __repr__ here, but now that we have it why not keeping it. I'll let you judge!

HGSilveri: I see. I think that now it's there, it doesn't hurt, so I'll just keep it

HGSilveri (approved):
Approved with @lvignoli 's blessing 🙏

PR #389, opened 2022-07-22
Keep register atom ids type in when encode decode

### LINKED ISSUE #380: Register ids type change after serialization and deserialization
When using numbers as atom ids, they are converted to strings when going through the serialization and deserialization. The consequence is that the outcome will differ from the original value.

For example:

```python
reg = Register(dict(enumerate([(2,3),(5,1),(10,0)])))
j=json.dumps(reg, cls=PulserEncoder)
r=json.loads(j, cls=PulserDecoder)
assert reg == r # raises because ids do not have the same types
```

This is due to json converting object ids to strings. This might lead to undesired behaviors, because we never know what consequences we might meet by having strings instead of integers for the ids.

I can find two solutions:

- either fix this type change by replacing the json objects with json lists, where the first element would represent the id: this way, the type will be kept
- or systmetically convert int ids to strings, even when not using serialization: the behavior with or without encoding/decoding will be identical

CdeTerra:
Closes #380


pulser-core/pulser/register/base_register.py
```diff
@@ -202,10 +206,28 @@ def _validate_layout(
...
+            self,
+            cls_dict,
+            [np.ndarray.tolist(qubit_coords) for qubit_coords in self._coords],
+            False,
+            None,
+            self._ids,
+            **kwargs,
+            _submodule=self.__class__.__name__,
+            _name="from_coordinates",
+        )
```
HGSilveri, line 230: Great! Can you maybe add just a comment explaining why we're using `from_coordinates`? I feel like this complexity might be confusing without an explanation

tests/test_json.py
```diff
@@ -89,6 +89,14 @@ def test_register_from_layout():
     assert new_reg._layout_info.trap_ids == (1, 0)
 
 
+def test_register_numbered_keys():
+    reg = Register(dict(enumerate([(2, 3), (5, 1), (10, 0)])))
+    j = json.dumps(reg, cls=PulserEncoder)
+    decoded_reg = json.loads(j, cls=PulserDecoder)
+    assert reg == decoded_reg
+    assert all([type(i) == int for i in decoded_reg.qubit_ids])
```
HGSilveri, line 97: Can you maybe add a test with the `Register3D`too? And on that one, perhaps use an unconventional order for the indices and make sure that is respected

HGSilveri (approved):
Looking great, thanks a lot! I have only a couple suggestions

pulser-core/pulser/register/base_register.py
```diff
@@ -233,7 +242,7 @@ def __eq__(self, other: Any) -> bool:
         if type(other) is not type(self):
             return False
 
-        return set(self._ids) == set(other._ids) and all(
+        return list(self._ids) == list(other._ids) and all(
```
HGSilveri, line 243: I'm not sure about this. What was wrong with the old way?

tests/test_register.py
```diff
@@ -405,14 +405,17 @@ def assert_ineq(left, right):
 def test_equality_function():
     reg1 = Register({"c": (1, 2), "d": (8, 4)})
     assert_eq(reg1, reg1)
-    assert_eq(reg1, Register({"d": (8, 4), "c": (1, 2)}))
```
HGSilveri, line 408: Shouldn't this still be true?

pulser-core/pulser/register/base_register.py
CdeTerra: The old way didn't check the ids order.

tests/test_register.py
CdeTerra: These register's ids don't have the same order, so they should not be equal.

pulser-core/pulser/register/base_register.py
HGSilveri: But that's ok, perhaps I expressed myself poorly in my previous suggestion. What I meant is the the ids should still match the same coordinates after encoding-decoding

CdeTerra: Actually that's quite not okay, because `find_indices` relies on the order of the ids. The coordinates are already checked in the `__eq__` function.

HGSilveri: Yes, you're right 🙂

PR #390, opened 2022-08-03
Simulation with modulated output

### LINKED ISSUE #315: Add option to simulate a Sequence with the modulated output

HGSilveri:
A continuation of #379, this PR:
- Adapts new sampler in `pulser.sampler` to extract the samples used by the `Simulation` class.
- Allows for simulation of sequences from their modulated output
- Introduces small modifications to the sampler that were missed in #388 
- Changes how noise is introduced to the samples. In particular, changes the application of amplitude noise to a more physically sound one.

Closes #315.


pulser-core/pulser/sampler/samples.py
```diff
@@ -100,9 +109,11 @@ def extend_duration(self, new_duration: int) -> ChannelSamples:
         Returns:
             The extend channel samples.
         """
-        extension = new_duration - len(self.amp)
-        if new_duration < self.duration:
+        extension = new_duration - self.duration
+        if extension < 0:
             raise ValueError("Can't extend samples to a lower duration.")
+        if not extension:
+            return self
```
lvignoli, line 116: Should it be a copy of self, to be consistent with returning a new object?

pulser-core/pulser/sampler/samples.py
```diff
@@ -166,13 +195,25 @@ def to_nested_dict(self) -> dict:
             addr = self._ch_objs[chname].addressing
             basis = self._ch_objs[chname].basis
             if addr == _GLOBAL:
-                d[_GLOBAL][basis][_AMP] += cs.amp
-                d[_GLOBAL][basis][_DET] += cs.det
-                d[_GLOBAL][basis][_PHASE] += cs.phase
+                start_t = self._slm_mask.end
+                d[_GLOBAL][basis][_AMP][start_t:] += cs.amp[start_t:]
+                d[_GLOBAL][basis][_DET][start_t:] += cs.det[start_t:]
+                d[_GLOBAL][basis][_PHASE][start_t:] += cs.phase[start_t:]
```
lvignoli, line 212: This is the consequence of the extension feature, so that left and right member may be arrays of different lengths?

pulser-core/pulser/sampler/samples.py
HGSilveri: You're right, I agree this is abusive... I just added this because `extend_duration` was being called blindly in `SequenceSamples.to_nested_dict()`. I think I'll remove this case and instead call `extend_duration()` in `SequenceSamples.to_nested_dict()` only when needed. Sounds good?

tests/conftest.py
```diff
@@ -0,0 +1,70 @@
...
+# limitations under the License.
+
+import numpy as np
+import pytest
+
+from pulser.channels import Raman, Rydberg
+from pulser.devices import Device
+
+
+@pytest.fixture
```
lvignoli, line 22: 👍

tests/conftest.py
```diff
@@ -0,0 +1,70 @@
...
+def mod_device() -> Device:
```
lvignoli, line 23: What about having a public ModMockDevice, or add modulation to the MockDevice?

tests/test_sequence_sampler.py
```diff
@@ -137,7 +135,48 @@ def test_modulation(mod_seq: pulser.Sequence) -> None:
...
+    out_ch_samples = output_samples.channel_samples["ch0"]
+    expected_slots = deepcopy(in_ch_samples.slots)
+    # The first slot should extend to the second
+    expected_slots[0].tf += partial_fall
+    assert expected_slots[0].tf == expected_slots[1].ti
+    # The next slots should fully account for fall time
+    expected_slots[1].tf += pulse2.fall_time(ch_obj)
+    expected_slots[2].tf += pulse1.fall_time(ch_obj)
+
+    assert out_ch_samples.slots == expected_slots
```
lvignoli, line 171: I have not idea of what going on.

tests/test_sequence_sampler.py
```diff
@@ -137,7 +135,48 @@ def test_modulation(mod_seq: pulser.Sequence) -> None:
...
+
+    # Check that the samples are fully extracted to the nested dict
+    samples_dict = output_samples.to_nested_dict()
+    for qty in ("amp", "det", "phase"):
+        combined = sum(
+            samples_dict["Local"]["ground-rydberg"][t][qty] for t in range(2)
+        )
+        np.testing.assert_array_equal(getattr(out_ch_samples, qty), combined)
```
lvignoli, line 179: Clever!

tests/test_simresults.py
```diff
@@ -152,12 +152,12 @@ def test_get_final_state_noisy():
...
-        final_state[0, 0] == 0.06666666666666667 + 0j
-        and final_state[2, 2] == 0.92 + 0j
+        final_state[0, 0] == 0.12 + 0j
+        and final_state[2, 2] == 0.8666666666666667 + 0j
     )
     assert res3.states[-1] == final_state
     assert res3.results[-1] == Counter(
-        {"10": 0.92, "00": 0.06666666666666667, "11": 0.013333333333333334}
+        {"10": 0.8666666666666667, "00": 0.12, "11": 0.013333333333333334}
     )
```
lvignoli, line 161: Woaw that's weird to have these floats hardcoded and checking against them like that.
I had no idea we do this.

What about using `math.isclose()`?

tests/test_simulation.py
```diff
@@ -25,53 +25,53 @@
...
-seq.target("control2", "ryd")
-seq.add(pi, "ryd")
-seq.target("control1", "ryd")
-seq.add(pi, "ryd")
-d += 5
-
-# Add a ConstantWaveform part to testout the drawing procedure
-seq.add(Pulse.ConstantPulse(duration, 1, 0, 0), "ryd")
-d += 1
+
```
lvignoli, line 28: Why not putting all the fixture in conftest.py?

lvignoli (commented):
@HGSilveri here are some comments on this extensive work! There are a few things that are unclear to me, otherwise looks pretty great.

pulser-core/pulser/sampler/samples.py
```diff
@@ -166,13 +195,25 @@ def to_nested_dict(self) -> dict:
             addr = self._ch_objs[chname].addressing
             basis = self._ch_objs[chname].basis
             if addr == _GLOBAL:
-                d[_GLOBAL][basis][_AMP] += cs.amp
-                d[_GLOBAL][basis][_DET] += cs.det
-                d[_GLOBAL][basis][_PHASE] += cs.phase
+                start_t = self._slm_mask.end
+                d[_GLOBAL][basis][_AMP][start_t:] += cs.amp[start_t:]
+                d[_GLOBAL][basis][_DET][start_t:] += cs.det[start_t:]
```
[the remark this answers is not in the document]
HGSilveri: This one is to account for the SLM - `start_t` marks the point when the SLM mask is turned off and so the sequence can be stored normally. For `:start_t`, we store the samples locally for the unmasked targets.

pulser-core/pulser/sequence/_schedule.py
```diff
@@ -94,19 +94,32 @@ def get_samples(self, modulated: bool = False) -> ChannelSamples:
...
-            slots.append(_TargetSlot(s.ti, s.tf, s.targets))
+            tf = s.tf
+            if modulated:
+                # Account for as much fall time as possible
+                fall_time = pulse.fall_time(self.channel_obj)
+                tf += (
+                    min(fall_time, channel_slots[ind + 1].ti - s.tf)
+                    if ind < len(channel_slots) - 1
+                    else fall_time
+                )
```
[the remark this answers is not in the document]
HGSilveri: I'm sorry about that, I'll add a comment. 
When there is modulation, the pulse gets elongated - this increase in duration is the `fall_time`, the extra time taken to go back to zero. If two pulses are adjacent or have a space between them that is smaller than the fall time, then they will be merged instead (ie the first one doesn't get to go back to zero before the second starts). 
The reason we need to account for this in the target slots is because they set the limits when a pulse starts and ends, which is used when separating the samples between targets in the nested dict. If we leave the target slots untouched, the modulated pulses will be truncated.

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -357,14 +358,22 @@ def phase_str(phi: float) -> str:
...
+                        alpha=0.3,
+                        hatch="////",
+                    )
+                else:
+                    ax.plot(
+                        t,
+                        ys_mod[i][:total_duration],
+                        color=COLORS[i],
+                        linestyle="dotted",
+                    )
```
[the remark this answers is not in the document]
HGSilveri: Indeed. The same goes for Simulation with modulation. 
Nonetheless, the users can already create their custom devices with modulation.

pulser-simulation/pulser_simulation/simconfig.py
```diff
@@ -92,6 +95,12 @@ class SimConfig:
     )
 
     def __post_init__(self) -> None:
+        if not 0.0 <= self.amp_sigma < 1.0:
+            raise ValueError(
+                "The standard deviation in amplitude (amp_sigma="
+                f"{self.amp_sigma}) must be greater than or equal"
+                " to 0. and smaller than 1."
+            )
         self._process_temperature()
```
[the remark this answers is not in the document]
HGSilveri: The lower bound should be clear, since this is a standard deviation.
I admit the upper bound is somewhat arbitrary, but since this is a normalized multiplicative factor that needs to be `<=0`, I figured that the case where that happens within 1 sigma was a reasonable upper-bound - realistically, we will never come even close to this.

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -100,7 +103,20 @@ def __init__(
         self._interaction = "XY" if self._seq._in_xy else "ising"
         self._qdict = self._seq.qubit_info
         self._size = len(self._qdict)
-        self._tot_duration = self._seq.get_duration()
+        self._modulated = bool(with_modulation)
```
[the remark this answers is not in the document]
HGSilveri: Since this is a public method, I figured I should at least try to account for the case where someone doesn't respect the type hint. Truth be told though, everything can be cast to a bool, so I can see it being the wrong approach here. Do you think I should explicitly check that it is a bool instead?

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -100,7 +103,20 @@ def __init__(
...
+                "Simulation of sequences combining an SLM mask and output "
+                "modulation is not supported."
+            )
+        self._tot_duration = self._seq.get_duration(
+            include_fall_time=self._modulated
+        )
+        self.samples_obj = sampler.sample(
+            self._seq,
+            modulation=self._modulated,
+            extended_duration=self._tot_duration + 1,
```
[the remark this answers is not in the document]
HGSilveri: This is the eternal discussion on the ODE solver convergence issue which was fixed by adding one extra sample at the end. You had previously addressed this by adding the `additional_final_sample: bool = False` option.

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -414,116 +442,51 @@ def draw(
...
-                    if qubit not in self.samples["Local"][basis]:
-                        continue
-                    for x in ("amp", "det", "phase"):
-                        self.samples["Local"][basis][qubit][x][0:tf] = 0
+            for qid in slot.targets:
+                if "doppler" in self.config.noise:
+                    noise_det = self._doppler_detune[qid]
+                    samples_dict[qid]["det"][slot.ti : slot.tf] += noise_det
+                # Gaussian beam loss in amplitude for global pulses only
+                # Noise is drawn at random for each pulse
```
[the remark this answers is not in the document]
HGSilveri: I considered it, but it was not being done before so I think it's out of scope here. 
The sampling being done per pulse rather than per qubit is already going a bit too far, but it didn't make sense to do it the old way and write tests for it that would just have to be rewritten later.
In any case, you can disable one or the other by changing the `laser_waist` and `amp_sigma` parameters.

tests/conftest.py
HGSilveri: `MockDevice` is supposed to be irrealistically good, so I don't think it would make sense to add a HW constraint to it. A `ModMockDevice` could make sense, but then what values of bandwith would we give it?

I think the direction we are going towards is rather to give more freedom for advanced users to define their one custom devices, so they are not restrict to the public ones.

tests/test_sequence_sampler.py
HGSilveri: Is it clearer after the explanation above?

tests/test_simulation.py
HGSilveri: Because I don't want these to be shared with other test files, especially when they're named `seq` and `reg`.

tests/test_simresults.py
HGSilveri: I'm not a fan either, but it's important to have some fixed values to catch changes like this.

How would you use `math.isclose()` when comparing `Counter`s?

pulser-core/pulser/sampler/samples.py
lvignoli: Sure!

pulser-core/pulser/sampler/samples.py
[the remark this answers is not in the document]
lvignoli: Alright

pulser-core/pulser/sequence/_schedule.py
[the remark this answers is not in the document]
lvignoli: Thanks it's much clearer, I was aware of this effect but it was unclear that was the reason behind these lines

pulser-simulation/pulser_simulation/simconfig.py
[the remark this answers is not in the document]
lvignoli: Okay, trust you on the "realistically", we could always relax this later if we need to explore uncommon regimes

pulser-simulation/pulser_simulation/simulation.py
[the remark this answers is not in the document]
lvignoli: Well by doing so then you then consider explicitly checking types of all arguments of public methods, in addition to the type hinting and the documentation.
This is python's peculiarity to have type hints unrelevant at runtime, they basically give no additional real guarantees, despite being quite useful to the users.

So I would say you should just assume that `with_modulation` is a boolean and add a type hint to `self._modulated` if it is inherited from `with_modulation`.

pulser-simulation/pulser_simulation/simulation.py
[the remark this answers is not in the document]
lvignoli: The usual suspect, alright.
Maybe add a comment so that we know where this pesky sample is introduced, I am sure it's far from the last time we deal with it.

pulser-simulation/pulser_simulation/simulation.py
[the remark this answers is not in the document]
lvignoli: Ok make sense

tests/test_sequence_sampler.py
lvignoli: Yes, much clearer!

tests/test_simresults.py
lvignoli: I understand these are hard guards providing strong guarantees, but float equality is touchy, so I am unsure about the lifespan of such guarantees. It may even be architecture dependent regarding the underlying float type.

Probably I would loop over the keys of the `Counter`.

That being said, this is outside the scope of this PR. And since it has been working until now on many many machines, I see no real reasons to address it.

tests/test_simulation.py
lvignoli: Make sense

pulser-simulation/pulser_simulation/simulation.py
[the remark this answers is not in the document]
HGSilveri: Ok, agreed.

lvignoli (approved):
Looks great, let's merge 😊

PR #396, opened 2022-08-10
Add step in publish workflow to confirm deployment

HGSilveri:
After uploading a new release to PyPI, this step will repeatedly attempt to download it. This will ensure that the subsequent tests will run after the new release is properly deployed.


CdeTerra (approved):
Thanks Henrique for this change!

PR #398, opened 2022-09-20
Switch device

Varda-star:
Add the switch_device method to the Sequence class.


pulser-core/pulser/sequence/sequence.py
```diff
@@ -370,6 +370,77 @@ def config_slm_mask(self, qubits: Iterable[QubitId]) -> None:
...
+                seq_channel = self._schedule[calls[i].args[0]]
+                for ch in new_device._channels:
+                    # Find the corresponding channel on the new device
+                    # We verify the channel class then
+                    # check wether it's Global or local
+                    if (
+                        seq_channel.channel_obj.basis == ch[1].basis
+                        and seq_channel.channel_obj.addressing
+                        == ch[1].addressing
+                    ):
```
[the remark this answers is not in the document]
HGSilveri: Naturally, the case described above should also be covered in the unit tests.

pulser-core/pulser/sequence/sequence.py
```diff
@@ -553,14 +624,14 @@ def add(
                 simultaneously.
 
                 - ``'min-delay'``: Before adding the pulse, introduces the
```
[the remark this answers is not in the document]
Varda-star: yes or a mistake. It wasn't my intention

pulser-core/pulser/sequence/sequence.py
```diff
@@ -370,6 +370,82 @@ def config_slm_mask(self, qubits: Iterable[QubitId]) -> None:
...
+                    if basis_match and addressing_match:
+                        if strict:
+                            phase_jump_time_check = (
+                                od_ch_obj.phase_jump_time
+                                != nd_ch_obj.phase_jump_time
+                            )
+                            clock_period_check = (
+                                od_ch_obj.clock_period
+                                != nd_ch_obj.clock_period
+                            )
```
HGSilveri, line 419: Thinking on this more deeply, these checks don't have to be so strict. It's a bit more involved but I think it's worth it:
- The clock period is only invalid if the old device clock period is not a multiple of the new device clock period. So for example, if you're changing from 4ns, clock periods of 1ns and 2ns would be valid, everything else would not. This is because the clock period enforces that every duration is a multiple of it, but a multiple of 4 is also a multiple of 1 and 2 so all elements in the sequence will still keep their duration unchanged. Note, however, that the inverse is not true: changing from 1ns to 4ns should fail.
- The phase jump time is only relevant when there are phase jumps in the sequence (ie consecutive pulses where the phase changes). An easy check you can do here is to see if the phase values as given by the [sampler](https://pulser.readthedocs.io/en/stable/apidoc/core.html#module-pulser.sampler.sampler) are constant - if they are, the phase jump time is irrelevant.

HGSilveri (changes requested):
Here are a few more comments. I think you forgot to separate the matching of the channels from the reconstruction of the sequence. Let me know if that wasn't so clear.

pulser-core/pulser/sequence/sequence.py
```diff
@@ -376,74 +377,130 @@ def switch_device(
...
+                                np.all(ch_sample_phase == ch_sample_phase[0])
                             )
-                            if phase_jump_time_check or clock_period_check:
+                        phase_jump_time_check = (
+                            od_ch_obj.phase_jump_time
+                            == nd_ch_obj.phase_jump_time
+                        )
+                        clock_period_check = (
+                            nd_ch_obj.clock_period % od_ch_obj.clock_period
+                            == 0
```
HGSilveri, line 438: I think this should be the inverse:
```suggestion
                            od_ch_obj.clock_period % nd_ch_obj.clock_period
                            == 0
```
You can check with an example: if `new_clock = 1` and `old_clock = 4`, it should be valid, but `1 % 4 = 1`.

pulser-core/pulser/sequence/sequence.py
```diff
@@ -376,74 +377,130 @@ def switch_device(
...
                     else:
-                        ch_type_er_mess = "No basis and addressing match."
+                        channel_matching[channel_name] = nd_ch_id
+                        matched = True
+                        break
+                else:
+                    matched = False
+                    ch_type_er_mess = (
+                        "No channel with both basis and addressing match."
+                    )
```
HGSilveri, line 472: This block also seems like it can be simplified. There is some unnecessary code replication. For example, there is no need to check `phase_jump_time_oracle` if `phase_jump_time_check is True`

pulser-core/pulser/sequence/sequence.py
```diff
@@ -376,74 +377,130 @@ def switch_device(
...
+                getattr(new_seq, "declare_channel")(
+                    *sw_channel_args, **call.kwargs
+                )
             else:
                 getattr(new_seq, call.name)(*call.args, **call.kwargs)
+
+        if new_device.rydberg_level != self._device.rydberg_level:
+            warnings.warn(
+                "The rydberg level of the new device is different"
+                + " take in account in your computations.",
```
HGSilveri, line 501: Instead of the `take in account in your computations` part, perhaps it's more useful to say what the old and new Rydberg levels are, eg
```suggestion
                f"The rydberg level of the new device ({new_device.rydberg_level})"
                f" is different from the current device's ({self._device.rydberg_level}).",
```
Also, I'm thinking that this should also be an error instead of a warning if `strict=True` and any of the new channels is a `Rydberg` channel.

pulser-core/pulser/sequence/sequence.py
```diff
@@ -553,14 +686,14 @@ def add(
...
                 - ``'no-delay'``: Adds the pulse to the channel, regardless of
-                  existing conflicts.
+                 existing conflicts.
 
                 - ``'wait-for-all'``: Before adding the pulse, adds a delay
-                  that idles the channel until the end of the other channels'
-                  latest pulse.
+                that idles the channel until the end of the other channels'
+                latest pulse.
         """
```
HGSilveri, line 697: For some reason, the whitespaces are still deleted...

tests/test_sequence.py
```diff
@@ -140,6 +140,200 @@ def test_magnetic_field():
...
+                ),
+            ),
+        ),
+    )
+
+    reg = Register.square(3, 5, prefix="q")
+
+    # Device checkout
+    seq = Sequence(reg, Chadoq2)
+    assert seq == seq.switch_device(Chadoq2)
```
HGSilveri, line 237: You should check that the warning message you expect is raised with `pytest.warns`

pulser-core/pulser/sequence/sequence.py
```diff
@@ -370,6 +371,138 @@ def config_slm_mask(self, qubits: Iterable[QubitId]) -> None:
...
+                                    + " & clock_period match."
+                                )
+                    else:
+                        channel_matching[channel_name] = nd_ch_id
+                        matched = True
+                        break
+                else:
+                    matched = False
+                    ch_type_er_mess = (
+                        "No channel with both basis and addressing match."
```
HGSilveri, line 471: I think this error message should mention what channel couldn't be matched

tests/test_sequence.py
```diff
@@ -140,6 +140,200 @@ def test_magnetic_field():
...
+
+    with pytest.raises(
+        TypeError, match="No channel with both basis and addressing match."
+    ):
+        seq.switch_device(IroiseMVP)
+
+    # Strict-Jump_phase_time & CLock-period
+    # Jump_phase_time: with post_phase_time
+    R_interatomic = Chadoq2.rydberg_blockade_radius(2 * np.pi)
+    reg = Register.square(3, R_interatomic, prefix="q")
```
HGSilveri, line 283: nit: Let's stick to python style conventions and not use an uppercase letter for a variable
```suggestion
    r_interatomic = Chadoq2.rydberg_blockade_radius(2 * np.pi)
    reg = Register.square(3, r_interatomic, prefix="q")
```

tests/test_sequence.py
```diff
@@ -140,6 +140,200 @@ def test_magnetic_field():
...
+    seq = Sequence(reg, IroiseMVP)
+    rise = Pulse.ConstantDetuning(
+        RampWaveform(250, 0.0, 2.3 * 2 * np.pi), -4 * np.pi, 0.0, True
+    )
+    sweep = Pulse.ConstantAmplitude(
+        2.3 * 2 * np.pi, RampWaveform(400, -4 * np.pi, 4 * np.pi), 1.0, True
+    )
+    fall = Pulse.ConstantDetuning(
+        RampWaveform(500, 2.3 * 2 * np.pi, 0.0), 4 * np.pi, 0.0, True
+    )
```
HGSilveri, line 293: `post_phase_shift` is a `float`, not a `bool`

tests/test_sequence.py
```diff
@@ -140,6 +140,200 @@ def test_magnetic_field():
...
+    seq = Sequence(reg, test_device2)
+    seq.declare_channel("ising", "rmn_local")
+
+    with pytest.raises(
+        TypeError, match="No channel with both basis and addressing match."
+    ):
+        seq.switch_device(IroiseMVP)
+
+    # Strict-Jump_phase_time & CLock-period
+    # Jump_phase_time: with post_phase_time
```
HGSilveri, line 281: Can you maybe make this comment clearer? And do you perhaps mean `post_phase_shift`?

tests/test_sequence.py
```diff
@@ -140,6 +140,200 @@ def test_magnetic_field():
...
+    )
+    fall = Pulse.ConstantDetuning(
+        RampWaveform(500, 2.3 * 2 * np.pi, 0.0), 4 * np.pi, 0.0
+    )
+    seq = Sequence(reg, IroiseMVP)
+    seq.declare_channel("ising", "rydberg_global")
+    seq.add(rise, "ising")
+    seq.add(sweep, "ising")
+    seq.add(fall, "ising")
+    assert seq.switch_device(Chadoq2, True)._device == Chadoq2
```
HGSilveri, line 322: You should check the warning is raised here as expected

HGSilveri (changes requested):
Here goes another batch of comments. It's getting there!

pulser-core/pulser/sequence/sequence.py
Varda-star: The oracle is there to identify whether it is relevant to take the phase_jump_time value. If it is false we don't care of phase_jump_time value, either False or True.

pulser-core/pulser/sequence/sequence.py
```diff
@@ -374,6 +375,132 @@ def config_slm_mask(self, qubits: Iterable[QubitId]) -> None:
...
+        new_seq = Sequence(self.register, new_device)
+
+        # Channel Matching
+
+        matched = False
+        strict_error_message = None
+        ch_type_er_mess = None
+        channel_matching = {}
+        sample_seq = sample(self)
+        alert_phase_jump = False
```
HGSilveri, line 416: This channel matching section will need to be updated and potentially moved to a method inside the device classes.

HGSilveri (changes requested):
Broadly, the incoming changes will require modifications in these parts

pulser-core/pulser/sequence/sequence.py
```diff
@@ -374,6 +375,132 @@ def config_slm_mask(self, qubits: Iterable[QubitId]) -> None:
...
+                "Switching a sequence to the same device"
+                + " returns the sequence unchanged.",
+                stacklevel=2,
+            )
+            return self
+        if new_device == MockDevice:
+            raise NotImplementedError(
+                "Switching the device of a sequence"
+                + " to 'MockDevice' is not supported."
+            )
```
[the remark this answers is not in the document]
Varda-star: So the condition must be `new_device.reusable_channels == True` right?

pulser-core/pulser/sequence/sequence.py
Varda-star: I will think about it and propose.

pulser-core/pulser/sequence/sequence.py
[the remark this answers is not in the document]
HGSilveri: You don't need to explicitly check with `== True` - in fact, it's recommended against in the [PEP8 style guide](https://peps.python.org/pep-0008/#programming-recommendations).

HGSilveri:
#409 is now merged, you can now update your changes here

pulser-core/pulser/sequence/sequence.py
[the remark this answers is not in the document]
HGSilveri: It's not pending anymore, it has been merged. You should merge the new changes into this branch

pulser-core/pulser/devices/_device_datacls.py
```diff
@@ -318,6 +319,90 @@ def _validate_coords(
     def _to_dict(self) -> dict[str, Any]:
         pass
 
+    def find_channel_match(
+        self,
+        channel_list: list[tuple[str, Channel]],
+        channel_names: dict,
+        sample_seq: SequenceSamples,
```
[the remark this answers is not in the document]
Varda-star: I tried to do so, but I had an issue when I tried to import Sequence in _device_datacls file. 
How can I deal with the sample step?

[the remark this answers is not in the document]
Varda-star: I also need the channel_id which is stored in the sequence class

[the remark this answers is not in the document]
Varda-star: I barely see how we can manage without taking sample_seq as an argument and still use the sampling to identify constant phases.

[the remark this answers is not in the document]
HGSilveri: Please check your Slack

pulser-core/pulser/sequence/sequence.py
```diff
@@ -364,6 +365,132 @@ def config_slm_mask(self, qubits: Iterable[QubitId]) -> None:
...
+                    continue
+                if phase_jump_time_check and clock_period_check:
+                    channel_match[o_d_ch_name] = n_d_ch_id
+                    break
+                else:
+                    channel_match[o_d_ch_name] = None
+                    channel_match["strict_error_message"] = (
+                        "No channel with phase_jump_time"
+                        + " & clock_period match."
+                    )
```
HGSilveri, line 458: I think there is some redundancy here. Doesn't this block do the same?

```suggestion
                if clock_period_check and (phase_is_constant or phase_jump_time_check):
                    channel_match[o_d_ch_name] = n_d_ch_id
                    channel_match["alert_phase_jump"] = (
                            not phase_jump_time_check
                        )
                    break
                channel_match[o_d_ch_name] = None
                channel_match["strict_error_message"] = (
                            "No channel with phase_jump_time"
                            + " & clock_period match."
                        )
                continue
```

pulser-core/pulser/sequence/sequence.py
```diff
@@ -364,6 +365,132 @@ def config_slm_mask(self, qubits: Iterable[QubitId]) -> None:
...
+        # Initialize the new sequence
+        new_seq = Sequence(self.register, new_device)
+
+        for call in self._calls[1:]:
+            if not (call.name == "declare_channel"):
+                getattr(new_seq, call.name)(*call.args, **call.kwargs)
+                continue
+            sw_channel_args = list(call.args)
+            # Switch the old id with the correct id
+            sw_channel_args[1] = channel_match[sw_channel_args[0]]
```
HGSilveri, line 474: Here you also have to account for the possibility that the channel ID was given as a keyword arg, so it won't be under `call.args` . Imagine the user declares the channel like this:
```python
seq.declare_channel("ch1", channel_id="rydberg_global")
```
Add an example like this to your tests and you'll see it will fail.
You can see an example on how this can be handled here: https://github.com/pasqal-io/Pulser/blob/1984e378069a1de536016aecf71195cd4c29f3e3/pulser-core/pulser/json/abstract_repr/serializer.py#L51-L67

HGSilveri (changes requested):

pulser-core/pulser/sequence/sequence.py
```diff
@@ -364,6 +365,132 @@ def config_slm_mask(self, qubits: Iterable[QubitId]) -> None:
...
+                stacklevel=2,
+            )
+            return self
+        if new_device.reusable_channels:
+            raise NotImplementedError(
+                "Switching the device of a sequence"
+                + " to 'MockDevice' is not supported."
+            )
+        # Channel match
+        channel_list = cast(list, self.declared_channels.items())
```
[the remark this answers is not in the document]
Varda-star: I did it because I merged the matching to the device class. Now it's useless

pulser-core/pulser/sequence/sequence.py
```diff
@@ -364,6 +365,132 @@ def config_slm_mask(self, qubits: Iterable[QubitId]) -> None:
...
+                    channel_match[o_d_ch_name] = n_d_ch_id
+                    break
+                ch_samples = sample_seq.channel_samples[o_d_ch_name]
+                ch_sample_phase = ch_samples.phase
+                # Find if there is phase change between pulses or not
+                phase_is_constant = True
+                if ch_sample_phase.size != 0:
+                    phase_is_constant = bool(
+                        np.all(ch_sample_phase == ch_sample_phase[0])
+                    )
```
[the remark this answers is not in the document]
Varda-star: I need it to be a boolean. May be using cast is smarter.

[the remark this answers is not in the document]
HGSilveri: But the output of `np.all` is already a boolean, no?

pulser-core/pulser/sequence/sequence.py
```diff
@@ -364,6 +365,132 @@ def config_slm_mask(self, qubits: Iterable[QubitId]) -> None:
...
+                "Switching the device of a sequence"
+                + " to 'MockDevice' is not supported."
+            )
+        # Channel match
+        channel_list = cast(list, self.declared_channels.items())
+        sample_seq = sample(self)
+        channel_match: dict[str, Any] = {}
+        channel_match["alert_phase_jump"] = False
+        channel_match["strict_error_message"] = ""
+        channel_match["ch_type_er_mess"] = ""
```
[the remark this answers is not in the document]
Varda-star: I did it in order to return only one object when I merged the channel match to the device class.

pulser-core/pulser/sequence/sequence.py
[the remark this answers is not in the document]
Varda-star: Yes you're right. An array of boolean. I will replace this line by `False in np.all(ch_sample_phase == ch_sample_phase[0]`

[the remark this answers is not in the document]
HGSilveri: No, in this case it's not an array of booleans, it's just a boolean. If mypy is complaining, you should replace it with a cast to bool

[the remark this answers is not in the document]
Varda-star: And it works as a bool so no need to convert.

pulser-core/pulser/sequence/sequence.py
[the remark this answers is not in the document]
Varda-star: The dictionary is not useful for `strict_error_message` and `ch_type_er_mess`. So I can break it and use two variables, what do you think?

pulser-core/pulser/sequence/sequence.py
```diff
@@ -364,6 +365,132 @@ def config_slm_mask(self, qubits: Iterable[QubitId]) -> None:
...
+            warnings.warn(
+                "Switching a sequence to the same device"
+                + " returns the sequence unchanged.",
+                stacklevel=2,
+            )
+            return self
+        if new_device.reusable_channels:
+            raise NotImplementedError(
+                "Switching the device of a sequence"
+                + " to 'MockDevice' is not supported."
```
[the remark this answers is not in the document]
Varda-star: " Switching a real device to virtual device is not supported" . Does it work well?

pulser-core/pulser/sequence/sequence.py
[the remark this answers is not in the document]
HGSilveri: Yes, sounds good

pulser-core/pulser/sequence/sequence.py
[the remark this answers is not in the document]
HGSilveri: You can be even more specific: "Switching the device of a sequence to one with reusable channels is not supported."

pulser-core/pulser/sequence/sequence.py
HGSilveri: When you call a python function `def f(a, b, c=0): ...` you can give positional arguments (ie args) as keyword arguments (e.g. `f(1, b=2, c=3)` is valid). 
If this happens in declare channel (e.g. `seq.declare_channel("ch1", channel_id="rydberg_global"`), the store _Call will have `args=("ch1",)` and `kwargs={"channel_id": "rydberg_global"}`. 
In contrast, a call like `seq.declare_channel("ch1", "rydberg_global"` would have `args={"ch1", "rydberg_global")` and `kwargs={}`. 
In your code, you are assuming that both arguments are given as args, but this is not necessarily true and it will fail if that is not the case.

Varda-star: Very clear.

Varda-star: So If I well  understand your code I'll get a value error if `name` is stored in `kwargs` and not in `args` ?

HGSilveri: Not `name`, `channel_id`

pulser-core/pulser/sequence/sequence.py
```diff
@@ -423,63 +420,63 @@ def switch_device(
...
-                    channel_match["strict_error_message"] = (
+                    strict_error_message = (
                         "No channel with phase_jump_time"
                         + " & clock_period match."
                     )
 
         if None in channel_match.values():
-            if channel_match["strict_error_message"] != "":
-                raise ValueError(channel_match["strict_error_message"])
+            if strict_error_message != "":
```
HGSilveri, line 447: A non-empty string evaluates to `True`, so the pyrhonic way to write this is actually
```suggestion
            if strict_error_message:
```

pulser-core/pulser/sequence/sequence.py
```diff
@@ -423,63 +420,63 @@ def switch_device(
...
         for call in self._calls[1:]:
             if not (call.name == "declare_channel"):
                 getattr(new_seq, call.name)(*call.args, **call.kwargs)
                 continue
-            sw_channel_args = list(call.args)
             # Switch the old id with the correct id
-            sw_channel_args[1] = channel_match[sw_channel_args[0]]
-            if strict and channel_match["alert_phase_jump"]:
+            sw_channel_args = list(call.args)
+            sw_channel_kw_args = call.kwargs
```
HGSilveri, line 460: You shouldn't modify this `dict` in place, since it will affect the original. 
```suggestion
            sw_channel_kw_args = call.kwargs.copy()
```

pulser-core/pulser/sequence/sequence.py
```diff
@@ -423,63 +420,63 @@ def switch_device(
...
+            if "name" in sw_channel_kw_args:
+                sw_channel_kw_args["channel_id"] = channel_match[
+                    sw_channel_kw_args["name"]
+                ]
+            elif "channel_id" in sw_channel_kw_args:
+                sw_channel_kw_args["channel_id"] = channel_match[
+                    sw_channel_args[0]
+                ]
+            else:
+                sw_channel_args[1] = channel_match[sw_channel_args[0]]
```
HGSilveri, line 470: This will work, nicely done!

tests/test_sequence.py
```diff
@@ -137,6 +137,221 @@ def test_magnetic_field():
     assert np.all(seq3_.magnetic_field == np.array((1.0, 0.0, 0.0)))
 
 
+def test_switch_device():
+    # Test devices
```
HGSilveri, line 141: Try to include at least one `VirtualDevice` in your tests

tests/test_sequence.py
```diff
@@ -137,6 +137,221 @@ def test_magnetic_field():
...
+    reg = Register.square(3, 5, prefix="q")
+
+    # Device checkout
+    seq = Sequence(reg, Chadoq2)
+    with pytest.warns(
+        UserWarning,
+        match="Switching a sequence to the same device"
+        + " returns the sequence unchanged.",
+    ):
+        seq.switch_device(Chadoq2)
```
HGSilveri, line 257: Here you should check that the returned sequence is indeed the same

tests/test_sequence.py
```diff
@@ -137,6 +137,221 @@ def test_magnetic_field():
...
+    # Different addressing channels
+    seq = Sequence(reg, test_device1)
+    seq.declare_channel("ising", "raman_global")
+
+    with pytest.raises(TypeError, match="No match for channel ising."):
+        seq.switch_device(test_device2)
+
+    seq = Sequence(reg, MockDevice)
+    seq.declare_channel("ising", "rydberg_global")
+    with pytest.warns(UserWarning):
```
HGSilveri, line 283: What warning is this?

tests/test_sequence.py
```diff
@@ -137,6 +137,221 @@ def test_magnetic_field():
     assert np.all(seq3_.magnetic_field == np.array((1.0, 0.0, 0.0)))
 
 
+def test_switch_device():
```
HGSilveri, line 140: There's an overarching issue with these tests: you are just trying to cover all the lines, rather than checking behaviour. The objective of the unit tests is to ensure everything is working properly. This includes checking that errors and warning happen when they should (which you are doing), but also asserting that the outcomes are correct when things go without errors, which involves adding `assert` statements - you are missing this part.

HGSilveri (changes requested):

tests/test_sequence.py
Varda-star: You're rigth

pulser-core/pulser/sequence/sequence.py
Varda-star: Thanks

tests/test_sequence.py
Varda-star: It's the one about the change on the Rydberg level

HGSilveri: You should specify it here

Varda-star: I was stick because when there are two warnings, I get a list of them, so that the macthing doesn't work as usual.

HGSilveri: What other warning are you getting here?

Varda-star: image

This is the error I met

HGSilveri: Well, it seems that the warning messages don't match. This might be because of regex special characters. Add `import re`at the top of the file and then put your match string inside `re.escape()`. Let's see if that works.

pulser-core/pulser/sequence/sequence.py
```diff
@@ -364,6 +365,129 @@ def config_slm_mask(self, qubits: Iterable[QubitId]) -> None:
...
+        new_seq = Sequence(self.register, new_device)
+
+        for call in self._calls[1:]:
+            if not (call.name == "declare_channel"):
+                getattr(new_seq, call.name)(*call.args, **call.kwargs)
+                continue
+            # Switch the old id with the correct id
+            sw_channel_args = list(call.args)
+            sw_channel_kw_args = call.kwargs.copy()
+            if "name" in sw_channel_kw_args:
```
HGSilveri, line 461:
```suggestion
            if "name" in sw_channel_kw_args:
```

pulser-core/pulser/sequence/sequence.py
```diff
@@ -364,6 +365,129 @@ def config_slm_mask(self, qubits: Iterable[QubitId]) -> None:
...
+                sw_channel_kw_args["channel_id"] = channel_match[
+                    sw_channel_kw_args["name"]
+                ]
+            elif "channel_id" in sw_channel_kw_args:
```
HGSilveri, line 465: First add a comment saying why these cases are not touched. 
Then add these lines:
```suggestion
            if "name" in sw_channel_kw_args:    # pragma: no cover
                sw_channel_kw_args["channel_id"] = channel_match[
                    sw_channel_kw_args["name"]
                ]
            elif "channel_id" in sw_channel_kw_args:    # pragam: no cover
```

tutorials/advanced_features/Simulating Sequences with Errors and Noise.ipynb
```diff
@@ -893,11 +893,11 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.8.5"
+   "version": "3.8.10"
   },
   "vscode": {
    "interpreter": {
-    "hash": "e088768f7ff7b4294439f8ed10f7eed9e3b885124bc20d9d06cc2a37b1883330"
+    "hash": "f824b8e4e16905929cafe3ad0d7552efed47de1342d6ced4330871a22269e77f"
```
HGSilveri, line 900: You committed these changes by accident, you should revert them

HGSilveri:
@Varda-star I think this has been dragging for quite a while now, let's push for having it merged by the end of the week, how does that sound?

tests/test_sequence.py
```diff
@@ -137,6 +137,280 @@ def test_magnetic_field():
...
+    sweep = Pulse.ConstantAmplitude(
+        2.3 * 2 * np.pi, RampWaveform(400, -4 * np.pi, 4 * np.pi), 1.0, 0.6
+    )
+    fall = Pulse.ConstantDetuning(
+        RampWaveform(500, 2.3 * 2 * np.pi, 0.0), 4 * np.pi, 0.0, 0.7
+    )
+    seq.declare_channel("ising", channel_id="rydberg_global")
+    seq.add(rise, "ising")
+    seq.add(sweep, "ising")
+    seq.add(fall, "ising")
```
HGSilveri, line 342: You should turn this into a fixture, it doesn't make sense to be repeating the same code block

tests/test_sequence.py
```diff
@@ -137,6 +137,280 @@ def test_magnetic_field():
...
+    test_device3.change_rydberg_level(test_device2.rydberg_level)
+    seq = Sequence(reg, test_device3)
+    seq.declare_channel("ising", "rmn_local1", "q0")
+    with pytest.raises(
+        ValueError,
+        match="No channel match for channel ising"
+        + " with the right fixed_retarget_t.",
+    ):
+        seq.switch_device(test_device2, True)
+
```
HGSilveri, line 412: You should make a successful strict switch on a non-empty sequence and check that the sequence samples all stay the same.

HGSilveri (commented):
On top of the comments, I think the UTs can still benefit from some extra assertions. For example: you should make a sequence with the same commands on two devices and check that switching the device on one sequence will give you the other one (even when sequence is `False`). This is the type of test that checks that your code is doing exactly what it should (note that you are not really verifying that yet)

Also, the test function is already too large, so it should be split into multiple, more modular tests.

pulser-core/pulser/sequence/sequence.py
```diff
@@ -364,6 +365,144 @@ def config_slm_mask(self, qubits: Iterable[QubitId]) -> None:
...
+                    phase_is_constant = bool(
+                        np.all(ch_sample_phase == ch_sample_phase[0])
+                    )
+                # Phase_jump_time and clock_period check
+                phase_jump_time_check = (
+                    o_d_ch_obj.phase_jump_time == n_d_ch_obj.phase_jump_time
+                )
+                clock_period_check = (
+                    o_d_ch_obj.clock_period % n_d_ch_obj.clock_period == 0
+                )
```
HGSilveri, line 456: It seems that this doesn't work after all, let's go back to:
```suggestion
                clock_period_check = (
                    o_d_ch_obj.clock_period == n_d_ch_obj.clock_period
                )
```

pulser-core/pulser/sequence/sequence.py
```diff
@@ -364,6 +365,144 @@ def config_slm_mask(self, qubits: Iterable[QubitId]) -> None:
...
+                sw_channel_kw_args["channel_id"] = channel_match[
+                    sw_channel_args[0]
+                ]
+            else:
+                sw_channel_args[1] = channel_match[sw_channel_args[0]]
+
+            if strict and alert_phase_jump:
+                warnings.warn(
+                    "The phase_jump_time of the matching channel"
+                    + "on the the new device is different, take it in account"
```
HGSilveri, line 499: typo:
```suggestion
                    + "on the the new device is different, take it into account"
```

tests/test_sequence.py
```diff
@@ -137,6 +138,288 @@ def test_magnetic_field():
     assert np.all(seq3_.magnetic_field == np.array((1.0, 0.0, 0.0)))
 
 
+def test_switch_device():
```
HGSilveri, line 141: What happened to splitting this into multiple functions? This test function is definitely too big now.

HGSilveri: You can make one function for testing all the errors and another for testing correct behaviour, for example. If you need the test devices in both, put them in one or more fixtures so you don't rewrite them

tests/test_sequence.py
```diff
@@ -137,6 +137,280 @@ def test_magnetic_field():
...
+    seq = Sequence(reg, MockDevice)
+    seq.declare_channel("ising", "rydberg_global")
+    with pytest.raises(
+        ValueError,
+        match="Device macth failed because of different rydberg levels.",
+    ):
+        seq.switch_device(IroiseMVP)
+
+    # Different Channels basis
+    test_device1.change_rydberg_level(IroiseMVP.rydberg_level)
```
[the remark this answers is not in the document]
HGSilveri: Why did you mark this as resolved without changing anything? What am I missing?

tests/test_sequence.py
```diff
@@ -137,6 +138,288 @@ def test_magnetic_field():
...
+    seq = Sequence(reg, test_device3)
+    twin_seq = Sequence(reg, test_device1)
+
+    seq.declare_channel("ising", channel_id="rydberg_global")
+    twin_seq.declare_channel("ising", channel_id="rydberg_global")
+    seq.add(rise, "ising")
+    twin_seq.add(rise, "ising")
+    seq.add(sweep, "ising")
+    twin_seq.add(sweep, "ising")
+    new_seq = seq.switch_device(test_device1, True)
```
HGSilveri, line 360: This is particularly redudant... These sequence calls can easily be put into a function

tests/test_sequence.py
```diff
@@ -137,6 +138,288 @@ def test_magnetic_field():
...
+
+    assert sample(new_seq).__repr__() == sample(twin_seq).__repr__()
```
HGSilveri, line 362: This doesn't seem to be a good way to compare samples. See how it can be done properly [here](https://github.com/pasqal-io/Pulser/blob/6b7a2f1e9715d702fae7c432983b4058eaeb7d82/tests/test_sequence_sampler.py#L48-L61), for example

tests/test_sequence.py
```diff
@@ -137,6 +138,288 @@ def test_magnetic_field():
...
+    seq.declare_channel("ising", "rydberg_global")
+
+    with pytest.raises(
+        ValueError,
+        match="No channel match for channel ising"
+        + " with the right phase_jump_time & clock_period.",
+    ):
+        seq.switch_device(test_device2, True)
+
+    test_device2.change_rydberg_level(test_device3.rydberg_level)
```
HGSilveri, line 397: Same issue here, please avoid using `change_rydberg_level()` on the `Device` class since this will be deprecated soon

tests/test_sequence.py
```diff
@@ -137,6 +138,288 @@ def test_magnetic_field():
...
+    seq = Sequence(reg, test_device3)
+    seq.declare_channel("ising", "rmn_local2", "q0")
+    assert seq.switch_device(test_device2, True)._device == test_device2
+    assert "ising" in seq.switch_device(test_device2, True).declared_channels
+
+    test_device2.change_rydberg_level(test_device1.rydberg_level)
```
HGSilveri, line 403: Same here

HGSilveri (changes requested):

tests/test_sequence.py
```diff
@@ -139,9 +139,9 @@ def test_magnetic_field():
 
 
 @pytest.fixture
-def test_devices():
+def devices():
 
-    test_device1 = Device(
+    device1 = Device(
```
HGSilveri, line 144: This is still a `Device` and you're calling `change_rydberg_level()` on it. Use a `VirtualDevice` instead

tests/test_sequence.py
```diff
@@ -174,15 +174,15 @@ def test_devices():
...
+                    clock_period=4,
                     phase_jump_time=500,
                     max_duration=2**26,
                 ),
             ),
         ),
     )
 
-    test_device2 = Device(
+    device2 = Device(
```
HGSilveri, line 185: Same here

tests/test_sequence.py
```diff
@@ -286,148 +286,171 @@ def pulses():
...
     seq = Sequence(reg, Chadoq2)
+    seq.declare_channel("ising", "rydberg_global")
+    seqs.append(seq)
+
+    seq = Sequence(reg, MockDevice)
+    seq.declare_channel("ising", "rydberg_global")
+    seqs.append(seq)
+
+    seq = Sequence(reg, devices[0])
+    seq.declare_channel("digital", "raman_global", "q1")
```
HGSilveri, line 302: Why are you declaring an `initial_target` on a global channel?

tests/test_sequence.py
```diff
@@ -286,148 +286,171 @@ def pulses():
...
+    seq.declare_channel("ising", "rmn_local2", "q0")
+    seqs.append(seq)
+
+    seq = Sequence(reg, devices[2])
+    seq.declare_channel(name="ising", channel_id="rydberg_global")
+    seq.add(pulses[0], "ising")
+    seq.add(pulses[2], "ising")
+    seqs.append(seq)
+
+    return seqs
```
HGSilveri, line 335: I'm not sure about this, it makes it very hard to know which sequence is which when reading the tests. 
Why don't you make a function like this instead (NOT a fixture)?
```python
def init_seq(device, ch_name, ch_id, *pulses, initial_target=None):
    seq = Sequence(reg, device)
    seq.declare_channel(ch_name, ch_id, initial_target=initial_target)
    for pulse in pulses:
        seq.add(pulse, ch_name)
    return seq
```
This way you can create a new sequence inside your test function in one line. That way it's also clear what's contained in the sequence from the call to the function.

tests/test_sequence.py
Varda-star: I am not calling `change_rydberg_level()` anymore

HGSilveri: You're right, my bad

tests/test_sequence.py
```diff
@@ -137,6 +138,325 @@ def test_magnetic_field():
...
+    s2 = sample(seq1)
+    s3 = sample(seq2)
+    nested_s1 = s1.to_nested_dict()["Global"]["ground-rydberg"]
+    nested_s2 = s2.to_nested_dict()["Global"]["ground-rydberg"]
+    nested_s3 = s3.to_nested_dict()["Global"]["ground-rydberg"]
+
+    # Check if the samples are the same
+    for key in ["amp", "det", "phase"]:
+        np.testing.assert_array_equal(nested_s1[key], nested_s3[key])
+        np.testing.assert_array_equal(nested_s1[key], nested_s2[key])
```
HGSilveri, line 446: Continuing the example above, I think you'll need to add this:
```suggestion
        np.testing.assert_array_equal(nested_s1[key], nested_s3[key])
        if strict:
            np.testing.assert_array_equal(nested_s1[key], nested_s2[key])
```

tests/test_sequence.py
```diff
@@ -137,6 +138,327 @@ def test_magnetic_field():
...
+    with pytest.raises(
+        ValueError,
+        match="No channel match for channel digital"
+        + " with the right fixed_retarget_t.",
+    ):
+        seq.switch_device(devices[1], True)
+
+
+@pytest.mark.parametrized("device_ind, strict", [(1, False), (2, True)])
+def test_switch_device_up(devices, pulses, device_ind, strict):
```
HGSilveri, line 413: The test might have vailed because of the order of the arguments, maybe try 
```suggestion
def test_switch_device_up(device_ind, strict, devices, pulses):
```
Test it locally first and the push, rather than just committing my suggestion.

Varda-star: Why is it taking `device_ind` as a fixture?

Varda-star: It's `..._parametrize` instead of `..._parametrized`

HGSilveri: Oh okay, good catch!

HGSilveri (approved):
LGTM, good job!

PR #400, opened 2022-09-23
Create the `pulser-pasqal` extension package

### LINKED ISSUE #399: Create the `pulser-pasqal` extension
Follow the same template used for `pulser-core` and `pulser-simulation`.

HGSilveri:
Closes #399 .

### DIFF
--- a/pulser-pasqal/setup.py
+++ b/pulser-pasqal/setup.py
@@ -1,0 +1,70 @@
+# Copyright 2022 Pulser Development Team
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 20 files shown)


pulser-pasqal/setup.py
```diff
@@ -0,0 +1,70 @@
...
+requirements.append(f"pulser-core=={__version__}")
+
+# Stashes the source code for the local version file
+local_version_fpath = Path(package_name) / "_version.py"
+with open(local_version_fpath, "r") as f:
+    stashed_version_source = f.read()
+
+# Overwrites the _version.py for the source distribution (reverted at the end)
+with open(local_version_fpath, "w") as f:
+    f.write(f"__version__ = '{__version__}'\n")
```
CdeTerra, line 45: Just a question: why do we need this, since "_version.py" already contains the code to get the right version?

CdeTerra (approved):
LGTM, thanks for creating this package!

pulser-pasqal/setup.py
HGSilveri: Ah yes, good question. This is for the distribution builds, because then there will be no `VERSION.txt` file to get the right version from. This overwrites the versions before it builds and then brings the original code back.

PR #402, opened 2022-10-07
Remove restriction that only Chadoq2 was supported for abstract sequence

awennersteen:
Remove restriction that only Chadoq2 was supported for abstract sequence. 

It would be useful if this could also be released as 0.7.1 since we're using it for validating sequences in the cloud service.

### DIFF
--- a/pulser-core/pulser/json/abstract_repr/schema.json
+++ b/pulser-core/pulser/json/abstract_repr/schema.json
@@ -150,6 +150,14 @@
       ],
       "type": "object"
     },
+    "Device": {
+      "enum": [
+        "Chadoq2",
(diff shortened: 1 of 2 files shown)


HGSilveri:
@awennersteen Ok, if want this to be released as 0.7.1 then we need to treat this change as an hotfix and target directly `master` instead. You can find the details [here](https://github.com/pasqal-io/Pulser/blob/develop/release.md#preparing-a-hotfix), let me know if you need any help.

HGSilveri (commented):
By the way, the code looks good to me, thanks!

awennersteen:
Closing in favor of #403.

PR #403, opened 2022-10-07
Make abstract sequence validation accept all devices

awennersteen:
Hotfix to allow abstract sequences with all currently defined `Device` types.

Would release v0.7.1


HGSilveri (approved):
LGTM, but I'll ask @bejito for review first to make sure this doesn't interfere with his projects

PR #412, opened 2022-10-28
Updates to the Abstract Representation

### LINKED ISSUE #397: Make abstract JSON representation feature complete
The abstract is really useful for multiple reasons:
- Not tied to python, easier to interpret in other languages
- Not tied to the implementation, easier to keep stable across time

In order to be able to use it fully, we need to add some of the missing features to it:
- RegisterLayout, MappableRegister
- Device serialization (independently from the sequence)

HGSilveri:
- Updates the JSON schema for the abstract representation, making it feature complete
- Updates the serializer and deserializer to work with the latest schema
- Fixes an internal bug in `Sequence.build()` where `Sequence._calls` were being wrongly modified when building a sequence with a mappable register

Fixes #397 .


pulser-core/pulser/sequence/sequence.py
```diff
@@ -825,6 +825,9 @@ def build(
                 >>> seq1 = seq.build(x=0.5, y=[1, 2, 3])
         """
         # Shallow copy with stored parametrized objects (if any)
+        # NOTE: While seq is a shallow copy, be extra careful with changes to
```
bejito, line 828: Do we actually need those references or should we use `deepcopy` instead of `copy` to be safer ?

pulser-core/pulser/sequence/sequence.py
```diff
@@ -861,6 +860,15 @@ def build(
         # Deepcopy the base sequence (what remains)
         seq = copy.deepcopy(seq)
 
+        # NOTE: Changes to seq are now safe to do
+
+        if qubits:
+            reg = cast(MappableRegister, self._register).build_register(qubits)
+            self._set_register(seq, reg)
+            assert self.is_register_mappable()
```
bejito, line 868: Why do we need this assertion ? Shouldn't `_set_register` with a `MappableRegister` ensure this ?

pulser-core/pulser/sequence/sequence.py
```diff
@@ -861,6 +860,15 @@ def build(
...
+            assert not seq.is_register_mappable()
+            assert isinstance(self._calls[0].args[0], MappableRegister)
```
bejito, line 870: This also looks like something that belongs to unit tests no ?

pulser-core/pulser/sequence/sequence.py
HGSilveri: Yes, I committed this by mistake, sorry...

pulser-core/pulser/sequence/sequence.py
HGSilveri: A `deepcopy` before `Sequence._reset_parametrized()` triggers a bunch of recursiveness errors. However, I can (and will) do is run `copy` -> `_reset_parametrized` -> `deepcopy` to ensure nothing fishy happens while the copy is shallow.

pulser-core/pulser/json/utils.py
```diff
@@ -60,3 +65,22 @@ def obj_to_dict(
 
     pulser.json.supported.validate_serialization(d)
     return d
+
+
+def stringify_qubit_ids(qubit_ids: Sequence[QubitId]) -> list[str]:
```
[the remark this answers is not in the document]
HGSilveri: Yes, but this was encoded in `Register._to_abstract_repr()` and now we also need it in the `MappableRegister` so I moved the logic to this util function

pulser-core/pulser/json/abstract_repr/serializer.py
```diff
@@ -102,36 +102,47 @@ def serialize_abstract_sequence(
...
+        res["variables"][var.name] = dict(type=var.dtype.__name__)
+
+    qubits_default = defaults.pop("qubits", None)
+    if defaults or qubits_default:
+        seq._cross_check_vars(defaults)
+        try:
+            seq.build(qubits=qubits_default, **defaults)
+        except Exception:
+            raise ValueError(
+                "The given 'defaults' produce an invalid sequence."
```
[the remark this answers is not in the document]
HGSilveri: Yes, that is actually much trickier to handle, so I'd rather leave it for later. For now, we either have all the defaults for Pulser Studio or no defaults for standard Pulser, which is sufficient imo

pulser-core/pulser/json/abstract_repr/schema.json
```diff
@@ -261,9 +262,11 @@
     "HardwareChannel": {
       "description": "Hardware channel name.",
       "enum": [
+        "raman_global",
         "raman_local",
         "rydberg_local",
-        "rydberg_global"
+        "rydberg_global",
+        "mw_global"
```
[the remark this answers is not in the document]
HGSilveri: Technically, yes, but we could still imagine we have it for emulation. It's not here because none of the Devices supports it (not even `MockDevice`).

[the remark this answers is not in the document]
bejito: Ok, I was just asking that out of curiosity.

pulser-core/pulser/sequence/sequence.py
bejito: This seems to be already implemented actually

pulser-core/pulser/json/abstract_repr/deserializer.py
```diff
@@ -236,16 +239,46 @@ def deserialize_abstract_sequence(obj_str: str) -> Sequence:
...
+        RegisterLayout(obj["layout"]["coordinates"])
+        if "layout" in obj
+        else None
+    )
+
     # Register
+    reg: Union[BaseRegister, MappableRegister]
     qubits = obj["register"]
-    reg = Register({q["name"]: (q["x"], q["y"]) for q in qubits})
+    if {"name", "x", "y"} == qubits[0].keys():
```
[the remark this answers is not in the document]
HGSilveri: I think it would have to be a field that is only available (and required) when the register is mappable (I'm thinking something like `with_mappable_reg: true`). 
Otherwise, it would be a breaking change unless it is optional (which would defeat the purpose). 
Should we go for it?

pulser-core/pulser/sequence/sequence.py
HGSilveri: Yes, I'll just place one after the other so that nothing is placed in between them in the fututre.

pulser-core/pulser/json/abstract_repr/deserializer.py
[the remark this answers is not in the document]
bejito: Arf, true. 
In this case, it's pretty easy so let's keep it this way. But next time we need to do `oneOf`s, using a `type` prop` can be very helpful.

bejito (approved):

PR #413, opened 2022-10-28
Extend support to Python 3.11

HGSilveri:
Makes CI/CD run in Python 3.11 to ensure compatibility


.github/workflows/ci.yml
```diff
@@ -57,7 +57,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.7', '3.9']
+        python-version: ["3.7", "3.11"]
```
CdeTerra, line 60: How about including all supported versions?

CdeTerra (approved):

.github/workflows/ci.yml
HGSilveri: We do that on the full test routine that runs when the PR is merged. I think it is excessive to test against all Python versions every time there is an update in a PR

PR #416, opened 2022-11-04
Update phase jump behaviour

HGSilveri:
- [x] Deduce the `phase_jump_time` from the modulation bandwidth instead of taking it as a free parameter
- [x] Include the previous pulse's fall time in the buffer where the phase between pulses changes
- [x] Make the buffering between pulses with different phases overridable? Right now it is always enforced


HGSilveri:
@CdeTerra I'm still waiting for a change coming in #421 but if you could, please start reviewing the existing changes already because it would make merging much quicker afterwards. Thanks in advance!

tests/test_sequence.py
```diff
@@ -996,13 +955,13 @@ def test_hardware_constraints():
     mid_delay = 40
     seq.delay(mid_delay, "ch0")
     seq.add(const_pls, "ch0")  # Phase = π
-    assert seq._last("ch0").ti - tf_ == rydberg_global.phase_jump_time
+    interval = seq._schedule["ch0"].adjust_duration(
+        rydberg_global.phase_jump_time + black_pls.fall_time(rydberg_global)
+    )
+    assert seq._last("ch0").ti - tf_ == interval
```
CdeTerra, line 961: This might not be ideal, since it is testing the behavior of `seq` against itself (if `adjust_duration` and `seq._last("ch0").ti - tf_` are both wrong, it may still pass).

CdeTerra (approved):

tests/test_sequence.py
HGSilveri: Changed to this:
```suggestion
    interval = seq._schedule["ch0"].adjust_duration(
        rydberg_global.phase_jump_time + black_pls.fall_time(rydberg_global)
    )
    assert seq._schedule["ch0"][-1].ti - tf_ == interval
```

CdeTerra: I would say that `seq` is still tested against itself here (between `seq._schedule["ch0"].adjust_duration` and `seq._schedule["ch0"][-1].ti`).

HGSilveri: Ok, how about this then? 
```python
    raw_interval = rydberg_global.phase_jump_time + black_pls.fall_time(
        rydberg_global
    )
    assert raw_interval >= rydberg_global.min_duration
    clock = rydberg_global.clock_period
    interval = raw_interval + (
        clock - raw_interval % clock if raw_interval % clock else 0
    )
    assert seq._schedule["ch0"][-1].ti - tf_ == interval
```
Feels a bit like overkill to me but I'm okay with it.

CdeTerra: This third option tests `Sequence` against `Pulse`, so I think that's equivalent.
I can guess this is hard to test cleanly anyway, because the calculations are complex. The best would probably to keep the first or the second option. And then maybe, add tests inside "test_pulse.py" to actually test `fall_time`, but with simple hard-coded values, instead of calculations (usually, it's better to avoid logics and calculations as much as possible in UTs because they can be as buggy as the tested code).

HGSilveri: Ok, I added a simple test for `fall_time` as suggested

pulser-core/pulser/sequence/_schedule.py
```diff
@@ -261,7 +261,9 @@ def add_pulse(
         t0 = last.tf
         current_max_t = max(t0, *phase_barrier_ts)
         for ch, ch_schedule in self.items():
-            if protocol == "no-delay" or ch == channel:
+            if protocol == "no-delay":
+                break
```
CdeTerra, line 265: This could be put outside of the for-loop in this way:
```python
if protocol != "no-delay":
    ... # code of the for-loop
```
Or, even encompassing the subsequent lines:
```python
phase_jump_buffer = 0
if protocol != "no-delay":
    ... # code of the for-loop
    ... # code of the try except
```
(Ideally, the for-loop would also be put in a separate function)

CdeTerra (approved):

pulser-core/pulser/sequence/_schedule.py
HGSilveri: Yes, I tried that but it pushed the code too much to the right...

HGSilveri: Ok, I followed your suggestion with the changes in the (hopefully) last commit. Wdyt @CdeTerra ?

CdeTerra (approved):

PR #417, opened 2022-11-09
Add the classes for EOM mode configuration

HGSilveri:
- Convert `channels.py` module into a module directory 
- Creates the classes for configuring operation in the EOM mode. Their usage by `Sequence` will come in a second PR, but is already done on branch [`eom-seq`](https://github.com/pasqal-io/Pulser/tree/eom-seq).


pulser-core/pulser/channels/channels.py
```diff
@@ -366,48 +368,62 @@ def validate_pulse(self, pulse: Pulse) -> None:
...
             )
             return input_samples
+        else:
+            mod_bandwidth = self.mod_bandwidth
+            rise_time = self.rise_time
 
         # The cutoff frequency (fc) and the modulation transfer function
         # are defined in https://tinyurl.com/bdeumc8k
-        fc = self.mod_bandwidth * 1e-3 / np.sqrt(np.log(2))
+        fc = mod_bandwidth * 1e-3 / np.sqrt(np.log(2))
```
madagra, line 408: Does the `1e-3` factor come from the units?

pulser-core/pulser/channels/eom.py
```diff
@@ -0,0 +1,170 @@
...
+            if not (isinstance(beam, RydbergBeam) and beam in RydbergBeam):
+                raise TypeError(
+                    "Every beam must be one of options of the `RydbergBeam`"
+                    f" enumeration, not {self.limiting_beam}."
+                )
+
+    def detuning_off_options(
+        self, rabi_frequency: float, detuning_on: float
+    ) -> list[float]:
+        """Calculates the possible detuning values when the amplitude is off.
```
madagra, line 121: This is the detuning that is applied because of the presence of the EOM during free evolution, right?

madagra (commented):
I have some minor comments (mainly due to a lack of context here which can be covered offline). 

The main point I do not get from this PR is how the EOM modulation is used in practice in a simulation. From what I understand now, the only things which are actually used in the `Channel.modulate()` method from the `EOM` class are the values of modulation bandwidth and rise time. I guess this will be part of the follow-up PR, right?

pulser-core/pulser/channels/channels.py
HGSilveri: Yes, exactly. The bandwidth is in MHz, but our timescale is ns, so we need to work in GHz.

pulser-core/pulser/channels/channels.py
```diff
@@ -366,48 +368,62 @@ def validate_pulse(self, pulse: Pulse) -> None:
...
         if keep_ends:
-            samples = np.pad(input_samples, 2 * self.rise_time, mode="edge")
+            samples = np.pad(input_samples, 2 * rise_time, mode="edge")
         else:
-            samples = np.pad(input_samples, self.rise_time)
+            samples = np.pad(input_samples, rise_time)
         freqs = fftfreq(samples.size)
         modulation = np.exp(-(freqs**2) / fc**2)
         mod_samples = ifft(fft(samples) * modulation).real
```
madagra, line 415: This operation could be potentially expensive depending on the sample size (FFT --> modulation --> IFFT). Is it executed every time before sequence execution?

HGSilveri: I'm not sure what you mean by "sequence execution", but I can tell you that it is not called in any time-sensitive part. Potential delays are confined to taking a bit longer to draw a sequence, pre-process it for simulation (ie when creating the time-dependent Hamiltonian) and potentially when adding a very long pulse.

madagra: Thanks, understood.

pulser-core/pulser/channels/channels.py
```diff
@@ -11,20 +11,21 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""The various hardware channel types."""
+"""Defines the Channel ABC and its subclasses."""
```
madagra, line 14: This file is very long. Wouldn't be better to try splitting it at least into two parts?

pulser-core/pulser/channels/channels.py
```diff
@@ -475,7 +497,10 @@ def __repr__(self) -> str:
         return self.name + config
 
     def _to_dict(self) -> dict[str, Any]:
-        return obj_to_dict(self, **asdict(self))
+        params = {
+            f.name: getattr(self, f.name) for f in fields(self) if f.init
+        }
+        return obj_to_dict(self, _module="pulser.channels", **params)
```
madagra, line 503: Cannot you use the `asdict` method provided by the `dataclass` library?

HGSilveri: You mean as a replacement for `obj_to_dict` or as it was being used before? I'll answer both, just in case.
- `obj_to_dict` is what is used by the JSON serializer and it includes some extra metadata that `asdict` won't.
- `asdict` will include all fields, even those that are not on `__init__`. This was then causing problems deserializing because `reusable_channels` is not in the `Device.__init__()`, so now we filter the fields accordingly.

pulser-core/pulser/channels/channels.py
HGSilveri: Sure, I guess I can move the subclasses to another file.

pulser-core/pulser/channels/channels.py
madagra: Got it, I understand now.

tests/test_eom.py
```diff
@@ -0,0 +1,120 @@
...
+    assert det_off_options[1] == zero_det  # All off
+    assert det_off_options[2] > zero_det  # BLUE on
+
+    detuning_on = 1.0
+    for beam, ind in [(RydbergBeam.RED, 2), (RydbergBeam.BLUE, 0)]:
+        params["controlled_beams"] = (beam,)
+        eom_ = RydbergEOM(**params)
+        off_options = eom_.detuning_off_options(amp, detuning_on)
+        assert len(off_options) == 1
+        assert off_options[0] == det_off_options[ind] + detuning_on
```
madagra, line 120: @HGSilveri Can you clarify what exactly are you testing in this one? As far as I understood, you are manually computing in the test the light shift and the detuning offset and checking that it is correctly returned by the `RydbergEOM` class, right?

madagra (commented):
I have a final comment on the unit tests, but it looks good to me after that. It might be useful to add in the documentation some succinct information on how the detuning offset and light shift is computed (with references on where it comes from).

Issue #418, opened 2022-11-09
Add tutorial and documentation for EOM mode operation


HGSilveri:
Should follow the full merge of the EOM mode features, which started on PR #417.

PR #421, opened 2022-11-28
Incorporate EOM mode operation in `Sequence`

### LINKED ISSUE #367: Tutorial on modulated output sequence creation and simulation
A tutorial for the docs showcasing how to use the output modulation features. 

Should be done after #315.

HGSilveri:
- Adds EOM mode operation to `Sequence`
- Incorporates EOM mode modulation in the `sampler`
- Adds a tutorial on output modulation and EOM mode operation (purposefully stored without output; to check the output, see the readthedocs CI build)

Closes #367 , #418.

### DIFF
--- a/pulser-core/pulser/sampler/samples.py
+++ b/pulser-core/pulser/sampler/samples.py
@@ -86,6 +86,7 @@ class ChannelSamples:
     det: np.ndarray
     phase: np.ndarray
     slots: list[_TargetSlot] = field(default_factory=list)
+    # (t_start, t_end) of each EOM mode block
     eom_intervals: list[tuple[int, int]] = field(default_factory=list)
 
     def __post_init__(self) -> None:
--- a/pulser-core/pulser/sampler/samples.py
+++ b/pulser-core/pulser/sampler/samples.py
@@ -2,7 +2,7 @@
 from __future__ import annotations
 
 from collections import defaultdict
-from dataclasses import dataclass, field
+from dataclasses import dataclass, field, replace
 from typing import Optional
 
 import numpy as np
@@ -86,6 +86,7 @@ class ChannelSamples:
     det: np.ndarray
     phase: np.ndarray
     slots: list[_TargetSlot] = field(default_factory=list)
+    eom_intervals: list[tuple[int, int]] = field(default_factory=list)
 
     def __post_init__(self) -> None:
         assert len(self.amp) == len(self.det) == len(self.phase)
@@ -120,7 +121,7 @@ class ChannelSamples:
             (0, extension),
             mode="edge" if self.phase.size > 0 else "constant",
         )
-        return ChannelSamples(new_amp, new_detuning, new_phase, self.slots)
+        return replace(self, amp=new_amp, det=new_detuning, phase=new_phase)
 
     def is_empty(self) -> bool:
         """Whether the channel is effectively empty.
@@ -148,11 +149,43 @@ class ChannelSamples:
         Returns:
             The modulated channel samples.
         """
-        times = slice(0, max_duration)
-        new_amp = channel_obj.modulate(self.amp)[times]
-        new_detuning = channel_obj.modulate(self.det)[times]
-        new_phase = channel_obj.modulate(self.phase, keep_ends=True)[times]
-        return ChannelSamples(new_amp, new_detuning, new_phase, self.slots)
+
+        def masked(samples: np.ndarray, mask: np.ndarray) -> np.ndarray:
+            new_samples = samples.copy()
+            new_samples[~mask] = 0
+            return new_samples
+
+        new_samples: dict[str, np.ndarray] = {}
+
+        if self.eom_intervals:
+            eom_mask = np.zeros(self.duration, dtype=bool)
+            for start, end in self.eom_intervals:
+                end = min(end, self.duration)  # This is defensive
+                eom_mask[np.arange(start, end)] = True
+
+            for key in ("amp", "det"):
+                samples = getattr(self, key)
+                std = channel_obj.modulate(masked(samples, ~eom_mask))
+                eom = channel_obj.modulate(masked(samples, eom_mask), eom=True)
+                sample_arrs = [std, eom]
+                sample_arrs.sort(key=len)
+                # Extend shortest array to match the longest
+                sample_arrs[0] = np.concatenate(
+                    (
+                        sample_arrs[0],
+                        np.zeros(sample_arrs[1].size - sample_arrs[0].size),
+                    )
+                )
+                new_samples[key] = sample_arrs[0] + sample_arrs[1]
+
+        else:
+            new_samples["amp"] = channel_obj.modulate(self.amp)
+            new_samples["det"] = channel_obj.modulate(self.det)
+
+        new_samples["phase"] = channel_obj.modulate(self.phase, keep_ends=True)
+        for key in new_samples:
+            new_samples[key] = new_samples[key][slice(0, max_duration)]
+        return replace(self, **new_samples)
 
 
 @dataclass
--- a/pulser-core/pulser/sequence/_schedule.py
+++ b/pulser-core/pulser/sequence/_schedule.py
@@ -77,12 +77,8 @@ class _ChannelSchedule:
                 max(duration, self.channel_obj.min_duration)
             )
 
-    def get_samples(self, modulated: bool = False) -> ChannelSamples:
-        """Returns the samples of the channel.
-
-        Args:
-            modulated: Whether to return the modulated samples.
-        """
+    def get_samples(self) -> ChannelSamples:
+        """Returns the samples of the channel."""
         # Keep only pulse slots
         channel_slots = [s for s in self.slots if isinstance(s.type, Pulse)]
         dt = self.get_duration()
@@ -102,25 +98,20 @@ class _ChannelSchedule:
             )
             phase[t_start:t_end] += pulse.phase
             tf = s.tf
-            if modulated:
-                # Account for the extended duration of the pulses
-                # after modulation, which is at most fall_time
-                fall_time = pulse.fall_time(self.channel_obj)
-                tf += (
-                    min(fall_time, channel_slots[ind + 1].ti - s.tf)
-                    if ind < len(channel_slots) - 1
-                    else fall_time
-                )
+            # Account for the extended duration of the pulses
+            # after modulation, which is at most fall_time
+            fall_time = pulse.fall_time(
+                self.channel_obj, in_eom_mode=self.in_eom_mode(t=s.ti)
+            )
+            tf += (
+                min(fall_time, channel_slots[ind + 1].ti - s.tf)
+                if ind < len(channel_slots) - 1
+                else fall_time
+            )
 
             slots.append(_TargetSlot(s.ti, tf, s.targets))
 
         ch_samples = ChannelSamples(amp, det, phase, slots)
-
-        if modulated:
-            ch_samples = ch_samples.modulate(
-                self.channel_obj,
-                max_duration=self.get_duration(include_fall_time=True),
-            )
 
         return ch_samples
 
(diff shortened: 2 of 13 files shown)


pulser-core/pulser/sequence/_schedule.py
```diff
@@ -257,3 +350,12 @@ def add_target(self, qubits_set: set[QubitId], channel: str) -> None:
...
         )
+
+    def wait_for_fall(self, channel: str) -> None:
+        """Adds a delay to let the channel's amplitude ramp down."""
+        fall_time = (
+            self[channel].get_duration(include_fall_time=True)
+            - self[channel].get_duration()
+        )
+        if fall_time > 0:
+            self.add_delay(self[channel].adjust_duration(fall_time), channel)
```
madagra, line 361: I assume this is to wait for a pulse that takes longer to go to zero because of modulation. Could you add a comment about it?

madagra, line 361: I assume this is to wait for a pulse that takes longer to go to zero because of modulation. Could you add a comment about it?

madagra (commented):
The review still needs to be finalized.

tutorials/advanced_features/Output Modulation and EOM Mode.ipynb
```diff
@@ -0,0 +1,317 @@
+{
```
madagra, line 1: <p>Has this new way of creating virtual devices been replaced everywhere across the documentation?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pasqal-io/Pulser/pull/421/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

pulser-core/pulser/sampler/samples.py
```diff
@@ -120,7 +121,7 @@ def extend_duration(self, new_duration: int) -> ChannelSamples:
             (0, extension),
             mode="edge" if self.phase.size > 0 else "constant",
         )
-        return ChannelSamples(new_amp, new_detuning, new_phase, self.slots)
+        return replace(self, amp=new_amp, det=new_detuning, phase=new_phase)
```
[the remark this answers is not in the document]
HGSilveri: No, it creates a new instance, leaving all the arguments that were not specified the same.

pulser-core/pulser/sampler/samples.py
```diff
@@ -86,6 +86,7 @@ class ChannelSamples:
     det: np.ndarray
     phase: np.ndarray
     slots: list[_TargetSlot] = field(default_factory=list)
+    eom_intervals: list[tuple[int, int]] = field(default_factory=list)
```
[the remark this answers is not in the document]
HGSilveri: Sure! Here's what I added: 
```suggestion
    # (t_start, t_end) of each EOM mode block
    eom_intervals: list[tuple[int, int]] = field(default_factory=list)
```

pulser-core/pulser/sequence/_schedule.py
```diff
@@ -102,25 +147,22 @@ def get_samples(self, modulated: bool = False) -> ChannelSamples:
             )
             phase[t_start:t_end] += pulse.phase
             tf = s.tf
-            if modulated:
```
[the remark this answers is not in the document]
HGSilveri: I got rid of the option of getting the channel samples modulated directly from `_ChannelSchedule.get_samples()` because it was giving inconsistent results. 
The main issue was that modulation followed by extension (ie increasing the length of the samples) is different then extension follow by modulation; I made it so modulation allows followed extension when creating the `SequenceSamples`.
The changes are isolated in ea3b3a73464879dac299d4974586fec085672b75.

As for this case in particular, the `Pulse.fall_time()` is zero when there is no modulation bandwidth specified, so there's no need for the `if` statement

pulser-core/pulser/sequence/_schedule.py
HGSilveri: Yes, but what would like me to add that's not already in the docstring?

tutorials/advanced_features/Output Modulation and EOM Mode.ipynb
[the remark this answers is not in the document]
HGSilveri: <p>Yeah, sure. What comparison do you think would be best? Against trying to do the same sequence with only the AOM (with modulation) or against the case were there is no modulation?</p>

pulser-core/pulser/sampler/samples.py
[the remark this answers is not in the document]
madagra: Great, thanks!

pulser-core/pulser/sequence/_schedule.py
```diff
@@ -177,6 +219,34 @@ def find_slm_mask_times(self) -> list[int]:
...
+        amp_on: float,
+        detuning_on: float,
+        detuning_off: float,
+    ) -> None:
+        channel_obj = self[channel_id].channel_obj
+        if any(isinstance(op.type, Pulse) for op in self[channel_id]):
+            # Wait for the last pulse to ramp down (if needed)
+            self.wait_for_fall(channel_id)
+            # Account for time needed to ramp to desired amplitude
+            self.add_delay(2 * channel_obj.rise_time, channel_id)
```
madagra, line 234: Where does the factor 2 come from here?

pulser-core/pulser/sequence/_schedule.py
madagra: Personally, I like to have some in-code comments to help developers understand the most tricky parts of an implementation. In this case, I would add one here.

pulser-core/pulser/sequence/_schedule.py
HGSilveri: In fact, by definition the `rise_time` is the time it takes to ramp from 10% to 90% of the maximum value. Visually, I find that roughly `2 * rise_time` is what it takes to go from 0% to 100%.

HGSilveri: I'll add a comment here too 😅

HGSilveri:
CI tests are failing due to an issue installing QuTip. Started a bug report in https://github.com/qutip/qutip/issues/2036

pulser-core/pulser/sequence/sequence.py
```diff
@@ -671,6 +716,188 @@ def declare_variable(
...
+            detuning_on: The detuning of the EOM pulses (in rad/µs).
+            detuning_off_choice: The criteria for choosing among the detuning
+                options for when there is no EOM pulse playing.
+
+                - ``'closest-to-zero'``: Picks the detuning value that's
+                    closest to zero.
+
+                - ``'highest'``: Picks the highest detuning value available.
+
+                - ``'lowest'``: Picks the lowest detuning value available.
```
HGSilveri, line 760: I must say, I'm not happy with this way of automatically selecting `detuning_off`.
I thought this three options would allow one to indirectly select each of the cases (both beams off, red on or blue on), but I have encountered an example where that is not the case.
What if, instead of having these protocols, we asked for an `optimal_detuning_off` value and then the selected `detuning_off` would be the closest to such a value?

HGSilveri: @madagra Do you have any thoughts on this?

tutorials/advanced_features/Output Modulation and EOM Mode.ipynb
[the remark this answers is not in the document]
madagra: @HGSilveri I would show it against the case without any modulation (which will likely be the case of most users of EOM mode I think) + maybe it is interesting to plot the states distribution as well.

madagra: I meant that you are now using the new `VirtualDevice` which is quite recent. I wanted to understand if this has been replaced across the whole documentation even though backward compatibility is maintained.

pulser-core/pulser/sequence/sequence.py
```diff
@@ -370,6 +398,7 @@ def config_slm_mask(self, qubits: Iterable[QubitId]) -> None:
         # If checks have passed, set the SLM mask targets
         self._slm_mask_targets = targets
 
+    @seq_decorators.screen
```
madagra, line 401: @HGSilveri What is this decorator for exactly?

pulser-core/pulser/sequence/sequence.py
madagra: @HGSilveri I was thinking on the same line. An `optimal_detuning_off` looks a more flexible and clear option to me. However, what default value has to be assigned to the detuning if no `optimal_detuning_off` is passed?

pulser-core/pulser/sequence/sequence.py
```diff
@@ -671,6 +716,188 @@ def declare_variable(
...
+            # Test the parameters
+            if not isinstance(duration, Parametrized):
+                channel_obj = self.declared_channels[channel]
+                channel_obj.validate_duration(duration)
+            for arg in (phase, post_phase_shift):
+                if not isinstance(arg, (float, int)):
+                    raise TypeError("Phase values must be a numeric value.")
+
+            return
+        eom_settings = self._schedule[channel].eom_blocks[-1]
```
madagra, line 891: Add a space here.

pulser-core/pulser/sequence/sequence.py
```diff
@@ -370,6 +398,7 @@ def config_slm_mask(self, qubits: Iterable[QubitId]) -> None:
         # If checks have passed, set the SLM mask targets
         self._slm_mask_targets = targets
 
+    @seq_decorators.screen
     def switch_device(
```
madagra, line 402: What is the advantage of this method over rebuilding the same sequence with a different device? Or is it just for convenience?

madagra (commented):
Some minor comments and clarifications but overall looks almost ready to me.

pulser-core/pulser/sequence/sequence.py
HGSilveri: I think `optimal_detuning_off=0` is the only thing that makes sense because in regular operation the detuning is zero during delays.

pulser-core/pulser/sequence/sequence.py
HGSilveri: It blocks call to the method when the sequence is parametrized. For now, at least, this method won't work when the sequence is parametrized, and since it's just a convenience method anyway, it's simpler to just block it for now. If there is the need for it, we can extend it later on.

pulser-core/pulser/sequence/sequence.py
```diff
@@ -1317,14 +1547,18 @@ def _last(self, channel: str) -> _TimeSlot:
...
+    ) -> None:
         if isinstance(channel, Parametrized):
             raise NotImplementedError(
                 "Using parametrized objects or variables to refer to channels "
                 "is not supported."
             )
         if channel not in self._schedule:
             raise ValueError("Use the name of a declared channel.")
+        if block_eom_mode and self.is_in_eom_mode(channel):
+            raise RuntimeError("The chosen channel is in EOM mode.")
```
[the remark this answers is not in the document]
HGSilveri: The `_validate_channel()` method is used on every call that provides a channel as an argument to check it is valid. I added the `block_eom_mode` flag that, when `True`, will also raise an error if the channel is currently in eom mode. I'm using this to block calls that are invalid during EOM mode, like `add()`, `target()`, ...

pulser-core/pulser/sequence/sequence.py
HGSilveri: It's just a convenience. In particular, it can make sure for you that the sequence remains the same on the new device, which can be useful.

tutorials/advanced_features/Output Modulation and EOM Mode.ipynb
HGSilveri: There is no tutorial explaining how to create a `VirtualDevice` (nor a `Device`, for that matter), so I think there is nothing to replace because this is just another way of getting a device for your sequence. 
That being said, the distinction between the two is already explained in the [docs](https://pulser.readthedocs.io/en/latest/apidoc/core.html#structure-of-a-device).

madagra (approved):
Looks ready to be merged for me.

PR #423, opened 2022-11-29
Fix type annotation in run() method of Simulation() class.

### LINKED ISSUE #422: Inconsistent type annotation in Simulation.run()
In the `run()` method of the `Simulation()` class there's an inconsistent type annotation for the options passed to the QuTiP solver ([see line](https://github.com/pasqal-io/Pulser/blob/develop/pulser-simulation/pulser_simulation/simulation.py#L819)). It says that the type should be a `qutip.solver.Options` object, but the code looks like it's treated as a dictionary, with the key-value pairs being splatted into the `Options` object when it's generated [a few lines later](https://github.com/pasqal-io/Pulser/blob/develop/pulser-simulation/pulser_simulation/simulation.py#L847).

No matter how I try, I cannot seem to get it to accept a QuTiP `Options` object, but things are fine with a regular `dict`.

dehond:
Fixes #422

### DIFF
--- a/pulser-simulation/pulser_simulation/simulation.py
+++ b/pulser-simulation/pulser_simulation/simulation.py
@@ -819,7 +819,7 @@ class Simulation:
     def run(
         self,
         progress_bar: Optional[bool] = False,
-        **options: dict[str, Any],
+        **options: Any,
     ) -> SimulationResults:
         """Simulates the sequence using QuTiP's solvers.
 
@@ -829,10 +829,10 @@ class Simulation:
         Args:
             progress_bar: If True, the progress bar of QuTiP's
                 solver will be shown. If None or False, no text appears.
-            options: If specified, will override
-                SimConfig solver_options. If no `max_step` value is provided,
-                an automatic one is calculated from the `Sequence`'s schedule
-                (half of the shortest duration among pulses and delays).
+            options: Used as arguments for qutip.Options(). If specified, will
+                override SimConfig solver_options. If no `max_step` value is
+                provided, an automatic one is calculated from the `Sequence`'s
+                schedule (half of the shortest duration among pulses and delays).
         """
         if "max_step" in options.keys():
             solv_ops = qutip.Options(**options)
--- a/pulser-simulation/pulser_simulation/simulation.py
+++ b/pulser-simulation/pulser_simulation/simulation.py
@@ -819,7 +819,7 @@ class Simulation:
     def run(
         self,
         progress_bar: Optional[bool] = False,
-        **options: qutip.solver.Options,
+        **options: dict[str, Any],
     ) -> SimulationResults:
         """Simulates the sequence using QuTiP's solvers.
 
--- a/pulser-simulation/pulser_simulation/simulation.py
+++ b/pulser-simulation/pulser_simulation/simulation.py
@@ -832,7 +832,8 @@ class Simulation:
             options: Used as arguments for qutip.Options(). If specified, will
                 override SimConfig solver_options. If no `max_step` value is
                 provided, an automatic one is calculated from the `Sequence`'s
-                schedule (half of the shortest duration among pulses and delays).
+                schedule (half of the shortest duration among pulses and
+                delays).
         """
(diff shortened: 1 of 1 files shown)


pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -819,7 +819,7 @@ def get_hamiltonian(self, time: float) -> qutip.Qobj:
     def run(
         self,
         progress_bar: Optional[bool] = False,
-        **options: qutip.solver.Options,
+        **options: dict[str, Any],
```
HGSilveri, line 822: Actually, `**kwargs` are type-hinted by specified the type hint of the values (which requires that all values are of the same type). [PEP 692](https://peps.python.org/pep-0692/) actually aims to improve this, but for now I'm afraid we'll have to fall back on doing it like this:
```suggestion
        **options: Any
```

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -819,7 +819,7 @@ def get_hamiltonian(self, time: float) -> qutip.Qobj:
...
     ) -> SimulationResults:
         """Simulates the sequence using QuTiP's solvers.
```
HGSilveri, line 824: Could you please also update the docstring to say that `options` expects the arguments to `qutip.Options()`?

dehond: Thanks for having a look. I've changed it, and I've also updated the docstring.

HGSilveri (approved):
Perfect, thank you for the contribution!

sebgrijalva (commented):
Thanks for the help here Julius!

Do you think there should be at least some explanation in the docstring about why would someone unfamiliar with QuTiP want to change settings and which are they?

One simple option would be to attach a link to the docstring (https://qutip.org/docs/latest/guide/dynamics/dynamics-options.html).

dehond:
Hi Sebastian, I like your suggestion, and have added a link to the docstring (hopefully so that it's rendered nicely in the docs). It's a bit of a challenge to do this while keeping the line short enough, so let me know if you'd rather I do it differently. (I couldn't find another example of where this was done in the docs.)

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -829,10 +829,13 @@ def run(
...
-                SimConfig solver_options. If no `max_step` value is provided,
-                an automatic one is calculated from the `Sequence`'s schedule
-                (half of the shortest duration among pulses and delays).
+            options: Used as arguments for qutip.Options(). If specified, will
+                override SimConfig solver_options. If no `max_step` value is
+                provided, an automatic one is calculated from the `Sequence`'s
+                schedule (half of the shortest duration among pulses and
+                delays).
+                Refer to the QuTiP docs_ for an overview of the parameters.
+.. _docs: https://qutip.org/docs/latest/guide/dynamics/dynamics-options.html
```
HGSilveri, line 838: Maybe using an URL shortener is simpler?

dehond: You're right - I wasn't sure whether you'd prefer a shortened URL over this.

HGSilveri: If the original URL had fit, it would indeed have been preferable

HGSilveri: Unfortunately, it seems it hasn't rendered correctly in the docs: https://pulser--423.org.readthedocs.build/en/423/apidoc/simulation.html#pulser_simulation.simulation.Simulation.run

dehond: Huh, that's odd. Apparently you need a newline between the end of the text and the start of the links. I've tested it [here](https://livesphinx.herokuapp.com/), so I hope it's fine now.

HGSilveri: Looks good now :+1:

dehond: Thanks for your patience!

PR #425, opened 2022-12-01
Make `Device.interaction_coeff_xy` optional by default

HGSilveri:
To accompany upcoming changes in the abstract representation schema, the `Device.interaction_coeff_xy` parameter type is now `Optional[float]` and defaults to `None`. If there is a `Microwave` channel in the device, however, it is enforced to be a `float`.


bejito (approved):
LGTM

PR #429, opened 2022-12-08
Fix automatic drawing range for detuning

HGSilveri:
When drawing a sequence where a channel has a constant non-zero detuning, the range of the y-axis was wrongly defined. 

e.g. With ∂=±10, the drawings were:

image

With these changes, they become:

image


CdeTerra (approved):
Looks good!

PR #434, opened 2022-12-12
Update method for Quantum Master Equation

### LINKED ISSUE #427: Update method for Quantum Master Equation
When calling the `_run_solver()` method in `Simulation` (line 880 in `Simulation.py`), a fix was introduced for using the `mesolve()` method from QuTiP, in which a `Liouvillian` object was created before calling the solver.

In the meantime, the creation of the Hamiltonian and initial state was updated in QuTiP, so the "fix" should be updated with the recommended creation.

Varda-star:
Fixes #427


Varda-star:
This PR is meant to solve the issue #427

sebgrijalva (commented):
Yes, this is what I meant! Since the `tlist` argument is interpolated from the `self._eval_times_array` there was some inconsistency appearing in an older version of QuTiP.

sebgrijalva (approved):
LGTM

PR #436, opened 2022-12-14
Maximum Layout Filling Fraction

HGSilveri:
This PR changes slightly how a register can be defined from a layout by the introducing the **maximum layout filling fraction** of a device.

Before:
- The maximum filling fraction was implicitly assumed to be 50% -> the layout knows the maximum amount of atoms it can hold
    - When defining a register on a layout, it checked right away if the number of atoms did not go over 0.5 * `number_of_traps`
    - When validated against a Device, a layout could have a maximum of (2 * `max_atom_num`) traps
    
Now:
- The maximum filling fraction is determined by the device -> the layout no longer knows the maximum amount of atoms it can hold
    - When defining a register on a layout, it only checks  if the number of atoms does not go over the `number_of_traps`
    - When validated against a Device, there is no longer a limit on the number of traps it can hold (though there is an implicit upper bound from enforcing `max_radial_distance` and `min_atom_distance`)
- When validating a register created from a layout against a device, we check that the (number of atoms <= number of traps in the layout* max_layout_filling)

### DIFF
--- a/pulser-core/pulser/register/register_layout.py
+++ b/pulser-core/pulser/register/register_layout.py
@@ -171,7 +171,7 @@ class RegisterLayout(RegDrawer):
             raise ValueError("Every 'trap_id' must be a unique integer.")
 
         if not trap_ids_set.issubset(self.traps_dict):
-            # This check makes it redudant to check  # qubits <= # traps
+            # This check makes it redundant to check # qubits <= # traps
             raise ValueError(
                 "All 'trap_ids' must correspond to the ID of a trap."
             )
--- a/pulser-core/pulser/register/register_layout.py
+++ b/pulser-core/pulser/register/register_layout.py
@@ -171,6 +171,7 @@ class RegisterLayout(RegDrawer):
             raise ValueError("Every 'trap_id' must be a unique integer.")
 
         if not trap_ids_set.issubset(self.traps_dict):
+            # This check makes it redudant to check  # qubits <= # traps
             raise ValueError(
                 "All 'trap_ids' must correspond to the ID of a trap."
             )
@@ -186,11 +187,6 @@ class RegisterLayout(RegDrawer):
                     f"provided 'trap_ids' ({len(trap_ids)})."
                 )
 
-        if len(trap_ids) > self.number_of_traps:
-            raise ValueError(
-                "The number of required traps is greater than the number "
-                f"of traps in this layout ({self.number_of_traps})."
-            )
         ids = (
             qubit_ids if qubit_ids else [f"q{i}" for i in range(len(trap_ids))]
         )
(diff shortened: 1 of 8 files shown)


pulser-core/pulser/register/register_layout.py
```diff
@@ -162,6 +171,7 @@ def define_register(
             raise ValueError("Every 'trap_id' must be a unique integer.")
 
         if not trap_ids_set.issubset(self.traps_dict):
+            # This check makes it redudant to check  # qubits <= # traps
```
CdeTerra, line 174: "redundant"

CdeTerra:
Looks good to me!

The only thing I'm not sure to understand is the following statement:
‶When validated against a Device, there is no longer a limit on the number of traps it can hold″
- Do you mean when the register is validated against a device? In this case, there are all the needed information to perform the full check, and that's what the review actually does, where `validate_register` is calling `validate_layout_filling`.
- Or do you mean when the layout is validated against a device? I don't see where this happens, but, if it did, it would be possible to limit the size of the layout with respect to `max_layout_filling`.

CdeTerra (approved):

HGSilveri:
> The only thing I'm not sure to understand is the following statement: ‶When validated against a Device, there is no longer a limit on the number of traps it can hold″
> 
> * Do you mean when the register is validated against a device? In this case, there are all the needed information to perform the full check, and that's what the review actually does, where `validate_register` is calling `validate_layout_filling`.
> * Or do you mean when the layout is validated against a device? I don't see where this happens, but, if it did, it would be possible to limit the size of the layout with respect to `max_layout_filling`.

I meant the layout. It used to be enforced that the layout had at most `2 * max_atom_num` traps, but this was unnecessarily restrictive. We could do the same with the layout filling, but this is actually not a limitation in real life so there's no reason to enforce it.

CdeTerra:
> I meant the layout. It used to be enforced that the layout had at most `2 * max_atom_num` traps, but this was unnecessarily restrictive. We could do the same with the layout filling, but this is actually not a limitation in real life so there's no reason to enforce it.
Okay, I think I understand: the layout can be arbitrarily large, but if it more traps than `max_layout_filling * max_atom_num`, then, a lower value than `max_layout_filling` will apply when filling in the atoms. Do you confirm?

HGSilveri:
> Okay, I think I understand: the layout can be arbitrarily large, but if it more traps than `max_layout_filling * max_atom_num`, then, a lower value than `max_layout_filling` will apply when filling in the atoms. Do you confirm?

Yes, if there are more traps than `max_atom_num / max_layout_filling`. Put in another way: a register made from a layout can have at most  `min(max_atom_num, max_layout_filling * number_of_traps)` atoms.

PR #437, opened 2022-12-15
Adding the device abstract representation + serialization updates

### LINKED ISSUE #430: Update the Abstract Representation JSON Schema, the serializer and the deserializer
Following all the features added since the last release, it is necessary to update the JSON schema, the serializer and the deserializer to include them before the next minor release.

- [x] Device serialization and deserialization
- [x] EOM operations
- [x] `Device.max_filling_fraction` parameter

HGSilveri:
Closes #430


pulser-core/pulser/json/abstract_repr/deserializer.py
```diff
@@ -218,6 +233,83 @@ def _deserialize_operation(seq: Sequence, op: dict, vars: dict) -> None:
...
+    for param in dataclasses.fields(channel_cls):
+        if param.init and param.name != "eom_config":
+            params[param.name] = obj[param.name]
+    return channel_cls(**params)
+
+
+def _deserialize_layout(layout_obj: dict[str, Any]) -> RegisterLayout:
+    return RegisterLayout(
+        layout_obj["coordinates"],
+        slug=layout_obj["slug"] if "slug" in layout_obj else None,
```
bejito, line 290: nit: you could use 
```python
layout_obj.get("glug")
```

pulser-core/pulser/json/abstract_repr/deserializer.py
```diff
@@ -218,6 +233,83 @@ def _deserialize_operation(seq: Sequence, op: dict, vars: dict) -> None:
...
+    )
+
+
+def _deserialize_device_object(obj: dict[str, Any]) -> Device | VirtualDevice:
+    device_cls: Type[Device] | Type[VirtualDevice] = (
```
bejito, line 295: nit: I don't understand why you need to specify `Type[Device] | Type[VirtualDevice]` here since you just assign one of the two values just afterwards ?

pulser-core/pulser/json/abstract_repr/deserializer.py
```diff
@@ -218,6 +233,83 @@ def _deserialize_operation(seq: Sequence, op: dict, vars: dict) -> None:
...
+        VirtualDevice if obj["is_virtual"] else Device
+    )
+    _channels = tuple(
+        (ch["id"], _deserialize_channel(ch)) for ch in obj["channels"]
+    )
+    params: dict[str, Any] = {"_channels": _channels}
+    for param in dataclasses.fields(device_cls):
+        if not param.init or param.name == "_channels":
```
bejito, line 303: What is param.init ?

pulser-core/pulser/json/abstract_repr/schemas/sequence-schema.json
```diff
@@ -445,6 +442,100 @@
...
+        "op",
+        "channel",
+        "duration",
+        "phase",
+        "post_phase_shift",
+        "protocol"
+      ],
+      "type": "object"
+    },
+    "OpEnableEOM": {
```
bejito, line 506: So in order to use the EOM, one has to
1. enable the EOM
2. Create EOMPulse operations
3. Disable the EOM.
Is there a reason we don't automatically enable/disable it on behalf of the user instead of an explicit switch ?

bejito (approved):
Minor questions, otherwise LGTM

pulser-core/pulser/json/abstract_repr/deserializer.py
HGSilveri: Will do, thanks for reminding me of this!

pulser-core/pulser/json/abstract_repr/deserializer.py
HGSilveri: It's because `mypy` doesn't execute the code, so it doesn't know which type `device_cls`. When static type checking, it has to check that either option works, hence the union. Does that make sense?

pulser-core/pulser/json/abstract_repr/deserializer.py
HGSilveri: It's an attribute of a dataclass field. Basically, it's a bool that specifies whether the field is in the `__init__` signature or not.

pulser-core/pulser/json/abstract_repr/schemas/sequence-schema.json
HGSilveri: Enabling and disabling will automatically add a buffer time necessary to switch in and out of the EOM mode, so we leave it to the user to decide when they're done adding EOM pulses. Plus, all EOM pulses share the parameters specified upon enabling the EOM mode (ie the amplitude and detuning), so it's useful that they are specified at a single point rather than on each pulse.

pulser-core/pulser/json/abstract_repr/deserializer.py
bejito: Yes totally, thanks

pulser-core/pulser/json/abstract_repr/deserializer.py
bejito: Aren't all fields of a dataclass in the __init__ ?

pulser-core/pulser/json/abstract_repr/schemas/sequence-schema.json
bejito: Ok makes sense, thanks

pulser-core/pulser/json/abstract_repr/deserializer.py
HGSilveri: By default yes, but you can choose otherwise. Take for example [`BaseDevice.reusable_channels`](https://github.com/pasqal-io/Pulser/blob/654a1d40ac8d1f7734ee50c619686a05e1a7e146/pulser-core/pulser/devices/_device_datacls.py#L80) - it is explicitly not in the init and always False in a subclass that does not specify otherwise (like `Device`). On `VirtualDevice`, it is in the `__init__` because we can actually choose it to be `True` or `False`.

PR #406, opened 2022-10-17
added QUBO tutorial

cdalyac:
Added the QUBO tutorial

PR #438, opened 2022-12-20
Fix UT bug creating wrong numpy array

CdeTerra:
Fixes the following UT in *test_register_layout.py* which is newly failing:

```python
def test_creation(layout, layout3d):
    with pytest.raises(
        ValueError, match="must be an array or list of coordinates"
    ):
RegisterLayout([[0, 0, 0], [1, 1], [1, 0], [0, 1]])
```

This has happened on trying to merge #406 on the job https://github.com/pasqal-io/Pulser/actions/runs/3741008851/jobs/6350057707, only on Python 3.11.1, which may have changed the behavior of creating numpy arrays with inconsistent list sizes.


HGSilveri (approved):
LGTM, thanks for the fix!

PR #442, opened 2022-12-27
Documentation updates

HGSilveri:
- Replaces the UD-MIS with the QUBO tutorial
- Replaces broken animation link with file
- Extends device classes documentation to include members inherited from `BaseDevice`


HGSilveri:
@cdalyac Are there any changes you are not happy with? 

By the way, the most convenient way to check how the tutorial will appear in the docs (input and output) is through the readthedocs build: https://pulser--442.org.readthedocs.build/en/442/tutorials/qubo.html

cdalyac:
Good for me to.

PR #233, opened 2021-07-09
First order dephasing

julien-bremont:
Adds a first-order (in dephasing probability) dephasing model for multiple qubits. 
Raises a warning if the user sets a high (> 0.2) dephasing probability.

PR #443, opened 2022-12-29
Generalize effective noise channels

### LINKED ISSUE #428: Generalize effective noise channels
So far we have added a dephasing channel for noisy `Simulation`. We can now generalize it to a depolarizing channel and more generally to any custom type of "collapse" operators for the Lindblad equation solver. Please look at QuTiP documentation [here](https://qutip.org/docs/latest/guide/dynamics/dynamics-master.html).

Varda-star:
Fixes #428


pulser-simulation/pulser_simulation/simconfig.py
```diff
@@ -54,6 +56,11 @@ class SimConfig:
             noise types:
 
             - "dephasing": Random phase (Z) flip
+            - "depolarizing": model of decohering where the qubit undergoes a
+              bit-flip error, phase-flip error or both errors.
+            - "gen_noise": general model of noise channels where the qubit
+            undergoes a modification according to the specified noise
+            operators, it encapsulates dephasing and depolarizing channels.
```
HGSilveri, line 63: From this description, it's not clear to me: what's the difference between having `gen_noise` vs `dephasing`+ `depolarizing`?

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -202,7 +209,7 @@ def set_config(self, cfg: SimConfig) -> None:
                 )
             # Probability of phase (Z) flip:
             # First order in prob
-            prob = self.config.dephasing_prob / 2
+            prob = self.config.dephasing_prob
```
HGSilveri, line 212: Why this change? Isn't this still a first-order approximation? Nothing else seems to have changed.

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -225,6 +232,97 @@ def set_config(self, cfg: SimConfig) -> None:
...
+
+            prob = self.config.depolarizing_prob
+            n = self._size
+            if prob > 0.1 and n > 1:
+                warnings.warn(
+                    "The depolarizing model is a first-order approximation"
+                    f" in the depolarizing probability. p = {2*prob}"
+                    " is too large for realistic results.",
+                    stacklevel=2,
+                )
```
HGSilveri, line 254: Shouldn't the probability be divided by 2, then?

HGSilveri (commented):
Hum, some things are not so clear to me

pulser-simulation/pulser_simulation/simconfig.py
Varda-star: The difference is that you can design any operator noises.
For example: dephasing and depolarizing channels are instances of the general noise channel

pulser-simulation/pulser_simulation/simulation.py
Varda-star: There is no reason to divide by two, the default probability is already less than 0.1, so I removed it.

HGSilveri: Are you sure that's the reason why it was being divided by two? It seems to me that it was rather because we are using a first order approximation (which assumes p is small) and the first term in the expansion has `p/2`. This is what the code below seems to imply.

HGSilveri: This was added in PR #233. 
`p/2` comes from the dephasing formula, actually - see eq. (2.3) in [this paper](https://arxiv.org/pdf/1509.02921.pdf).

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -225,6 +232,97 @@ def set_config(self, cfg: SimConfig) -> None:
...
+            ]
+            self._collapse_ops += [
+                1j
+                * k
+                * (
+                    self.build_operator([("sigma_gr", [qid])]).dag()
+                    - self.build_operator([("sigma_gr", [qid])])
+                )
+                for qid in self._qid_index
+            ]
```
HGSilveri, line 285: Where did you get these Kraus operators from? They don't seem right to me

HGSilveri: Or rather, they make sense if you make `p -> p*3/4`, I think you are missing that

Varda-star: image

Varda-star: From chapter 8 on a MIT lecture on radiation.
Link: https://ocw.mit.edu/courses/22-51-quantum-theory-of-radiation-interactions-fall-2012/resources/mit22_51f12_ch8/

Varda-star: I tried then to add, M1, M2, and M3 according the uncorrelated noise approximation for many-body systems, i.e. assuming one bit is modified at the time.

HGSilveri: Ok, it seems that we have contradicting sources. Yours matches what is in Nielsen and Chuang, so I think we should stick to it

HGSilveri: But wait, wikipedia says otherwise... https://en.wikipedia.org/wiki/Quantum_depolarizing_channel

HGSilveri: Ok, I think the issue is the definition. My sources define "depolarization" as _"an operation that replaces the input state with a completely mixedstate with probability p"_, while yours state that _"with probability p an “error” occurs.
The error can be of any one of three types, where each type of error is equally likely."_ 

These are different things, so it's natural that they have different formulas.

Varda-star: I browse the Nielsen and Chuang book on Quantum COmputation and Quantum Information and they have a slight different set of operators.

HGSilveri: And look what they say right after that: 
image

This is essentially our problem here. I think we just have to pick one and clearly state what we mean by "depolarizing" in the docstring.

Varda-star: Defining the "depolarization" as "an operation that replaces the input state with a completely mixed state with probability p" is not wrong. We can start from that definition and end up with the Krauss operators I sent first.

HGSilveri: Let's see what @sebgrijalva thinks

Varda-star: @sebgrijalva told me he prefers the equation (8.102) which is in accordance with the dephasing model already chosen.

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -225,6 +232,97 @@ def set_config(self, cfg: SimConfig) -> None:
...
+            for i in range(1, m):
+                k = np.sqrt(
+                    self.config.gen_noise_probs[i] * prob_id ** (n - 1)
+                )
+                c_op = k * self.config.gen_noise_opers[i]
+                self._collapse_ops += [
+                    self.build_operator([(c_op, [qid])])
+                    for qid in self._qid_index
+                ]
```
HGSilveri, line 324: I think you could simplify the code by making the `dephasing` and `depolarizing` cases define the proper Kraus operators and then they would always be built here. 
The current implementation introduces a lot of unnecessary duplication.

HGSilveri (changes requested):

pulser-simulation/pulser_simulation/simconfig.py
```diff
@@ -86,6 +92,13 @@ class SimConfig:
     epsilon: float = 0.01
     epsilon_prime: float = 0.05
     dephasing_prob: float = 0.05
+    depolarizing_prob: float = 0.05
+    gen_noise_probs: list[float] = field(
+        init=True, default_factory=list, repr=False
+    )
+    gen_noise_opers: list[qutip.Qobj] = field(
+        init=True, default_factory=list, repr=False
```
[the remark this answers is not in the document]
Varda-star: I got an error when I set it `init = False`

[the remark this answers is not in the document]
Varda-star: I removed the init arg and now it works well.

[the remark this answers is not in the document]
HGSilveri: What I mean is that you don't need to specifiy it because the default is `True`

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -225,6 +232,97 @@ def set_config(self, cfg: SimConfig) -> None:
...
+                    + self.build_operator([("sigma_gr", [qid])])
+                )
+                for qid in self._qid_index
+            ]
+            self._collapse_ops += [
+                1j
+                * k
+                * (
+                    self.build_operator([("sigma_gr", [qid])]).dag()
+                    - self.build_operator([("sigma_gr", [qid])])
```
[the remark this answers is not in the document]
Varda-star: You're right

pulser-simulation/pulser_simulation/simconfig.py
```diff
@@ -176,3 +205,62 @@ def _calc_sigma_doppler(self) -> None:
...
+                if not isinstance(prob, float):
+                    raise TypeError(
+                        "gen_noise_probs is a list of floats"
+                        f" it must not contain a {type(prob)}."
+                    )
+
+            prob_distr = np.array(self.gen_noise_probs)
+            lower_bound = np.any(prob_distr < 0.0)
+            upper_bound = np.any(prob_distr > 1.0)
+            sum_p = sum(prob_distr) != 1.0
```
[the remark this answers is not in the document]
Varda-star: Comparing to 0.0 will create numerical errors?

pulser-simulation/pulser_simulation/simconfig.py
```diff
@@ -54,6 +56,10 @@ class SimConfig:
             noise types:
 
             - "dephasing": Random phase (Z) flip
+            - "depolarizing": model of decohering where the qubit undergoes a
+              bit-flip error, phase-flip error or both errors.
```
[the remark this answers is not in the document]
Varda-star: Yes as it is in the book
image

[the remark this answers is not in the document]
Varda-star: Then from this we derive the equation (8.102)
image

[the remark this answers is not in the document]
Varda-star: You're right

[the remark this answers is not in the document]
Varda-star: I will try something else.

pulser-simulation/pulser_simulation/simconfig.py
```diff
@@ -54,6 +56,10 @@ class SimConfig:
...
+            - "gen_noise": general noise channel through which we can custom
```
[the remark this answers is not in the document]
Varda-star: You mean the way I wrote the code only allow on of the three to exist?

pulser-simulation/pulser_simulation/simconfig.py
```diff
@@ -86,6 +92,13 @@ class SimConfig:
     epsilon: float = 0.01
     epsilon_prime: float = 0.05
     dephasing_prob: float = 0.05
+    depolarizing_prob: float = 0.05
+    gen_noise_probs: list[float] = field(
+        init=True, default_factory=list, repr=False
+    )
```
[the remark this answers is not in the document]
Varda-star: It's because of the approximation we use for many bodies. 
All effective channels are not probability-wise as the amplitude damping and the phase damping channel.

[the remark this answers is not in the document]
Varda-star: Our noise model is based on the fact that only one bit string is modified at the time, and it is only justified when "rates" of collapse operators are low.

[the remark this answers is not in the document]
Varda-star: This is why I wanted to seperate collapse operators from their rates to make sure the approx is verified. Normally I must not allow any set of operators where all rates aren't all low. It will not give realistic result. Or as done for the dephasing and depolarizing channel, warn the user that the effective channel he proposed will not yield realistic results since all rates are not low enough.

pulser-simulation/pulser_simulation/simconfig.py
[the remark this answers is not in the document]
HGSilveri: Yes, you are overriding the `_colase_ops` on each case

pulser-simulation/pulser_simulation/simconfig.py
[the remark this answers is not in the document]
Varda-star: And as it's done for the dephasing and depolarizing channels, I warn the user whenever the rate of an operator is not low enough that it would yield unrealistic results according to the approx we used.

pulser-simulation/pulser_simulation/simulation.py
Varda-star: I must remove them from the set of noise models?

HGSilveri: No no, they are convenient to have. But you could store their noise operators under `gen_noise_opers` so that the collapse operators are built always in the same place

pulser-simulation/pulser_simulation/simconfig.py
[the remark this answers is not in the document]
Varda-star: How can I?

pulser-simulation/pulser_simulation/simconfig.py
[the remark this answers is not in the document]
Varda-star: So if I put a `+=` instead of `=` it's solved?

pulser-simulation/pulser_simulation/simulation.py
Varda-star: Because the only place where collapse_operators are defined is in `dephasing`, `depolarizing` or `gen_noise` mode

Varda-star: I think `effective_nosie` is more convenient than `gen_noise`. What do you think?

pulser-simulation/pulser_simulation/simconfig.py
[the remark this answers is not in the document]
HGSilveri: No, you can just add more and more Kraus operators. It's tricky, to be honest. I think to start the best thing is to allow only one of the three. If the user wants to combine multiple noises, they still can do so with the general noise operators option

pulser-simulation/pulser_simulation/simulation.py
HGSilveri: I agree, I prefer `eff_noise` over `gen_noise` too

pulser-simulation/pulser_simulation/simconfig.py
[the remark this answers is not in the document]
HGSilveri: Well, the probability associated with each operator (if the operators already include the probablity, as I'm suggesting) will simply be `op.dag() * op` right?

pulser-simulation/pulser_simulation/simulation.py
Varda-star: No no, they are convenient to have. But you could store their noise operators under gen_noise_opers so that the collapse operators are built always in the same place// If I do so it will create confusion since I iterate it in a loop to build Kraus operators.

pulser-simulation/pulser_simulation/simconfig.py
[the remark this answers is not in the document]
Varda-star: Ah ok, so `dephasing + depolarizing` will be prohibited?

pulser-simulation/pulser_simulation/simconfig.py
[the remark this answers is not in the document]
HGSilveri: If the probability is incorporated in the operator, it wouldn't be just `sigma_x`, it would be `op = sqrt(p)*sigma_x`

pulser-simulation/pulser_simulation/simconfig.py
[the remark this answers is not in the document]
HGSilveri: Yes, for now we can allow only one of `dephasing`, `depolarizing` or `gen_noise`, but never more than one at a time

pulser-simulation/pulser_simulation/simulation.py
HGSilveri: I think you are missing the point. In all three cases, all you are doing is taking the probabilities, building the single-qubit Kraus operators and then building the collapse operators from them. The difference in each case is contained in the single-qubit Kraus operators, so the collapse operators could be built from them in one place rather than repeated every time

pulser-simulation/pulser_simulation/simconfig.py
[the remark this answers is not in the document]
Varda-star: So it is hard to extract p. In the case of pauli matrices it's simple but for a general Kraus Operator it isn't possible. The only relation we know about them is that `sum(op_k) = id`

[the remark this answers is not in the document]
HGSilveri: Yes, I think you are right. I was assuming the operator was a unitary, but this is not a requirement. Nevermind my suggestion then.

pulser-simulation/pulser_simulation/simconfig.py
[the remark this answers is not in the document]
Varda-star: What do you think about adding this restriction in the method `_check_noise_types` in simconfig?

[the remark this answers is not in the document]
HGSilveri: I think it's a good idea, it's the right place for that check

pulser-simulation/pulser_simulation/simulation.py
HGSilveri: And why not?

Varda-star: In either `dephasing` , `depolarizing` or `eff_noise`, i will build Kraus operators then build all collapse operators at once.

HGSilveri: Exactly, don't you think it's cleaner this way?

Varda-star: Since Kraus operators are built only in one of these modes, I need that storage variable to be defined beforehand, why not defining it as an attribute of the class?

Varda-star: `self._Kraus_ops`

HGSilveri: Can't you do it before the first `if`?

HGSilveri: I think an attribute is not necessary if you build the collapse operators in the same call

Varda-star: Ok, I will do it before the first `if`.

pulser-simulation/pulser_simulation/simconfig.py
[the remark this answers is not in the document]
Varda-star: Let's cook some code then

tests/test_simulation.py
```diff
@@ -590,19 +613,17 @@ def test_dephasing():
...
+    pulse = Pulse.ConstantPulse(duration, np.pi, 0, 0)
     seq.add(pulse, "ch0")
     sim = Simulation(
         seq, sampling_rate=0.01, config=SimConfig(noise="dephasing")
     )
     assert sim.run().sample_final_state() == Counter({"0": 595, "1": 405})
     assert len(sim._collapse_ops) != 0
     with pytest.warns(UserWarning, match="first-order"):
-        reg = Register.from_coordinates([(0, 0), (0, 10)], prefix="q")
+        reg = Register.from_coordinates([(0, 0), (0, 7)], prefix="q")
```
HGSilveri, line 624: Why this change?

tests/test_simulation.py
```diff
@@ -611,7 +632,67 @@ def test_dephasing():
...
+    reg = Register.from_coordinates([(0, 0)], prefix="q")
+    seq = Sequence(reg, Chadoq2)
+    seq.declare_channel("ch0", "rydberg_global")
+    duration = 2500
+    pulse = Pulse.ConstantPulse(duration, np.pi, 0, 0)
+    seq.add(pulse, "ch0")
+    sim = Simulation(
+        seq, sampling_rate=0.01, config=SimConfig(noise="depolarizing")
+    )
+    assert sim.run().sample_final_state() == Counter({"0": 587, "1": 413})
```
HGSilveri, line 646: Check also that the state is not pure anymore, otherwise you can't distinguish between a mixed state and an equal superpostion

tests/test_simulation.py
```diff
@@ -611,7 +632,67 @@ def test_dephasing():
...
+    seq.add(pulse, "ch0")
+    sim = Simulation(
+        seq,
+        sampling_rate=0.01,
+        config=SimConfig(
+            noise="eff_noise",
+            eff_noise_opers=[matrices["I"], matrices["Z"]],
+            eff_noise_probs=[0.95, 0.05],
+        ),
+    )
```
HGSilveri, line 676: If I'm not mistaken, this should be the same as dephasing with `p=0.10`. Why don't you run a simulation with that config and check that the final states match?

HGSilveri (commented):
Much better!

tests/test_simulation.py
Varda-star: No real reason in reality

tests/test_simulation.py
Varda-star: I took the trace of the square of the density matrix at the end of the evolution and checked if it's lower than one without being close to one.

HGSilveri: Correct me if I'm wrong, but it should actually be close to `1-3p/4`, no? Perhaps you can check for that

Varda-star: Is it enough? Because I made the same test for the dephasing channel and the test below worked. Does it mean the state is a mixed state?

Varda-star: The absolute difference is 0.03, is it close enough?

HGSilveri:
The docs are failing because of an update to the `sphinx_autodoc_typehints` extension. Would be best to pin it to the previous release (ie `sphinx_autodoc_typehints == 1.21.3`) in `docs/requirements.txt`

tests/test_simulation.py
HGSilveri: Are you struggling with this?

Varda-star: Yes. I don't understand why `1- 3*P`. With `rho = p/2 I + (1-p) 1` I found `tr(rho**2) = 2 (1-3/4p) ** 2 + 3/8 p**2`

Varda-star: By expanding `rho**2` and using `tr(sigma_i) = 0` and `tr(sigma_i * sigma_j) = 0` for `i!=j`

tests/test_simconfig.py
```diff
@@ -70,26 +70,23 @@ def test_init(matrices):
...
-        NotImplementedError,
-        match="Depolarizing, dephasing and eff_noise channels",
-    ):
-        SimConfig(noise=("dephasing", "eff_noise"))
-    with pytest.raises(
-        NotImplementedError,
-        match="Depolarizing, dephasing and eff_noise channels",
-    ):
-        SimConfig(noise=("dephasing", "depolarizing", "eff_noise"))
+        SimConfig(noise=noise_sample)
```
HGSilveri, line 89: These should be isolated in a single test, otherwise you are repeating the rest of the function needlessly

tests/test_simconfig.py
```diff
@@ -70,26 +70,23 @@ def test_init(matrices):
...
+@pytest.mark.parametrize(
+    "noise_sample,",
+    [
+        ("dephasing", "depolarizing"),
+        ("eff_noise", "depolarizing"),
+        ("eff_noise", "dephasing"),
+        ("depolarizing", "eff_noise", "dephasing"),
+    ],
+)
+def test_init2(matrices, noise_sample):
```
HGSilveri, line 84: Give a more descriptive name, like `test_multiple_kraus_noise`

tests/test_simulation.py
HGSilveri: Then you should revert it

tests/test_simconfig.py
```diff
@@ -40,3 +70,79 @@ def test_init():
...
+            noise=("eff_noise"),
+            eff_noise_opers=[matrices["I"], matrices["X"]],
+            eff_noise_probs=[0.5, 2.0],
+        )
+    with pytest.raises(ValueError, match="is not a probability distribution."):
+        SimConfig(
+            noise=("eff_noise"),
+            eff_noise_opers=[matrices["I"], matrices["X"]],
+            eff_noise_probs=[0.3, 0.2],
+        )
```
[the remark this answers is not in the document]
HGSilveri: You didn't separate these in the end

pulser-simulation/pulser_simulation/simconfig.py
```diff
@@ -206,29 +213,36 @@ def _calc_sigma_doppler(self) -> None:
...
-            if self.gen_noise_opers == [] or self.gen_noise_probs == []:
+        if "eff_noise" in self.noise:
+            if len(self.eff_noise_opers) != len(self.eff_noise_probs):
+                raise ValueError(
+                    f"The operators list length({len(self.eff_noise_opers)}) "
+                    "and probabilities list length"
+                    f"({len(self.eff_noise_probs)}) must be equal."
+                )
+            if self.eff_noise_opers == [] or self.eff_noise_probs == []:
                 raise ValueError("Fill the general noise parameters.")
```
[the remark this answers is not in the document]
HGSilveri: You didn't change this

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -253,75 +248,56 @@ def set_config(self, cfg: SimConfig) -> None:
...
                 k = np.sqrt(
-                    self.config.gen_noise_probs[i] * prob_id ** (n - 1)
+                    self.config.eff_noise_probs[i] * prob_id ** (n - 1)
                 )
-                c_op = k * self.config.gen_noise_opers[i]
-                self._collapse_ops += [
-                    self.build_operator([(c_op, [qid])])
-                    for qid in self._qid_index
-                ]
+                K_op = k * self.config.eff_noise_opers[i]
```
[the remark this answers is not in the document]
HGSilveri: Didn't change this either

tests/test_simconfig.py
```diff
@@ -40,3 +70,84 @@ def test_init():
...
+        match="Depolarizing, dephasing and eff_noise channels",
+    ):
+        SimConfig(noise=noise_sample)
+
+
+@pytest.mark.parametrize(
+    "oper_list",
+    [
+        [qeye(2), sigmax()],
+    ],
```
HGSilveri, line 96: If `oper_list` is always the same, there's no reason for it to be a variable

tests/test_simconfig.py
```diff
@@ -40,3 +70,84 @@ def test_init():
...
+    with pytest.raises(ValueError, match="is not a probability distribution."):
+        SimConfig(
+            noise=("eff_noise"),
+            eff_noise_opers=oper_list,
+            eff_noise_probs=prob_distr,
+        )
+    with pytest.raises(TypeError, match="eff_noise_probs is a list of floats"):
+        SimConfig(
+            noise=("eff_noise"), eff_noise_probs=[""], eff_noise_opers=[""]
+        )
```
HGSilveri, line 116: This should be in `test_eff_noise_opers`, I guess, just so it's not repeated

HGSilveri (approved):
LGTM!

PR #446, opened 2023-01-06
 Add saving possibility to register layout draw method 

a-corni:
Add variable fig_name and kwargs_savefig to RegisterLayout.draw
Tested to solve @LucasGitQ error
Further tests needed


HGSilveri (approved):
LGTM, thanks!

PR #455, opened 2023-01-13
Enable RTD fail on warning

HGSilveri:
As we have RTD automatically build the docs, sometimes there are warnings we miss that could have let us know something is not quite right.

Since our build runs without warnings under normal circumstances, we can let it fail in the presence of warnings so that we can catch them.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -12,13 +12,14 @@ build:
 
 # Build documentation in the docs/ directory with Sphinx
 sphinx:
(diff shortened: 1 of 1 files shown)


a-corni (approved):
From what I have read, it is the best solution to have the build going on and raising all the warnings !

https://stackoverflow.com/questions/61414199/see-all-warnings-as-errors-when-building-sphinx-docs-on-readthedocs-org

PR #459, opened 2023-01-20
Replace Iterable type checking in variable validation

HGSilveri:
When validating the arguments to calls where the arguments are potentially parametrized, there was a check for recursiveness for cases where e.g a parametrized element is inside an iterable. This check relied on a type check of the kind  `isinstance(x, Iterable)`, which essentially relied on the existence of an `__iter__` method in the object. This posed two issues:
1. Some objects are iterable because they define a `__getitem__` but not an `__iter__`
2. Crucially, some objects (in particular 0-D numpy arrays) have an `__iter__` method but then raise an error when it is called.

Because of these issues, I propose we replace the type check with a duck typing approach, where we simply try to iterate over the object and handle the case where it fails with a `try...except` block. This fixes both of the issues outlined above.

### DIFF
--- a/tests/test_paramseq.py
+++ b/tests/test_paramseq.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import copy
+from collections.abc import Iterable
 
 import numpy as np
 import pytest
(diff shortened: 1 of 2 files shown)


tests/test_paramseq.py
```diff
@@ -325,3 +326,15 @@ def test_parametrized_before_eom_mode(mod_device):
...
+
+def test_iterable_variable_check():
+    seq = Sequence(reg, device)
+    seq.declare_channel("ch0", "rydberg_global")
+    #
+    delay_t = np.array(100)
+    # A 0-D array is identified as an Iterable but then fails iteration
+    # Giving this as an argument to a sequence call should still work
+    assert isinstance(delay_t, Iterable)
+    seq.delay(delay_t, "ch0")
```
a-corni, line 339: It works because `delay` has the decorator `store` that calls the function `verify_parametrization` that call `verify_variable`, which we want to test.

a-corni (approved):
Looks good to me ! I agree with the initial problem, the solution and its associated test

PR #464, opened 2023-02-14
Refactoring and feature isolation to ease subclassing

HGSilveri:
These changes add no extra features and should have no impact on existing behaviour. However, they help with the subclassing and feature extension in external projects.


pulser-core/pulser/sequence/_schedule.py
```diff
@@ -156,15 +158,18 @@ def get_samples(self) -> ChannelSamples:
             slots.append(_TargetSlot(s.ti, tf, s.targets))
 
             # The phase of detuned delays is not considered
```
a-corni, line 160: I guess this comment should be changed `# The phase of detuned delays can be not considered`

pulser-core/pulser/sequence/sequence.py
```diff
@@ -1338,14 +1337,17 @@ def draw(
             draw_modulation="output" in mode,
             draw_phase_curve=draw_phase_curve,
         )
-        if fig_name is not None and draw_register:
+        if fig_name is not None and fig_reg is not None:
             name, ext = os.path.splitext(fig_name)
             fig.savefig(name + "_pulses" + ext, **kwargs_savefig)
             fig_reg.savefig(name + "_register" + ext, **kwargs_savefig)
         elif fig_name:
             fig.savefig(fig_name, **kwargs_savefig)
```
a-corni, line 1345: Is there a reason why we don't always have the same name for the figure files ? `fig.savefig(name + "_pulses" + ext, **kwargs_savefig)` ?

a-corni (commented):
Looks good to me on the whole, I have just left few comments related to the way we modulate.

pulser-core/pulser/channels/base_channel.py
```diff
@@ -391,22 +400,41 @@ def modulate(
...
-            rise_time = self.rise_time
+            mod_padding = self._modulation_padding
 
-        # The cutoff frequency (fc) and the modulation transfer function
-        # are defined in https://tinyurl.com/bdeumc8k
-        fc = mod_bandwidth * 1e-3 / np.sqrt(np.log(2))
         if keep_ends:
-            samples = np.pad(input_samples, 2 * rise_time, mode="edge")
+            samples = np.pad(
+                input_samples, mod_padding + self.rise_time, mode="edge"
```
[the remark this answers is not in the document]
HGSilveri: I'm not sure about spectral leakage. The input samples are padded because the output signal is longer by about `rise_time` on each side. 
By default, we pad the signal with zeros, but when `keep_ends=True`, we want to make it seem like the signal started out from the first value and at the end stays at the last value, so we pad with `mode="edge"` instead. Plus, we use a longer padding time that we then truncate such that the modulation doesn't show the signal ramping to the starting and ending values at the edges.

pulser-core/pulser/channels/base_channel.py
```diff
@@ -391,22 +400,41 @@ def modulate(
             return input_samples
         else:
             mod_bandwidth = self.mod_bandwidth
-            rise_time = self.rise_time
+            mod_padding = self._modulation_padding
```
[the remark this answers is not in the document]
HGSilveri: Here no, but you will see it is important on `LaserChannel` because we will modify it.

pulser-core/pulser/sequence/sequence.py
HGSilveri: Well, we figured there was no need to change the name of the figure provided by the user unless we needed to distinguish between the register and the pulse sequence.

a-corni (approved):
Thanks for the explanations !

PR #467, opened 2023-02-23
Bug fixes in serialization to the abstract representation

HGSilveri:
- Fixes serialization of a sequence with the default measurement basis
- Fixes serialization of parametrized sequences without default values for the variables


pulser-core/pulser/json/abstract_repr/serializer.py
```diff
@@ -216,36 +221,28 @@ def get_all_args(
             )
         elif call.name == "measure":
             data = get_all_args(("basis",), call)
-            res["measurement"] = data["basis"]
+            res["measurement"] = data.get("basis", "ground-rydberg")
```
MatthieuMoreau0, line 224: Does this mean that this will need to be updated if we change the default value for the kwargs in the sequence methods ? 

If yes, don't you think that's a risk for maintainability ? Can we retrieve the default value here "dynamically" ?

HGSilveri: I think it's not a risk only because I'm covering it explicitly in the tests, otherwise I would agree. 
I think a dynamic solution should be possible, though it will probably look a bit ugly. Let me give it a shot, I would also prefer it if it's not too hacky

HGSilveri: Take a look at 9e5e5ddc3f8dc148a8ae8a47afc1b67b6f9e8da7 @MatthieuMoreau0 , I think it's not so ugly after all. Thanks for the suggestion!

MatthieuMoreau0: Looking good!

a-corni (commented):
Everything looks good to me.

PR #469, opened 2023-02-27
Draw register add subfigure capacity

CdeTerra:
Changes that allow to combine a register plot with other plots.


pulser-core/requirements.txt
```diff
@@ -1,5 +1,5 @@
 jsonschema
-matplotlib
+matplotlib >= 3.6.2
```
HGSilveri, line 2: This breaks installation for Python 3.7. Can't we avoid all these version requirements by making the plotting happen on custom axes rahter than custom figures?

CdeTerra: Thanks for the comment! It also works, it is simpler, and even more flexible because it allows to draw other plots on the register.

CdeTerra: The only downside I can see is that the `draw` function won't be able to switch to more complex plotting now, combining multiple plots. Let's hope it won't be needed.

HGSilveri (approved):

PR #473, opened 2023-03-01
 Set an upper bound on the supported modulation bandwidth

a-corni:
I have set the upper bound to `MODBW_TO_TR*1e3` because this guarantees that the rise time `int(MODBW_TO_TR*1e3/mod_bandwidth)` is always above or equal to 1.


HGSilveri (approved):
LGTM, nice job!

PR #474, opened 2023-03-01
Update Pulser-Pasqal to the latest cloud-sdk syntax

awennersteen:
The PulserPasqal packge has an old version of the cloud-sdk. This MR bumps the version and implements the required interface for the latest version.

### DIFF
--- a/pulser-pasqal/pulser_pasqal/pasqal_cloud.py
+++ b/pulser-pasqal/pulser_pasqal/pasqal_cloud.py
@@ -56,8 +56,9 @@ class PasqalCloud:
         seq: Sequence,
         jobs: list[JobParameters],
         device_type: sdk.DeviceType = sdk.DeviceType.QPU,
-        configuration: Optional[Union[sdk.device.configuration.BaseConfig, Dict[str, Any]]] = None,
+        configuration: Optional[sdk.device.configuration.BaseConfig] = None,
         wait: bool = False,
+        fetch_results: bool = False,
     ) -> sdk.Batch:
         """Create a new batch and send it to the API.
 
@@ -70,9 +71,9 @@ class PasqalCloud:
             device_type: The type of device to use, either an emulator or a QPU
                 If set to QPU, the device_type will be set to the one
                 stored in the serialized sequence.
-            configuration: A dictionary with extra configuration for the
-                emulators that accept it.
-            wait: Whether to wait for results to be sent back.
+            configuration: Optional xtra configuration for emulators.
+            wait: Whether to wait for the batch to be done.
+            fetch_results: Whether to download the results. Implies waiting for the batch.
 
         Returns:
             Batch: The new batch that has been created in the database.
@@ -96,6 +97,7 @@ class PasqalCloud:
             device_type=device_type,
             configuration=configuration,
             wait=wait,
+            fetch_results=fetch_results,
         )
 
     def _serialize_seq(
@@ -105,7 +107,7 @@ class PasqalCloud:
             return seq.serialize()
         return seq.to_abstract_repr()
 
-    def get_batch(self, id: int, fetch_results: bool = False) -> sdk.Batch:
+    def get_batch(self, id: str, fetch_results: bool = False) -> sdk.Batch:
         """Retrieve a batch's data and all its jobs.
 
         Args:
--- a/pulser-pasqal/pulser_pasqal/pasqal_cloud.py
+++ b/pulser-pasqal/pulser_pasqal/pasqal_cloud.py
@@ -11,10 +11,10 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Allows to connect to the cloud powered by Pasqal to run sequences."""
+"""Allows to connect to PASQAL's cloud platform to run sequences."""
 from __future__ import annotations
 
-from typing import Any, Optional
+from typing import Any, Dict, Optional, Union
 
 import sdk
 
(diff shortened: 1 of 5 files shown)


pulser-pasqal/pulser_pasqal/pasqal_cloud.py
```diff
@@ -104,7 +107,7 @@ def _serialize_seq(
             return seq.serialize()
```
HGSilveri, line 107: Do you happen to know if the QPU has moved to the abstract representation? If so, this could be a good opportunity to change this part

pulser-pasqal/requirements.txt
```diff
@@ -1 +1 @@
-pasqal-sdk == 0.1.6
+pasqal-sdk == 0.1.12
```
HGSilveri, line 1: Do you reckon we will still need a strict version pin or could we potentially go with `pasqal-sdk >= 0.1.12` or `pasqal-sdk ~= 0.1.12` ?

HGSilveri (approved):
LGTM, thanks!

pulser-pasqal/pulser_pasqal/pasqal_cloud.py
awennersteen: I believe not, but will check in the morning.

pulser-pasqal/requirements.txt
awennersteen: good point, I'll discuss it tomorrow and let you know before merging.

Perhaps we could just add a simple test using pulser-pasqal to the cloud-e2e tests, that way the cloud team will be aware and can raise the issue?

pulser-pasqal/pulser_pasqal/pasqal_cloud.py
MatthieuMoreau0: I don't see any blockers for not using the abstract representation, all our stack up to the QPU execution is compatible with it

pulser-pasqal/requirements.txt
MatthieuMoreau0: I think we should stick to a strict version pin, we recently overhauled the authentication system and I predict we'll have to rework a few things which might break compatibility with pulser-pasqal.

In the long run I definitely think we should have no breaking changes introduced and hence this could target the latest pasqal-sdk version. But I'm worried it's a bit too early for that

awennersteen: My only issue is that PulserPasqal is going to be running "in production", it is how some quantum libraries connect to the cloud apparently.  We can perhaps also change that. But I'm not sure if that would be taking a step backwards considering future plans. For new external users the most natural thing might become to use the PulserPasqal module too. 

In a production system we'll of course also pin the cloud-sdk version down to whatever is needed to support the respective environment, but it would be nice for these versions to be valid, and not something we have to forcefully override. 

We could run some simple test in our end to end tests using PulserPasqal, which should allow us to detect changes more easily.
If we then also introduce the dev releases on cloud-sdk we might be able to keep our released packages in sync.

I admit it is a bit annoying and will cause some overhead, but at the same time so does the current way of doing things. At least this way we'll always be on top of things?

I would suggest that we for now add it as a `~=`, dependency, so we can merge this now. When there is a new release shortly on the cloud-sdk, but that won't make any changes to the interface we'll get that one for free, and the next time the cloud-sdk has to be updated externally facing we can bump the minor to `0.2.x`?

@bejito Any thoughts?

anneclairelh:
@awennersteen  reminder to switch to use core-fast!! :) 
(Maybe I missed it ?)

awennersteen:
> @awennersteen reminder to switch to use core-fast!! :) (Maybe I missed it ?)

I think the best for default behavior will be that we merge your MR making the default value switch.
For internally we also set the Endpoints specifically  for the envs

awennersteen:
@anneclairelh 
>     * Yes, QPU has moved to abstract representation as for the cloud services part (and integration with Fresnel is planned to be done with abtract repr)
Then I've just updated PulserPasqal then to always submit with `seq.to_abstract_repr()`.

>     * Version pinning: agreed with Matthieu for the current situation. We need to organize anyway a meeting in the next weeks to discuss, change if needed, and write down once for all the versioning plan
Alright

pulser-pasqal/pulser_pasqal/pasqal_cloud.py
awennersteen: made into always abstract

pulser-pasqal/requirements.txt
MatthieuMoreau0: Adding pulser-pasqal to our e2e tests battery sound like a good idea!

Not sure I understand what you imply in the two first paragraphs !

awennersteen: "late" night rambling. 

We're connecting to "the cloud"  via PulserPasqal for the upcoming release of quantum discovery module 2.

tests/test_pasqal.py
```diff
@@ -141,7 +144,7 @@ def test_pasqal_cloud_qpu(fixt):
         fixt=fixt,
```
HGSilveri, line 144: To fix the failing test, you can replace 
```python
reg = Register(dict(enumerate([(0, 0), (0, 10)])))
```
with 
```python
reg = Register.from_coordinates([(0, 0), (0, 10)], prefix="q")
```

pulser-pasqal/requirements.txt
MatthieuMoreau0: Ah right and you're saying that this is going to be painful becase to update the version of the SDK we'll need to release a new version of pulser, is that correct ?

awennersteen: correct, this dependency will be manifested inside the PCS infra.

HGSilveri: +1 for the `~=` option, if possible.

MatthieuMoreau0: Yes ok let's go for this then

awennersteen: Alright, I've pushed. And I've added an issue in the (internal) end-to-end tests.

PR #480, opened 2023-03-09
Enable to choose custom colors of qubits when drawing Register

### LINKED ISSUE #479: Allow to customize the colors when drawing a register
Currently, when we plot a register, they are all shown in green. Just like it's possible to do some plotting customization, as specifying masked qubits, I would also like to be able to specify individual colors for each qubit.

The signature of `Register.draw` and `RegDrawer._draw_2D` would be added a new parameter:

```python
qubit_colors: dict[QubitId, str]
```

This parameter would basically apply where "darkgreen" currently does, hence the point of the qubit as well as the transparent disk and the circle when the "draw_half_radius" is set to `True`.

CdeTerra:
Closes #479 

Example of use:

```python
r = Register.from_coordinates([[0,0], [10,0], [15,3], [2,4], [5, -3]])
r.draw(
    blockade_radius=8,
    draw_half_radius=True,
    qubit_colors={3: "darkred", 2: "darkred", 4: "#45b9fe"}
)
```

@HGSilveri : should I also apply this to `Register3D` to keep consistency?


HGSilveri:
Looking good!

> @HGSilveri : should I also apply this to `Register3D` to keep consistency?

Yes, if possible that would be preferable.

pulser-core/pulser/register/_reg_drawer.py
```diff
@@ -39,15 +54,20 @@ def _draw_2D(
         blockade_radius: Optional[float] = None,
         draw_graph: bool = True,
         draw_half_radius: bool = False,
+        qubit_colors: Optional[dict[QubitId, str]] = None,
```
HGSilveri, line 57: This is minutia, by I guess this doesn't have to be optional, it can rather be:
```suggestion
        qubit_colors: dict[QubitId, str] = {}, 
```
It would simplify the rest of the code a bit too

pulser-core/pulser/register/register.py
```diff
@@ -295,6 +295,7 @@ def draw(
         blockade_radius: Optional[float] = None,
         draw_graph: bool = True,
         draw_half_radius: bool = False,
+        qubit_colors: Optional[dict[QubitId, str]] = None,
```
HGSilveri, line 298: Same here

pulser-core/pulser/register/register3d.py
```diff
@@ -178,6 +178,7 @@ def draw(
         blockade_radius: Optional[float] = None,
         draw_graph: bool = True,
         draw_half_radius: bool = False,
+        qubit_colors: Optional[dict[QubitId, str]] = None,
```
HGSilveri, line 181: And here

HGSilveri (commented):
Looking good! Can you include a small test just to make sure the case where qubit colors are specified is at least executed?

pulser-core/pulser/register/_reg_drawer.py
CdeTerra: > This is minutia, by I guess this doesn't have to be optional, it can rather be:
> 
> It would simplify the rest of the code a bit too

It is generally considered as a bad practice by the community to take a mutable as a default parameter. The pitfall is that the parameter may be modified, and subsequent calls will be passed the muted parameter. Even if it's not currently modified, future modifications may not notice about it and trigger bugs.

I have found a way to circumvent this recommendation: I've added a commit where I'm declaring the parameter with `Mapping` so that mypy checks it's not written. Let me know what you think about it.

pulser-core/pulser/register/_reg_drawer.py
```diff
@@ -17,7 +17,7 @@
 from collections import defaultdict
 from collections.abc import Sequence as abcSequence
 from itertools import combinations
-from typing import Optional
+from typing import Mapping, Optional
```
HGSilveri, line 20: I think you're supposed to import `Mapping` from `collections.abc` (importing from `typing` has been deprecated).

HGSilveri (approved):
Looking good, thanks @CdeTerra !

pulser-core/pulser/register/_reg_drawer.py
HGSilveri: Looks good, thanks!

tests/test_sequence.py
```diff
@@ -677,6 +677,7 @@ def test_str(mod_device):
 
 
 def test_sequence():
+    plt.close()
```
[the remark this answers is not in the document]
CdeTerra: I've added it everywhere it was needed for the UTs to pass.

tests/test_register.py
```diff
@@ -337,22 +357,32 @@ def test_cubic():
...
+            draw_half_radius=True,
+            draw_graph=True,
+            **draw_params,
+        )
 
     with pytest.raises(ValueError, match="'blockade_radius' to draw."):
-        reg.draw(draw_half_radius=True)
+        reg.draw(draw_half_radius=True, **draw_params)
+
+    plt.close()
```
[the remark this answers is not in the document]
HGSilveri: Yes, I think if you change that here you can delete all the others you added. That way we can merge this first

PR #481, opened 2023-03-09
Fix requirements conflict and improve `plt.show()` patching

HGSilveri:
- Excludes `numpy == 1.24.0` to avoid conflicts with `matplotlib < 3.6.1`
- Replaces `unitest.patch()` calls for `plt.show()` with a monkeypatch in a fixture that also closes the figure

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import matplotlib.pyplot as plt
 import numpy as np
 import pytest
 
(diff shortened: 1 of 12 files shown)


tests/conftest.py
```diff
@@ -72,3 +73,11 @@ def mod_device() -> Device:
...
         ),
     )
+
+
+@pytest.fixture()
+def patch_plt_show(monkeypatch):
+    # Close residual figures
+    plt.close("all")
+    # Closes a figure instead of showing it
+    monkeypatch.setattr(plt, "show", plt.close)
```
a-corni, line 83: So we no longer test the diplaying of the plot ? As far I have understood, monkey patching is different as patching https://stackoverflow.com/questions/5626193/what-is-monkey-patching

HGSilveri: We never did test it, because we never called `plt.show()`. The only difference to what I'm doing here is that I'm taking the opportunity to close the figure, instead of just suppressing the call.

a-corni (approved):
I am convinced, everything is fine by me.

PR #482, opened 2023-03-10
Implement custom exception for device deserialization

### LINKED ISSUE #478: Exception handling - raise more consistent errors in deserialize_device
In pulser/json/deserializer:

```
def deserialize_device(obj_str: str) -> BaseDevice:
    """Deserialize a device from an abstract JSON object.

    Args:
        obj_str: the JSON string representing the device encoded
            in the abstract JSON format.

    Returns:
        BaseDevice: The Pulser device.
    """
    obj = json.loads(obj_str)
    # Validate the format of the data against the JSON schema.
    jsonschema.validate(instance=obj, schema=schemas["device"])
    return _deserialize_device_object(obj)
 
```
There are multiple ways this function can fail, and raising multiple different classes of Error.
We use this function to validate a serialised abstract representation of a device. Because of the unpredictability of the exceptions raised in this function, we have to cast a wide net with a `try ... except Exception... ` which is less than ideal.

- First point is obviously the json load that could fail and that is not caught. (could raise `TypeError` or `JSONDecodeError`)
    
- Second point the the `jsonschema.validate(instance=obj, schema=schemas["device"]` function that can fail in multiple ways: at least:
    
```
     `jsonschema.exceptions.ValidationError` if the instance
            is invalid

        `jsonschema.exceptions.SchemaError` if the schema itself
            is invalid
```
   
   according to the documentation. Though I've not dug deeper to see if other uncaught exceptions could arise.
   
   - Lastly the `_deserialize_device_object`, which I suspect could also throw other errors, especially when registering the layouts.
   
   Ultimately what would be nice is a unique Exception class that we could catch for when using this deserialize_device function.

HGSilveri:
- Creates `DeserializeDeviceError` and makes it the only error raised by `deserialize_device()`
- Adds an additional check to `RegisterLayout()` to forbid duplicates in trap coordinates

Fixes #478 .

### DIFF
--- a/pulser-core/pulser/json/abstract_repr/deserializer.py
+++ b/pulser-core/pulser/json/abstract_repr/deserializer.py
@@ -288,22 +288,20 @@ def _deserialize_channel(obj: dict[str, Any]) -> Channel:
         if param.init and param.name != "eom_config":
             params[param.name] = obj[param.name]
     try:
-        ch_obj = channel_cls(**params)
+        return channel_cls(**params)
     except (ValueError, NotImplementedError) as e:
         raise AbstractReprError("Channel deserialization failed.") from e
-    return ch_obj
 
 
 def _deserialize_layout(layout_obj: dict[str, Any]) -> RegisterLayout:
     try:
-        reg_layout = RegisterLayout(
+        return RegisterLayout(
             layout_obj["coordinates"], slug=layout_obj.get("slug")
         )
     except ValueError as e:
         raise AbstractReprError(
             "Register layout deserialization failed."
         ) from e
-    return reg_layout
 
 
 def _deserialize_device_object(obj: dict[str, Any]) -> Device | VirtualDevice:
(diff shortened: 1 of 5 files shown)


pulser-core/pulser/json/abstract_repr/deserializer.py
```diff
@@ -263,30 +264,46 @@ def _deserialize_channel(obj: dict[str, Any]) -> Channel:
...
     elif obj["basis"] == "XY":
         channel_cls = Microwave
+    # No other basis allowed by the schema
 
     for param in dataclasses.fields(channel_cls):
         if param.init and param.name != "eom_config":
             params[param.name] = obj[param.name]
-    return channel_cls(**params)
+    try:
+        ch_obj = channel_cls(**params)
```
MatthieuMoreau0, line 291: nit: no need to create a ch_obj variable you can directly return it here, and the exception will be caught regardless

MatthieuMoreau0 (approved):
LGTM !

HGSilveri:
@Augustinio Should no further objections be raised, this will be merged tomorrow.

Augustinio (approved):

PR #488, opened 2023-03-28
Fix for sporadic CI fail on Windows

HGSilveri:
The CI test runs on `windows-os` occasionally fail due claiming "This probably means that Tcl wasn't installed properly.". 

If I understand correctly, this happens when matplotlib is involved, as it tries to use a backend for plotting that requires Tcl, which apparently is poorly installed on some of Github's Windows runners. 
To circumvent this, we can set the backend to `agg`, which no longer requires Tcl and so should work every time.

Note: I'll be temporarily enabling the `full-tests` workflow to run on this PR to check things are running smoothly. I'll then revert that change before merging.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,7 +5,6 @@ on:
     branches:
       - master
       - develop
-  pull_request:
 
 jobs:
   full-tests:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,6 +5,7 @@ on:
     branches:
       - master
       - develop
+  pull_request:
 
 jobs:
   full-tests:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -5,10 +5,14 @@ on:
     branches:
       - master
       - develop
+  pull_request:
```
a-corni, line 8: So if I get it right that's temporary, just to run the tests on this pull request

a-corni (commented):
I agree with these changes. So you will make a next PR where you will revert the `pull_request` ?

.github/workflows/test.yml
HGSilveri: Exactly, I will revert it now

PR #491, opened 2023-03-29
Fix ParamObj serialization with missing keyword arguments

### LINKED ISSUE #489: Error when deserializing sequence
On 0.11.dev0, the following sequence builds and serializes fine:
```python
from pulser import Sequence, Register
from pulser.waveforms import RampWaveform, ConstantWaveform
from pulser.devices import Chadoq2

seq = Sequence(register=Register(dict(a=[0,0],b=[10,0])), device=Chadoq2)
seq.declare_channel("ch0", "rydberg_global")

my_variable = seq.declare_variable("my_variable", size=1)
seq.add(pulser.Pulse(
    RampWaveform(duration=1000, start=0, stop=15),
    ConstantWaveform(duration=1000, value=my_variable),
    0,
), "ch0")
seq.measure("ground-rydberg")

json_seq = seq.to_abstract_repr()
```

However, deserializing it back does not work:
```python
seq2 = Sequence.from_abstract_repr(json_seq)
```

The error is:

```
ValidationError: Additional properties are not allowed ('variable' was unexpected)

Failed validating 'additionalProperties' in schema[0]:
    {'additionalProperties': False,
     'description': 'Simple binary expression involving variables and '
                    'constants.\n'
                    '\n'
                    'The array access behaviour depends on expression:\n'
                    '- index:   - the lhs array is indexed using rhs '
                    'indices, resulting in an array of the same length as '
                    'rhs.   - out of bounds indexing is a runtime error   '
                    '- NOTE: Pulser only supports variable references on '
                    'lhs of index expression.           This limitation '
                    'might be lifted in the future.\n'
                    '- everything else:   - the expression is applied '
                    'element-wise   - operating on arrays of different '
                    'lengths is a runtime error',
     'properties': {'expression': {'description': 'Expresion operation',
                                   'enum': ['add',
                                            'sub',
                                            'mul',
                                            'div',
                                            'mod',
                                            'pow',
                                            'index'],
                                   'type': 'string'},
                    'lhs': {'$ref': '#/definitions/ExprArgument',
                            'description': 'Left-hand side of an '
                                           'operation'},
                    'rhs': {'$ref': '#/definitions/ExprArgument',
                            'description': 'Right-hand side of an '
                                           'operation'}},
     'required': ['expression', 'lhs', 'rhs'],
     'type': 'object'}

On instance:
    {'variable': 'my_variable'}
```

HGSilveri:
- Uses the default values for unspecified keyword arguments when serializing a `ParamObj` to the abstract representation
- Handles the special case of a parametrized `InterpolatedWaveform` without a `times` arguments

Fixes #489 .

### DIFF
--- a/pulser-core/pulser/register/register_layout.py
+++ b/pulser-core/pulser/register/register_layout.py
@@ -302,6 +302,22 @@ class RegisterLayout(RegDrawer):
         hash.update(self.coords.tobytes())
         return hash.digest()
 
+    def static_hash(self) -> str:
+        """Returns the layout's idempotent hash.
+
+        Python's standard hash is not idempotent as it changes between
+        sessions. This hash can be used when an idempotent hash is
+        required.
+
+        Returns:
+            str: An hexstring encoding the hash.
+
+        Note:
+            This hash will be returned as an hexstring without
+            the '0x' prefix (unlike what is returned by 'hex()').
+        """
+        return self._safe_hash().hex()
+
     def __eq__(self, other: Any) -> bool:
         if not isinstance(other, RegisterLayout):
             return False
--- a/pulser-core/pulser/register/special_layouts.py
+++ b/pulser-core/pulser/register/special_layouts.py
@@ -32,11 +32,11 @@ class SquareLatticeLayout(RegisterLayout):
         spacing: The distance between neighbouring traps (in µm).
     """
 
-    def __init__(self, rows: int, columns: int, spacing: int):
+    def __init__(self, rows: int, columns: int, spacing: float):
         """Initializes a SquareLatticeLayout."""
         self._rows = int(rows)
         self._columns = int(columns)
-        self._spacing = int(spacing)
+        self._spacing = float(spacing)
         slug = (
             f"SquareLatticeLayout({self._rows}x{self._columns}, "
             f"{self._spacing}µm)"
@@ -102,9 +102,9 @@ class TriangularLatticeLayout(RegisterLayout):
         spacing: The distance between neighbouring traps (in µm).
     """
 
-    def __init__(self, n_traps: int, spacing: int):
+    def __init__(self, n_traps: int, spacing: float):
         """Initializes a TriangularLatticeLayout."""
-        self._spacing = int(spacing)
+        self._spacing = float(spacing)
         slug = f"TriangularLatticeLayout({int(n_traps)}, {self._spacing}µm)"
         super().__init__(
             patterns.triangular_hex(int(n_traps)) * self._spacing, slug=slug
--- a/tests/test_devices.py
+++ b/tests/test_devices.py
@@ -321,13 +321,13 @@ def test_calibrated_layouts():
         min_atom_distance=4,
         channel_objects=(),
         pre_calibrated_layouts=(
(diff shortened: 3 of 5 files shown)


tests/test_abstract_repr.py
```diff
@@ -490,18 +490,18 @@ def test_paramobj_serialization(self, sequence):
         )
 
         s = json.dumps(
-            Pulse.ConstantDetuning(wf, 0.0, var, post_phase_shift=1.0),
+            Pulse.ConstantDetuning(wf, 0.0, var),
```
a-corni, line 493: The idea is first to test without the post_phase_shift and then with a post_phase_shift of 1 ?

tests/test_abstract_repr.py
```diff
@@ -518,6 +518,58 @@ def test_paramobj_serialization(self, sequence):
         ):
             method_call._to_abstract_repr()
 
+        # Check the defaults are added when not specified
+        s = json.dumps(
+            KaiserWaveform.from_max_val(1.0, var), cls=AbstractReprEncoder
```
a-corni, line 523: Test of a class method

a-corni (commented):
There are things I have not understood.

pulser-core/pulser/parametrized/paramobj.py
```diff
@@ -241,19 +252,18 @@ def _to_abstract_repr(self) -> dict[str, Any]:
...
-                        "post_phase_shift",
-                    )
-                    all_args = {
-                        **dict(zip(signature, self.args[1:])),
-                        **self.kwargs,
-                    }
-                    if "post_phase_shift" not in all_args:
-                        all_args["post_phase_shift"] = 0.0
+                signature = SIGNATURES[
+                    "Pulse" if cls_name == "Pulse" else name
```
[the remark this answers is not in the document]
HGSilveri: This first case is specific to classmethods (it's why we check if the first argument is a class).
On the signatures, we don't have `Pulse.ConstantAmplitude` and `Pulse.ConstantDetuning`, so we have to "convert" the arguments to the `Pulse` signature. The only cases that remain are `BlackmanWaveform.from_max_val` and `KaiserWaveform.from_max_val`.

pulser-core/pulser/parametrized/paramobj.py
```diff
@@ -241,19 +252,18 @@ def _to_abstract_repr(self) -> dict[str, Any]:
...
+                ]
+                # No existing classmethod has *args in its signature
+                assert (
+                    signature.var_pos is None
+                ), "Unexpected signature with VAR_POSITIONAL arguments."
+                all_args = {
+                    **self._default_kwargs,
+                    **dict(zip(signature.all_pos_args(), self.args[1:])),
+                    **self.kwargs,
+                }
```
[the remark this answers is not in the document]
HGSilveri: Yes, would you like me to add a comment with that?

tests/test_abstract_repr.py
HGSilveri: It was just to cover multiple cases: 
1. the value is not given 
2.the value is given as an arg 
3. the value is given as a kwarg

a-corni (approved):
Thanks for the explanations !

PR #492, opened 2023-03-29
Add RegisterLayout.static_hash()

### LINKED ISSUE #490: [Improvement] Public deterministic hash method for RegisterLayout
`RegisterLayout`'s `__hash__` method is not idempotent as it is defined as:
```
def __hash__(self):
    return hash(self._safe_hash()) 
```

The `_safe_hash()` method is idempotent but is private and returns a byte string.

For our public APIs, if we want to make use of public methods from Pulser there is currently only one way to target a `RegisterLayout` from a `Device`, which is to rebuild a whole `RegisterLayout` instance (passed from the API) and check for equality with each of the layouts registered on the device.

We would like a deterministic, idempotent, public method on `RegisterLayout` that returns an `int` or `hexstring`. Which we could then easily add as path param to our `delete` method to remove a layout from a device.

We could for example add the following method:
```
def __index__(self) -> int:
        return int(self._safe_hash().hex(), 16)
```
which could then allow us to call `hex` on an instance of `RegisterLayout`.

HGSilveri:
Closes #490 .

### DIFF
--- a/pulser-core/pulser/register/register_layout.py
+++ b/pulser-core/pulser/register/register_layout.py
@@ -302,6 +302,21 @@ class RegisterLayout(RegDrawer):
         hash.update(self.coords.tobytes())
         return hash.digest()
 
+    def static_hash(self) -> int:
+        """Returns the layout's idempotent hash.
+
+        Python's standard hash is not idempotent as it changes between
+        sessions. This hash can be used when an idempotent hash is
+        required.
+
+        Note:
+            This hash will be returned as an 'int'. When converting to
+            to 'hex', 'hex()' will prepend '0x' to the string,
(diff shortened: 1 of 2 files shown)


Augustinio:
Nit: I'm not sure I see a benefit to returning an int rather than the original hex.

HGSilveri:
> Nit: I'm not sure I see a benefit to returning an int rather than the original hex.

I agree we're splitting hairs at this point, but don't like how sometimes the hexstring has the '0x' and others it doesn't. With an `int`, the uniqueness is guaranteed.

MatthieuMoreau0 (approved):
LGTM, thank you for the quick implementation @HGSilveri !

Augustinio (approved):

PR #506, opened 2023-04-17
Adding the Result class

HGSilveri:
- [x] Defines the `Result` base class to represent the result of any run
- [x] Defines `SampledResult` to represent a result from a series of samples
- [x] Defines`QutipResult` to represent the result of a simulation run at a given evaluation time
- [x] Turns `SimulationResults` into sequences of `Result` objects
- [ ] Defines `Result.plot_histogram()` to plot a result in an histogram
- [ ] Completes the unit tests (with some refactoring to use fixtures and better practices)

### DIFF
--- a/pulser-core/pulser/result.py
+++ b/pulser-core/pulser/result.py
@@ -1,0 +1,110 @@
+# Copyright 2023 Pulser Development Team
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Classes to store measurement results."""
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+from collections import Counter
+from dataclasses import dataclass
+
+import numpy as np
+
+from pulser.register import QubitId
+
+
+@dataclass  # type: ignore
+class Result(ABC):
+    """Base class for storing the result of a sequence run."""
+
+    atom_order: tuple[QubitId, ...]
+    meas_basis: str
+
+    @property
+    def probability_dist(self) -> dict[str, float]:
+        """Probability distribution of each bitstring."""
+        n = self._size
+        return {
+            np.binary_repr(ind, width=n): prob
+            for ind, prob in enumerate(self._weights())
+            if prob != 0
+        }
+
+    @property
+    @abstractmethod
+    def sampling_errors(self) -> dict[str, float]:
+        """The sampling error associated to each bitstring probability."""
+        pass
+
+    @property
+    def _size(self) -> int:
+        return len(self.atom_order)
+
+    @abstractmethod
+    def _weights(self) -> np.ndarray:
+        """Contains the probability of every state in an ordered array."""
+        pass
+
+    def sample(self, n_samples: int) -> Counter[str]:
+        """Samples the probabilite distribution multiple times.
+
+        Args:
+            n_samples: Number of samples to return.
+
+        Returns:
+            Sample distribution of bitstrings corresponding to
+            measured quantum states.
+        """
+        dist = np.random.multinomial(n_samples, self._weights())
+        return Counter(
+            {
+                np.binary_repr(i, self._size): dist[i]
+                for i in np.nonzero(dist)[0]
+            }
+        )
+
+    def plot_histogram(self) -> None:
+        """Plots the result in an histogram."""
+        raise NotImplementedError
+
+
+@dataclass
+class SampledResult(Result):
+    """Represents the result of a run from a series of samples.
+
+    Args:
+        atom_order: The order of the atoms in the bitstrings that
+            represent the measured states.
+        meas_basis: The measurement basis.
+        bitstring_counts: The number of times each bitstring was
+            measured.
+    """
+
+    bitstring_counts: dict[str, int]
+
+    def __post_init__(self) -> None:
+        self.n_samples = sum(self.bitstring_counts.values())
+
+    @property
+    def sampling_errors(self) -> dict[str, float]:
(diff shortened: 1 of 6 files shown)


pulser-core/pulser/result.py
```diff
@@ -0,0 +1,133 @@
...
+import numpy as np
+
+from pulser.register import QubitId
+
+
+@dataclass  # type: ignore
+class Result(ABC):
+    """Base class for storing the result of a sequence run."""
+
+    atom_order: tuple[QubitId, ...]
```
madagra, line 31: What is this variable `atom_order` used for, exactly? Is it some sort of qubit ordering?

pulser-core/pulser/result.py
```diff
@@ -0,0 +1,133 @@
...
+    @property
+    def _size(self) -> int:
+        return len(self.atom_order)
+
+    @abstractmethod
+    def _weights(self) -> np.ndarray:
+        """Contains the probability of every state in an ordered array."""
+        pass
+
+    def sample(self, n_samples: int) -> Counter[str]:
```
madagra, line 59: Don't we want to also have access to the state directly (for simulators) rather than only the samples? This would likely mean adding another (I guess abstract) method.

pulser-simulation/pulser_simulation/qutip_result.py
```diff
@@ -0,0 +1,184 @@
...
+                weights[dec_val] = np.sum(probs[tuple(ind)])
+        else:
+            raise NotImplementedError(
+                "Cannot sample system with single-atom state vectors of "
+                "dimension > 3."
+            )
+        # Takes care of numerical artefacts in case sum(weights) != 1
+        return cast(np.ndarray, weights / sum(weights))
+
+    def get_state(
```
madagra, line 116: This is likely the function I was mentioning could be added to the base interface.

pulser-simulation/pulser_simulation/simresults.py
```diff
@@ -16,26 +16,35 @@
...
 from qutip.piqs import isdiagonal
 
+from pulser.result import Result, SampledResult
+from pulser_simulation.qutip_result import QutipResult
 
-class SimulationResults(ABC):
+ResultType = TypeVar("ResultType", bound=Result)
+
+
+class SimulationResults(ABC, typing.Sequence[ResultType]):
```
madagra, line 37: So now the `SimulationsResults` will be the return type for all the backends? And each backend will have, instead, a different `Result` class implementation. Is this the idea? If so, I think it works fine. I would have hoped to remove this additional `SimulationResults` and just have a single `Result` class (implemented differently for each backend) but I guess it was not really possible.

pulser-simulation/pulser_simulation/simresults.py
```diff
@@ -373,7 +381,7 @@ class CoherentResults(SimulationResults):
 
     def __init__(
```
madagra, line 382: Isn't this one specific to Qutip? It should likely go to the `qutip_result.py` module, no?

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -970,8 +972,17 @@ def _run_solver() -> CoherentResults:
...
                 )
+            results = [
+                QutipResult(
+                    tuple(self._qdict),
+                    self._meas_basis,
+                    state,
+                    self._meas_basis == self.basis_name,
+                )
+                for state in result.states
+            ]
```
madagra, line 983: I guess this one will also be part of the Qutip backend, or at least that's the idea, right?

madagra (commented):
I have some initial comments but I guess the structure of the final structure of the code is not there yet.

pulser-core/pulser/result.py
HGSilveri: It's the qubit order in the bitstring. Since we don't enforce any type of convention on the qubit IDs, I figured this information is needed to avoid ambiguity.

pulser-core/pulser/result.py
HGSilveri: That's the thing, for now we can only get the state when it is coming from Qutip (plus it is given as a `qutip.Qobj`), so I defined `get_state()` inside `QutipResult` instead.

pulser-simulation/pulser_simulation/qutip_result.py
HGSilveri: Maybe I'm looking at things the wrong way, but why we want a method that is particular to some result types to be defined in the base interface?

madagra: Because getting the state should be a common operation for all the backends, imo.

HGSilveri: But then how would that work for QPUs?

pulser-core/pulser/result.py
madagra: Ah got it. It makes sense, thanks @a-corni.

pulser-simulation/pulser_simulation/qutip_result.py
madagra: Either we create two classes from the parent `Results` or, for the QPU, we simply return a `NotImplementedError`, both are fine for me (the first one is likely more elegant but a bit more cumbersome to implement).

pulser-simulation/pulser_simulation/simresults.py
madagra: Yes, perfect. This won't be an issue if the conceptual difference is well documented. However, I surmise we might also have, at some point, results at each evaluation time for our emulators. How would you treat that? In the end, I guess you could just do a `Sequence[Sequence[Result]]` and keep the generic result more flexible. It won't be a big change.

pulser-simulation/pulser_simulation/simresults.py
madagra: Just for sake of bundling together stuff which are related to Qutip. But it can also stay separate, as long as you mention in the comments that it is used in Qutip only (for backward compatibility).

madagra:
@HGSilveri looks almost there to me, just need to fix the pipeline. Do I understand correctly that backward compatibility is maintained?

HGSilveri:
> @HGSilveri looks almost there to me, just need to fix the pipeline. Do I understand correctly that backward compatibility is maintained?

Yes, backward compatibility is maintained. I think I'm done with the changes, I'll move on to the tests now.

madagra (approved):
Nice work!

PR #507, opened 2023-04-20
Impose qubit order in MappableRegister

### LINKED ISSUE #502: Impose qubit order in `MappableRegister`
Starting from a `RegisterLayout`, a `MappableRegister` can be created to pre-declare qubit IDs in a `Sequence` without specifying their positions. These qubits are then only mapped to positions when `Sequence.build()` is called, where trap IDs can be associated to a subset of the declared qubits. 

Currently, there is no imposition on which qubit IDs are selected at build time. For example, a mappable register declared with qubit IDs `("q0", "q1, "q2", "q3")` can then be built with `qubits={"q2": 14, "q0": 8}`. This will create a register with qubits ("q2", "q0"), _in this order_.

This means that qubit IDs in a `MappableRegister` can't be unequivocally associated with indices, which poses a few problems:
- It prevents the use of index-dependent methods, like `Sequence.target_index()` or `Sequence.phase_shift_index()` when the sequence is not parametrized (when it is parametrized, the `MappableRegister` is turned into a `Register` at build time, which will determined the required qubit order). However, this can be circumvented by using `Sequence.target()` and `Sequence.phase_shift()`.
- Most relevantly, it prevents the serialization to the abstract representation when the sequence has a mappable register and local operations (ie `target` or `phase_shift`). This is because the abstract representation format encodes the qubits involved in these operations by their index, which is undefined when the register is mappable.

It seems to me that the only way around this is to impose an order on the `MappableRegister` 's qubit IDs that is preserved when it is converted into a regular register. This implies that, when building a `Register` with N qubits out of a mappable register with N+M qubit IDs, the mapped qubit IDs must be `Register.qubit_ids[:N]` . This will allow us to associate an index to each qubit ID, ie `index i -> qubit_ids[i]` .

Furthermore, the order of the qubits given at build time no longer matters. Returning to the example above, the register could be declared with `qubits={"q1": 14, "q0": 8}"`, but the qubits in the `Register` should still be ordered `("q0", "q1")`.

a-corni:
- I propose to do it by defining a default `Sequence._register` if the provided register is a `MappableRegister`, mapping each qubit id from the `qubit_ids` of the mappable register with the trap of same index in the `layout` of the mappable register.
- `_register` is used as a dummy register during all the operations prior to the `build` operation.
- `layout` of the mappable register is given as an optional parameter of `_register`, such that it can be used in `build` to perform the same operation as before.
Closes #502

### DIFF
--- a/pulser-core/pulser/json/abstract_repr/serializer.py
+++ b/pulser-core/pulser/json/abstract_repr/serializer.py
@@ -153,7 +153,7 @@ def serialize_abstract_sequence(
         og_dim = target_array.ndim
         if og_dim == 0:
             target_array = target_array[np.newaxis]
-        indices = seq._register.find_indices(target_array.tolist())
+        indices = seq.get_register().find_indices(target_array.tolist())
         return indices[0] if og_dim == 0 else indices
 
     def get_kwarg_default(call_name: str, kwarg_name: str) -> Any:
--- a/pulser-core/pulser/json/abstract_repr/serializer.py
+++ b/pulser-core/pulser/json/abstract_repr/serializer.py
@@ -153,14 +153,7 @@ def serialize_abstract_sequence(
         og_dim = target_array.ndim
         if og_dim == 0:
             target_array = target_array[np.newaxis]
-        try:
-            indices = seq.register.find_indices(target_array.tolist())
-        # RuntimeError raised when calling seq.register for a MappableRegister
-        except RuntimeError:
-            raise NotImplementedError(
-                "Serialization of sequences with local operations and"
-                " a mappable register is currently not supported."
-            )
+        indices = seq._register.find_indices(target_array.tolist())
         return indices[0] if og_dim == 0 else indices
 
     def get_kwarg_default(call_name: str, kwarg_name: str) -> Any:
--- a/pulser-core/pulser/json/abstract_repr/serializer.py
+++ b/pulser-core/pulser/json/abstract_repr/serializer.py
@@ -153,7 +153,9 @@ def serialize_abstract_sequence(
         og_dim = target_array.ndim
         if og_dim == 0:
             target_array = target_array[np.newaxis]
-        indices = seq.get_register().find_indices(target_array.tolist())
+        indices = seq.get_register(include_mappable=True).find_indices(
+            target_array.tolist()
+        )
         return indices[0] if og_dim == 0 else indices
 
     def get_kwarg_default(call_name: str, kwarg_name: str) -> Any:
--- a/pulser-core/pulser/register/mappable_reg.py
+++ b/pulser-core/pulser/register/mappable_reg.py
@@ -72,6 +72,11 @@ class MappableRegister:
             raise ValueError(
                 "All qubits must be labeled with pre-declared qubit IDs."
             )
+        elif set(chosen_ids) != set(self.qubit_ids[: len(chosen_ids)]):
+            raise ValueError(
+                f"'qubits' should contain the {len(qubits.keys())} first "
+                "elements of the 'qubit_ids'."
+            )
         register_ordered_qubits = {
             id: qubits[id] for id in self._qubit_ids if id in chosen_ids
         }
--- a/pulser-core/pulser/register/mappable_reg.py
+++ b/pulser-core/pulser/register/mappable_reg.py
@@ -74,8 +74,9 @@ class MappableRegister:
             )
         elif set(chosen_ids) != set(self.qubit_ids[: len(chosen_ids)]):
             raise ValueError(
-                f"'qubits' should contain the {len(qubits.keys())} first "
-                "elements of the 'qubit_ids'."
+                f"To declare {len(qubits.keys())} qubits, 'qubits' should "
+                f"contain the first {len(qubits.keys())} elements of the "
+                "'qubit_ids'."
             )
         register_ordered_qubits = {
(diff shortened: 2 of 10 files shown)


pulser-core/pulser/json/abstract_repr/serializer.py
```diff
@@ -153,14 +153,7 @@ def convert_targets(
...
             target_array = target_array[np.newaxis]
-        try:
-            indices = seq.register.find_indices(target_array.tolist())
-        # RuntimeError raised when calling seq.register for a MappableRegister
-        except RuntimeError:
-            raise NotImplementedError(
-                "Serialization of sequences with local operations and"
-                " a mappable register is currently not supported."
-            )
+        indices = seq._register.find_indices(target_array.tolist())
```
HGSilveri, line 156: Sweet! Once `Sequence.get_register()` is defined, it would be nice to use it here instead of accessing `_register`

a-corni: `Sequence.get_register()` is now added ;)

pulser-core/pulser/register/mappable_reg.py
```diff
@@ -72,6 +72,11 @@ def build_register(self, qubits: Mapping[QubitId, int]) -> BaseRegister:
             raise ValueError(
                 "All qubits must be labeled with pre-declared qubit IDs."
             )
+        elif set(chosen_ids) != set(self.qubit_ids[: len(chosen_ids)]):
+            raise ValueError(
+                f"'qubits' should contain the {len(qubits.keys())} first "
+                "elements of the 'qubit_ids'."
+            )
```
HGSilveri, line 79: I think this message is a bit misleading because it makes you think you need to declare a given number of qubits (which coincidentally matches the number you are trying). Something a bit more verbose but clearer could be:
```suggestion
        elif set(chosen_ids) != set(self.qubit_ids[: len(chosen_ids)]):
            raise ValueError(
                f"To declare {len(qubits.keys())} qubits, 'qubits' should contain"
                f" the first {len(qubits.keys())} elements of the 'qubit_ids'."
            )
```

tutorials/advanced_features/Register Layouts.ipynb
```diff
@@ -525,7 +525,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "To define the register, we can then call `Sequence.build()`, indicated in the `qubits` argument the map between qubit IDs and trap IDs (note that not all the qubit IDs need to be associated to a trap ID). \n",
+    "To define the register, we can then call `Sequence.build()`, indicating in the `qubits` argument the map between qubit IDs and trap IDs. Note that not all the qubit IDs need to be associated to a trap ID, and that the qubit IDs have to be defined in their order of appearance in `MappableRegister.qubit_ids` (it is not possible to associate a trap ID to qubit ID \"q4\" if no trap ID was assigned to qubit ID \"q3\").\n",
```
HGSilveri, line 528: Nice!

tests/test_sequence.py
```diff
@@ -922,76 +926,54 @@ def test_sequence(reg, device, patch_plt_show):
     assert str(seq) == str(seq_)
 
 
-def test_config_slm_mask(device):
-    reg_s = Register({"q0": (0, 0), "q1": (10, 10), "q2": (-10, -10)})
-    seq_s = Sequence(reg_s, device)
-
+@pytest.mark.parametrize("qubit_ids", [["q0", "q1", "q2"], [0, 1, 2]])
+def test_config_slm_mask(qubit_ids, device):
+    reg: Register | MappableRegister
```
HGSilveri, line 931: What is mypy complaining about without this?

pulser-core/pulser/sequence/sequence.py
```diff
@@ -185,6 +185,10 @@ def register(self) -> BaseRegister:
             )
         return cast(BaseRegister, self._register)
 
+    def get_register(self) -> BaseRegister | MappableRegister:
```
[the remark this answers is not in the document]
HGSilveri: Btw, you can then also do an overload to get better typing, something like:
```
@overload
def get_register(self, include_mappable: Literal[False]) -> Register:
    ...
```

tests/test_sequence.py
a-corni: Oh it's just a simple rewriting of the test. I found it was better to write it with a parametrize call.

I implemented this when I thought I could also modify the slm. With my changes it was making the initial test super long and I found it necessary to synthetize it. Now these changes are unnecessary but I find it nicer this way. If you don't find so I will cancel these changes.

HGSilveri: No, that's ok, I was just wondering

pulser-core/pulser/sequence/sequence.py
```diff
@@ -185,9 +185,21 @@ def register(self) -> BaseRegister:
...
-    def get_register(self) -> BaseRegister | MappableRegister:
+    @overload
+    def get_register(self, include_mappable: Literal[False]) -> BaseRegister:
+        pass
+
+    @overload
+    def get_register(
+        self, include_mappable: Literal[True]
+    ) -> BaseRegister | MappableRegister:
+        pass
```
HGSilveri, line 196: Mypy complains without this one?

HGSilveri (approved):
LGTM!

PR #519, opened 2023-05-11
Add a QutipEmulator class to perform simulations from samples instead of sequence

a-corni:


pulser-core/pulser/sampler/samples.py
```diff
@@ -292,6 +292,10 @@ class SequenceSamples:
     samples_list: list[ChannelSamples]
     _ch_objs: dict[str, Channel]
     _slm_mask: _SlmMask = field(default_factory=_SlmMask)
+    _interaction_coeff_xy: float | None = None
+    _interaction_coeff: float | None = None
```
a-corni, line 296: 2 other approaches here:
- store the device of the `Sequence`
- provide the device in the init of the Simulation (I favored keeping the signature as simple as possible)

pulser-core/pulser/sampler/samples.py
```diff
@@ -303,6 +307,7 @@ def max_duration(self) -> int:
         """The maximum duration among the channel samples."""
         return max(samples.duration for samples in self.samples_list)
 
+    @property
```
a-corni, line 310: Felt like this was actually a property

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -716,17 +688,24 @@ def make_xy_term(q1: QubitId, q2: QubitId) -> qutip.Qobj:
             """
             dist = np.linalg.norm(self._qdict[q1] - self._qdict[q2])
             coords_dim = len(self._qdict[q1])
-            mag_norm = np.linalg.norm(self._seq.magnetic_field[:coords_dim])
+            mag_norm = np.linalg.norm(
+                cast(
+                    np.ndarray,
+                    self.samples_obj._magnetic_field,
+                )[:coords_dim]
```
HGSilveri, line 695: Since you are casting again below, maybe cast once and put it in a variable instead

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -1062,3 +1044,231 @@ def _run_solver() -> CoherentResults:
...
+
+            - An ArrayLike object of times in µs if you wish to only include
+              those specific times.
+
+            - A float to act as a sampling rate for the resulting state.
+        with_modulation: Whether to simulated the sequence with the programmed
+            input or the expected output.
+    """
+
+    def __init__(
```
HGSilveri, line 1147: Should we deprecate `Simulation` too?

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -1062,3 +1044,231 @@ def _run_solver() -> CoherentResults:
...
+        return self._emulator.evaluation_times
+
+    @evaluation_times.setter
+    def evaluation_times(self, value: Union[str, ArrayLike, float]) -> None:
+        """Sets times at which the results of this simulation are returned."""
+        warnings.warn(
+            DeprecationWarning(
+                "Setting `evaluation_times` is deprecated,"
+                " use `set_evaluation_times` instead."
+            ),
```
HGSilveri, line 1189: The `DeprecationWarnings` are not shown by default, it's better to change the warnings filter inside a context manager:
```suggestion
        with warnings.catch_warnings():
            warnings.simplefilter("always")
            warnings.warn(
                DeprecationWarning(
                    "Setting `evaluation_times` is deprecated,"
                    " use `set_evaluation_times` instead."
                ),
```

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -64,58 +67,76 @@ class Simulation:
...
-        with_modulation: bool = False,
     ) -> None:
         """Instantiates a Simulation object."""
-        if not isinstance(sequence, Sequence):
+        # Initializing the samples obj
+        if not isinstance(sampled_seq, SequenceSamples):
             raise TypeError(
                 "The provided sequence has to be a valid "
-                "pulser.Sequence instance."
+                "pulser.SequenceSamples instance."
```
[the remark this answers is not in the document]
HGSilveri: It's a bit long, I think just `SequenceSamples` is fine

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -64,58 +67,76 @@ class Simulation:
...
             )
-        if sequence.is_parametrized() or sequence.is_register_mappable():
+        if sampled_seq.max_duration == 0:
+            raise ValueError("SequenceSamples is empty.")
+        # Check compatibility of register and device
+        self.device = device
+        self.device.validate_register(register)
+        self.register = register
```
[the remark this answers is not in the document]
a-corni: I agree they have to be made private. However we could make methods `set_register` and `set_device`, that would let the user change them (and make the small tests on them) without starting a new `QutipEmulator` (just the way we do with initial states/evaluation_times and run)

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -1062,3 +1044,231 @@ def _run_solver() -> CoherentResults:
...
+            plt.savefig(fig_name, **kwargs_savefig)
+        plt.show()
+
+    def __getattr__(self, name: str) -> Any:
+        return getattr(self._emulator, name)
+
+    # def __setattr__(self, name: str, value: Any) -> Any:
+    #     if hasattr(self, name):
+    #         return super().__setattr__(name, value)
+    #     return setattr(self._emulator, name, value)
```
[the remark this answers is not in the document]
a-corni: I am guessing so

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -64,58 +67,76 @@ class Simulation:
...
+            targets_slots: set[QubitId] = set()
+            slots = []
+            for slot in ch_samples.slots:
+                if sampled_seq._ch_objs[ch].addressing == "Local":
+                    # Check that register defines targets of Local Channels
+                    targets_slots = targets_slots.union(
+                        *[slot.targets for slot in ch_samples.slots]
+                    )
+                    slots = ch_samples.slots
+                    break
```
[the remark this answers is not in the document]
a-corni: At the beginning it was outside, and then I was afraid it was a bit complex to understand + it was taking lots of place so I put it there... Let me explain below what is the purpose.

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -64,58 +67,76 @@ class Simulation:
...
+                elif slot.tf <= sampled_seq._slm_mask.end:
+                    # Check that targets of SLM mask are defined in register
+                    slots.append(slot)
+                    targets_slots = targets_slots.union(slot.targets)
+                else:
+                    # Replace targets of Global channels by qubits of register
+                    slots.append(
+                        replace(slot, targets=set(register.qubit_ids))
+                    )
+            samples_list.append(replace(ch_samples, slots=slots))
```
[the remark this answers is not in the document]
a-corni: In this `for` loop, I am doing two things:
- Checking that `slot.targets <= set(register.qubit_ids)` for slots in local channels and in slm mask
- Replacing the targets of the slots that are in global channels and not masked by the slm by all the qubit_ids in the register. That refers to one of our discussion about sequence built with only `Global` channels: in this case we don't care about the register (it's a dummy register) and it should only be provided in the simulation (which is the register taken here)

pulser-simulation/pulser_simulation/simulation.py
a-corni: I hesitated, I think so too.

pulser-simulation/pulser_simulation/simulation.py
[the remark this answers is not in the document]
HGSilveri: `set_register` will force you to rebuild the whole Hamiltonian anyway, as will `set_device` if the interaction coefficient changes for example. I'm inclined to not having to deal with that complexity

pulser-simulation/pulser_simulation/simulation.py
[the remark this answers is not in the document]
HGSilveri: Ahh yes, I remember now...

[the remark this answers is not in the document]
a-corni: But it's True I don't care about the SLM. I should just check that `SequenceSamples._slm_masl.targets <= set(register.qubit_ids)`

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -83,58 +83,63 @@ def __init__(
...
             )
         if sampled_seq.max_duration == 0:
             raise ValueError("SequenceSamples is empty.")
         # Check compatibility of register and device
-        self.device = device
-        self.device.validate_register(register)
-        self.register = register
+        self.__device = device
+        self.__device.validate_register(register)
+        self.__register = register
```
HGSilveri, line 93: Double leading underscore will mangle the attribute, won't it? Are you sure this is what you want? And if so, why?

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -64,58 +67,81 @@ class Simulation:
...
-            extended_duration=self._tot_duration + 1,
-        )
+        samples_list = []
+        for ch, ch_samples in sampled_seq.channel_samples.items():
+            if sampled_seq._ch_objs[ch].addressing == "Local":
+                # Check that targets of Local Channels are defined
+                # in register
+                if not set().union(
+                    *[slot.targets for slot in ch_samples.slots]
+                ) <= set(register.qubit_ids):
```
HGSilveri, line 119: Did you try this form?
```suggestion
                if not set().union(
                    slot.targets for slot in ch_samples.slots
                ) <= set(register.qubit_ids):
```

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -64,58 +67,81 @@ class Simulation:
...
-        self._modulated = with_modulation
-        if self._modulated and sequence._slm_mask_targets:
-            raise NotImplementedError(
-                "Simulation of sequences combining an SLM mask and output "
-                "modulation is not supported."
+        # Check compatibility of masked samples and register
+        if not sampled_seq._slm_mask.targets <= set(register.qubit_ids):
+            raise ValueError(
+                "The ids of qubits targeted in SLM mask"
+                " should be defined in register"
```
HGSilveri, line 110: nit: The error messages in this section could use a clean-up (punctuation and stuff like that)

pulser-simulation/pulser_simulation/simulation.py
a-corni: image
I went a bit too quick on an article.

pulser-simulation/pulser_simulation/simulation.py
a-corni: I got (and get after re-try) `TypeError: unhashable type: 'set'`

HGSilveri (approved):
LGTM!

PR #520, opened 2023-05-12
Clarify the units of the interaction coefficient

HGSilveri:


pulser-core/pulser/devices/interaction_coefficients/__init__.py
```diff
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""C_6/hbar (in  um^6 / us`), coeffs for Rydberg levels between 50 and 100."""
+"""C_6/hbar (in  µm^6 / µs), coeffs for Rydberg levels between 50 and 100."""
```
lvignoli, line 14: Maybe precise its for Rubidium 87 (right?) and the source of these coeffs—if known…

lvignoli (approved):
Just a nit on origin of the coeffs, otherwise a nice addition, thanks!

pulser-core/pulser/devices/interaction_coefficients/__init__.py
LucasGitQ: A better expression would be C_6/hbar (in 2pi x MHz x um^6) yes. I got those coefficients from calculations I did with ARC package (arc-alkali-rydberg-calculator.readthedocs.io) doublechecked with PairInteraction package (https://www.pairinteraction.org/)

HGSilveri: Much obliged!

HGSilveri: Actually, these units are bugging me a bit... It's true we work in rad/µs which implies multiplying the value in MHz by 2π. However, the value should absorb the 2π and I think specifying the units like this is contrary to that. 
Let's say you have some value `x` in MHz, then indeed `x MHz = x*2π rad/µs`. What we are storing is actually `x*2π`, but if the units are `2π rad/µs`, to me it indicates that we are still storing only `x`. Do you agree with this @LucasGitQ ?

LucasGitQ: You can write rad/us * um^6 if you prefer (maybe while recalling rad/us = 2pi * MHz). 
You can maybe specify that in the Hamiltonian written in Pulser https://pulser.readthedocs.io/en/stable/review.html the U_ij = h*C_6 = h * MHz * um^6 so that you need to compare hbar*Omega to h*C_6

HGSilveri: I think we are making a mistake here, `2π rad/s = 1 Hz`, so `rad/µs = 1/(2π) MHz`, right? I feel like I'm going crazy here

LucasGitQ: Now I'm confused too

HGSilveri: - [C_6 / h] = MHz µm^6 -> I agree with this
- [C_6 / hbar] = rad/µs µm^6  -> This is what we are storing as the interaction coefficients
The units agree with the Rabi frequency units we use for `Omega` (basically, if our whole Hamiltonian is divided by \hbar all the terms should be in rad/µs)

LucasGitQ: Ok then we can keep the way it was written before !

HGSilveri: I agree, I just added `rad` to be more explicit.

PR #523, opened 2023-05-17
Removing ParamObj.__getattr__()

HGSilveri:
This functionality was basically not used and was causing weird (and not yet understood) recursion errors when multiprocessing, so it was decided it was best to just remove it.


a-corni (approved):
Looks good to me !

madagra (approved):
Looks good to me as well.

PR #526, opened 2023-05-29
Add an error when a Channel is called with an eom_config but without …

### LINKED ISSUE #497: Add an error when a `Channel` is called with an `eom_config` but without a `modulation bandwidth`
No error is raised when a `Channel` is called with an `eom_config` but without a `modulation bandwidth`.
This is currently only the case for a `Rydberg` channel. It should be generalized to all channels.

dakk:
…a modulation bandwidth

Fixes #497

### DIFF
--- a/tests/test_channels.py
+++ b/tests/test_channels.py
@@ -193,7 +193,7 @@ def test_eom_channel():
         TypeError,
         match="When defined, 'eom_config' must be a valid 'RydbergEOM'",
     ):
-        Rydberg.Global(None, None, eom_config=BaseEOM(50))
+        Rydberg.Global(None, None, mod_bandwidth=3, eom_config=BaseEOM(50))
 
     with pytest.raises(
         ValueError,
--- a/tests/test_channels.py
+++ b/tests/test_channels.py
@@ -190,17 +190,17 @@ _eom_config = RydbergEOM(
 
 def test_eom_channel():
     with pytest.raises(
-        TypeError,
-        match="When defined, 'eom_config' must be a valid 'RydbergEOM'",
-    ):
-        Rydberg.Global(None, None, mod_bandwidth=3, eom_config=BaseEOM(50))
-
-    with pytest.raises(
         ValueError,
         match="'eom_config' can't be defined in a Channel without a"
         " modulation bandwidth",
     ):
         Rydberg.Global(None, None, eom_config=_eom_config)
+
+    with pytest.raises(
+        TypeError,
+        match="When defined, 'eom_config' must be a valid 'RydbergEOM'",
+    ):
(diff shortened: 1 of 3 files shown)


a-corni (commented):
Very nice @dakk !

a-corni:
Here are a couple of comments:
- Have a look to the Continuous Integration Requirements section in the file [Contributing](https://github.com/pasqal-io/Pulser/blob/master/CONTRIBUTING.md). You have to install and run black, flake8 to solve the two first failing CI requirements.
- To solve the two last failing requirements, you should run the tests using `pytest --cov` (or only a text file using `pytest tests/mytestfile.py`). You should also modify the tests files to take into account your modifications.

dakk:
> Here are a couple of comments:
> 
>     * Have a look to the Continuous Integration Requirements section in the file [Contributing](https://github.com/pasqal-io/Pulser/blob/master/CONTRIBUTING.md). You have to install and run black, flake8 to solve the two first failing CI requirements.
> 
>     * To solve the two last failing requirements, you should run the tests using `pytest --cov` (or only a text file using `pytest tests/mytestfile.py`). You should also modify the tests files to take into account your modifications.

Sorry for that, I forgot to run it before commit. Now should be ok.
Anyway, when I try to run flake8 I'm getting hundreds of errors, while the CI only returns two; do you know what is the problem with it?
image

tests/test_channels.py
```diff
@@ -193,7 +193,7 @@ def test_eom_channel():
         TypeError,
         match="When defined, 'eom_config' must be a valid 'RydbergEOM'",
     ):
-        Rydberg.Global(None, None, eom_config=BaseEOM(50))
+        Rydberg.Global(None, None, mod_bandwidth=3, eom_config=BaseEOM(50))
```
a-corni, line 196: nit: Perhaps you can invert the place of this error with the one below (first condition is to have mod_bandwidth defined, second is to have eom_config defined) ?

a-corni (commented):
Well done fixing the CIs !

tests/test_channels.py
a-corni: Hum no I am okay with the current implementation, it was just about the order in the test file:
```
    with pytest.raises(
        ValueError,
        match="'eom_config' can't be defined in a Channel without a"
        " modulation bandwidth",
    ):
        Rydberg.Global(None, None, eom_config=_eom_config)

    with pytest.raises(
        TypeError,
        match="When defined, 'eom_config' must be a valid 'RydbergEOM'",
    ):
        Rydberg.Global(None, None, mod_bandwidth=3, eom_config=BaseEOM(50))
```

dakk: Got it, done

a-corni (approved):
Looks good to me !

Issue #522, opened 2023-05-17
Adding a function to draw `SequenceSamples`


a-corni:
At the moment, we can draw instances of `Sequence`  using the `draw_sequence` function of `pulser.sequence._seq_drawer`, but there is no function to draw `SequenceSamples` (the result of the sampling of a `Sequence` instance, that can be obtained using the `sample` method from `pulser.sampler.sampler`). We could need such a function to draw `SequenceSamples` in the recently built `QutipEmulator` of `pulser_simulation`.

In fact, `draw_sequence` mainly relies on the `sample` method, and handles almost exclusively `SequenceSamples`. I propose to have an intermediate function (`draw_samples`) that would draw samples, and that would be called by `draw_sequence` with the sampled sequence.

This can be implemented by modifying `gather_data` such that it works with `SequenceSamples`. The features of `Sequence` that cannot be accessed in `SequenceSamples` (like interpolation of waveforms) can normally still be performed in the `draw_sequence`, such that the previous results of `draw_sequence` are unchanged.

neogyk:
Hey all, I would like to work on this issue as part of the Unitary Hack. 
 Is it possible for me to be assigned to this challenge?

a-corni:
Hey @neogyk,
Thanks for you interest in this issue ! It is now assigned to you, good luck !

dakk:
@a-corni I have a doubt; the parameter list of draw_samples can have also the sequence?

a-corni:
There is indeed an issue in rebuilding the `targets` from a `ChannelSample` of a `SequenceSamples`:
- you can get the "initial" target from `ChannelSample.initial_targets` (so no issues here).
- you can get an estimation of the other "targets" by checking if the `targets` between two `_TargetSlot` varies. If so, there was a `target` slot in between the two pulses (only `_TimesSlot` of type `Pulse` of the `sequence._schedule[ch].slots` gives a `TargetSlot` in the `SequenceSamples.channel_samples[ch].slots`, meaning that the time between two `TargetSlot`s is for `_TimeSlot`s of type `delay` and `target`). So the times of the `targets` you will find from a `ChannelSample` will be composed of targets and delays.
You need these `targets` slots in `draw_samples`, as they are used from line 537 to 634 to build `target_regions`, and I think that retarget operations should be displayed on samples. You can note that the operations of lines 537-576 are not impacted thanks to `ChannelSample.initial_targets`.

I see two options:
[...]

PR #528, opened 2023-05-30
Fix Register drawing out of window

### LINKED ISSUE #501: Drawing of `Register` goes out of window boundaries
When plotting a very simple register, the axes of the plot are going out of the boundaries of the figure, making it unreadable without manual resizing : 

image

Code to reproduce the bug : 

```python
from pulser import Register

register = Register.rectangle(rows=1, columns=2)

register.draw()

```

This issue appears on the following platforms and modules versions : 

| Platform           | Python | Pulser | Matplotlib |
| ------------------ | ------ | ------ | ---------- |
| Windows 10         | 3.10.9 | 0.10.0 | 3.7.1      |
| Ubuntu 20.04 WSL 2 | 3.10.9 | 0.11.0 | 3.7.1      |

dakk:
- Fixes #501 
- Use tight_layout and adjust plot size

### DIFF
--- a/pulser-core/pulser/register/_reg_drawer.py
+++ b/pulser-core/pulser/register/_reg_drawer.py
@@ -330,7 +330,7 @@ class RegDrawer:
         big_side = max(diffs)
         proportions = diffs / big_side
         Ls = proportions * min(
-            big_side / 4, 10
+            big_side / 1.9, 10
         )  # Figsize is, at most, (10,10)
         fig, axes = plt.subplots(figsize=Ls)
 
--- a/pulser-core/pulser/register/register.py
+++ b/pulser-core/pulser/register/register.py
@@ -361,6 +361,9 @@ class Register(BaseRegister, RegDrawer):
             draw_half_radius=draw_half_radius,
             qubit_colors=qubit_colors,
         )
+
+        plt.tight_layout()
+
         if fig_name is not None:
             plt.savefig(fig_name, **kwargs_savefig)
 


HGSilveri (commented):
I agree that this solution solves the issue. However, I took a look at the docs and it also makes some registers unnecessarily large - eg from [this](https://pulser.readthedocs.io/en/stable/tutorials/creating.html#1.-Creating-the-Register) to [this](https://pulser--528.org.readthedocs.build/en/528/tutorials/creating.html).
I suspect that this is related to the scaling factor having been increased. Why did you choose to change it? Do you think something more moderate can be reached?

dakk:
That value was a break point value; I will try to find a better solution considering also the other snippets you linked me

a-corni:
You can have a look to the effects of your changes in the docs in the checks after you push them or by running in the source of the repository:
```
pip install -r docs/requirements.txt
cd docs
make html
```
In addition to Henrique's comment, the scaling factor is being used in other places in `_reg_drawer`, notably in `_initialize_fig_axes_projection`. I think it should be modified as well.

a-corni:
Hey @dakk,
You are right, the size of the figure is the breaking point. I think you can do it by setting a minimum figure size (arbitrarily, such that it works).
Also `tight_layout()` is not advised and deprecated according the the matplotlib documentation. I suggest you declare the `layout` keyword in `subplots` (in both `_initialize_fig_axes` and `_initialize_fig_axes_projection`). From my tests, it looks better using a `"constrained"` layout than a `"tight"` one (see more [on this tutorial](https://matplotlib.org/stable/tutorials/intermediate/constrainedlayout_guide.html#simple-example))

dakk:
> Hey @dakk, You are right, the size of the figure is the breaking point. I think you can do it by setting a minimum figure size (arbitrarily, such that it works). Also `tight_layout()` is not advised and deprecated according to the matplotlib documentation. I suggest you declare the `layout` keyword in `subplots` (in both `_initialize_fig_axes` and `_initialize_fig_axes_projection`). From my tests, it looks better using a `"constrained"` layout than a `"tight"` one (see more [on this tutorial](https://matplotlib.org/stable/tutorials/intermediate/constrainedlayout_guide.html#simple-example))

I'm back on it, I will try other solutions in the next days

dakk:
Your solution seems to work perfectly (min size and constrained layout); these are the results.

Figure_1
image
image
image

a-corni (commented):
Nice ! I still see couple of issues in the docs:
- In the [intro to the Rydberg blockade](https://pulser--528.org.readthedocs.build/en/528/intro_rydberg_blockade.html) there is an issue with the graph there 
image
- In the [SLM tutorial](https://pulser--528.org.readthedocs.build/en/528/tutorials/slm_mask.html), the quality of the first plot is also decreased 
image

I have two ideas that could eventually fix these plots:
- set a minimal ylim (of `(-1, 1)` ?)
- Perhaps could we fix the values for which to show the ticks when all the atoms are on the same line and the rydberg radius is not displayed (I find this weird that we never get the same ticks for all these plots)

[EDIT]: I am not sure these will solve the issues in the end. It seems more related to the optimization performed by the "constrained" layout. Perhaps would it work better with a `tight` layout ? Or by extending the size of the figure.
I also get that the difference between the ticks of the various graph come from the labels of the register. Perhaps could you try to optimize the asked figure (1 line, 2 columns) with a prefix `"atom"` instead of None (as in the introduction to the Rydberg blockade tutorial)

dakk:
The problem seems to be the min height; if I set 1.0 for the minimum height it seems to work good everywhere. I'm doing some other investigation to assure it works in every context

dakk:
> Looks good to me ! Well done solving this issue @dakk ! We still have [one issue left](https://github.com/pasqal-io/Pulser/issues/522) that does not seem tackled, would you like to contribute to it ? If so, I will let you put a comment on it such that I assign it to you.

Sure, I will take a look in the next days

PR #529, opened 2023-05-30
Fix SPAM errors introducing overhead

### LINKED ISSUE #352: SPAM errors introduce large simulation overhead
Doing simulations with SPAM errors introduces a significant time overhead dependent on the `samples_per_run` argument, which was intended to give extra statistics (almost) for free.

The reason is that detection errors (false positives and false negatives) cycle on the sampled bitstrings and randomly flip the bits. In the following simple example execution time increases tenfold when increasing `samples_per_run` from 10 to 1000:
```python
reg = Register.square(side=2)
pulse = Pulse.ConstantDetuning(BlackmanWaveform(duration=1000, area=np.pi),0,0)
seq = Sequence(reg, MockDevice)
seq.declare_channel('ch', 'rydberg_global')
seq.add(pulse, 'ch')
sim = Simulation(seq)
noise_config = SimConfig(noise=('SPAM', 'doppler', 'amplitude'),
                             eta=0.005, epsilon=0.03, epsilon_prime=0.08,
                             temperature=30, laser_waist=148,
                             runs=10, samples_per_run=10)
sim.set_config(noise_config)
sim.show_config()
res = sim.run()
```
A first issue is that reducing the `epsilon` and `epsilon_prime` parameters to zero does not get rid of the overhead.
Secondly, the time overhead could probably be improved following a suggestion of @Louis-PaulHenry.

As @sebgrijalva pointed out, reducing the evaluation times helps dramatically, so it could be used as a workaround.

dakk:
- Fixes #352 
- Handle eps_p = eps = 0 case

### DIFF
--- a/pulser-simulation/pulser_simulation/simresults.py
+++ b/pulser-simulation/pulser_simulation/simresults.py
@@ -528,29 +528,28 @@ class CoherentResults(SimulationResults):
         eps = self._meas_errors["epsilon"]
         eps_p = self._meas_errors["epsilon_prime"]
 
+        if eps_p == 0.0 and eps == 0.0:
+            return sampled_state
+
         for shot, n_detects in sampled_state.items():
             # Shot as an array of 1s and 0s
             shot_arr = np.array(list(shot), dtype=int)
 
-            if eps_p != 0.0 and eps != 0.0:
-                # Probability of flipping each bit
-                flip_probs = np.array(
-                    [eps_p if x == "1" else eps for x in shot]
-                )
-                # 1 if it flips, 0 if it stays the same
-                flips = (
-                    np.random.uniform(size=(n_detects, len(flip_probs)))
-                    < flip_probs
-                ).astype(int)
-                # XOR betwen the original array and the flips
-                # Gives an array of n_detects individual shots
-                new_shots = shot_arr ^ flips
-            else:
-                new_shots = shot_arr
+            # Probability of flipping each bit
+            flip_probs = np.array([eps_p if x == "1" else eps for x in shot])
+            # 1 if it flips, 0 if it stays the same
+            flips = (
+                np.random.uniform(size=(n_detects, len(flip_probs)))
+                < flip_probs
+            ).astype(int)
+            # XOR betwen the original array and the flips
+            # Gives an array of n_detects individual shots
+            new_shots = shot_arr ^ flips
 
             # Count all the new_shots
-            detected_sample_dict += Counter(
+            new_shots_count = [
                 "".join(map(str, measured)) for measured in new_shots
-            )
+            ]
+            detected_sample_dict.update(new_shots_count)
 
         return detected_sample_dict
--- a/pulser-simulation/pulser_simulation/simresults.py
+++ b/pulser-simulation/pulser_simulation/simresults.py
@@ -521,15 +521,14 @@ class CoherentResults(SimulationResults):
             quantum states at time t.
         """
         sampled_state = super().sample_state(t, n_samples, t_tol)
-        if self._meas_errors is None:
+        if self._meas_errors is None or (
+            self._meas_errors["epsilon"] == 0.0
+            and self._meas_errors["epsilon_prime"] == 0
+        ):
             return sampled_state
 
         eps = self._meas_errors["epsilon"]
         eps_p = self._meas_errors["epsilon_prime"]
-
-        if eps_p == 0.0 and eps == 0.0:
-            return sampled_state
-
         shots = list(sampled_state.keys())
         n_detects_list = list(sampled_state.values())
 
--- a/pulser-simulation/pulser_simulation/simresults.py
+++ b/pulser-simulation/pulser_simulation/simresults.py
@@ -18,7 +18,7 @@ from __future__ import annotations
 import collections.abc
 import typing
 from abc import ABC, abstractmethod
-from collections import Counter, defaultdict
+from collections import Counter
 from functools import lru_cache
 from typing import Mapping, Optional, Tuple, Union, cast
 
@@ -524,33 +524,31 @@ class CoherentResults(SimulationResults):
         if self._meas_errors is None:
             return sampled_state
 
-        detected_sample_dict: defaultdict = defaultdict(int)
         eps = self._meas_errors["epsilon"]
         eps_p = self._meas_errors["epsilon_prime"]
 
         if eps_p == 0.0 and eps == 0.0:
             return sampled_state
 
-        for shot, n_detects in sampled_state.items():
-            # Shot as an array of 1s and 0s
-            shot_arr = np.array(list(shot), dtype=int)
+        shots = list(sampled_state.keys())
+        n_detects_list = list(sampled_state.values())
 
-            # Probability of flipping each bit
-            flip_probs = np.where(shot_arr == 1, eps_p, eps)
-
-            # 1 if it flips, 0 if it stays the same
-            flips = (
-                np.random.uniform(size=(n_detects, len(flip_probs)))
-                < flip_probs
-            )
-
-            # XOR betwen the original array and the flips
-            # Gives an array of n_detects individual shots
-            new_shots = shot_arr ^ flips
-
-            # Count all the new_shots
-            for x in new_shots:
-                detected_sample_dict[tuple(x)] += 1
+        # Convert shots to a 2D array
+        shot_arr = np.array([list(shot) for shot in shots], dtype=int)
+        # Compute flip probabilities
+        flip_probs = np.where(shot_arr == 1, eps_p, eps)
+        # Repeat flip_probs based on n_detects_list
+        flip_probs_repeated = np.repeat(flip_probs, n_detects_list, axis=0)
+        # Generate random matrix of shape (sum(n_detects_list), len(shot))
+        random_matrix = np.random.uniform(
+            size=(np.sum(n_detects_list), len(shot_arr[0]))
+        )
+        # Compare random matrix with flip probabilities
+        flips = random_matrix < flip_probs_repeated
+        # Perform XOR between original array and flips
+        new_shots = shot_arr.repeat(n_detects_list, axis=0) ^ flips
+        # Count all the new_shots
+        detected_sample_dict: Counter = Counter(map(tuple, new_shots))
 
         return Counter(
             {"".join(map(str, k)): v for k, v in detected_sample_dict.items()}
--- a/pulser-simulation/pulser_simulation/simresults.py
+++ b/pulser-simulation/pulser_simulation/simresults.py
@@ -525,21 +525,29 @@ class CoherentResults(SimulationResults):
             return sampled_state
 
         detected_sample_dict: Counter = Counter()
+        eps = self._meas_errors["epsilon"]
+        eps_p = self._meas_errors["epsilon_prime"]
+
         for shot, n_detects in sampled_state.items():
-            eps = self._meas_errors["epsilon"]
-            eps_p = self._meas_errors["epsilon_prime"]
             # Shot as an array of 1s and 0s
             shot_arr = np.array(list(shot), dtype=int)
-            # Probability of flipping each bit
-            flip_probs = np.array([eps_p if x == "1" else eps for x in shot])
-            # 1 if it flips, 0 if it stays the same
-            flips = (
-                np.random.uniform(size=(n_detects, len(flip_probs)))
-                < flip_probs
-            ).astype(int)
-            # XOR betwen the original array and the flips
-            # Gives an array of n_detects individual shots
-            new_shots = shot_arr ^ flips
+
+            if eps_p != 0.0 and eps != 0.0:
+                # Probability of flipping each bit
+                flip_probs = np.array(
+                    [eps_p if x == "1" else eps for x in shot]
+                )
+                # 1 if it flips, 0 if it stays the same
+                flips = (
+                    np.random.uniform(size=(n_detects, len(flip_probs)))
+                    < flip_probs
+                ).astype(int)
+                # XOR betwen the original array and the flips
+                # Gives an array of n_detects individual shots
+                new_shots = shot_arr ^ flips
+            else:
+                new_shots = shot_arr
+
             # Count all the new_shots
             detected_sample_dict += Counter(
                 "".join(map(str, measured)) for measured in new_shots
--- a/pulser-simulation/pulser_simulation/simresults.py
+++ b/pulser-simulation/pulser_simulation/simresults.py
@@ -18,7 +18,7 @@ from __future__ import annotations
 import collections.abc
 import typing
 from abc import ABC, abstractmethod
-from collections import Counter
+from collections import Counter, defaultdict
 from functools import lru_cache
 from typing import Mapping, Optional, Tuple, Union, cast
 
@@ -524,7 +524,7 @@ class CoherentResults(SimulationResults):
         if self._meas_errors is None:
             return sampled_state
 
-        detected_sample_dict: Counter = Counter()
+        detected_sample_dict: defaultdict = defaultdict(int)
         eps = self._meas_errors["epsilon"]
         eps_p = self._meas_errors["epsilon_prime"]
 
@@ -536,20 +536,22 @@ class CoherentResults(SimulationResults):
             shot_arr = np.array(list(shot), dtype=int)
 
             # Probability of flipping each bit
-            flip_probs = np.array([eps_p if x == "1" else eps for x in shot])
+            flip_probs = np.where(shot_arr == 1, eps_p, eps)
+
             # 1 if it flips, 0 if it stays the same
             flips = (
                 np.random.uniform(size=(n_detects, len(flip_probs)))
                 < flip_probs
-            ).astype(int)
+            )
+
             # XOR betwen the original array and the flips
             # Gives an array of n_detects individual shots
             new_shots = shot_arr ^ flips
 
             # Count all the new_shots
-            new_shots_count = [
-                "".join(map(str, measured)) for measured in new_shots
-            ]
-            detected_sample_dict.update(new_shots_count)
+            for x in new_shots:
+                detected_sample_dict[tuple(x)] += 1
 
-        return detected_sample_dict
+        return Counter(
+            {"".join(map(str, k)): v for k, v in detected_sample_dict.items()}
+        )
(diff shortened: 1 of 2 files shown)


HGSilveri (commented):
Good start, but I think the improvement from this alone will be quite small. The main goal is to avoid unecessarily repeating the time-consuming part where the `flips` are calculated

pulser-simulation/pulser_simulation/simresults.py
```diff
@@ -525,11 +525,16 @@ def sample_state(
             return sampled_state
 
         detected_sample_dict: Counter = Counter()
+        eps = self._meas_errors["epsilon"]
+        eps_p = self._meas_errors["epsilon_prime"]
+
+        if eps_p == 0.0 and eps == 0.0:
+            return sampled_state
+
```
a-corni, line 533: I think you can just add this condition to the condition line 524:
`if self._meas_errors is None and eps_p == 0.0 and eps == 0.0:`

dakk: @a-corni if _meas_errors is None, I don't have any eps_p / eps value; I could do something like this:

```python
if self._meas_errors is None or (self._meas_errors["epsilon"] == 0.0 and self._meas_errors["epsilon_prime"] == 0)
```

But IMO it is more readable as it is now

HGSilveri: But if `self._meas_errors is None` there is no `eps` and  `eps_p` though

HGSilveri: To avoid the duplicated return, I think it needs to be something like this:
```python
if self._meas_errors is None or (self._meas_errors["epsilon_p"] == 0.0 and self._meas_errors["epsilon"] == 0.0):
```

dakk: Do you mean like this? If you confirm I proceed with this edit.

```python
if self._meas_errors is None or (self._meas_errors["epsilon"] == 0.0 and self._meas_errors["epsilon_prime"] == 0)
```

@a-corni I requested HG review because was the only name showing in the request review list, feel free to take the task.

HGSilveri (commented):
This is already quite a nice optimization, but I think we can go further. I'll write up another comment detailing how

pulser-simulation/pulser_simulation/simresults.py
a-corni: Yes that's what I meant ;)

pulser-simulation/pulser_simulation/simresults.py
```diff
@@ -540,9 +545,11 @@ def sample_state(
...
             # Gives an array of n_detects individual shots
             new_shots = shot_arr ^ flips
+
             # Count all the new_shots
-            detected_sample_dict += Counter(
+            new_shots_count = [
                 "".join(map(str, measured)) for measured in new_shots
-            )
+            ]
+            detected_sample_dict.update(new_shots_count)
```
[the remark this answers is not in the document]
dakk: Indeed, I made a new optimization in the last commit, by using tuple instead of strings for sample_dict indexing. Please take a look.

(I removed the screenshot from the issue, since I noticed there was an error in my profiling script)

dakk:
> A question for you @dakk : How is the runtime scaling with `samples_per_run` now?

Now scales sub-linearly; this is a test run with different `samples_per_run`:
Samples_per_run: 1	Elapsed: 2.8112339973449707 seconds
Samples_per_run: 10	Elapsed: 3.293823003768921 seconds
Samples_per_run: 100	Elapsed: 5.302839040756226 seconds
Samples_per_run: 500	Elapsed: 11.655674457550049 seconds
Samples_per_run: 1000	Elapsed: 19.241939544677734 seconds
Samples_per_run: 2000	Elapsed: 36.630615234375 seconds
Samples_per_run: 5000	Elapsed: 84.35432386398315 seconds
Samples_per_run: 10000	Elapsed: 156.6097002029419 seconds

```python
import time
import sys
import numpy as np
from pulser import Register, Pulse, Sequence, waveforms, devices
from pulser_simulation import Simulation, SimConfig

reg = Register.square(side=2)
pulse = Pulse.ConstantDetuning(waveforms.BlackmanWaveform(duration=1000, area=np.pi),0,0)
seq = Sequence(reg, devices.MockDevice)
seq.declare_channel('ch', 'rydberg_global')
seq.add(pulse, 'ch')
sim = Simulation(seq)

def run(r):
        noise_config = SimConfig(noise=('SPAM', 'doppler', 'amplitude') if spam else ('doppler', 'amplitude'),
                             eta=0.005, epsilon=0.03, epsilon_prime=0.08,
                             temperature=30, laser_waist=148,
                             runs=10, samples_per_run=r)
        sim.set_config(noise_config)
        #sim.show_config()
        t = time.time()
        res = sim.run(False)
        tt = time.time() - t
        print(f'Samples_per_run: {r}\tElapsed: {tt} seconds')

for x in [1, 10, 100, 500, 1000, 2000, 5000, 10000]:
        run(x)
```

HGSilveri:
Ok, here is the alternative that I think it's worth investigating: when `CoherentResults` has measurement error information, `CoherentResults._calc_pseudo_density()` generates a pseudo-density matrix (diagonal only) where the measurement errors are **included**. This means that if the `_weights` used in `Result.get_samples()` came from the diagonal of this pseudo-density matrix instead, there would be no need to post-process the samples to include the bitflips.

The potential downside of this approach is that calculating the density matrix is expensive, as you go from a state of size N to a matrix of size N**2, so I'm not sure it will be benefitial. I have a feeling this is the conclusion I arrived at a while ago, but I can't remember with certainty anymore. On the other hand, I suspect that calculating the full pseudo-density matrix might even not be necessary as we are only using the N elements in its diagonal.

HGSilveri:
Let's take this step by step:

1. Isolate the contents of `Result.get_sample()` in a staticmethod so we can use it:
```python
    def get_samples(self, n_samples: int) -> Counter[str]:
        """Takes multiple samples from the sampling distribution.

        Args:
            n_samples: Number of samples to return.

        Returns:
            Samples of bitstrings corresponding to measured quantum states.
        """
        return self.get_samples_from_weights(n_samples, self._weights())

    @staticmethod
    def get_samples_from_weights(n_samples: int, weights: np.ndarray):
        # TODO: Docstring
        dist = np.random.multinomial(n_samples, weights)
        return Counter(
            {
                np.binary_repr(i, self._size): dist[i]
                for i in np.nonzero(dist)[0]
            }
        )
```

2. Modifiy `CoherentResults.sample_state()` to use the weights from the pseudo-density matrix when there are measurement errors:
```python
        ...
        if eps_p == 0.0 and eps == 0.0:
            return sampled_state
        t_index = self._get_index_from_time(t, t_tol)
        custom_weights = np.abs(self._calc_pseudo_density(t_index).diag())
        return Result.get_samples_from_weights(n_samples, custom_weights)
```
 
3. Compare the time it takes to run this solution with your current proposal. Try for different values of `samples_per_run` and also number of atoms (by changing the size of the register).

My guess is that there will be regimes where this performs better (ie large number of `samples_per_run` and or low number of atoms) and others where it will perform worse due to the cost of calculating the pseudo-density matrices. I think avoiding the pseudo-density matrix calculation is not straightforward at all, so let's drop that and stick to this.

PS: It's very likely that there are bugs in the code above, I just wanted to give some pointers so feel free to adjust it as you see fit

dakk:
> Let's take this step by step:
> 
>     1. Isolate the contents of `Result.get_samples()` in a staticmethod so we can use it:
> 
> 
> ```python
>     def get_samples(self, n_samples: int) -> Counter[str]:
>         """Takes multiple samples from the sampling distribution.
> 
>         Args:
>             n_samples: Number of samples to return.
> 
>         Returns:
>             Samples of bitstrings corresponding to measured quantum states.
>         """
>         return self.get_samples_from_weights(n_samples, self._weights())
> 
>     @staticmethod
>     def get_samples_from_weights(n_samples: int, weights: np.ndarray):
>         # TODO: Docstring
>         dist = np.random.multinomial(n_samples, weights)
>         return Counter(
>             {
>                 np.binary_repr(i, self._size): dist[i]
>                 for i in np.nonzero(dist)[0]
>             }
>         )
> ```
> 
>     2. Modifiy `CoherentResults.sample_state()` to use the weights from the pseudo-density matrix when there are measurement errors:
> 
> 
> ```python
>         ...
>         if eps_p == 0.0 and eps == 0.0:
>             return sampled_state
>         t_index = self._get_index_from_time(t, t_tol)
>         custom_weights = np.abs(self._calc_pseudo_density(t_index).diag())
>         return Result.get_samples_from_weights(n_samples, custom_weights)
> ```
> 
>     3. Compare the time it takes to run this solution with your current proposal. Try for different values of `samples_per_run` and also number of atoms (by changing the size of the register).
> 
> 
> My guess is that there will be regimes where this performs better (ie large number of `samples_per_run` and or low number of atoms) and others where it will perform worse due to the cost of calculating the pseudo-density matrices. I think avoiding the pseudo-density matrix calculation is not straightforward at all, so let's drop that and stick to this.
> 
> PS: It's very likely that there are bugs in the code above, I just wanted to give some pointers so feel free to adjust it as you see fit

I just integrated your code into Pulsar, but it takes way more time to run simulations; for way more I mean several minutes instead of seconds even with the smallest samples_per_run. Maybe I'm missing something

https://github.com/dakk/Pulser/commit/8c5f519bf6fe25f30f9d4bc256a42fff87d7109b

HGSilveri:
Your implementation seems correct, it's very likely that `_calc_pseudo_density()` is just prohibitively expensive.  Let's drop that idea and go with your solution. Could you also post the elapsed time you were getting before your changes (for comparison sake)?

dakk:
> Your implementation seems correct, it's very likely that `_calc_pseudo_density()` is just prohibitively expensive. Let's drop that idea and go with your solution. Could you also post the elapsed time you were getting before your changes (for comparison sake)?

Sure; this is before:

```
Samples_per_run: 1	Elapsed: 2.9893434047698975 seconds
Samples_per_run: 10	Elapsed: 3.8784408569335938 seconds
Samples_per_run: 100	Elapsed: 8.337844133377075 seconds
Samples_per_run: 500	Elapsed: 23.297382354736328 seconds
Samples_per_run: 1000	Elapsed: 44.66521739959717 seconds
Samples_per_run: 2000	Elapsed: 80.20746278762817 seconds
Samples_per_run: 5000	Elapsed: 177.2137794494629 seconds
Samples_per_run: 10000	Elapsed: 367.4108419418335 seconds
```

and after:
```
Samples_per_run: 1	Elapsed: 3.863002061843872 seconds
Samples_per_run: 10	Elapsed: 4.330973863601685 seconds
Samples_per_run: 100	Elapsed: 6.834812164306641 seconds
Samples_per_run: 500	Elapsed: 14.479618549346924 seconds
Samples_per_run: 1000	Elapsed: 23.99378752708435 seconds
Samples_per_run: 2000	Elapsed: 42.217456579208374 seconds
Samples_per_run: 5000	Elapsed: 100.99924063682556 seconds
Samples_per_run: 10000	Elapsed: 203.1134431362152 seconds
```

pulser-simulation/pulser_simulation/simresults.py
```diff
@@ -524,25 +524,34 @@ def sample_state(
...
             # 1 if it flips, 0 if it stays the same
             flips = (
                 np.random.uniform(size=(n_detects, len(flip_probs)))
                 < flip_probs
-            ).astype(int)
+            )
+
             # XOR betwen the original array and the flips
             # Gives an array of n_detects individual shots
             new_shots = shot_arr ^ flips
```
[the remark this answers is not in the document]
a-corni: Note: I don't think we really care about the order of the lines in `shot_matrix`, it is perhaps easier to build a matrix of containing `n_detect` times the list of `shots`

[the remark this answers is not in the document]
dakk: > I think this code can be vectorized. To me, what costs a lot is the multiple generation of a random matrix (one per instance in the for loop), and its comparison to compute `flips` and `new_shots`.
> 
> My idea is to create a `shot_matrix` that would be composed of `len(shot)*n_detect` lines and `len(shot)` columns (a matrix composed of `n_detect` times the array `shot` for each `shot` in `sampled_state.keys()`) `flip_probs` then becomes a matrix. `flips` is created by comparing a random matrix of size `(len(shot)*n_detect, len(shot))` with `flip_probs`. `new_shots` becomes a matrix and last for loop around `new_shots` is unchanged.
> 
> To perform this vectorization, it might be faster to use exclusively numpy functions (`np.bitwise_xor` is eventually faster than `^`, and combining `np.bincount` and `np.array2string` seems more relevant to build the final Counter than using `dectected_sample_dict`)

I can make a try, the performance will increase (not so much IMO): in addition will require more ram.  I don't understand the last part of `np.bincount` / `np.array2string` (btw ^ is a short-hand for `np.bitwise_xor`), can you explain further?

EDIT: I implemented it

```python
        shots = list(sampled_state.keys())
        n_detects_list = list(sampled_state.values())

        # Convert shots to a 2D array
        shot_arr = np.array([list(shot) for shot in shots], dtype=int)
        # Compute flip probabilities
        flip_probs = np.where(shot_arr == 1, eps_p, eps)
        # Repeat flip_probs based on n_detects_list
        flip_probs_repeated = np.repeat(flip_probs, n_detects_list, axis=0)
        # Generate random matrix of shape (sum(n_detects_list), len(shot))
        random_matrix = np.random.uniform(size=(np.sum(n_detects_list), len(shot_arr[0])))
        # Compare random matrix with flip probabilities
        flips = random_matrix < flip_probs_repeated
        # Perform XOR between original array and flips
        new_shots = shot_arr.repeat(n_detects_list, axis=0) ^ flips
        # Count all the new_shots
        detected_sample_dict = Counter(map(tuple, new_shots))
        
        return Counter(
            {"".join(map(str, k)): v for k, v in detected_sample_dict.items()}
        )
```

Performance are improved on increasing samples_per_run, but it increases also the RAM usage. 

Before:
```
Samples_per_run: 1	Elapsed: 3.863002061843872 seconds
Samples_per_run: 10	Elapsed: 4.330973863601685 seconds
Samples_per_run: 100	Elapsed: 6.834812164306641 seconds
Samples_per_run: 500	Elapsed: 14.479618549346924 seconds
Samples_per_run: 1000	Elapsed: 23.99378752708435 seconds
Samples_per_run: 2000	Elapsed: 42.217456579208374 seconds
Samples_per_run: 5000	Elapsed: 100.99924063682556 seconds
Samples_per_run: 10000	Elapsed: 203.1134431362152 seconds
```

After:
```
Samples_per_run: 1	Elapsed: 4.041079044342041 seconds
Samples_per_run: 10	Elapsed: 4.325520038604736 seconds
Samples_per_run: 100	Elapsed: 6.078380107879639 seconds
Samples_per_run: 500	Elapsed: 13.140138626098633 seconds
Samples_per_run: 1000	Elapsed: 21.684226989746094 seconds
Samples_per_run: 2000	Elapsed: 38.53966021537781 seconds
Samples_per_run: 5000	Elapsed: 90.3535590171814 seconds
Samples_per_run: 10000	Elapsed: 169.49966073036194 seconds
```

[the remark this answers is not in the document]
a-corni: Thanks for having a look to it ! Looks nice to me, what about you @HGSilveri ? I don't see any issue in using more RAM for a little speedup, but perhaps the tradeoff is not interesting.
Forget about the bincount/array2string thing, `Counter` is what you want ;)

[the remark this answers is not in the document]
HGSilveri: It's faster and pretty well documented, so I don't have objections at first glance.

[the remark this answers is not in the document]
dakk: Yep, just pushed!

pulser-simulation/pulser_simulation/simresults.py
```diff
@@ -524,25 +524,32 @@ def sample_state(
...
+        # Compare random matrix with flip probabilities
+        flips = random_matrix < flip_probs_repeated
+        # Perform XOR between original array and flips
+        new_shots = shot_arr.repeat(n_detects_list, axis=0) ^ flips
+        # Count all the new_shots
+        detected_sample_dict: Counter = Counter(map(tuple, new_shots))
+
+        return Counter(
+            {"".join(map(str, k)): v for k, v in detected_sample_dict.items()}
+        )
```
HGSilveri, line 555: Is this preferable over simply converting directly to an iterable of strings and giving it to `Counter`?

pulser-simulation/pulser_simulation/simresults.py
HGSilveri: Don't forget about this :)

pulser-simulation/pulser_simulation/simresults.py
HGSilveri: Something like this:
```suggestion
        return Counter(map("".join, new_shots))
```

dakk: I did this because it is more efficient; this is with your code modified like this:

```python
return Counter(map(lambda x: "".join(map(str, x)), new_shots))
```

```
Samples_per_run: 10	Elapsed: 4.808044672012329 seconds
Samples_per_run: 100	Elapsed: 9.669185876846313 seconds
Samples_per_run: 500	Elapsed: 33.64236116409302 seconds
```

Instead of :
```
Samples_per_run: 1	Elapsed: 4.041079044342041 seconds
Samples_per_run: 10	Elapsed: 4.325520038604736 seconds
Samples_per_run: 100	Elapsed: 6.078380107879639 seconds
Samples_per_run: 500	Elapsed: 13.140138626098633 seconds
```

dakk: > Right, you have to convert to `str`, I forgot about that. What about (this is my final suggestion, I promise!):

Bad as well; the bootleneck here is the indexing: tuple indexing is faster than str one (don't worry :D)

```
Samples_per_run: 10	Elapsed: 5.543439865112305 seconds
Samples_per_run: 100	Elapsed: 12.317859172821045 seconds
Samples_per_run: 500	Elapsed: 35.426414489746094 seconds
```

HGSilveri: Ok, you win 😅 
Maybe just leave a comment explaining why you did it like this because I think at first glance it looks redundant

HGSilveri (approved):
Very nice, thank your for your work!

PR #530, opened 2023-06-01
Adding effective channel noises to digital basis

### LINKED ISSUE #496: Adding effective channel noises to digital basis
Currently, effective channel noises are implemented only for the Rydberg basis. A `NotImplementedError` is raised when a simulation is ran with a `Simconfig` having `detuning` or `dephasing` noise.
Implementing this feature would have an impact on simresults: the reduction of a density matrix to a basis is not implemented and cannot be tested because of the fact that no simulation can be performed with dimensions 3. This should be implemented using a [partial_trace](https://physics.stackexchange.com/questions/179671/how-to-take-partial-trace)

Aaron-Robertson:
Fixes #496

### DIFF
--- a/pulser-core/pulser/backend/noise_model.py
+++ b/pulser-core/pulser/backend/noise_model.py
@@ -134,12 +134,6 @@ class NoiseModel:
                     + "Valid noise types: "
                     + ", ".join(get_args(NOISE_TYPES))
                 )
-        dephasing_on = "dephasing" in self.noise_types
-        depolarizing_on = "depolarizing" in self.noise_types
-        eff_noise_on = "eff_noise" in self.noise_types
-        eff_noise_conflict = dephasing_on + depolarizing_on + eff_noise_on > 1
-        if eff_noise_conflict:
-            print(eff_noise_conflict)
 
     def _check_eff_noise(self) -> None:
         if len(self.eff_noise_opers) != len(self.eff_noise_probs):
@@ -185,20 +179,31 @@ class NoiseModel:
                     f"not {operator.shape}."
                 )
         # Identity position
-        dim = len(self.noise_types)
-        identity = np.eye(dim)
-        if not np.allclose(self.eff_noise_opers[0], np.eye(dim)):
+        identity2 = np.eye(2)
+        identity3 = np.eye(3)
+        if not (np.array_equal(self.eff_noise_opers[0], identity2) or np.array_equal(self.eff_noise_opers[0], identity3)):
             raise NotImplementedError(
                 "You must put the identity matrix at the "
                 "beginning of the operator list."
             )
         # Completeness relation checking
-        sum_op = np.zeros((dim, dim), dtype=complex)
-        for prob, op in zip(self.eff_noise_probs, self.eff_noise_opers):
-            sum_op += prob * op @ op.conj().transpose()
+        elif not np.array_equal(self.eff_noise_opers[0], identity2):
+            sum_op = np.zeros((2, 2), dtype=complex)
+            for prob, op in zip(self.eff_noise_probs, self.eff_noise_opers):
+                sum_op += prob * op @ op.conj().transpose()
 
-        if not np.all(np.isclose(sum_op, identity)):
-            raise ValueError(
-                "The completeness relation is not verified."
-                f" Ended up with {sum_op} instead of {identity}."
-            )
+            if not np.all(np.isclose(sum_op, identity2)):
+                raise ValueError(
+                    "The completeness relation is not verified."
+                    f" Ended up with {sum_op} instead of {identity2}."
+                )
+        else:
+            sum_op = np.zeros((3, 3), dtype=complex)
+            for prob, op in zip(self.eff_noise_probs, self.eff_noise_opers):
+                sum_op += prob * op @ op.conj().transpose()
+
+            if not np.all(np.isclose(sum_op, identity3)):
+                raise ValueError(
+                    "The completeness relation is not verified."
+                    f" Ended up with {sum_op} instead of {identity2}."
+                )
--- a/pulser-core/pulser/backend/noise_model.py
+++ b/pulser-core/pulser/backend/noise_model.py
@@ -134,6 +134,15 @@ class NoiseModel:
                     + "Valid noise types: "
                     + ", ".join(get_args(NOISE_TYPES))
                 )
+        dephasing_on = "dephasing" in self.noise_types
+        depolarizing_on = "depolarizing" in self.noise_types
+        eff_noise_on = "eff_noise" in self.noise_types
+        eff_noise_conflict = dephasing_on + depolarizing_on + eff_noise_on > 2
+        if eff_noise_conflict:
+            raise NotImplementedError(
+                "Depolarizing, dephasing and effective noise channels "
+                "cannot be simultaneously selected."
+            )
 
     def _check_eff_noise(self) -> None:
         if len(self.eff_noise_opers) != len(self.eff_noise_probs):
@@ -202,5 +211,5 @@ class NoiseModel:
         if not np.all(np.isclose(sum_op, identity[size])):
             raise ValueError(
                 "The completeness relation is not verified."
-                f" Ended up with {sum_op} instead of {identity[size]}."
+                f" Ended up with {sum_op.real} instead of {identity[size]}."
             )
--- a/pulser-core/pulser/backend/noise_model.py
+++ b/pulser-core/pulser/backend/noise_model.py
@@ -178,35 +178,26 @@ class NoiseModel:
                     "Operator's shape must be (2,2) or (3,3), "
                     f"not {operator.shape}."
                 )
-        # Identity position
-        identity2 = np.eye(2)
-        identity3 = np.eye(3)
-        if not (
-            np.array_equal(self.eff_noise_opers[0], identity2)
-            or np.array_equal(self.eff_noise_opers[0], identity3)
-        ):
+        # Identity matrices
+        identity = {2: np.eye(2), 3: np.eye(3)}
+
+        # Check if the first operator is an identity matrix
+        if not any(np.array_equal(self.eff_noise_opers[0], identity[i]) for i in identity):
             raise NotImplementedError(
                 "You must put the identity matrix at the "
                 "beginning of the operator list."
             )
+
+        # Determine the size of the identity matrix
+        size = len(self.eff_noise_opers[0])
+
         # Completeness relation checking
-        elif not np.array_equal(self.eff_noise_opers[0], identity2):
-            sum_op = np.zeros((2, 2), dtype=complex)
-            for prob, op in zip(self.eff_noise_probs, self.eff_noise_opers):
-                sum_op += prob * op @ op.conj().transpose()
+        sum_op = np.zeros((size, size), dtype=complex)
+        for prob, op in zip(self.eff_noise_probs, self.eff_noise_opers):
+            sum_op += prob * op @ op.conj().transpose()
 
-            if not np.all(np.isclose(sum_op, identity2)):
-                raise ValueError(
-                    "The completeness relation is not verified."
-                    f" Ended up with {sum_op} instead of {identity2}."
-                )
-        else:
-            sum_op = np.zeros((3, 3), dtype=complex)
-            for prob, op in zip(self.eff_noise_probs, self.eff_noise_opers):
-                sum_op += prob * op @ op.conj().transpose()
-
-            if not np.all(np.isclose(sum_op, identity3)):
-                raise ValueError(
-                    "The completeness relation is not verified."
-                    f" Ended up with {sum_op} instead of {identity2}."
-                )
+        if not np.all(np.isclose(sum_op, identity[size])):
+            raise ValueError(
+                "The completeness relation is not verified."
+                f" Ended up with {sum_op} instead of {identity[size]}."
+            )
--- a/pulser-core/pulser/backend/noise_model.py
+++ b/pulser-core/pulser/backend/noise_model.py
@@ -139,10 +139,7 @@ class NoiseModel:
         eff_noise_on = "eff_noise" in self.noise_types
         eff_noise_conflict = dephasing_on + depolarizing_on + eff_noise_on > 1
         if eff_noise_conflict:
-            raise NotImplementedError(
-                "Depolarizing, dephasing and effective noise channels"
-                "cannot be simultaneously selected."
-            )
+            print(eff_noise_conflict)
 
     def _check_eff_noise(self) -> None:
         if len(self.eff_noise_opers) != len(self.eff_noise_probs):
@@ -182,19 +179,21 @@ class NoiseModel:
             # type checking
             if not isinstance(operator, np.ndarray):
                 raise TypeError(f"{operator} is not a Numpy array.")
-            if operator.shape != (2, 2):
+            if operator.shape not in [(2, 2), (3, 3)]:
                 raise NotImplementedError(
-                    "Operator's shape must be (2,2) " f"not {operator.shape}."
+                    "Operator's shape must be (2,2) or (3,3), "
+                    f"not {operator.shape}."
                 )
         # Identity position
-        identity = np.eye(2)
-        if np.any(self.eff_noise_opers[0] != identity):
+        dim = len(self.noise_types)
+        identity = np.eye(dim)
+        if not np.allclose(self.eff_noise_opers[0], np.eye(dim)):
             raise NotImplementedError(
                 "You must put the identity matrix at the "
                 "beginning of the operator list."
             )
         # Completeness relation checking
-        sum_op = np.zeros((2, 2), dtype=complex)
+        sum_op = np.zeros((dim, dim), dtype=complex)
         for prob, op in zip(self.eff_noise_probs, self.eff_noise_opers):
             sum_op += prob * op @ op.conj().transpose()
 
--- a/pulser-core/pulser/backend/noise_model.py
+++ b/pulser-core/pulser/backend/noise_model.py
@@ -182,7 +182,10 @@ class NoiseModel:
         identity = {2: np.eye(2), 3: np.eye(3)}
 
         # Check if the first operator is an identity matrix
-        if not any(np.array_equal(self.eff_noise_opers[0], identity[i]) for i in identity):
+        if not any(
+            np.array_equal(self.eff_noise_opers[0], identity[i])
+            for i in identity
+        ):
             raise NotImplementedError(
                 "You must put the identity matrix at the "
                 "beginning of the operator list."
--- a/pulser-core/pulser/backend/noise_model.py
+++ b/pulser-core/pulser/backend/noise_model.py
@@ -137,7 +137,7 @@ class NoiseModel:
         dephasing_on = "dephasing" in self.noise_types
         depolarizing_on = "depolarizing" in self.noise_types
         eff_noise_on = "eff_noise" in self.noise_types
-        eff_noise_conflict = dephasing_on + depolarizing_on + eff_noise_on > 2
+        eff_noise_conflict = dephasing_on + depolarizing_on + eff_noise_on > 1
         if eff_noise_conflict:
             raise NotImplementedError(
                 "Depolarizing, dephasing and effective noise channels "
--- a/pulser-core/pulser/backend/noise_model.py
+++ b/pulser-core/pulser/backend/noise_model.py
@@ -181,7 +181,10 @@ class NoiseModel:
         # Identity position
         identity2 = np.eye(2)
         identity3 = np.eye(3)
-        if not (np.array_equal(self.eff_noise_opers[0], identity2) or np.array_equal(self.eff_noise_opers[0], identity3)):
+        if not (
+            np.array_equal(self.eff_noise_opers[0], identity2)
+            or np.array_equal(self.eff_noise_opers[0], identity3)
+        ):
             raise NotImplementedError(
                 "You must put the identity matrix at the "
                 "beginning of the operator list."
--- a/pulser-simulation/pulser_simulation/qutip_result.py
+++ b/pulser-simulation/pulser_simulation/qutip_result.py
@@ -15,7 +15,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import Union, cast
+from typing import Union, cast, List
 
 import numpy as np
 import qutip
@@ -125,6 +125,69 @@ class QutipResult(Result):
             )
         # Takes care of numerical artefacts in case sum(weights) != 1
         return cast(np.ndarray, weights / sum(weights))
+    
+    def _get_relevant_subsystem_indices(self, to_basis: str, tol: float) -> List[int]:
+        """Finds the subsystems to keep based on the target basis.
+
+        Args:
+            to_basis: The basis to which the state should be transformed.
+
+        Returns:
+            A list of indices representing the relevant subsystems.
+        """
+        if to_basis == "ground-rydberg":
+            ex_state = "2"
+        elif to_basis == "digital":
+            ex_state = "0"
+        else:
+            raise ValueError("'to_basis' must be 'ground-rydberg' or 'digital'.")
+
+        # Find the subsystems to keep.
+        ex_inds = [
+            i
+            for i in range(3**self._size)
+            if ex_state in np.base_repr(i, base=3).zfill(self._size)
+        ]
+
+        ex_probs = np.abs(self.state.extract_states(ex_inds).full()) ** 2
+        if not np.all(np.isclose(ex_probs, 0, atol=tol)):
+            raise TypeError(
+                "Can't reduce to chosen basis because the population of a "
+                "state to eliminate is above the allowed tolerance."
+            )
+
+        return ex_inds
+
+    def _reduce_state_vector_to_basis(self, ex_inds: List[int], normalize: bool) -> qutip.Qobj:
+        """Reduces the state vector by eliminating the states at given indices.
+
+        Args:
+            ex_inds: The indices of the states to eliminate.
+
+        Returns:
+            A Qobj instance representing the state vector with the states eliminated.
+        """
+        state = self.state.copy()
+        state = state.eliminate_states(ex_inds, normalize=normalize)
+        return state
+    
+    def _reduce_density_matrix_to_basis(self, ex_inds: List[int], normalize: bool) -> qutip.Qobj:
+        """Reduces the density matrix by tracing out the subsystems at given indices.
+
+        Args:
+            ex_inds: The indices of the subsystems to trace out.
+
+        Returns:
+            The subsystem state, reduced by tracing out the subsystems at given indices.
+        """
+        state = self.state.copy()
+        reduced_state = state.ptrace(ex_inds)
+        
+        # Normalize the density matrix by ensuring that its trace equals 1.
+        if normalize:
+            reduced_state = reduced_state / reduced_state.tr()
+
+        return reduced_state
 
     def get_state(
         self,
@@ -133,26 +196,6 @@ class QutipResult(Result):
         tol: float = 1e-6,
         normalize: bool = True,
     ) -> qutip.Qobj:
-        """Gets the state with some optional post-processing.
-
-        Args:
-            reduce_to_basis: Reduces the full state vector
-                to the given basis ("ground-rydberg" or "digital"), if the
-                population of the states to be ignored is negligible. Doesn't
-                apply to XY mode.
-            ignore_global_phase: If True and if the final state is a vector,
-                changes the final state's global phase such that the largest
-                term (in absolute value) is real.
-            tol: Maximum allowed population of each eliminated state.
-            normalize: Whether to normalize the reduced state.
-
-        Returns:
-            The resulting state.
-
-        Raises:
-            TypeError: If trying to reduce to a basis that would eliminate
-                states with significant occupation probabilites.
-        """
         state = self.state.copy()
         is_density_matrix = state.isoper
         if ignore_global_phase and not is_density_matrix:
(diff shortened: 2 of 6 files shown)


Aaron-Robertson:
@HGSilveri I still need to ensure my tests (the local ones) are robust, but I think it's much closer than I expected to get on my own! I'd appreciate any feedback, but otherwise I'll continue with test refinement, remove the now obsolete tests, and 🤞 may be able to wrap tomorrow.

pulser-simulation/pulser_simulation/qutip_result.py
```diff
@@ -126,33 +126,84 @@ def _weights(self) -> np.ndarray:
...
+        """Reduces the density matrix by tracing out the subsystems at given indices.
+
+        Args:
+            ex_inds: The indices of the subsystems to trace out.
+
+        Returns:
+            The subsystem state, reduced by tracing out the subsystems at given indices.
         """
+        state = self.state.copy()
+        reduced_state = state.ptrace(ex_inds)
```
HGSilveri, line 192: Two things here:
1. I think that `ptrace()` **keeps** the provided indices, rather than exclude them.
2. This does not have the same behaviour as `eliminate_states()`. The partial trace removes part of a multipartite system (eg if you have a two-qubit system and want the state of only one) while `eliminate_states()` will remove specifc indices (eg turning a qutrit into a qubit). Also, it seems that you can use `eliminate_states()` on density matrices too, so there's no need to use the partial trace here.

HGSilveri (commented):
It's a good start, but there's still quite a way to go. I suggest you familiarize yourself with noise channels and Kraus operators first.

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -619,6 +623,18 @@ def _update_noise(self) -> None:
...
+                0, self.config.depolarizing_prob, size=len(self._qid_index)
+            )
+            self._depolarizing_noise = dict(
+                zip(self._qid_index, depolarizing_error)
+            )
+        if "dephasing" in self.config.noise:
+            dephasing_error = np.random.normal(
+                0, self.config.dephasing_prob, size=len(self._qid_index)
+            )
+            self._dephasing_noise = dict(zip(self._qid_index, dephasing_error))
```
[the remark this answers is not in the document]
Aaron-Robertson: Great, thanks @HGSilveri! Pushed up another attempt, this time making use of the `_collapse_ops`. Please let me know if I'm going wrong anywhere else, or if it isn't what you had in mind!

pulser-simulation/pulser_simulation/qutip_result.py
```diff
@@ -126,33 +126,64 @@ def _weights(self) -> np.ndarray:
...
+
+        Args:
+            ex_inds: The indices of the states to eliminate.
+
+        Returns:
+            A Qobj instance representing the state vector with the states eliminated.
+        """
+        state = self.state.copy()
+        state = state.eliminate_states(ex_inds, normalize=normalize)
+        return state
```
HGSilveri, line 178: I'd say this is probably too simple to require being isolated now, I'd move it back to `get_state()`. Also, I think the object is not changed in place so there's no need to copy it

pulser-simulation/pulser_simulation/qutip_result.py
```diff
@@ -126,33 +126,64 @@ def _weights(self) -> np.ndarray:
...
-            tol: Maximum allowed population of each eliminated state.
-            normalize: Whether to normalize the reduced state.
-
-        Returns:
-            The resulting state.
-
-        Raises:
-            TypeError: If trying to reduce to a basis that would eliminate
-                states with significant occupation probabilites.
-        """
```
HGSilveri, line 155: I'd say we want to keep this docstring, any reason why you deleted it?

pulser-simulation/pulser_simulation/qutip_result.py
```diff
@@ -126,33 +126,64 @@ def _weights(self) -> np.ndarray:
         # Takes care of numerical artefacts in case sum(weights) != 1
         return cast(np.ndarray, weights / sum(weights))
 
+    def _get_relevant_subsystem_indices(
+        self, to_basis: str, tol: float
+    ) -> List[int]:
```
HGSilveri, line 131: We actually don't need `List` here, we can (and should since python 3.9 was released) use `list` instead

pulser-simulation/pulser_simulation/qutip_result.py
```diff
@@ -126,33 +126,64 @@ def _weights(self) -> np.ndarray:
         # Takes care of numerical artefacts in case sum(weights) != 1
         return cast(np.ndarray, weights / sum(weights))
 
+    def _get_relevant_subsystem_indices(
```
HGSilveri, line 129: I'm not sure any of this refactoring is needed at all, since it turns out the density matrices supported `eliminate_states()` from the start.

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -524,6 +504,21 @@ def add_noise(
...
+                    ] *= dephasing_noise
+                if "depolarizing" in self.config.noise:
+                    # Add depolarizing noise
+                    depolarizing_noise = self._depolarizing_noise[qid]
+                    samples_dict[qid]["amp"][
+                        slot.ti : slot.tf
+                    ] *= depolarizing_noise
+                    samples_dict[qid]["det"][
+                        slot.ti : slot.tf
+                    ] *= depolarizing_noise
```
HGSilveri, line 521: Maybe you forgot to delete this?

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -620,6 +615,43 @@ def _update_noise(self) -> None:
...
+                    "The dephasing model is a first-order approximation in the"
+                    f" dephasing probability. p = {2*prob} is too large for "
+                    "realistic results.",
+                    stacklevel=2,
+                )
+            k = np.sqrt(prob * (1 - prob) ** (n - 1))
+            self._collapse_ops[
+                idx_dephasing
+            ] *= k  # Update Z collapse operator
+
```
HGSilveri, line 654: I'm struggling to understand what you are trying here, but I can give you a few points as to why it's not the way to go:

1. `_update_noise()` is reserved for noises that rely on some form of random distrbution sampling (which is not the case for the collapse operators)
2. `self._collapse_ops` stores `qutip.Qobj`s, so I wouldn't rely on their `str()` like I see you did here.
3. I identify substantial code duplication here that I'm pretty sure can be avoided

I'm curious, did you test this? Did it work?

Aaron-Robertson: Ah, I see, I interpreted the first review incorrectly, along with the requirements. It didn't work functionally, to do what we want (though I see now that I missed a save in the commit that made it fail entirely), but I started having doubts so I'm glad I checked in.

The latest comment certainly clarifies my confusion, so thanks! One thing of note, is that my initial approach added to the confusion. The first thing I did was remove the `NotImplementedError` blocks and apply the state elimination to the `density_matrix` (for which all tests pass), but that didn't seem to make sense in the context of the requirements.

HGSilveri: Ok, that makes sense! I was a bit puzzled so thanks for clarifying :)

HGSilveri:
@Aaron-Robertson if you have questions regarding my attempted clarification in #496, feel free to ask! 

Also, in case you feel like you misjudged the difficulty of this issue and now feel at a loss, I'd completely understand if you think it's not worth the time investment anymore. If this is not the case, then I'll be here to help out along the way, I leave that decision entirely up to you. Just be aware that it might require quite a bit of learning and research from your part, I feel like I owe you this disclaimer.

HGSilveri:
> > If this is not the case, then I'll be here to help out along the way, I leave that decision entirely up to you. Just be aware that it might require quite a bit of learning and research from your part, I owe you this disclaimer.
> 
> As long as you're willing to help out, then I'm here to learn! Nothing better than being wrong on repeat to achieve that goal 🤯 (for me at least).
> 
> For now I'm looking into potential representations of the three-level noise, and learning with an attempt to implement the Kraus operators according to the generalized Pauli representation.

Awesome! Feel free to post what you find before moving on to the implementation

Aaron-Robertson:
Alright @HGSilveri, that clarification yesterday really helped, so thanks! Let me know what you think of using the spin-1 matrices, or if I'm still off base. Based on this work[[1](https://arxiv.org/abs/0806.1174)][[2](https://arxiv.org/abs/2011.04672)], here's what I'm thinking as far as the crux:
```
def _build_basis_and_op_matrices(self) -> None:
    """Determine dimension, basis and projector operators."""
    if self._interaction == "XY":
        self.basis_name = "XY"
        self.dim = 2
        basis = ["u", "d"]
        projectors = ["uu", "du", "ud", "dd"]
    else:
        if "digital" not in self.samples_obj.used_bases:
            self.basis_name = "ground-rydberg"
            self.dim = 2
            basis = ["r", "g"]
            projectors = ["gr", "rr", "gg"]
        elif "ground-rydberg" not in self.samples_obj.used_bases:
            self.basis_name = "digital"
            self.dim = 2
            basis = ["g", "h"]
            projectors = ["hg", "hh", "gg"]
        else:
            self.basis_name = "all"  # All three states
            self.dim = 3
            basis = ["r", "g", "h"]
            projectors = ["gr", "hg", "rr", "gg", "hh"]

    self.basis = {b: qutip.basis(self.dim, i) for i, b in enumerate(basis)}

    # Setting up the op_matrix with identity and generalized Pauli operators for a qutrit
    self.op_matrix = {"I": qutip.qeye(self.dim)}

    if self.dim == 3:  # if we are working with a qutrit system
        jx, jy, jz = [0.5 * qutip.jmat(1, basis) for basis in "xyz"]
        self.op_matrix.update({"Jx": jx, "Jy": jy, "Jz": jz})
        self.op_matrix.update({f"J{i}2": op*op for i, op in zip("xyz", [jx, jy, jz])})
        self.op_matrix.update({"Jxy": jx * jy + jy * jx, "Jxz": jx * jz + jz * jx, "Jyz": jy * jz + jz * jy})

    for proj in projectors:
        self.op_matrix["sigma_" + proj] = (
            self.basis[proj[0]] * self.basis[proj[1]].dag()
        )

```

HGSilveri:
Hi @Aaron-Robertson , I'm sorry, I somehow missed your last message. I'm taking a look right now.

HGSilveri:
@Aaron-Robertson I'm not sure I follow your reasoning behind the generation of the Kraus operators from the Gell-Mann matrices, do you have a reference for this? 
Speaking of reference, I found this article: https://arxiv.org/abs/1407.6804 
In it, they define the Kraus operators for dephasing and depolarizing noise in qutrits. Do they match yours? It seems to me that this is all we need, do you agree?

Aaron-Robertson:
Alright @HGSilveri, ready for a look! Only thing left is to change the `NotImplemented` tests, but for now they're showing results.

HGSilveri:
Hey @Aaron-Robertson ! 
So, I was still unhappy with how complex and opaque the Kraus operators were in the article I sent you yesterday, so I took the time to define the simplest possible form of the Kraus operators for both noise channels. I compiled them in this document. I didn't have the time to go write down the derivations, but I you encourage to verify that the completeness relation is respected and that the operators actually do the desired noise map. I can also point you to some of the references I looked into to arrive at these results.

[Qudit_Noise_Channels.pdf](https://github.com/pasqal-io/Pulser/files/11706696/Qudit_Noise_Channels.pdf)

 I'm sorry for all the back and forth. I'm confident this will make the implementation much cleaner, so I think it's worth the effort. Thank you for patience in advance and for all the hard work you've put into this so far.

Aaron-Robertson:
@HGSilveri with fresh eyes I found some oversight on my part, should be closer now. I'm not exactly sure I follow why the test should warn here, when comparing to other tests.

HGSilveri:
> Hey @Aaron-Robertson ! So, I was still unhappy with how complex and opaque the Kraus operators were in the article I sent you yesterday, so I took the time to define the simplest possible form of the Kraus operators for both noise channels. I compiled them in this document. I didn't have the time to go write down the derivations, but I you encourage to verify that the completeness relation is respected and that the operators actually do the desired noise map. I can also point you to some of the references I looked into to arrive at these results.
> 
> [Qudit_Noise_Channels.pdf](https://github.com/pasqal-io/Pulser/files/11706765/Qudit_Noise_Channels.2.pdf)
> 
> I'm sorry for all the back and forth. I'm confident this will make the implementation much cleaner, so I think it's worth the effort. Thank you for patience in advance and for all the hard work you've put into this so far.

@Aaron-Robertson I noticed you sent a message right after this one, so I'm pinging you just to be sure you see this.

Aaron-Robertson:
@HGSilveri

> I'm sorry for all the back and forth. I'm confident this will make the implementation much cleaner, so I think it's worth the effort. Thank you for patience in advance and for all the hard work you've put into this so far.

No, worries on the back and forth! I appreciate the opportunity to see a few different models and learn as much as I can.

> So, I was still unhappy with how complex and opaque the Kraus operators were in the article I sent you yesterday, so I took the time to define the simplest possible form of the Kraus operators for both noise channels. I compiled them in this document. I didn't have the time to go write down the derivations, but I you encourage to verify that the completeness relation is respected and that the operators actually do the desired noise map. I can also point you to some of the references I looked into to arrive at these results.
> 
> [Qudit_Noise_Channels.pdf](https://github.com/pasqal-io/Pulser/files/11706696/Qudit_Noise_Channels.pdf)

Sounds good to me. We could keep the "experimentally informed" kraus operators as an option if you like, but for now I'll move forward replacing with the simpler ones (my initial attempt was similar, to generate a correct mapping before I got concerned about physicality, but I like that this approach is simpler still).

One thing I found interesting about the paper you shared is the exponential decay rate of dephasing, if you'd like to include an option for a decay function, that may also capture some of the experimentally informed approach.

HGSilveri:
> One thing I found interesting about the paper you shared is the exponential decay rate of dephasing, if you'd like to include an option for a decay function, that may also capture some of the experimentally informed approach.

I think this exponential decay can be incorporated in the probability `p`, if we adjust it depending on the duration of the sequence. It's a good idea, but I would leave it to a next iteration and keep things simple for now.

pulser-core/pulser/backend/noise_model.py
```diff
@@ -134,15 +134,6 @@ def _check_noise_types(self) -> None:
...
                 )
-        dephasing_on = "dephasing" in self.noise_types
-        depolarizing_on = "depolarizing" in self.noise_types
-        eff_noise_on = "eff_noise" in self.noise_types
-        eff_noise_conflict = dephasing_on + depolarizing_on + eff_noise_on > 1
-        if eff_noise_conflict:
-            raise NotImplementedError(
-                "Depolarizing, dephasing and effective noise channels"
-                "cannot be simultaneously selected."
-            )
```
HGSilveri, line 145: This is actually still true, these changes won't allow it so you can keep it.

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -753,6 +776,15 @@ def build_coeffs_ops(basis: str, addr: str) -> list[list]:
...
                 op_ids = ["sigma_du", "sigma_dd"]
+            elif basis == "all":
+                op_ids = [
+                    "sigma_rg",
+                    "sigma_rh",
+                    "sigma_gr",
+                    "sigma_gh",
+                    "sigma_hr",
+                    "sigma_hg",
+                ]
```
HGSilveri, line 787: Why do you need these?

HGSilveri (commented):
I won't be able to give support over the weekend, so here are a couple of minor remarks on things I think you won't change. I'll be as reactive as possible on Monday so that we can have this done before the hackathon is over.
Good luck!

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -221,37 +221,35 @@ def set_config(self, cfg: SimConfig) -> None:
...
+                kraus_ops.append(M_i)
+            else:  # two-level system
+                prob = self.config.dephasing_prob / 2
+                if prob > 0.1 and n > 1:
+                    warnings.warn(
+                        "The dephasing model is a first-order approximation in the"
+                        f" dephasing probability. p = {2*prob} is too large for "
+                        "realistic results.",
+                        stacklevel=2,
+                    )
```
HGSilveri, line 246: This warning can probably be moved to after the `if-else` clause to avoid repetition

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -642,7 +648,7 @@ def _build_basis_and_op_matrices(self) -> None:
                 self.basis_name = "all"  # All three states
                 self.dim = 3
                 basis = ["r", "g", "h"]
-                projectors = ["gr", "hg", "rr", "gg", "hh"]
+                projectors = ["gr", "hg", "hr", "gg", "hh", "rr"]
```
HGSilveri, line 651: I thing you don't need this either

HGSilveri (changes requested):

pulser-simulation/pulser_simulation/simulation.py
Aaron-Robertson: Tests error without

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -221,37 +221,35 @@ def set_config(self, cfg: SimConfig) -> None:
...
+                prob = self.config.dephasing_prob / 3
+                if prob > 0.1 and n > 1:
+                    warnings.warn(
+                        "The dephasing model is a first-order approximation in the"
+                        f" dephasing probability. p = {3*prob} is too large for "
+                        "realistic results.",
+                        stacklevel=2,
+                    )
+                M_0 = np.sqrt(prob/(1-prob)) * qutip.qeye(3)
+                M_i = [np.sqrt(prob) * qutip.ket(str(i), 3) * qutip.bra(str(i), 3) for i in range(3)]
```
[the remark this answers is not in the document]
Aaron-Robertson: We do need the `str` conversion unfortunately, as `qutip.ket` expects a sequence for its first arg.

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -222,68 +221,75 @@ def set_config(self, cfg: SimConfig) -> None:
...
+                    warnings.warn(
+                        "The depolarizing model is a first-order approximation"
+                        f" in the depolarizing probability. p = {9*prob}"
+                        " is too large for realistic results.",
+                        stacklevel=2,
+                    )
+                M_0 = np.sqrt(prob/(1-prob)) * qutip.qeye(3)
+                M_ij = [np.sqrt(prob/3) * qutip.ket(str(i), 3) * qutip.bra(str(j), 3) for i in range(3) for j in range(3)]
+                kraus_ops.append(M_0)
+                kraus_ops.append(M_ij)
```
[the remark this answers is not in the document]
Aaron-Robertson: Not sure if I quite hit the mark on this one, but the rest should be ready to go! I kept the prob scaling as in the other depolarizing case, while also fixing the weight.

pulser-simulation/pulser_simulation/simulation.py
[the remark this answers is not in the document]
HGSilveri: You're right, I confused it with `quitp.basis()`, which you should use instead

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -221,29 +221,26 @@ def set_config(self, cfg: SimConfig) -> None:
...
+                    "The dephasing model is a first-order approximation in the"
+                    f" dephasing probability. p = {prob} is too large for "
+                    "realistic results.",
+                    stacklevel=2,
+                )
+            if (
+                self.basis_name == "all" and self.dim == 3
+            ):  # three-level system
+                k = np.sqrt(prob * (1 - prob) ** (n - 1))
+                m_0 = np.sqrt(1 - prob) * qutip.qeye(3)
```
HGSilveri, line 237: This is already being added to the `_collapse_ops` in line 246, so you shouldn't do it again

pulser-simulation/pulser_simulation/simulation.py
HGSilveri: I'm curious what tests fail because I can't see where it is being used

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -222,68 +221,71 @@ def set_config(self, cfg: SimConfig) -> None:
...
+                k = np.sqrt((prob) * (1 - 8 * prob) ** (n - 1))
+                m_0 = np.sqrt(1 - prob) * qutip.qeye(3)
+                for i in range(3):
+                    ket_ = qutip.ket(str(i), 3)
+                    kraus_ops.append(k * ket_ * ket_.dag())
+                kraus_ops.append(m_0)
+                self._collapse_ops += [
+                    np.sqrt((1 - 8 * prob) ** n)
+                    * qutip.tensor([self.op_matrix["I"] for _ in range(n)])
+                ]
```
HGSilveri: Okay, so on this one if you divide by 3 you can use the same line for adding to `_collapse_ops` as is being used for the qubit case. The rest should be like this:
```suggestion
                prob = prob / 3

                k = np.sqrt((prob) * (1 - 3 * prob) ** (n - 1))
                for i in range(3):
                    ket_ = qutip.basis(3, 1)
                    kraus_ops.append(k * ket_ * ket_.dag())
```

pulser-simulation/pulser_simulation/simulation.py
Aaron-Robertson: Quite a few tests fail. It happens when trying to call build_operator() without the operator in the op_matrix. Eg,
```
    def build_operator(self, operations: Union[list, tuple]) -> qutip.Qobj:
        """Creates an operator with non-trivial actions on some qubits.

        Takes as argument a list of tuples ``[(operator_1, qubits_1),
        (operator_2, qubits_2)...]``. Returns the operator given by the tensor
        product of {``operator_i`` applied on ``qubits_i``} and Id on the rest.
        ``(operator, 'global')`` returns the sum for all ``j`` of operator
        applied at ``qubit_j`` and identity elsewhere.

        Example for 4 qubits: ``[(Z, [1, 2]), (Y, [3])]`` returns `ZZYI`
        and ``[(X, 'global')]`` returns `XIII + IXII + IIXI + IIIX`

        Args:
            operations: List of tuples `(operator, qubits)`.
                `operator` can be a ``qutip.Quobj`` or a string key for
                ``self.op_matrix``. `qubits` is the list on which operator
                will be applied. The qubits can be passed as their
                index or their label in the register.

        Returns:
            The final operator.
        """
        op_list = [self.op_matrix["I"] for j in range(self._size)]

        if not isinstance(operations, list):
            operations = [operations]

        for operator, qubits in operations:
            if qubits == "global":
                return sum(
                    self.build_operator([(operator, [q_id])])
                    for q_id in self._qdict
                )
            else:
                qubits_set = set(qubits)
                if len(qubits_set) < len(qubits):
                    raise ValueError("Duplicate atom ids in argument list.")
                if not qubits_set.issubset(self._qdict.keys()):
                    raise ValueError(
                        "Invalid qubit names: "
                        f"{qubits_set - self._qdict.keys()}"
                    )
                if isinstance(operator, str):
                    try:
                        operator = self.op_matrix[operator]
                    except KeyError:
>                       raise ValueError(f"{operator} is not a valid operator")
E                       ValueError: sigma_rr is not a valid operator
```

HGSilveri: For sure we need `"rr"`, it was there before after all, I just don't see why we need `"hr"` (the one you added)

tests/test_backend.py
```diff
@@ -149,6 +149,15 @@ def matrices(self):
...
 
+    def test_eff_noise_three_init(self, matrices):
+        print(
+            NoiseModel(
+                noise_types=("eff_noise",),
+                eff_noise_opers=[matrices["I3"]],
+                eff_noise_probs=[1.0],
+            )
+        )
+
```
HGSilveri, line 161: I guess you forgot to delete this, right?

tests/test_simulation.py
```diff
@@ -82,6 +82,10 @@ def matrices():
     pauli["X"] = qutip.sigmax()
     pauli["Y"] = qutip.sigmay()
     pauli["Z"] = qutip.sigmaz()
+    pauli["I3"] = qutip.qeye(3)
+    pauli["Z3"] = qutip.jmat(1, "z")
+    pauli["X3"] = qutip.jmat(1, "x")
+    pauli["Y3"] = qutip.jmat(1, "y")
```
HGSilveri, line 88: Do you still need all of these?

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -222,68 +221,66 @@ def set_config(self, cfg: SimConfig) -> None:
...
             self._collapse_ops += [
                 np.sqrt((1 - 3 * prob) ** n)
                 * qutip.tensor([self.op_matrix["I"] for _ in range(n)])
             ]
-            kraus_ops.append(k * qutip.sigmax())
-            kraus_ops.append(k * qutip.sigmay())
-            kraus_ops.append(k * qutip.sigmaz())
 
         if "eff_noise" in self.config.noise:
-            if self.basis_name == "digital" or self.basis_name == "all":
```
HGSilveri, line 280: In this case you should now check if the dimension of the Kraus operators is compatible with `self.dim`

tests/test_simulation.py
```diff
@@ -764,6 +756,35 @@ def test_eff_noise(matrices):
...
+@pytest.mark.parametrize(
+    "noise_sample,",
+    [
+        ("depolarizing"),
+        ("dephasing"),
+    ],
+)
+def test_three_level_noise(noise_sample, matrices):
+    np.random.seed(123)
+    reg = Register3D.from_coordinates([(0, 0, 0), (0, 0, 1)], prefix="q")
```
HGSilveri, line 768: Why `Register3D`? The position coordinates of the atoms and the dimensions are actually not related

tests/test_simulation.py
```diff
@@ -764,6 +756,35 @@ def test_eff_noise(matrices):
...
+    sim = QutipEmulator(
+        sampler.sample(seq),
+        reg,
+        MockDevice,
+        sampling_rate=0.01,
+        config=SimConfig(
+            noise=noise_sample,
+        ),
+    )
+    assert all([op.shape == (9, 9) for op in sim._collapse_ops])
```
HGSilveri, line 785: This is a fine thing to check but we can (and should) go much further. Must of the tests done with qubits should be extensible to qutrits. 
It's important to run simple simulations and assert that the results match our expectations.

HGSilveri (commented):
Looking good! All that's missing now is extensive testing. I sugest you take a look at what is done for qubits and try to extend to qutrits. When possible, you can even keep the same test function and just extend it with `pytest.mark.parametrize`

pulser-simulation/pulser_simulation/simulation.py
Aaron-Robertson: Here did you mean the `eff_noise_opers`? Something like:
```
if not np.all(op.shape == (self.dim, self.dim) for op in self.config.eff_noise_opers):
    # Go back to previous config
    self.set_config(prev_config)
    raise NotImplementedError(
        "Cannot mix noise operator dimensions."
    )
```

HGSilveri: Not sure it's necessary to revert the config. You are already checking that all the operators have the same dimension inside `NoiseModel`, here you just have to check that it is compatible with `self.dim` (eg the first operator has dimension `(self.dim, self.dim)`). 

I would change the error message to inform that the `eff_noise_opers` dimensions are incompatible with the state's dimensions. Including what those dimensions are in the error message is even better.

Aaron-Robertson:
Alright @HGSilveri, getting closer on testing. I still need to work through some more assertions and the three level dephasing match, but are these the tests and approach you have in mind? Any others we'd like to include?

tests/test_simulation.py
```diff
@@ -671,106 +764,113 @@ def test_noise_with_zero_epsilons(seq, matrices):
...
-    duration = 2500
-    pulse = Pulse.ConstantPulse(duration, np.pi, 0.0 * 2 * np.pi, 0)
-    seq.add(pulse, "ch0")
+@pytest.mark.parametrize(
+    "sequence_index, eff_noise_opers",
+    [((1,), ["I", "Z"]), ((0,), ["I3", "P23-1"])],
+)
+def test_add_config(sequence_index, eff_noise_opers, matrices, sequences):
+    sequences = [sequences[i] for i in sequence_index]
+    seq = sequences[0]
```
HGSilveri, line 872: Same nit:

```suggestion
    "sequence_index, eff_noise_opers",
    [(1, ["I", "Z"]), (0, ["I3", "P23-1"])],
)
def test_add_config(sequence_index, eff_noise_opers, matrices, sequences):
    seq = sequences[sequence_index]
```

HGSilveri (commented):
Still missing some simulation execution and result verification with qutrits in particular. Why not check that the noise channel is doing what is expected of it? You can look at the density matrix directly, it will be easier to check the effect

tests/test_simulation.py
```diff
@@ -671,106 +764,113 @@ def test_noise_with_zero_epsilons(seq, matrices):
...
-    pulse = Pulse.ConstantPulse(duration, np.pi, 0, 0)
-    seq.add(pulse, "ch0")
+    sequences = [sequences[i] for i in sequence_index]
+    seq = sequences[0]
+    seq2 = sequences[1]
     sim = Simulation(
         seq, sampling_rate=0.01, config=SimConfig(noise="dephasing")
     )
-    assert sim.run().sample_final_state() == Counter({"0": 595, "1": 405})
+    assert sim.run().sample_final_state() == counters[counter_key]
```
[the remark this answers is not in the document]
HGSilveri: You're right, you are running it on `seq`, nevermind

PR #531, opened 2023-06-05
Fixes for emulation of empty sequences 

HGSilveri:
- Allows for emulation of sequences with only delays by adding `intial_targets` information to `ChannelSamples`
- Fixes determination of `max_step` in `QutipEmulator.run()` when there is no non-zero pulse


a-corni (approved):
Looks good to me ! At first I was questioning the purpose of this changes, but I understood with the tests that it can be interesting for noise simulation.

PR #533, opened 2023-06-07
separate draw_samples and draw_sequence

### LINKED ISSUE #522: Adding a function to draw `SequenceSamples`

dakk:
Fix #522

### DIFF
--- a/pulser-core/pulser/sampler/samples.py
+++ b/pulser-core/pulser/sampler/samples.py
@@ -91,7 +91,24 @@ class ChannelSamples:
     phase: np.ndarray
     slots: list[_TargetSlot] = field(default_factory=list)
     eom_blocks: list[_EOMSettings] = field(default_factory=list)
-    initial_targets: set[QubitId] = field(default_factory=set)
+
+    def get_eom_mode_intervals(self) -> list[tuple[int, int]]:
+        return [
+            (
+                block.ti,
+                block.tf if block.tf is not None else self.duration,
+            )
+            for block in self.eom_blocks
+        ]
+
+    def in_eom_mode(self, time_slot: Optional[_TimeSlot] = None) -> bool:
+        """States if a time slot is inside an EOM mode block."""
+        if time_slot is None:
+            return bool(self.eom_blocks) and (self.eom_blocks[-1].tf is None)
+        return any(
+            start <= time_slot.ti < end
+            for start, end in self.get_eom_mode_intervals()
+        )
 
     def __post_init__(self) -> None:
         assert len(self.amp) == len(self.det) == len(self.phase)
@@ -374,10 +391,6 @@ class SequenceSamples:
                     d[_LOCAL][basis][t][_DET][:start_t] += cs.det[:start_t]
                     d[_LOCAL][basis][t][_PHASE][:start_t] += cs.phase[:start_t]
             else:
-                if not cs.slots:
-                    # Fill the defaultdict entries to not return an empty dict
-                    for t in cs.initial_targets:
-                        d[_LOCAL][basis][t]
                 for s in cs.slots:
                     for t in s.targets:
                         ti = s.ti
--- a/pulser-core/pulser/sequence/_seq_drawer.py
+++ b/pulser-core/pulser/sequence/_seq_drawer.py
@@ -29,6 +29,7 @@ import pulser
 from pulser import Register, Register3D
 from pulser.channels.base_channel import Channel
 from pulser.pulse import Pulse
+from pulser.sampler.sampler import sample
 from pulser.sampler.samples import ChannelSamples
 from pulser.waveforms import InterpolatedWaveform
 
@@ -161,7 +162,7 @@ class ChannelDrawContent:
         ]
 
 
-def gather_data(seq: pulser.sequence.Sequence, gather_output: bool) -> dict:
+def gather_data(seqsamples: pulser.samples.SequenceSamples, gather_output: bool) -> dict:
     """Collects the whole sequence data for plotting.
 
     Args:
@@ -173,17 +174,18 @@ def gather_data(seq: pulser.sequence.Sequence, gather_output: bool) -> dict:
     """
     # The minimum time axis length is 100 ns
     total_duration = max(
-        seq.get_duration(include_fall_time=gather_output), 100
+        seqsamples.max_duration, 100
     )
     data: dict[str, Any] = {}
-    for ch, sch in seq._schedule.items():
-        # List of interpolation points
+
+    for ch, ch_samples in seqsamples.channel_samples.items():
+        # # List of interpolation points
         interp_pts: defaultdict[str, list[list[float]]] = defaultdict(list)
         target: dict[Union[str, tuple[int, int]], Any] = {}
-        # Extracting the EOM Buffers
+        # # Extracting the EOM Buffers
         eom_intervals = [
             EOMSegment(eom_interval[0], eom_interval[1])
-            for eom_interval in sch.get_eom_mode_intervals()
+            for eom_interval in ch_samples.get_eom_mode_intervals()
         ]
         nb_eom_intervals = len(eom_intervals)
         eom_start_buffers = [EOMSegment() for _ in range(nb_eom_intervals)]
@@ -191,52 +193,55 @@ def gather_data(seq: pulser.sequence.Sequence, gather_output: bool) -> dict:
         in_eom_mode = False
         eom_block_n = -1
         # Last eom interval is extended if eom mode not disabled at the end
-        if nb_eom_intervals > 0 and seq.get_duration() == eom_intervals[-1].tf:
+        if nb_eom_intervals > 0 and ch_samples.duration == eom_intervals[-1].tf:
             eom_intervals[-1].tf = total_duration
         # sampling the channel schedule
-        samples = sch.get_samples()
-        extended_samples = samples.extend_duration(total_duration)
-        for slot in sch:
-            if slot.ti == -1:
-                target["initial"] = slot.targets
-                continue
-            else:
-                # If slot is not the first element in schedule
-                if sch.in_eom_mode(slot):
-                    # EOM mode starts
-                    if not in_eom_mode:
-                        in_eom_mode = True
-                        eom_block_n += 1
-                elif in_eom_mode:
-                    # Buffer when EOM mode is disabled and next slot has 0 amp
-                    in_eom_mode = False
-                    if extended_samples.amp[slot.ti] == 0:
-                        eom_end_buffers[eom_block_n] = EOMSegment(
-                            slot.ti, slot.tf
-                        )
-                if (
-                    eom_block_n + 1 < nb_eom_intervals
-                    and slot.tf == eom_intervals[eom_block_n + 1].ti
-                    and extended_samples.det[slot.tf - 1]
-                    == sch.eom_blocks[eom_block_n + 1].detuning_off
-                ):
-                    # Buffer if next is eom and final det matches det_off
-                    eom_start_buffers[eom_block_n + 1] = EOMSegment(
-                        slot.ti, slot.tf
-                    )
+        extended_samples = ch_samples.extend_duration(total_duration)
 
-            if slot.type == "target":
-                target[(slot.ti, slot.tf - 1)] = slot.targets
-                continue
-            if slot.type == "delay":
-                continue
-            pulse = cast(Pulse, slot.type)
-            for wf_type in ["amplitude", "detuning"]:
-                wf = getattr(pulse, wf_type)
-                if isinstance(wf, InterpolatedWaveform):
-                    pts = wf.data_points
-                    pts[:, 0] += slot.ti
-                    interp_pts[wf_type] += pts.tolist()
+
+        # TODO: move this to draw_sequence, since we are unable to create targets
+        # TODO: but eom_ buffers are set here
+        # for slot in sch:
+        #     if slot.ti == -1:
+        #         target["initial"] = slot.targets
+        #         continue
+        #     else:
+        #         # If slot is not the first element in schedule
+        #         if sch.in_eom_mode(slot):
+        #             # EOM mode starts
+        #             if not in_eom_mode:
+        #                 in_eom_mode = True
+        #                 eom_block_n += 1
+        #         elif in_eom_mode:
+        #             # Buffer when EOM mode is disabled and next slot has 0 amp
+        #             in_eom_mode = False
+        #             if extended_samples.amp[slot.ti] == 0:
+        #                 eom_end_buffers[eom_block_n] = EOMSegment(
+        #                     slot.ti, slot.tf
+        #                 )
+        #         if (
+        #             eom_block_n + 1 < nb_eom_intervals
+        #             and slot.tf == eom_intervals[eom_block_n + 1].ti
+        #             and extended_samples.det[slot.tf - 1]
+        #             == sch.eom_blocks[eom_block_n + 1].detuning_off
+        #         ):
+        #             # Buffer if next is eom and final det matches det_off
+        #             eom_start_buffers[eom_block_n + 1] = EOMSegment(
+        #                 slot.ti, slot.tf
+        #             )
+
+        #     if slot.type == "target":
+        #         target[(slot.ti, slot.tf - 1)] = slot.targets
+        #         continue
+        #     if slot.type == "delay":
+        #         continue
+        #     pulse = cast(Pulse, slot.type)
+        #     for wf_type in ["amplitude", "detuning"]:
+        #         wf = getattr(pulse, wf_type)
+        #         if isinstance(wf, InterpolatedWaveform):
+        #             pts = wf.data_points
+        #             pts[:, 0] += slot.ti
+        #             interp_pts[wf_type] += pts.tolist()
 
         # Store everything
         data[ch] = ChannelDrawContent(
@@ -248,41 +253,38 @@ def gather_data(seq: pulser.sequence.Sequence, gather_output: bool) -> dict:
         )
         if interp_pts:
             data[ch].interp_pts = dict(interp_pts)
-    if hasattr(seq, "_measurement"):
-        data["measurement"] = seq._measurement
+
+    # TODO: move to draw_sequence
+    # if hasattr(seq, "_measurement"):
+    #     data["measurement"] = seq._measurement
     data["total_duration"] = total_duration
     return data
 
 
-def draw_sequence(
-    seq: pulser.sequence.Sequence,
+
+
+
+def draw_samples(
+    seqsamples: pulser.samples.SequenceSamples,
     sampling_rate: Optional[float] = None,
-    draw_phase_area: bool = False,
     draw_interp_pts: bool = True,
     draw_phase_shifts: bool = False,
-    draw_register: bool = False,
     draw_input: bool = True,
     draw_modulation: bool = False,
     draw_phase_curve: bool = False,
 ) -> tuple[Figure | None, Figure]:
-    """Draws the entire sequence.
+    """Draws a SequenceSamples
 
     Args:
-        seq: The input sequence of operations on a device.
+        seqsamples: The input sequence of operations on a device.
         sampling_rate: Sampling rate of the effective pulse used by
             the solver. If present, plots the effective pulse alongside the
             input pulse.
-        draw_phase_area: Whether phase and area values need to be shown
-            as text on the plot, defaults to False. If `draw_phase_curve=True`,
-            phase values are ommited.
         draw_interp_pts: When the sequence has pulses with waveforms of
             type InterpolatedWaveform, draws the points of interpolation on
             top of the respective waveforms (defaults to True).
         draw_phase_shifts: Whether phase shift and reference information
             should be added to the plot, defaults to False.
-        draw_register: Whether to draw the register before the pulse
-            sequence, with a visual indication (square halo) around the qubits
-            masked by the SLM, defaults to False.
         draw_input: Draws the programmed pulses on the channels, defaults
             to True.
         draw_modulation: Draws the expected channel output, defaults to
@@ -292,23 +294,14 @@ def draw_sequence(
             if the phase doesn't change throughout the channel).
     """
 
-    def phase_str(phi: float) -> str:
-        """Formats a phase value for printing."""
-        value = (((phi + np.pi) % (2 * np.pi)) - np.pi) / np.pi
-        if value == -1:
-            return r"$\pi$"
-        elif value == 0:
-            return "0"  # pragma: no cover - just for safety
-        else:
-            return rf"{value:.2g}$\pi$"
-
-    n_channels = len(seq.declared_channels)
+    n_channels = len(seqsamples.channels)
     if not n_channels:
         raise RuntimeError("Can't draw an empty sequence.")
-    data = gather_data(seq, gather_output=draw_modulation)
+
+    data = gather_data(seqsamples, gather_output=draw_modulation)
     total_duration = data["total_duration"]
     time_scale = 1e3 if total_duration > 1e4 else 1
-    for ch in seq._schedule:
+    for ch in seqsamples.channels:
         if np.count_nonzero(data[ch].samples.det) > 0:
             data[ch].curves_on["detuning"] = True
         if draw_phase_curve and np.count_nonzero(data[ch].samples.phase) > 0:
@@ -317,55 +310,11 @@ def draw_sequence(
     # Boxes for qubit and phase text
     q_box = dict(boxstyle="round", facecolor="orange")
     ph_box = dict(boxstyle="round", facecolor="ghostwhite")
-    area_ph_box = dict(boxstyle="round", facecolor="ghostwhite", alpha=0.7)
-    slm_box = dict(boxstyle="round", alpha=0.4, facecolor="grey", hatch="//")
     eom_box = dict(boxstyle="round", facecolor="lightsteelblue")
-
-    # Draw masked register
-    if draw_register:
-        pos = np.array(seq.register._coords)
-        if isinstance(seq.register, Register3D):
-            labels = "xyz"
-            fig_reg, axes_reg = seq.register._initialize_fig_axes_projection(
-                pos,
-                blockade_radius=35,
-                draw_half_radius=True,
-            )
-            fig_reg.get_layout_engine().set(w_pad=6.5)
-
-            for ax_reg, (ix, iy) in zip(
-                axes_reg, combinations(np.arange(3), 2)
-            ):
-                seq.register._draw_2D(
-                    ax=ax_reg,
-                    pos=pos,
-                    ids=seq.register._ids,
-                    plane=(ix, iy),
-                    masked_qubits=seq._slm_mask_targets,
-                )
-                ax_reg.set_title(
-                    "Masked register projected onto\n the "
-                    + labels[ix]
-                    + labels[iy]
-                    + "-plane"
-                )
-
-        elif isinstance(seq.register, Register):
-            fig_reg, ax_reg = seq.register._initialize_fig_axes(
-                pos,
-                blockade_radius=35,
-                draw_half_radius=True,
-            )
-            seq.register._draw_2D(
-                ax=ax_reg,
-                pos=pos,
-                ids=seq.register._ids,
-                masked_qubits=seq._slm_mask_targets,
-            )
-            ax_reg.set_title("Masked register", pad=10)
+    slm_box = dict(boxstyle="round", alpha=0.4, facecolor="grey", hatch="//")
 
     ratios = [
-        SIZE_PER_WIDTH[data[ch].n_axes_on] for ch in seq.declared_channels
+        SIZE_PER_WIDTH[data[ch].n_axes_on] for ch in seqsamples.channels
     ]
     fig = plt.figure(
         constrained_layout=False,
@@ -374,7 +323,7 @@ def draw_sequence(
     gs = fig.add_gridspec(n_channels, 1, hspace=0.075, height_ratios=ratios)
 
     ch_axes = {}
-    for i, (ch, gs_) in enumerate(zip(seq.declared_channels, gs)):
+    for i, (ch, gs_) in enumerate(zip(seqsamples.channels, gs)):
         ax = fig.add_subplot(gs_)
         for side in ("top", "bottom", "left", "right"):
             ax.spines[side].set_color("none")
@@ -412,8 +361,8 @@ def draw_sequence(
     t_max = final_t * 1.05
 
     for ch, axes in ch_axes.items():
-        ch_obj = seq.declared_channels[ch]
         ch_data = data[ch]
+        ch_obj = seqsamples._ch_objs[ch]
         ch_eom_intervals = data[ch].eom_intervals
         ch_eom_start_buffers = data[ch].eom_start_buffers
         ch_eom_end_buffers = data[ch].eom_end_buffers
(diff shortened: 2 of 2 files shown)


pulser-core/pulser/sampler/samples.py
```diff
@@ -3,10 +3,11 @@
 
 from collections import defaultdict
 from dataclasses import dataclass, field, replace
-from typing import TYPE_CHECKING, Optional, cast
+from typing import TYPE_CHECKING, Any, Optional, cast
 
 import numpy as np
 
+# from pulser.sequence._schedule import _TimeSlot
```
a-corni, line 10: As you are only using it for type checking you can move the import to `if TYPE_CHECKING`. This will solve the circular import and one of your mypy issues.

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -487,6 +412,276 @@ def phase_str(phi: float) -> str:
             special_kwargs = dict(labelpad=10) if i == 0 else {}
             ax.set_ylabel(LABELS[i], fontsize=14, **special_kwargs)
 
+        if draw_target_regions:
```
a-corni, line 415: I don't think it is an option to display it.

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -487,6 +412,276 @@ def phase_str(phi: float) -> str:
...
+                        ax.axvspan(ti, tf, alpha=0.4, color="grey", hatch="//")
+                    axes[0].text(
+                        tf + final_t * 5e-3,
+                        tgt_txt_y,
+                        tgt_str,
+                        ha="left",
+                        fontsize=12,
+                        bbox=q_box,
+                    )
+
```
a-corni, line 461: what about `draw_phase_shifts` ? Are you not showing them because of `seq._basis_ref` ? I think you can get them from `ChannelSamples.phase`... (I have to dive a bit further in the code for this)

pulser-core/pulser/sampler/samples.py
```diff
@@ -90,8 +91,8 @@ class ChannelSamples:
     det: np.ndarray
     phase: np.ndarray
     slots: list[_TargetSlot] = field(default_factory=list)
+    time_slots: list[Any] = field(default_factory=list)
```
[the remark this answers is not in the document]
dakk: > I have not finished reviewing everything yet, here are some comments that can be tackled for the moment. It's already a good job ;)

Great, thank you for the review @a-corni , I'll will take care of it

pulser-core/pulser/sequence/_seq_drawer.py
dakk: The "optional" is because if we are drawing a sequence, we will draw it in `draw_sequence` instead of drawing in `draw_samples`

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -248,133 +244,63 @@ def gather_data(seq: pulser.sequence.Sequence, gather_output: bool) -> dict:
...
         if np.count_nonzero(data[ch].samples.det) > 0:
             data[ch].curves_on["detuning"] = True
         if draw_phase_curve and np.count_nonzero(data[ch].samples.phase) > 0:
             data[ch].curves_on["phase"] = True
 
     # Boxes for qubit and phase text
     q_box = dict(boxstyle="round", facecolor="orange")
-    ph_box = dict(boxstyle="round", facecolor="ghostwhite")
-    area_ph_box = dict(boxstyle="round", facecolor="ghostwhite", alpha=0.7)
-    slm_box = dict(boxstyle="round", alpha=0.4, facecolor="grey", hatch="//")
```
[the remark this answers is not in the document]
a-corni: Yes I realized, I wanted to delete this comment before publishing the review sorry I forgot.

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -161,7 +162,7 @@ def _give_curves_from_samples(
         ]
 
 
-def gather_data(seq: pulser.sequence.Sequence, gather_output: bool) -> dict:
+def gather_data(seqsamples: SequenceSamples, gather_output: bool) -> dict:
```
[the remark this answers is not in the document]
dakk: do you mean a `SequenceSamples.draw` method like the `Sequence.draw`?

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -248,133 +244,63 @@ def gather_data(seq: pulser.sequence.Sequence, gather_output: bool) -> dict:
...
-                blockade_radius=35,
-                draw_half_radius=True,
-            )
-            seq.register._draw_2D(
-                ax=ax_reg,
-                pos=pos,
-                ids=seq.register._ids,
-                masked_qubits=seq._slm_mask_targets,
-            )
-            ax_reg.set_title("Masked register", pad=10)
```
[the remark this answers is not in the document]
a-corni: Yes I think so, see my comment above in the declaration of `draw_samples`, I think you should replace `draw_register` by an optional `BaseRegister`: if it is provided then you show the register, otherwise you don't. In `draw_sequence`, you will then simply call `draw_samples` and provide the `Sequence.register` to it.

pulser-core/pulser/sequence/_seq_drawer.py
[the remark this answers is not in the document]
dakk: ok (y) I resolve this, remember then about `QutipEmulator.draw()`

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -487,6 +408,278 @@ def phase_str(phi: float) -> str:
...
+
+        for slot in sch:
+            if slot.ti == -1:
+                target["initial"] = slot.targets
+                continue
+            if slot.type == "target":
+                target[(slot.ti, slot.tf - 1)] = slot.targets
+                continue
+            if slot.type == "delay":
+                continue
```
a-corni, line 629: All of this is already taking place in `gather_data` and are stored in `ChannelDrawContent`.

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -487,6 +408,278 @@ def phase_str(phi: float) -> str:
...
+            axes[-1].text(
+                tgt_txt_x,
+                tgt_txt_y,
+                tgt_str,
+                fontsize=12,
+                ha="left",
+                bbox=slm_box,
+            )
+
+    return (fig, ch_axes, data)
```
a-corni: Hum, I am not sure we want to return data, but I see why you did it. The structure has to be a bit changed I guess because I think we want the same output as a `draw_sequence`.

I guess this is what you meant by adding a `draw` method to `SequenceSamples`.

a-corni: I think you should have a `draw_sequence`, a `draw_sequence_samples` and a `_draw_channel_content`.

`draw_channel_content` would be the same as the `draw_samples` you have at the moment, and would return (fig, ch_axes, data) (eventually it can be a `draw` method of `DrawChannelContent` ?)

`draw_samples` calls `gather_data` and calls `draw_channel_content` with the `ChannelDrawContent` obtained.

`draw_sequence` does the same thing as `draw_samples`, but add additional features that can only be shown on a `Sequence`.

What do you think ?

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -487,6 +408,278 @@ def phase_str(phi: float) -> str:
...
+                continue
+            pulse = cast(Pulse, slot.type)
+            for wf_type in ["amplitude", "detuning"]:
+                wf = getattr(pulse, wf_type)
+                if isinstance(wf, InterpolatedWaveform):
+                    pts = wf.data_points
+                    pts[:, 0] += slot.ti
+                    interp_pts[wf_type] += pts.tolist()
+
+        data[ch].target = target
```
a-corni, line 638: Unchanged

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -487,6 +408,278 @@ def phase_str(phi: float) -> str:
...
+        ax_lims = [
+            (amp_bottom, amp_top),
+            (det_bottom, det_top),
+            (min(0.0, *ref_ys[2]), max(1.1, *ref_ys[2])),
+        ]
+        ax_lims = [ax_lims[i] for i in ch_data.curves_on_indices()]
+
         if draw_phase_area:
             top = False  # Variable to track position of box, top or center.
             print_phase = not draw_phase_curve and any(
```
a-corni, line 691: All the for loops with an if selecting slots that are `TimeSlots` of type "Pulse" can be replaced by a for loop over the `_targetSlot` of `ChannelSamples.slots`. The phase can be obtained by looking at `ChannelSamples.phase[slot.ti: slot.tf]`

a-corni (commented):
It seems we still have some changes in the structure of the code to perform, I let you see my propositions in the comment. I would like to have as many options in draw_samples as in draw_sequence, while avoiding repetitions between the two :)

pulser-core/pulser/sequence/_seq_drawer.py
dakk: ok

pulser-core/pulser/sequence/_seq_drawer.py
dakk: ok

pulser-core/pulser/sequence/_seq_drawer.py
dakk: yes, this encapsulation is a good solution; I'm going for it

a-corni:
Hey @dakk, I am about to start reviewing your latest changes.
I have noticed differences between the documentation built with your code and our latest documentation in the notebook on [Output Modulation and EOM monde)(https://pulser--533.org.readthedocs.build/en/533/tutorials/output_mod_eom.html). It seems that you don't take care about the fall time when there is modulation. I let you have a look to it.

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -161,48 +163,46 @@ def _give_curves_from_samples(
...
     Args:
         seq: The input sequence of operations on a device.
-        gather_output: Whether to gather the modulated output curves.
 
     Returns:
         The data to plot.
     """
     # The minimum time axis length is 100 ns
-    total_duration = max(
-        seq.get_duration(include_fall_time=gather_output), 100
```
a-corni, line 176: The issue in the documentation comes from here: the samples have to be extended if `gather_output` is set to True. 
- `gather_output` should be always False in `draw_samples`.
- `gather_output` is equal to `draw_modulation` in `draw_sequence` (because samples have to be extended if `draw_modulation` is set to True, because they will be modulated afterwards)

a-corni: Better idea: You should add a `shown_duration` option, that represents the duration shown on the plots. You should call `extend_duration` with this `shown duration`. 
- In `draw_samples`, `shown_duration` is None
- In `draw_sequence`, `shawn_duration` is `seq.get_duration(include_fall_time=draw_modulation)`

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -246,43 +238,31 @@ def gather_data(seq: pulser.sequence.Sequence, gather_output: bool) -> dict:
...
-    if hasattr(seq, "_measurement"):
-        data["measurement"] = seq._measurement
+
     data["total_duration"] = total_duration
     return data
 
 
-def draw_sequence(
-    seq: pulser.sequence.Sequence,
+def draw_channel_content(
```
a-corni, line 246: I think you can make it a private function

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -487,6 +408,278 @@ def phase_str(phi: float) -> str:
...
+    if not n_channels:
+        raise RuntimeError("Can't draw an empty sequence.")
+
+    # Boxes for qubit and phase text
+    area_ph_box = dict(boxstyle="round", facecolor="ghostwhite", alpha=0.7)
+    q_box = dict(boxstyle="round", facecolor="orange")
+    ph_box = dict(boxstyle="round", facecolor="ghostwhite")
+
+    # Sample the sequence
+    sampled_seq = sample(seq)
```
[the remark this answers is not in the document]
a-corni: Okay sorry this was a mistake from my part

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -670,6 +534,306 @@ def phase_str(phi: float) -> str:
...
+            to True.
+        draw_modulation: Draws the expected channel output, defaults to
+            False. If the channel does not have a defined 'mod_bandwidth', this
+            is skipped unless 'draw_input=False'.
+        draw_phase_curve: Draws the changes in phase in its own curve (ignored
+            if the phase doesn't change throughout the channel).
+    """
+    n_channels = len(sampled_seq.channels)
+    if not n_channels:
+        raise RuntimeError("Can't draw an empty sequence.")
```
a-corni, line 568: I think this can be placed in `gather_data`.

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -670,6 +534,306 @@ def phase_str(phi: float) -> str:
...
+            pulse = cast(Pulse, slot.type)
+            for wf_type in ["amplitude", "detuning"]:
+                wf = getattr(pulse, wf_type)
+                if isinstance(wf, InterpolatedWaveform):
+                    pts = wf.data_points
+                    pts[:, 0] += slot.ti
+                    interp_pts[wf_type] += pts.tolist()
+
+        if interp_pts:
+            data[ch].interp_pts = dict(interp_pts)
```
a-corni, line 657: This can be done with the `data` from `draw_channel_content`

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -670,6 +534,306 @@ def phase_str(phi: float) -> str:
...
+
+    if hasattr(seq, "_measurement"):
+        data["measurement"] = seq._measurement
```
a-corni, line 660: if `hasattr(seq, "_measurement")` then `sampled_seq` has an attribute `_measurement`. I propose you move this to `draw_channel_content`.

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -670,6 +534,306 @@ def phase_str(phi: float) -> str:
...
+
+    # Boxes for qubit and phase text
+    area_ph_box = dict(boxstyle="round", facecolor="ghostwhite", alpha=0.7)
+    ph_box = dict(boxstyle="round", facecolor="ghostwhite")
+
```
a-corni, line 665: As all of this can be moved after / put in `draw_channel_content`, I suggest we do so and have `gather_data` called in `draw_channel_content`. That way `channel_draw_content` only takes a `SequenceSamples` and not a `SequenceSamples` and a `data` dict.

a-corni (commented):
One more suggestion, but I think we are closing in ;)

pulser-core/pulser/sequence/_seq_drawer.py
dakk: do you mean to `gather_data` as it was before?

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -670,6 +534,306 @@ def phase_str(phi: float) -> str:
...
+                    if sampling_rate:
+                        area_val = (
+                            np.sum(yseff[0][seq_.ti : seq_.tf]) * 1e-3 / np.pi
+                        )
+                    else:
+                        area_val = seq_.type.amplitude.integral / np.pi
+                    phase_val = seq_.type.phase
+                    x_plot = (seq_.ti + seq_.tf) / 2 / time_scale
+                    if (
+                        seq._schedule[ch][pulse_num - 1].type == "target"
```
a-corni, line 739: [draw_phase_area] Replace by `slot.ti in [target_slot.tf for target_slot in sampled_seq.channel_samples[ch].time_slots_target]`

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -670,6 +534,306 @@ def phase_str(phi: float) -> str:
...
+                        or not top
+                    ):
+                        y_plot = np.max(seq_.type.amplitude.samples) / 2
```
a-corni, line 742: `sampled_seq.channel_samples[ch].amp[slot.ti:slot.tf]`

a-corni (commented):
I think that [sampling_rate] should be provided to `draw_samples`. I propose an implementation to add `draw_phase_area` to `draw_samples`

pulser-core/pulser/sequence/_seq_drawer.py
a-corni: Yes indeed !

a-corni: The check becomes `if sampled_seq._measurement is not None`

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -670,6 +534,306 @@ def phase_str(phi: float) -> str:
...
+        if draw_phase_area:
+            top = False  # Variable to track position of box, top or center.
+            print_phase = not draw_phase_curve and any(
+                seq_.type.phase != 0
+                for seq_ in seq._schedule[ch]
+                if isinstance(seq_.type, Pulse)
+            )
+            for pulse_num, seq_ in enumerate(seq._schedule[ch]):
+                # Select only `Pulse` objects
+                if isinstance(seq_.type, Pulse):
```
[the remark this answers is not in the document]
dakk: but using `sampled_seq.slots`  isn't this `isinstance(seq_.type, Pulse)` always false?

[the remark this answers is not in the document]
dakk: and also here do you mean `sampled_seq.channel_samples[ch].slots`?

pulser-core/pulser/sequence/_seq_drawer.py
dakk: I tried to do this (last commit), but the problem seems to remain

a-corni: It's now solved ;)

pulser-core/pulser/sequence/_seq_drawer.py
a-corni: Sorry I have not been very clear, I meant that since we have (if you move the `measurement` part in `_draw_channel_content`)
```
    data = gather_data(sampled_seq)
    for ch, sch in seq._schedule.items():
        interp_pts: defaultdict[str, list[list[float]]] = defaultdict(list)

        for slot in sch:
            if slot.ti == -1 or slot.type in ["target", "delay"]:
                continue

            pulse = cast(Pulse, slot.type)
            for wf_type in ["amplitude", "detuning"]:
                wf = getattr(pulse, wf_type)
                if isinstance(wf, InterpolatedWaveform):
                    pts = wf.data_points
                    pts[:, 0] += slot.ti
                    interp_pts[wf_type] += pts.tolist()

        if interp_pts:
            data[ch].interp_pts = dict(interp_pts)
    # Boxes for qubit and phase text
    area_ph_box = dict(boxstyle="round", facecolor="ghostwhite", alpha=0.7)
    ph_box = dict(boxstyle="round", facecolor="ghostwhite")
    (fig_reg, fig, ch_axes, data) = draw_channel_content(
        data,
        sampled_seq,
        seq.register if draw_register else None,
        sampling_rate,
        draw_input,
        draw_modulation,
        draw_phase_curve,
    )
```
And that `interp_pts` of data is unused in `draw_channel_content`, you can re-write this as
```
    data = gather_data(sampled_seq)
    (fig_reg, fig, ch_axes, data) = draw_channel_content(
        data,
        sampled_seq,
        seq.register if draw_register else None,
        sampling_rate,
        draw_input,
        draw_modulation,
        draw_phase_curve,
    )
    for ch, sch in seq._schedule.items():
        interp_pts: defaultdict[str, list[list[float]]] = defaultdict(list)

        for slot in sch:
            if slot.ti == -1 or slot.type in ["target", "delay"]:
                continue

            pulse = cast(Pulse, slot.type)
            for wf_type in ["amplitude", "detuning"]:
                wf = getattr(pulse, wf_type)
                if isinstance(wf, InterpolatedWaveform):
                    pts = wf.data_points
                    pts[:, 0] += slot.ti
                    interp_pts[wf_type] += pts.tolist()

        if interp_pts:
            data[ch].interp_pts = dict(interp_pts)
    # Boxes for qubit and phase text
    area_ph_box = dict(boxstyle="round", facecolor="ghostwhite", alpha=0.7)
    ph_box = dict(boxstyle="round", facecolor="ghostwhite")
```
And as you have the structure
```
    data = gather_data(sampled_seq)
    (fig_reg, fig, ch_axes, data) = draw_channel_content(
        data,
        sampled_seq,
        ...
    )
...
```
for `draw_sequence` and `draw_samples`, I propose you modify `draw_channel_content` such that it calls `gather_data` (first operation it does after testing validity of `sampled_seq`) and that it only takes `sampled_seq` (no longer need `data` in the list of attributes).
Such that in the end you only have
```

    (fig_reg, fig, ch_axes, data) = draw_channel_content(
        sampled_seq,
        ...
    )
...
```

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -670,6 +534,306 @@ def phase_str(phi: float) -> str:
...
+                    if sampling_rate:
+                        area_val = (
+                            np.sum(yseff[0][seq_.ti : seq_.tf]) * 1e-3 / np.pi
+                        )
+                    else:
+                        area_val = seq_.type.amplitude.integral / np.pi
+                    phase_val = seq_.type.phase
```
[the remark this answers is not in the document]
a-corni: From the perspective of the phase area, it is great !
It is also working well from the phase value perspective, but I am noticing a difference between yours and the current version, your version is displaying all the phase whereas we are displaying only the phase shifts.
Nit: You can try `sampled_seq.channel_samples[ch].phase[slot.tf-1]`, it should work as well (and be simpler)
image

image

a-corni (commented):
It's looking good, you have solved the issue with the documentation, I see one issue to solve with draw_phase_area and one simplification you can implement

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -718,61 +837,6 @@ def phase_str(phi: float) -> str:
...
-                        y_plot,
-                        txt,
-                        fontsize=10,
-                        ha="center",
-                        va="center",
-                        bbox=area_ph_box,
-                    )
-
         # Draw target regions phase_shifts
         if draw_phase_shifts:
```
a-corni, line 841: In the end, why don't you pass `_basis_ref` to `SequenceSamples` when you sample a `Sequence` (just like `_measurement` and `_magnetic_field`) ?

pulser-core/pulser/sequence/_seq_drawer.py
dakk: oh, got it!

pulser-core/pulser/sequence/_seq_drawer.py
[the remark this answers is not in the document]
dakk: By using `phase_val = sampled_seq.channel_samples[ch].phase[slot.tf-1]` I was getting outofbound; I fixed by passing `shown_duration` as `extended_duration` to sample().

The other issue (displaying all the phase) was a wrong condition, I think I've solved it in the next commit.

pulser-core/pulser/sequence/_seq_drawer.py
dakk: Yes, I can do this.

dakk: I did; if I didn't miss anything should be ok now; it is still missing coverage, but I wait for your approval first

pulser-core/pulser/sequence/_seq_drawer.py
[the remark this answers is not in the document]
a-corni: That's weird, to me `sampled_seq.channel_samples[ch].phase[slot.ti:slot.tf][-1]` should be equal to `sampled_seq.channel_samples[ch].phase[slot.tf-1]` - and it's weird that one raises an error without the other 😅

pulser-core/pulser/sequence/_seq_drawer.py
a-corni: I am noticing an issue in the tutorial [Output Modulation and EOM](https://pulser.readthedocs.io/en/stable/tutorials/output_mod_eom.html#). The buffers before and after the EOM mode is switched on (the time slots that are just before and just after the eom interval) are not represented...
Obtained:
image
Should be:
image

pulser-core/pulser/sequence/_seq_drawer.py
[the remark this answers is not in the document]
dakk: They are not equal in python; [A:B] get a slice from index A to index B OR len() if B > len() (try `[1,2][0:1000]`)

pulser-core/pulser/sequence/_seq_drawer.py
dakk: I see; I think the problem is in the `gather_data` part where we create eom buffers starting from `ChannelSamples.target_time_slots`; the drawing part is almost unchanged

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -707,11 +726,136 @@ def phase_str(phi: float) -> str:
...
+
+
+def draw_samples(
+    sampled_seq: SequenceSamples,
+    register: Optional[BaseRegister] = None,
+    sampling_rate: Optional[float] = None,
+    draw_phase_area: bool = False,
+    draw_phase_shifts: bool = False,
+    draw_input: bool = True,
+    draw_modulation: bool = False,
```
a-corni, line 739: I don't think the user can have a control over these: 
`draw_samples` is displaying the sample he is providing and only this sample (if he wants to draw the output he should modulate it by himself), otherwise we will end up with user modulating twice their samples.

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -707,11 +726,136 @@ def phase_str(phi: float) -> str:
...
+    draw_phase_curve: bool = False,
+) -> tuple[Figure | None, Figure]:
+    """Draws a SequenceSamples.
+
+    Args:
+        sampled_seq: The input sequence of operations on a device.
```
a-corni, line 745: has to be changed. As above "sampled_seq: The input samples of a sequence of operations."

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -707,11 +726,136 @@ def phase_str(phi: float) -> str:
...
+        register: If present, draw the register before the pulse
+            sequence, with a visual indication (square halo) around the qubits
```
a-corni, line 747: register: If present, draw the register before the pulse
            sequence samples, with a visual indication (square halo) around the qubits

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -246,50 +242,48 @@ def gather_data(seq: pulser.sequence.Sequence, gather_output: bool) -> dict:
...
-    """Draws the entire sequence.
+    shown_duration: Optional[int] = None,
+) -> tuple[Figure | None, Figure, Any, dict]:
+    """Draws the computed data for a SequenceSamples.
 
     Args:
-        seq: The input sequence of operations on a device.
+        sampled_seq: The input sequence of operations on a device.
+        register: If present, draw the register before the pulse
+            sequence, with a visual indication (square halo) around the qubits
```
a-corni, line 268: register: If present, draw the register before the pulse
            sequence samples, with a visual indication (square halo) around the qubits

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -707,11 +726,136 @@ def phase_str(phi: float) -> str:
...
+            if the phase doesn't change throughout the channel).
+    """
+    (fig_reg, fig, ch_axes, data) = _draw_channel_content(
+        sampled_seq,
+        register,
+        sampling_rate,
+        draw_phase_area,
+        draw_phase_shifts,
+        draw_input,
+        draw_modulation,
```
a-corni, line 772: `draw_input=True,
draw_modulation=False
`

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -707,11 +726,136 @@ def phase_str(phi: float) -> str:
...
+            to True.
+        draw_modulation: Draws the expected channel output, defaults to
+            False. If the channel does not have a defined 'mod_bandwidth', this
+            is skipped unless 'draw_input=False'.
+        draw_phase_curve: Draws the changes in phase in its own curve (ignored
+            if the phase doesn't change throughout the channel).
+    """
+    # Sample the sequence and get the data to plot
+    shown_duration = seq.get_duration(include_fall_time=draw_modulation)
+    sampled_seq = sample(seq, extended_duration=shown_duration)
```
a-corni, line 818: By extending it here, you know you will get different results from the previous implementation, because the duration of `sampled_seq` is not the same as the `seq` you were providing before.
Normally all the commands related to extension is handled in `gather_data`, why did you feel you had to put it here as well ? 
I suggest you take out `extended_duration=shown_duration`.

a-corni (commented):
Looking great indeed !

pulser-core/pulser/sequence/_seq_drawer.py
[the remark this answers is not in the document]
a-corni: Sure, thanks for reminding me ;)

pulser-core/pulser/sequence/_seq_drawer.py
dakk: I further investigate it, but I cannot figure out what is the problem.

dakk:
With the last commit I think I solved the EOM drawing; the coverage is almost complete, it's missing two lines: 

pulser-core/pulser/sampler/samples.py                               183      1    99%   188
pulser-core/pulser/sequence/_seq_drawer.py                      347      1    99%   225

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -218,25 +219,19 @@ def gather_data(seq: pulser.sequence.Sequence, gather_output: bool) -> dict:
...
-            if slot.type == "delay":
-                continue
-            pulse = cast(Pulse, slot.type)
-            for wf_type in ["amplitude", "detuning"]:
-                wf = getattr(pulse, wf_type)
-                if isinstance(wf, InterpolatedWaveform):
-                    pts = wf.data_points
-                    pts[:, 0] += slot.ti
-                    interp_pts[wf_type] += pts.tolist()
+            if time_slot.type == "target":
```
a-corni, line 233: Aren't they all supposed to be of type "target" ? ;)
It's almost an assert

a-corni (commented):
All nice ! 
@dakk could you perhaps post some output of draw_samples for samples of the sequences defined in [the Output Modulation and EOM Mode notebook](https://pulser.readthedocs.io/en/stable/tutorials/output_mod_eom.html) and in (the tutorial about Phase Shifts)[https://pulser.readthedocs.io/en/stable/tutorials/phase_shifts_vz_gates.html]. It could be nice to verify that providing a register renders nicely, (for example with an SLM)[https://pulser.readthedocs.io/en/stable/tutorials/slm_mask.html].
@HGSilveri do you want to have a look ?
Core concepts and ideas subject to discussion: 
- `draw_sequence` and `draw_sequence` both calls `draw_channel_content` (draw_samples can take a register to draw a register, and only shows the input sample - it is not possible to draw a modulated sample from a sample)
- ChannelSamples has a list of TimeSlots of type "target", and can compute `is_eom_mode` and `get_eom_mode_intervals`.
- SequenceSamples has a new attribute _basis_ref that stores seq._basis_ref. There is surely a way to circumvent this but I don't know if it's worth spending time on it. 
- the value of the phase shown with `draw_area` is the last of the slot (works with non-modulated sample as phase is constant, should work with modulated samples as well as only beginning of the phase might be impacted).

pulser-core/pulser/sequence/_seq_drawer.py
dakk: yep, totally useless!

dakk:
> All nice ! @dakk could you perhaps post some output of draw_samples for samples of the sequences defined in [the Output Modulation and EOM Mode notebook](https://pulser.readthedocs.io/en/stable/tutorials/output_mod_eom.html) and in (the tutorial about Phase Shifts)[https://pulser.readthedocs.io/en/stable/tutorials/phase_shifts_vz_gates.html]. It could be nice to verify that providing a register renders nicely, (for example with an SLM)[https://pulser.readthedocs.io/en/stable/tutorials/slm_mask.html]. 

Sure, these are images generated from the new branch; I'm comparing them with the stable branch, if there are differences I'm posting both.

## `tutorials/slm_mask.html`
image
image
image

## `tutorials/phase_shifts_vz_gates.html`
image
image
image

## `tutorials/output_mod_eom.html`
image

This EOM rendering still have differences:
WAS:
image
NOW:
image

And the problem for this is that `if extended_samples.amp[slot.ti] == 0` is always `False` on line 214, and so eom_end_buffers is never populated. @a-corni can you figure out why?

HGSilveri:
I'm taking a look right now. A few preliminary comments?

> * `draw_sequence` and `draw_sequence` both calls `draw_channel_content` (draw_samples can take a register to draw a register, and only shows the input sample - it is not possible to draw a modulated sample from a sample)

Do you mean `draw_samples` and `draw_sequence`?

> * ChannelSamples has a list of TimeSlots of type "target", and can compute `is_eom_mode` and `get_eom_mode_intervals`.
> * SequenceSamples has a new attribute _basis_ref that stores seq._basis_ref. There is surely a way to circumvent this but I don't know if it's worth spending time on it.

I don't think so either, it's fine to include it imo.

> * the value of the phase shown with `draw_area` is the last of the slot (works with non-modulated sample as phase is constant, should work with modulated samples as well as only beginning of the phase might be impacted).

Sounds ok, though I would probably go with the middle instead (ie @ (ti + tf)/2)

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -490,49 +482,73 @@ def phase_str(phi: float) -> str:
         if draw_phase_area:
             top = False  # Variable to track position of box, top or center.
             print_phase = not draw_phase_curve and any(
-                seq_.type.phase != 0
-                for seq_ in seq._schedule[ch]
-                if isinstance(seq_.type, Pulse)
+                any(
```
HGSilveri, line 485: nit: I'd use `np.any` here

a-corni:
@HGSilveri I indeed mean `draw_samples` and `draw_sequence`.
@dakk on EOM I think we forgot something, it should not draw the output. That something might be the check if a channel of the device has a modulation bandwidth, I am going to dive into it. 
About your question; perhaps a `if np.isclose(extended_samples.amp[slot.ti], 0)` will be better suited.

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -490,49 +482,73 @@ def phase_str(phi: float) -> str:
...
+                if (
+                    slot.ti
+                    in [
+                        target_slot.tf
+                        for target_slot in sampled_seq.channel_samples[
+                            ch
+                        ].target_time_slots
+                    ]
+                    or not top
+                ):
```
HGSilveri, line 518: Won't this miss target slots surrounded by delays?
It's true it's not very important, it's just to decide where to place the phase&area box. Perhaps if there is a delay, the box doesn't have to go down (the idea is just to avoid an overlap with the "target" box).

Still, this also seems very inefficient. Why not compute the list of `target_slot.tf`s once beforehand and use it here?

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -161,48 +163,47 @@ def _give_curves_from_samples(
...
+        extended_samples = ch_samples.extend_duration(total_duration)
+
+        for slot in ch_samples.slots:
+            if slot.ti != -1:
                 # If slot is not the first element in schedule
-                if sch.in_eom_mode(slot):
+                if ch_samples.in_eom_mode(slot):
                     # EOM mode starts
                     if not in_eom_mode:
                         in_eom_mode = True
```
a-corni, line 209: So indeed I would first try with `np.isclose(extended_samples.amp[slot.ti], 0)`. However, if the sample is the result of a modulated sample then it is not going to be close to 0 at `slot.ti`. I think it's best to check that `np.isclose(extended_samples.amp[slot.tf-1], 0)` or `(slot.ti + slot.tf)/2`.

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -412,8 +404,8 @@ def phase_str(phi: float) -> str:
     t_max = final_t * 1.05
 
     for ch, axes in ch_axes.items():
-        ch_obj = seq.declared_channels[ch]
         ch_data = data[ch]
+        ch_obj = sampled_seq._ch_objs[ch]
         ch_eom_intervals = data[ch].eom_intervals
         ch_eom_start_buffers = data[ch].eom_start_buffers
         ch_eom_end_buffers = data[ch].eom_end_buffers
```
a-corni, line 411: line 416 can you check that `draw_output` is false in the EOM tutorial for `draw_samples` ?

HGSilveri:
> And the problem for this is that `if extended_samples.amp[slot.ti] == 0` is always `False` on line 214, and so eom_end_buffers is never populated. @a-corni can you figure out why?

This is a bit concerning to me actually, how come it was working before but not now? There seems to be something slightly different about the `extended_samples` compared to before and there shouldn't be

pulser-core/pulser/sequence/_seq_drawer.py
dakk: Neither one is working; the missing smooth_draw area in eom_end_buffers is from 940 to 1180, which was this delay _TimeSlot in the original schedule `_TimeSlot(type='delay', ti=940, tf=1180, targets={0, 1, 2, 3})`  (Still using EOM Mode Operation https://pulser.readthedocs.io/en/stable/tutorials/output_mod_eom.html as test case)
So the check on amp value should be correct, but it is not including the delay slot.

The output is restored as the original if I do this; if the current slot ti is different from the current eom_block end (tf), then it means there is a delay and we can add an end_buffer for this missing interval:

```python
                    if slot.ti != eom_intervals[eom_block_n].tf:
                        if extended_samples.amp[eom_intervals[eom_block_n].tf] == 0:
                            eom_end_buffers[eom_block_n] = EOMSegment(
                                eom_intervals[eom_block_n].tf, slot.ti
                            )
                    elif extended_samples.amp[slot.ti] == 0:
                        eom_end_buffers[eom_block_n] = EOMSegment(
                            slot.ti, slot.tf
                        )
```

I'm not sure if the elif part is still necessary and if this version is semantically identical to the original.

And this is the result:
Figure_1

pulser-core/pulser/sequence/_seq_drawer.py
dakk: > Won't this miss target slots surrounded by delays?
> It's true it's not very important, it's just to decide where to place the phase&area box. Perhaps if there is a delay, the box doesn't have to go down (the idea is just to avoid an overlap with the "target" box).

Yes, it may be drawn down if there is a delay; handling this case is not really straightforward and raises the doubt if the data we have in slots is enough

> Still, this also seems very inefficient. Why not compute the list of target_slot.tfs once beforehand and use it here?

You are right.

pulser-core/pulser/sequence/_seq_drawer.py
dakk: Yes, if I use draw_samples for the EOM tutoral, draw_output is False

pulser-core/pulser/sequence/_seq_drawer.py
a-corni: Good job finding this ! My question was related to using `draw_samples` on a modulated sample: imagine the user provides the hashed green curve to display (`draw_samples(sampler.sample(seq, modulation=True))`), how do you display the EOM mode and its buffer ? I think the first buffer will be shown, the main EOM part as well, but that you will have to check that your amplitude is not equal to 0 right after tf but after the fall time of the EOM (`2 * cast(BaseEOM, extended_samples._ch_objs[ch].eom_config).rise_time`, see in `ChannelSamples.modulate()`)
```
                    if slot.ti != eom_intervals[eom_block_n].tf:
                        if extended_samples.amp[eom_intervals[eom_block_n].tf + 2 * cast(BaseEOM, extended_samples._ch_objs[ch].eom_config).rise_time] == 0:
                            eom_end_buffers[eom_block_n] = EOMSegment(
                                eom_intervals[eom_block_n].tf, slot.ti
                            )
```

pulser-core/pulser/sequence/_seq_drawer.py
HGSilveri: Don't worry about eventual overlaps, I think it's not a big deal.

pulser-core/pulser/sequence/_seq_drawer.py
dakk: You were right, the ending part was not show when drawing with `draw_samples(sampler.sample(seq, modulation=True))`: your solution works for the ending part.
The problem remains with the eom_start_buffers; the condition `extended_samples.det[slot.tf - 1] == ch_samples.eom_blocks[eom_block_n + 1].detuning_off` never becomes True, but the values are close:
-0.49731363802743167 instead of -0.5

pulser-core/pulser/sequence/_seq_drawer.py
dakk: Images are: first is `draw_samples(sampler.sample(seq))` and second is `draw_samples(sampler.sample(seq, modulation = True))`; I'm also passing the parameters passed to the draws in the notebook (ie: draw_phase_shifts)

## Phase shifts and vz:

image
image

image
image

image
image


## Output Modulation
image
image

## EOM mode (using the isclose with the relative factor I shown before)
image
image


## SLM Mask:
image
image

image
image

image
image

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -707,11 +719,135 @@ def phase_str(phi: float) -> str:
...
+        draw_phase_shifts: Whether phase shift and reference information
+            should be added to the plot, defaults to False.
+        draw_phase_curve: Draws the changes in phase in its own curve (ignored
+            if the phase doesn't change throughout the channel).
+    """
+    slot_tfs = [
+        ch_samples.slots[-1].tf
+        for ch_samples in sampled_seq.channel_samples.values()
+    ]
+    max_slot_tf = max(slot_tfs) if len(slot_tfs) > 0 else None
```
dakk, line 755: This is introduced since if the total_duration is less that the last slot.tf, it raise an out of bound while accessing .amp, .phase and .det

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -161,82 +164,87 @@ def _give_curves_from_samples(
...
+                    ):
+                        eom_end_buffers[eom_block_n] = EOMSegment(tf, slot.ti)
+
+            if (
+                eom_block_n + 1 < nb_eom_intervals
+                and slot.tf == eom_intervals[eom_block_n + 1].ti
+                and np.isclose(
+                    extended_samples.det[slot.tf - 1],
+                    ch_samples.eom_blocks[eom_block_n + 1].detuning_off,
+                    0.006,
```
dakk, line 235: This is the factor introduced to preserve the EOM drawing of start_buffer; I think is one of the last thing to take care of

pulser-core/pulser/sequence/_seq_drawer.py
a-corni: I think that using `np.isclose` and defining a `rtol` is definitely the way to go. Eventually I would make it dependent on the modulation_bandwidth (I think the higher the modulation bandwidth, the higher the error will be, so I guess you can define the relative tolerance as an empirical factor times the modulation bandwidth)

dakk: Ok, I'll try making some experiment on it

pulser-core/pulser/sequence/_seq_drawer.py
a-corni: Thanks for having a look at this, looks good to me !

pulser-core/pulser/sequence/_seq_drawer.py
dakk: Doesn't the mod_bandwidth change only the rising time? I don't understand how it can affect the amplitude or the detuning

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -161,82 +164,87 @@ def _give_curves_from_samples(
...
+                        extended_samples.amp[
+                            tf
+                            + 2
+                            * cast(
+                                BaseEOM, sampled_seq._ch_objs[ch].eom_config
+                            ).rise_time
+                        ],
+                        0,
+                    ):
+                        eom_end_buffers[eom_block_n] = EOMSegment(tf, slot.ti)
```
HGSilveri, line 227: So, you are assuming that the entire time between the end of the EOM block and the next pulse is an end buffer, right? If one adds a delay after disabling EOM mode, then this delay will be included in the end buffer, right?
I think it's quite hard (if not impossible) to be accurately inferring the EOM end buffer like this. Personally, I'd be in favor of adding extra metadata to the `ChannelSamples` at the moment they are created to include this information. The `EOMSegment`s could even come already defined.

pulser-core/pulser/sequence/_seq_drawer.py
HGSilveri: This is less likely to run into the same issue as the EOM end buffer because in general it will be a Pulse, but even that is not guaranteed.

HGSilveri (commented):
Looks like it's almost there! I agree that it's only the EOM buffers detection that's missing. I have a suggestion for that, let me know what you think @dakk !

pulser-core/pulser/sequence/_seq_drawer.py
dakk: Ok, I will do it this afternoon

dakk: @a-corni `eom_end_buffers` and `eom_start_buffers` are list of EOMSegments, but essentially they are list of pairs (ti,tf). Can you provide me further details about what you are thinking?

HGSilveri: @dakk Indeed, I think the information of when a segment starts and ends should do, so storing the start and buffers as separate lists of tuples inside `ChannelSamples` should be enough. Then, when drawing, you can extract this information to recreate the `eom_start_buffers` and `eom_end_buffers` as lists of `EOMSegment`s

HGSilveri:
@dakk The failing CI tests seem to be due to the latest release of numpy (1.25.0). We'll look into it on our end, but you can pin the numpy version in the requirements to `<1.25` so that you're not held back.

a-corni (commented):
Looks good to me ! Just one not important comment

dakk:
> Great ! Thanks for your hard work on this issue and all your collaborations during the unitary hack @dakk ! It has been a pleasure to work with you !

Thank you too @a-corni , it was a pleasure also for me, I learned a lot!

PR #537, opened 2023-06-20
Fix Numpy 1.25 compatibility

### LINKED ISSUE #536: Fix compatibility with Numpy 1.25
The latest Numpy release (1.25.0) broke some our CI tests in Python >= 3.9. 
I didn't check why exactly, but we should fix these issues if possible or avoid this release otherwise.

HGSilveri:
Fixes #536.


pulser-simulation/pulser_simulation/qutip_result.py
```diff
@@ -157,7 +157,7 @@ def get_state(
         is_density_matrix = state.isoper
         if ignore_global_phase and not is_density_matrix:
             full = state.full()
-            global_ph = float(np.angle(full[np.argmax(np.abs(full))]))
+            global_ph = float(np.angle(full[np.argmax(np.abs(full))])[0])
```
a-corni, line 160: Wow they made a major change

> Only ndim-0 arrays are treated as scalars. NumPy used to treat all arrays of size 1 (e.g., np.array([3.14])) as scalars. In the future, this will be limited to arrays of ndim 0 (e.g., np.array(3.14)). The following expressions will report a deprecation warning:

HGSilveri:
> I was wondering if we should not modify `Variable` instead of modifying the tests, because I thought that some users might see their code modified as well, but in the mean time I guess that if numpy is changing then all the users are already impacted and it's best we accompany this change.

Our variables already make this distinction. `Variable` is always an array, even if of size 1, and then its elements are `VariableItem`s. It's true that the users sometimes declare variables of `size=1` (which is an array of size 1) and successfully use it as if they had declared it with `size=None` (which would be a single value) because the casting of [x] to x works, but I'm not sure what you are proposing we change here.

a-corni (commented):
I think you have to change the requirements.txt file of pulser-core as dakk modified it to impose numpy <1.25

a-corni:
> > I was wondering if we should not modify `Variable` instead of modifying the tests, because I thought that some users might see their code modified as well, but in the mean time I guess that if numpy is changing then all the users are already impacted and it's best we accompany this change.
> 
> Our variables already make this distinction. `Variable` is always an array, even if of size 1, and then its elements are `VariableItem`s. It's true that the users sometimes declare variables of `size=1` (which is an array of size 1) and successfully use it as if they had declared it with `size=None` (which would be a single value) because the casting of [x] to x works, but I'm not sure what you are proposing we change here.

Yes I completely agree I have just found understood this.

a-corni (approved):

PR #539, opened 2023-06-23
Adding  DetuningMap, DMM

a-corni:
In this draft, I have tried to define a DetuningMap, a DMM and the way to define them via Register, Register3D, MappableRegister, RegisterLayout for the DetuningMap, and in Device for the DMM.
I think the easiest way to draw a DetuningMap is via RegDrawer. The fact that I defined DetuningMap in the module dmm, separately from RegDrawer and Register creates a circular import: To define a Register you need a DetuningMap that needs a RegDrawer that needs the register module to finish intializing.
Conclusion: with the current draft, the next step would be to move the DetuningMap to the register module, and -I guess- DMM to the channel module.


HGSilveri:
I refactored as you suggested. I also made `DetuningMap` and subclass of `WeightMap` and made some changes to fix broken UTs. 
Since this changes `BaseDevice`, we will need to update the JSON schema accordingly...

a-corni:
Here is the rendering of a DetuningMap built from a RegisterLayout:
```
layout = RegisterLayout([[0, 0], [1, 0], [0, 1], [1, 1]])
det_map_1 = layout.define_detuning_map({0:0.09, 1:0.8, 2:0.11, 3:0})
```
image
The rendering is the same for :
```
layout = RegisterLayout([[0, 0], [1, 0], [0, 1], [1, 1]])
det_map_1 = layout.define_detuning_map({0:0.089, 1:0.8, 2:0.111, 3:0})
```
If the detuning map represents the one of a SLM Mask (all weights are the same), the representation is the same as before:
image

Here is the representation when two points are too close:
```
layout = RegisterLayout([[0, 0], [1, 0], [0, 1e-2], [1, 1]])
det_map_1 = layout.define_detuning_map({0:0.8, 1:0.0, 2:0.2, 3:0})
```

pulser-core/pulser/devices/_devices.py
```diff
@@ -82,6 +83,15 @@
...
     ),
+    dmm_objects=(
+        DMM(
+            clock_period=4,
+            min_duration=16,
+            max_duration=2**26,
+            mod_bandwidth=16,
+            bottom_detuning=-20,
+        ),
+    ),
```
HGSilveri, line 94: I would add it to `Chadoq2` and `MockDevice` rather than `IroiseMVP`

pulser-core/pulser/json/abstract_repr/deserializer.py
```diff
@@ -346,12 +347,15 @@ def _deserialize_device_object(obj: dict[str, Any]) -> Device | VirtualDevice:
     params: dict[str, Any] = dict(
         channel_ids=tuple(ch_ids), channel_objects=tuple(ch_objs)
     )
-    ex_params = ("channel_objects", "channel_ids")
+    if "dmm_channels" in obj:
+        params["dmm_objects"] = tuple(
+            _deserialize_channel(dmm_ch) for dmm_ch in obj["dmm_channels"]
+        )
```
HGSilveri, line 353: I think you can still cover this if you bypass the JSON schema validation

pulser-core/pulser/channels/dmm.py
```diff
@@ -0,0 +1,75 @@
+# Copyright 2020 Pulser Development Team
```
HGSilveri, line 1: Let's go with 2023

pulser-core/pulser/register/_reg_drawer.py
```diff
@@ -120,26 +136,44 @@ def _draw_2D(
...
+                            (
+                                q
+                                + (
+                                    f": {dmm_qubits[int(q)]:.2f}"
+                                    if not is_mask
+                                    else ""
+                                )
+                            )
+                            for q in plot_ids[i][j:]
+                        ]
```
HGSilveri, line 162: The formatting is making this very hard to read... Here's an alternative:
```suggestion
                        qubit_det = []
                        for q in plot_ids[i][j:]:
                            extra_label = f":{dmm_qubits[int(q)]:.2f}" if not is_mask else ""
                            qubit_det.append(f"q{extra_label}")
```
I can't tell if this will look better after formatting though

pulser-core/pulser/register/base_register.py
```diff
@@ -204,6 +205,30 @@ def _validate_layout(
...
+                and detuning weights (between 0 and 1, their sum must be equal
+                to 1).
+
+        Returns:
+            A DetuningMap associating detuning weights to the trap coordinates
+                of the targeted qubits.
+        """
+        if not set(detuning_weights.keys()) <= set(self.qubit_ids):
+            raise ValueError(
+                "The qubit ids of detuning weights have to be defined in the"
```
HGSilveri, line 224:
```suggestion
                "The qubit ids linked to detuning weights have to be defined in the"
```

pulser-core/pulser/register/register_layout.py
```diff
@@ -185,6 +188,30 @@ def define_register(
...
+            A DetuningMap associating detuning weights to the trap coordinates
+                of the targeted traps.
+        """
+        if not set(detuning_weights.keys()) <= set(self.traps_dict):
+            raise ValueError(
+                "The trap ids of detuning weights have to be integers"
+                f" between 0 and {self.number_of_traps}."
+            )
+        return DetuningMap(
+            itemgetter(*detuning_weights.keys())(self.traps_dict),
```
HGSilveri, line 211: Not sure the compromised readability is worth it, but it works nonetheless

tests/test_abstract_repr.py
```diff
@@ -691,6 +698,7 @@ def test_mappable_register(self, triangular_lattice):
         ]
         assert abstract["variables"]["var"] == dict(type="int", value=[0])
 
+    @pytest.mark.xfail
```
HGSilveri, line 701: Why are these failing? If things were done in backwards compatible way, they should still pass

tests/test_dmm.py
```diff
@@ -0,0 +1,167 @@
...
+
+from pulser.channels.dmm import DMM
+from pulser.register.base_register import BaseRegister
+from pulser.register.mappable_reg import MappableRegister
+from pulser.register.register_layout import RegisterLayout
+from pulser.register.weight_maps import DetuningMap
+
+
+@pytest.fixture
+def trap_coordinates():
```
HGSilveri, line 29: It seems like you're not using this outside of `layout`, so maybe it's not worth it to make it a fixture

tests/test_dmm.py
```diff
@@ -0,0 +1,167 @@
...
+
+def test_init(
+    layout: RegisterLayout, register: BaseRegister, map_reg: MappableRegister
+):
+    str_key = {"1": 1.0}
+    wrong_key = {4: 1.0}
+    bad_weights = {0: -1.0, 1: 1.0, 2: 1.0}
+    bad_sum = {0: 0.1, 2: 0.9, 3: 0.1}
+    det_map = {0: 0.7, 1: 0.3, 2: 0}
+    slm_map = {0: 1 / 3, 1: 1 / 3, 2: 1 / 3}
```
HGSilveri, line 56: I would define these closer to the check they are used in, but I realize it's a matter of personal preference

tests/test_dmm.py
```diff
@@ -0,0 +1,167 @@
...
+            assert np.all(
+                [map[i] == detuning_map.weights[i] for i in range(len(map))]
+            )
+            assert np.all(
+                [
+                    layout.coords[i]
+                    == np.array(detuning_map.trap_coordinates)[i]
+                    for i in range(len(map))
+                ]
+            )
```
HGSilveri, line 101: Since you're defining fixtures for `det_map` and `slm_map` anyway, why not put this in a separate test function and use them?
Or even use them here, if you'd prefer that

HGSilveri:
I realized that we don't need the DMMs to be added to the pre-defined devices yet and it actually causes issues, so I just removed it from `IroiseMVP` and added some TODOs instead.

pulser-core/pulser/channels/base_channel.py
```diff
@@ -78,6 +78,7 @@ class Channel(ABC):
     min_avg_amp: int = 0
     mod_bandwidth: Optional[float] = None  # MHz
     eom_config: Optional[BaseEOM] = field(init=False, default=None)
+    _has_fixed_addressing: bool = field(init=False, default=False)
```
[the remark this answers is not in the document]
a-corni: So I agree that it looks a lot like a ClassVar, I will make it so. 

This is to prevent the generation of DMM by `DMM.Global` and `DMM.Local`. I thought we would like this for all the child classes of Channel having only one type of addressing (that's why I named it has_fixed_adressing)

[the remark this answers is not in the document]
a-corni: In fact there is another solution: I should use typing.get_args to get the argument of addressing.

pulser-core/pulser/devices/_device_datacls.py
```diff
@@ -155,6 +161,10 @@ def type_check(
         for ch_obj in self.channel_objects:
             type_check("All channels", Channel, value_override=ch_obj)
 
+        for dmm_obj in self.dmm_objects:
+            type_check("All dmm channels", DMM, value_override=dmm_obj)
+        # self.supports_slm_mask = True if self.dmm_objects else False
```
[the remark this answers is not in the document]
a-corni: Good question, there is at the moment no plan.
We cancelled the idea of defining automatically supports_slm_mask. I think we should allow the definition of dmm_objects and supports_slm_mask independantly, but enforce that at least one dmm channel is defined if supports_slm_mask is True

pulser-core/pulser/devices/_device_datacls.py
```diff
@@ -34,7 +35,8 @@
 
 DIMENSIONS = Literal[2, 3]
 
-ALWAYS_OPTIONAL_PARAMS = ("max_sequence_duration", "max_runs")
+ALWAYS_OPTIONAL_PARAMS = ("max_sequence_duration", "max_runs", "dmm_objects")
+EX_PARAMS = ("channel_objects", "channel_ids", "dmm_objects")
```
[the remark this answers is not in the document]
a-corni: I guess `PARAMS_WITH_ABSTR_REPR` is a bit too long...
`PARAMS_WITH_REPR` is shorter, otherwise `CHANNEL_PARAMS` ?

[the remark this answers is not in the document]
a-corni: Actually `PARAMS_WITH_ABSTR_REPR` is as long as `ALWAYS_OPTIONAL_PARAMS`

pulser-core/pulser/devices/_device_datacls.py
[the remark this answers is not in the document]
HGSilveri: I agree with that approach

pulser-core/pulser/devices/_device_datacls.py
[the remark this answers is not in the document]
HGSilveri: It's ok for it to be long I think. How about `SPECIAL_REPR_PARAMS`?

pulser-core/pulser/channels/base_channel.py
```diff
@@ -263,7 +271,8 @@ def Local(
             min_avg_amp: The minimum average amplitude of a pulse (when not
                 zero).
         """
-        if cls._has_fixed_addressing:
+        print(get_type_hints(cls)["addressing"])
```
HGSilveri, line 274: You forgot this print

pulser-core/pulser/channels/base_channel.py
```diff
@@ -263,7 +271,8 @@ def Local(
...
+        if len(get_args(get_type_hints(cls)["addressing"])) == 1:
```
HGSilveri, line 275: This is a smart solution, but I'm not sure relying so much on type hints is such a good idea... 
Did you consider using the dataclass fields instead? You could check for `Field.init` being `False` and `default` not being `MISSING`.

pulser-core/pulser/channels/base_channel.py
```diff
@@ -304,7 +313,8 @@ def Global(
             min_avg_amp: The minimum average amplitude of a pulse (when not
                 zero).
         """
-        if cls._has_fixed_addressing:
+        print(get_args(get_type_hints(cls)["addressing"]))
+        if len(get_args(get_type_hints(cls)["addressing"])) == 1:
```
HGSilveri, line 317: Same comments here

pulser-core/pulser/devices/_device_datacls.py
```diff
@@ -162,8 +162,14 @@ def type_check(
...
-            type_check("All dmm channels", DMM, value_override=dmm_obj)
-        # self.supports_slm_mask = True if self.dmm_objects else False
+            type_check("All DMM channels", DMM, value_override=dmm_obj)
+
+        # TODO: Check that device has dmm objects if it supports SLM mask
+        # once DMM is supported for serialization
+        # if self.supports_slm_mask and not self.dmm_objects:
+        #     raise ValueError(
+        #         "One DMM object should be defined to support SLM mask."
+        #     )
```
HGSilveri, line 172: Why do we need to wait for the serialization support?

pulser-core/pulser/channels/base_channel.py
```diff
@@ -262,6 +262,14 @@ def Local(
             min_avg_amp: The minimum average amplitude of a pulse (when not
                 zero).
         """
+        # Can't initialize a channel whose addressing is determined internally
+        for cls_field in fields(cls):
+            if cls_field.name == "addressing":
+                break
+        if not cls_field.init and not cls_field.default == MISSING:
```
HGSilveri, line 269: nit: The more pythonic way is
```suggestion
        if not cls_field.init and cls_field.default is not MISSING:
```

tests/test_abstract_repr.py
```diff
@@ -63,7 +64,14 @@
...
 class TestDevice:
-    @pytest.fixture(params=[Chadoq2, IroiseMVP, MockDevice, AnalogDevice])
+    @pytest.fixture(
+        params=[
+            Chadoq2,
+            IroiseMVP,
+            MockDevice,
+            AnalogDevice,
+        ]
+    )
```
HGSilveri, line 74: nit: This is just formatting

a-corni (commented):
Perfeito !

HGSilveri (approved):

PR #540, opened 2023-06-26
Restrict `scipy` version to avoid breaking `qutip`

HGSilveri:
The latest Scipy release (1.11), broke `qutip` on import (see https://github.com/qutip/qutip/issues/2182). While this is resolved, this patch avoids `scipy 1.11` to avoid breaking `qutip` inside `pulser-simulation`.


madagra (approved):
It looks good to me. Let's maybe add a comment on the requirements file to make sure we relax this constraint as soon as it is possible.

PR #543, opened 2023-07-03
Update QAOA and QAA to solve a QUBO problem.ipynb

andreabasiliorava:
Edited the total number of bit string to check as 2**len(Q) in the 3rd code block


HGSilveri (approved):
LGTM, thanks for the fix!

HGSilveri:
Ok, the CI test fails are not on you, there seems to be a problem with a dependency, but you should install `black[jupyter]` and run it to fix the other one.

PR #549, opened 2023-07-07
Upgrade qutip to drop restriction on scipy

### LINKED ISSUE #542: Upgrade qutip to drop the restriction on scipy
In #540, we restricted the SciPy version to be `<1.11` because it was breaking Qutip. Now that Qutip 4.7.2 was released with a fix for this, we can relax that restriction by enforcing that `qutip ~= 4.7.2`.

a-corni:
As mentioned in the issue, I hesitated with `qutip >= 4.7.2`.
Fixes #542


HGSilveri (approved):
LGTM

PR #555, opened 2023-07-10
Restore backwards compatibility on the legacy serializer

HGSilveri:
The removal of `BaseDevice._channels` broke the deserialization of sequences made with `pulser-core < 0.10`. These changes should fix that.


a-corni (approved):
Looks good to me !

PR #564, opened 2023-07-25
Adding dmm config and modulation to sequence

a-corni:
In #565:
- [x] we introduced `Traps` to define `RegisterLayout` and `WeightMap`.
- [x] Write unit tests that don't depend on changes in `Sequence`.

In #576:
- [x] We added the possibility to define `slug` in `define_detuning_map` of `Register`, `MappableRegister`, `RegisterLayout`

Here:
- [x] Configure DMM using `config_det_map` (impossible to declare a DMM using `declare_channel`
- [x] Configure SLM using `config_slm_mask`: 
     - if in xy: it works as before
     - if in ising: it configures a DMM and either modulates the DMM from the first pulse found on the first declared Global channel having one, or (if no such pulse has been defined) the DMM will be modulated from the first pulse added to a Global Channel. 
     - If in neither of them, the dmm to use for SLM configuration is stored and behaves in xy/ising if next channel is in xy/ising.
     - **Modification of find_slm_mask_times**: it does not stop on a detuned_delay pulse.
- [x] `add_dmm_detuning`: Add the waveform of the detuning to a DMM. Same as using `add` to add a Pulse zero amplitude and negative detuning to a DMM. If the DMM is configured as an SLM Mask, this is only possible if a Pulse has been defined on a Global channel before.

In #576:
- [x] Parametrized sequences are now taken into account. The SLM Mask and config detuning map are seen as operations. They are stored to be performed at build. This has an impact on the way declared channels work since we now go into the to_build_calls to find the declared dmms.
- [x] Tests for compatibility between mappable register and detuning maps 

In #568 and #576:
- [x] Update serializer and deserializer for devices with DMM channels
- [x] Write UTs for DMM channel serialization
- [x] Update serializer and deserializer for the sequence using detuning map modulation
- [x] Write UTs for sequence serialization and deserialization with detuning map modulation
- [x] Update to the final sequence JSON schema
- [x] Stores Config_det_map as an operation in the schema.

In #565, we develop a Special handling of samples from a DMM channel:
- [x] We add a `DMMSamples` subclass of `ChannelSamples`, that stores the `DetuningMap` and the qubits of the register.
- Note: it is not possible to sample a Sequence with a Mappable Register and a configured DMM.

In #569:
- Fixes on the drawing of Register, RegisterLayout, DetuningMap:
     - The minimal height is extended and a label is added, "atom" for Register and "trap" for the others. 
     - Display the DetuningMaps as traps (like a RegisterLayout, because it is not enforced to have an atom at the coordinates at which there is a weight). Label 'traps' is added to plot (label "atoms" added for general registers).
     - labels in DetuningMaps could only take integers, now it can handle QubitId.
- Addition to Sequence:
     - Add possibility to draw detuning maps (2D or 3D). Can be drawn with/without register.
     - Add possibility to draw local amplitude and detuning per group of qubits, that are represented on the register.
- Tutorials:
     - Creates a tutorial explaining how to create a DetuningMap, a DMM channel, how to modulate a detuning map in a Schedule and how to display amplitude and detuning per qubits.
     - Modifies the tutorial on state preparation with the SLM Mask to show difference between XY and Ising mode.
     - Updates some notebooks to take into account changes on Register


tests/test_simulation.py
```diff
@@ -1032,9 +1036,13 @@ def test_mask_equals_remove():
             ValueError,
             match="The ids of qubits targeted in SLM mask",
         ):
-            QutipEmulator(sampler.sample(seq_masked), reg_two, MockDevice)
+            QutipEmulator(
+                sampler.sample(seq_masked),
+                reg_two,
+                replace(MockDevice, dmm_objects=(DMM(),)),
+            )
```
HGSilveri, line 1043: Some of these `replace`s are conflicting with changes in my PR. Since we are going to have to delete them anyway, may I suggest we do that now and instead add this at the begining:
```
MockDevice = replace(MockDevice, dmm_objects=(DMM(),))
```

This way we only have to delete one line

tests/test_json.py
```diff
@@ -57,7 +58,7 @@ def test_device(mod_device):
 
 def test_virtual_device(mod_device):
     assert encode_decode(MockDevice) == MockDevice
-    virtual_mod = mod_device.to_virtual()
+    virtual_mod = replace(mod_device.to_virtual(), dmm_objects=())
```
HGSilveri, line 61: Is this just because we don't support the legacy serialization yet? It should be an xfail instead then

pulser-core/pulser/sampler/sampler.py
```diff
@@ -28,8 +28,20 @@ def sample(
         raise NotImplementedError("Parametrized sequences can't be sampled.")
 
     samples_list = []
+    if seq._slm_mask_dmm:
+        seq.insert_slm_mask()
```
HGSilveri, line 32: I'm not sure I understand why this is being added at sampling time. Imo, sampling should not alter the contents of the sequence and it seems that this does that

pulser-core/pulser/sequence/_schedule.py
```diff
@@ -252,6 +253,50 @@ def __iter__(self) -> Iterator[_TimeSlot]:
...
+
+    def insert_slm(self, duration: int, det: float) -> None:
+        """Inserts a constant pulse as first operation."""
+        new_slot = [self.slots[0]]
+        new_slot.append(
+            _TimeSlot(
+                Pulse.ConstantAmplitude(0, ConstantWaveform(duration, det), 0),
+                0,
+                duration,
+                self.slots[0].targets,
```
HGSilveri, line 286: Why not use `Pulse.ConstantPulse`?

pulser-core/pulser/sequence/_schedule.py
```diff
@@ -252,6 +253,50 @@ def __iter__(self) -> Iterator[_TimeSlot]:
...
+        )
+        for slot in self:
+            new_slot.append(
+                _TimeSlot(
+                    slot.type,
+                    slot.ti + duration,
+                    slot.tf + duration,
+                    slot.targets,
+                )
+            )
```
HGSilveri, line 297: So, if I already have things in this `DMMSchedule`, which are likely aligned with what's in the other channels, won't this operation screw up that alignment?

tests/test_json.py
HGSilveri: I fixed it in my PR

pulser-core/pulser/sampler/sampler.py
HGSilveri: Also, isn't this being added every time the sequence is sampled?

HGSilveri (commented):
I still have to review the changes in `Sequence` by already have some comments

pulser-core/pulser/sequence/sequence.py
```diff
@@ -205,21 +213,45 @@ def declared_variables(self) -> dict[str, Variable]:
...
+        xy_channels = {
+            id: ch for id, ch in all_channels.items() if ch.basis == "XY"
+        }
+        if not self._schedule:
+            # Show all channels if none are declared
+            if not self._in_xy:
+                return all_channels
+            # If already in XY mode, filter right away but let the possibility
+            # to define an SLM Mask using a DMM channel
+            return {**xy_channels, **self._device.dmm_channels}
```
HGSilveri, line 226: I agree that the DMM channels should appear as available, but I don't think it should be possible to declare them in XY mode

pulser-core/pulser/sequence/sequence.py
```diff
@@ -205,21 +213,45 @@ def declared_variables(self) -> dict[str, Variable]:
...
+            non_occupied_channels = {
                 id: ch
-                for id, ch in self._device.channels.items()
+                for id, ch in all_channels.items()
                 if (
                     id not in occupied_ch_ids or self._device.reusable_channels
                 )
-                and (ch.basis == "XY" if self._in_xy else ch.basis != "XY")
+            }
+            if occupied_ch_ids == [self._slm_mask_dmm]:
```
HGSilveri, line 237: In XY mode, don't we skip `config_detuning_map()` in `config_slm_mask()`, meaning the chosen `dmm` won't be in `occupied_ch_ids`?

pulser-core/pulser/sequence/sequence.py
```diff
@@ -437,8 +475,102 @@ def config_slm_mask(self, qubits: Iterable[QubitId]) -> None:
         if self._slm_mask_targets:
             raise ValueError("SLM mask can be configured only once.")
 
-        # If checks have passed, set the SLM mask targets
         self._slm_mask_targets = targets
+        self._slm_mask_dmm = dmm_id
+        if self._in_xy and dmm_id not in self._device.dmm_channels:
```
HGSilveri, line 480: Doesn't the `dmm_id` need to be available too?
```suggestion
        if self._in_xy and dmm_id not in self.available_channels:
```

pulser-core/pulser/sequence/sequence.py
```diff
@@ -437,8 +475,102 @@ def config_slm_mask(self, qubits: Iterable[QubitId]) -> None:
...
+                    for qubit in self.register.qubit_ids
+                }
+            )
+            self.config_detuning_map(detuning_map, dmm_id)
+            for key in reversed(self.declared_channels.keys()):
+                if dmm_id in key:
+                    self._slm_mask_dmm = key
+                    break
+
+    def insert_slm_mask(self) -> None:
```
HGSilveri, line 496: Should this be something the user can call? I think once `config_slm_mask` is called, we should handle the insertion internally

pulser-core/pulser/sequence/sequence.py
```diff
@@ -437,8 +475,102 @@ def config_slm_mask(self, qubits: Iterable[QubitId]) -> None:
...
+        if not self._slm_mask_dmm:
+            raise ValueError("No SLM was defined.")
+        elif not self._slm_mask_time or self._in_xy:
+            return
+        slm_mask_end = self._slm_mask_time[1]
+        slm_mask_det = -10 * max(
+            np.max(ch_schedule.get_samples().amp[:slm_mask_end])
+            for ch_schedule in self._schedule.values()
+            if not isinstance(ch_schedule.channel_obj, DMM)
+            and ch_schedule.get_duration() > 0
```
HGSilveri, line 507: if the sequence is empty this will fail

pulser-core/pulser/sequence/sequence.py
```diff
@@ -437,8 +475,102 @@ def config_slm_mask(self, qubits: Iterable[QubitId]) -> None:
...
+
+        Args:
+            detuning_map: A DetuningMap defining atoms to act on and bottom
+                detuning to modulate.
+            dmm_id: How the channel is identified in the device.
+                See in ``Sequence.available_channels`` which DMM IDs are still
+                available (start by "dmm_" ) and the associated description.
+        """
+        if dmm_id not in self._device.dmm_channels:
+            raise ValueError("No DMM %s in the device." % dmm_id)
```
HGSilveri, line 542: nit: f-strings are preferred for readability and performance reasons

pulser-core/pulser/sequence/sequence.py
```diff
@@ -437,8 +475,102 @@ def config_slm_mask(self, qubits: Iterable[QubitId]) -> None:
         if self._slm_mask_targets:
             raise ValueError("SLM mask can be configured only once.")
 
-        # If checks have passed, set the SLM mask targets
         self._slm_mask_targets = targets
+        self._slm_mask_dmm = dmm_id
```
HGSilveri, line 479: You should do these assignments only after all checks have passed

pulser-core/pulser/sequence/sequence.py
```diff
@@ -669,6 +804,13 @@ def declare_channel(
         if ch.basis == "XY" and not self._in_xy:
             self._in_xy = True
             self.set_magnetic_field()
+            # If schedule contains a DMM channel, delete it.
+            if len(self._schedule) > 0:
+                print(self._schedule)
+                assert len(self._schedule) == 1 and isinstance(
+                    list(self._schedule.values())[0], _DMMSchedule
+                )
+                self._schedule.pop(list(self._schedule.keys())[0])
```
HGSilveri, line 813: So you are allowing for a DMM to be declared and then deleting it if you go into XY mode? To me, declaring a DMM should tie you down to Ising mode from that moment on

tests/test_sequence.py
```diff
@@ -47,9 +48,19 @@ def reg():
     return layout.rectangular_register(4, 7, prefix="q")
 
 
+@pytest.fixture
+def det_map(reg: Register):
+    return reg.define_detuning_map(
+        {"q" + str(i): (1 / 4 if i in [0, 1, 3, 4] else 0) for i in range(28)}
```
HGSilveri, line 54: Why the long range?

pulser-core/pulser/sequence/sequence.py
```diff
@@ -159,6 +160,22 @@ def _slm_mask_time(self) -> list[int]:
...
+    def _in_ising(self, value: bool) -> None:
+        if not isinstance(value, bool):
+            raise TypeError("_in_ising must be a bool.")
+        if self._in_ising and not value:
+            raise ValueError("Cannot quit ising.")
+        if self._in_xy and value:
+            raise ValueError("Cannot be in ising if in xy.")
+        if self._slm_mask_dmm and self._in_ising != value:
+            self._in_ising_value = value
+            self._set_slm_mask_dmm()
```
HGSilveri, line 177: This fails to set `_in_ising_value = True` when `self._slm_mask_dmm` is not defined. Here's an alternative formulation:
```suggestion
        if self._in_ising == value:
            # If the value doesn't change, do nothing
            return 
        if self._in_ising:    # ie value = False
            # Trying to switch off ising
            raise ValueError("Cannot quit ising.")
        # At this point, value = True
        if self._in_xy:
            raise ValueError("Cannot be in ising if in xy.")
        self._in_ising_value = True
        if self._slm_mask_dmm:
            self._set_slm_mask_dmm()
```

pulser-core/pulser/sequence/sequence.py
```diff
@@ -801,17 +799,12 @@ def declare_channel(
...
-                    list(self._schedule.values())[0], _DMMSchedule
-                )
-                self._schedule.pop(list(self._schedule.keys())[0])
-
+        if ch.basis == "XY":
+            if not self._in_xy:
+                self.set_magnetic_field()
+                self._in_xy = True
+        else:
+            self._in_ising = True
```
HGSilveri, line 807: Cool, it's a pretty minimal change

pulser-core/pulser/sequence/sequence.py
```diff
@@ -1116,6 +1109,28 @@ def add(
         """
         self._validate_channel(channel, block_eom_mode=True)
         self._add(pulse, channel, protocol)
+        if (
+            self._in_ising
+            and self._slm_mask_dmm
+            and self._schedule[self._slm_mask_dmm].get_duration() != 0
```
HGSilveri, line 1115: Shouldn't the duration be 0?

pulser-core/pulser/sequence/sequence.py
```diff
@@ -1802,6 +1817,14 @@ def _validate_channel(
             raise ValueError("Use the name of a declared channel.")
         if block_eom_mode and self.is_in_eom_mode(channel):
             raise RuntimeError("The chosen channel is in EOM mode.")
+        if (
+            channel == self._slm_mask_dmm
+            and self._schedule[self._slm_mask_dmm].get_duration() == 0
```
HGSilveri, line 1822: Two issues:
- `_validate_channel` is more of a way to check if a channel is valid, I think going further to block adding things to it is assuming it is only being called before a change. I would be more comfortable if that were controlled by an argument (like `block_eom_mode`) so we don't modify the default behaviour.
- Calling `_ChannelSchedule.get_duration()` repeatedly might be a bit time consuming, maybe it'd make sense to have a flag registering whether the first pulse has been added instead? (Only if it is clear this flag will always be reliably switched, otherwise it's not worth it)

tests/test_sequence.py
```diff
@@ -63,6 +63,15 @@ def device():
     )
 
 
+@pytest.fixture
+def mock_dev_dmm():
+    # Will be replaced by MockDevice once it has a DMM.
+    return dataclasses.replace(
+        MockDevice,
+        dmm_objects=(DMM(),),
+    )
```
HGSilveri, line 72: My suggestion was to avoid you replacing the `MockDevice` everywhere, but this will work too, it's just more work

pulser-core/pulser/sequence/sequence.py
```diff
@@ -1116,6 +1109,28 @@ def add(
         """
         self._validate_channel(channel, block_eom_mode=True)
         self._add(pulse, channel, protocol)
+        if (
+            self._in_ising
+            and self._slm_mask_dmm
+            and self._schedule[self._slm_mask_dmm].get_duration() != 0
+            and not _ChannelSchedule.is_detuned_delay(pulse)
```
HGSilveri, line 1116: The channel needs to be `Global` too

pulser-core/pulser/sequence/sequence.py
```diff
@@ -1116,6 +1109,28 @@ def add(
...
+            ).bottom_detuning
+            min_det = -10 * max(pulse.detuning.samples())
+            min_det = (
+                bottom_detuning
+                if (bottom_detuning and min_det < bottom_detuning)
+                else min_det
+            )
+            self._add(
+                Pulse.ConstantPulse(
+                    self._schedule.get_duration(), 0, min_det, 0
```
HGSilveri, line 1129: This should take the duration of the channel instead.

pulser-core/pulser/sequence/sequence.py
HGSilveri, line 1115: Doing this here will disregard EOM pulses, so it should probably be moved to inside `self._add()` instead. I realize this will mean that you will be calling `self._add()` from within itself, which can enter a recursive loop if not handled carefuly. This is another argument in favour of having a variable registering whether the first pulse has been added - you can have something like this inside `_add()`:

```python
        if (
            self._waiting_for_first_pulse
            and self._in_ising
            and self._slm_mask_dmm
            and self.declared_channels[channel].addressing == "Global"
            and self._schedule[self._slm_mask_dmm].get_duration() != 0
            and not _ChannelSchedule.is_detuned_delay(pulse)
        ):
            ...
            self._waiting_for_first_pulse = False
            self._add(...)
```
This way, this `if` will only be entered once **and** you will be skipping all the unnecessary time-consuming checks with `get_duration()`

pulser-core/pulser/sequence/sequence.py
a-corni: At the moment, this is already checked above as if we are in xy mode, either no slm have been defined yet and then available_channels = dmm_channels and xy channels, or a slm has been defined but then _slm_mask_targets is defined and an error is raised before.

tests/test_sequence.py
HGSilveri: Ah, so you want to have 0 explicitly on all others qubit IDs. How about cutting it in half and see if still works? Undefined weigths should be assumed to be 0, after all

pulser-core/pulser/sequence/sequence.py
```diff
@@ -205,21 +230,46 @@ def declared_variables(self) -> dict[str, Variable]:
     @property
     def available_channels(self) -> dict[str, Channel]:
```
HGSilveri, line 231: I think this method can be refactored now that we have `_in_ising` and `_in_xy`

pulser-core/pulser/sequence/sequence.py
a-corni: I would rather not use `self.available_channels` because it uses `self._slm_mask_dmm`.

tests/test_sequence.py
a-corni: Yes indeed, let's go for this.

pulser-core/pulser/sequence/sequence.py
a-corni: I was closing in on this issue as well, that is emerging when setting `self._slm_mask__dmm` after the tests... Your proposition solves it well, I am going for it ;)

pulser-core/pulser/sequence/sequence.py
a-corni: Indeed 😅

pulser-core/pulser/sequence/sequence.py
a-corni: Changes are coming ;)

pulser-core/pulser/sequence/sequence.py
a-corni: Thanks for these observations, errors could be induced from this indeed.

tests/test_sequence.py
a-corni: Sorry I think I am missing what your suggestion was ?

HGSilveri: Just what I mentioned here: https://github.com/pasqal-io/Pulser/pull/564#discussion_r1276478182

I ended up using it in `test_simulation.py`, actually.

a-corni: Okay sorry I missed that comment 😢 I decided to go for this because it seemed less work than replacing all the MockDevice by the replace (and because later it will be easier to search and replace all the `mock_dev_dmm` by `MockDevice`, whereas I think it will be harder to search for `replace(MockDevice, dmm_objects=(DMM(),))`

pulser-core/pulser/sequence/sequence.py
```diff
@@ -803,6 +824,7 @@ def declare_channel(
             if not self._in_xy:
                 self.set_magnetic_field()
                 self._in_xy = True
+                self._in_ising = False
```
HGSilveri, line 827: Why do we need this line? If it's just to make sure, I would put it before `self._in_xy=True`

pulser-core/pulser/sequence/sequence.py
```diff
@@ -1632,6 +1632,33 @@ def draw(
     def _plot(self, **draw_options: bool) -> tuple[Figure | None, Figure]:
         return draw_sequence(self, **draw_options)
 
+    def _add_slm_pulse(self, pulse: Pulse, channel: str) -> None:
+        if (
+            self._in_ising
+            and self._slm_mask_dmm
+            and self.declared_channels[channel].addressing == "Global"
+            and isinstance(self.declared_channels[channel], DMM)
```
HGSilveri, line 1640: This doesn't seem right... I think you probably meant the opposite, no?

pulser-core/pulser/sequence/_schedule.py
```diff
@@ -56,6 +56,7 @@ class _ChannelSchedule:
     def __post_init__(self) -> None:
         self.slots: list[_TimeSlot] = []
         self.eom_blocks: list[_EOMSettings] = []
+        self._has_pulse_slot = False
```
HGSilveri, line 59: Instead of doing this, I would add a variable only in `_DMMSchedule` called `self._waiting_for_first_pulse = False`, which would be flipped to `True` in `Sequence._set_slm_mask_dmm()` and back to `False` on the first successful call to `Sequence._add_slm_pulse()`

pulser-core/pulser/sequence/sequence.py
```diff
@@ -1678,6 +1705,7 @@ def _add(
             self._phase_shift(
                 pulse.post_phase_shift, *last.targets, basis=basis
             )
+        self._add_slm_pulse(pulse, channel)
```
HGSilveri, line 1708: I find this very misleading, it makes it seem like you are adding an SLM pulse every time. I would move some checks outside of the function to make clear this is only taking effect in rare occasions, eg (in line with my suggestion above)
```suggestion
        if self._slm_mask_dmm and self._schedule[self._slm_mask_dmm]._waiting_for_first_pulse:
            self._add_slm_pulse(pulse, channel)
```

HGSilveri (commented):
Things are a bit confusing with this latest commit

pulser-core/pulser/sequence/sequence.py
```diff
@@ -231,44 +236,39 @@ def declared_variables(self) -> dict[str, Variable]:
...
-        }
-        if not self._schedule:
-            # Show all channels if none are declared
-            if not self._in_xy:
-                return all_channels
-            # If in xy_mode, the DMMs are available if no SLM has been defined
-            elif not self._slm_mask_dmm:
-                return {**xy_channels, **self._device.dmm_channels}
-            return xy_channels
+        if not self._schedule and not self._in_xy and not self._in_ising:
```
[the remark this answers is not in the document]
a-corni: It's true that `if not self._in_xy and not self._in_ising:` would be enough.

pulser-core/pulser/sequence/sequence.py
```diff
@@ -231,44 +236,39 @@ def declared_variables(self) -> dict[str, Variable]:
...
+            if (
+                self._slm_mask_dmm is not None
+                and not self._device.reusable_channels
+            ):
+                return {
+                    id: ch
+                    for id, ch in all_channels.items()
+                    if id != self._slm_mask_dmm
+                }
+            # Otherwise return all the channels without the defined channel
```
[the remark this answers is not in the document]
a-corni: Sorry that's a typo, I will delete this comment

HGSilveri:
> * Note 1: If in ising, I am wondering if I should not be simply using `find_slm_mask_times`/ an equivalent strategy.

I think this is a good idea. Instead of manually searching for the first pulse, we can call `find_slm_mask_times`, see if the list of times comes out empty or not and take the first non-empty list we get. Just remember that we forgot so far to filter out detuned delays in `find_slm_mask_times()`, so this should be changed.

pulser-core/pulser/sequence/sequence.py
```diff
@@ -231,44 +236,39 @@ def declared_variables(self) -> dict[str, Variable]:
...
-            # Show all channels if none are declared
-            if not self._in_xy:
-                return all_channels
-            # If in xy_mode, the DMMs are available if no SLM has been defined
-            elif not self._slm_mask_dmm:
-                return {**xy_channels, **self._device.dmm_channels}
-            return xy_channels
+        if not self._schedule and not self._in_xy and not self._in_ising:
+            # If no channel has been declared nor any DMM configured
+            # Take out the DMM for the SLM Mask if channels are not reusable
```
a-corni, line 241: Perhaps `Don't show` would be better than `Take out` ?
```suggestion
            # If no channel has been declared nor any DMM configured, and if channels 
            # are not reusable, don't show the DMM used for the SLM Mask 
```

HGSilveri: Yes, I agree

a-corni:
> > * Note 1: If in ising, I am wondering if I should not be simply using `find_slm_mask_times`/ an equivalent strategy.
> 
> I think this is a good idea. Instead of manually searching for the first pulse, we can call `find_slm_mask_times`, see if the list of times comes out empty or not and take the first non-empty list we get. Just remember that we forgot so far to filter out detuned delays in `find_slm_mask_times()`, so this should be changed.

I think the latter should be addressed in another PR, I have tried to modify this and saw an unexpected error that did not seem so trivial (see Note 2).
image

HGSilveri:
> > > * Note 1: If in ising, I am wondering if I should not be simply using `find_slm_mask_times`/ an equivalent strategy.
> > 
> > 
> > I think this is a good idea. Instead of manually searching for the first pulse, we can call `find_slm_mask_times`, see if the list of times comes out empty or not and take the first non-empty list we get. Just remember that we forgot so far to filter out detuned delays in `find_slm_mask_times()`, so this should be changed.
> 
> I think the latter should be addressed in another PR, I have tried to modify this and saw an unexpected error that did not seem so trivial (see Note 2). image

It's just a bad test because the first pulse is empty, so when you start ignoring it the qubits are no longer masked and you see the interaction terms appearing. If you put some amplitude on the `rise` pulse, this should go away

pulser-core/pulser/sequence/sequence.py
a-corni: Unuseful indeed, I just thought it would make things clearer (like an assert). Let me delete it.

pulser-core/pulser/sequence/_schedule.py
```diff
@@ -440,6 +419,7 @@ def add_pulse(
         tf = ti + pulse.duration
         self._check_duration(tf)
         self[channel].slots.append(_TimeSlot(pulse, ti, tf, last.targets))
+        self[channel]._has_pulse_slot = True
```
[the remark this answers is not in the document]
HGSilveri: My idea of a flag was more to register whether we need to look for the first pulse, which is only true if the DMM is waiting for one. The actually looking for the first pulse could actually be done with `find_slm_mask_times()`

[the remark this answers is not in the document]
a-corni: I just saw the comment above

pulser-core/pulser/sequence/sequence.py
a-corni: I was not a big fan of my name `add_slm_pulse` neither

pulser-core/pulser/sequence/_schedule.py
a-corni: I hesitated in doing so because I found it a bit heavy to modify the value of a parameter that would be in `DMMSchedule` without being in `ChannelSchedule` from the type perspective (it would require cast I guess, or to add #type:ignore)

pulser-core/pulser/sequence/sequence.py
a-corni: Yes indeed, it is to avoid having it used during modulate_dmm (althought I figured out later that it was not mandatory as the pulse added via this method are detuned_delays)

pulser-core/pulser/sequence/_schedule.py
HGSilveri: The thing is that you'd be checking for different things. The idea behind the `_waiting_for_first_pulse` is that it is switched by the relevant methods, and it is indeed something specific to `_DMMSchedule`. Note that by default, an empty `_DMMSchedule` should have `_waiting_for_first_pulse=False`, even though it doesn't have any pulse. The attribute is explicitly flipped to `True` only when the DMM is used to make an SLM mask

a-corni:
> > > > * Note 1: If in ising, I am wondering if I should not be simply using `find_slm_mask_times`/ an equivalent strategy.
> > > 
> > > 
> > > I think this is a good idea. Instead of manually searching for the first pulse, we can call `find_slm_mask_times`, see if the list of times comes out empty or not and take the first non-empty list we get. Just remember that we forgot so far to filter out detuned delays in `find_slm_mask_times()`, so this should be changed.
> > 
> > 
> > I think the latter should be addressed in another PR, I have tried to modify this and saw an unexpected error that did not seem so trivial (see Note 2). image
> 
> It's just a bad test because the first pulse is empty, so when you start ignoring it the qubits are no longer masked and you see the interaction terms appearing. If you put some amplitude on the `rise` pulse, this should go away

So I have been adding an amplitude to the first pulse but the test is failing anyway, there are off-diagonal elements in the hamiltonian that depends on the amplitude of the pulse. I can't figure out where it is coming from (I have looked at the qobj_list in get_hamiltonian() but nothing easy came up)
image

HGSilveri:
>

HGSilveri:
> So I have been adding an amplitude to the first pulse but the test is failing anyway, there are off-diagonal elements in the hamiltonian that depends on the amplitude of the pulse. I can't figure out where it is coming from (I have looked at the qobj_list in get_hamiltonian() but nothing easy came up)

I can look into this, don't waste too much time on it

a-corni:
> > So I have been adding an amplitude to the first pulse but the test is failing anyway, there are off-diagonal elements in the hamiltonian that depends on the amplitude of the pulse. I can't figure out where it is coming from (I have looked at the qobj_list in get_hamiltonian() but nothing easy came up)
> 
> The off-diagonal terms appear precisely because we changed the Rabi frequency of the first pulse. Makes sense, since now the only unmasked atom (`atom3`) does feel something at `t=0`.

Ookay "atom3" is the only atom feeling something because "atom1" is masked, so "atom0", "atom2" and "atom3" can feel something but due to state preparation error "atom0" and "atom_2" are not present (hence don't feel anything), is that right ?

HGSilveri:
> Ookay "atom3" is the only atom feeling something because "atom1" is masked, so "atom0", "atom2" and "atom3" can feel something but due to state preparation error "atom0" and "atom_2" are not present (hence don't feel anything), is that right ?

Yup, that's it :)

pulser-core/pulser/sequence/sequence.py
```diff
@@ -469,22 +478,20 @@ def _set_slm_mask_dmm(self, dmm_id: str, targets: set[QubitId]) -> None:
...
+                [
+                    np.max(ch_schedule.get_samples().amp[: slm_mask_times[1]])
+                    for ch_schedule in self._schedule.values()
+                    if not isinstance(ch_schedule, _DMMSchedule)
+                    and ch_schedule.channel_obj.addressing == "Global"
+                ]
+            )
+            self._add_slm_pulse(
+                Pulse.ConstantPulse(slm_mask_times[1], max_amp, 0, 0)
+            )
```
HGSilveri, line 494: Indeed, this makes it seem like you are adding this pulse to the SLM, it's a confusing name

pulser-core/pulser/sequence/sequence.py
```diff
@@ -469,22 +478,20 @@ def _set_slm_mask_dmm(self, dmm_id: str, targets: set[QubitId]) -> None:
...
+        slm_mask_times = self._schedule.find_slm_mask_times()
+        if slm_mask_times:
+            max_amp = max(
+                [
+                    np.max(ch_schedule.get_samples().amp[: slm_mask_times[1]])
+                    for ch_schedule in self._schedule.values()
+                    if not isinstance(ch_schedule, _DMMSchedule)
+                    and ch_schedule.channel_obj.addressing == "Global"
+                ]
+            )
```
HGSilveri, line 491: Why not just go get the first pulse instead? Sampling is a bit expensive and doesn't seem necessary here

pulser-core/pulser/sequence/sequence.py
```diff
@@ -1632,29 +1638,25 @@ def draw(
...
+            min_det = -10 * np.max(pulse.amplitude.samples)
             min_det = (
                 bottom_detuning
                 if (bottom_detuning and min_det < bottom_detuning)
                 else min_det
             )
-            self._schedule[self._slm_mask_dmm]._has_pulse_slot = True
+            self._schedule[
+                self._slm_mask_dmm
+            ]._waiting_for_first_pulse = False  # type: ignore
```
HGSilveri, line 1657: Tired of all the casting? 😅 
Since you are casting further down anyway, why not do it once and store it? Like this
```python
dmm_schedule = cast(_DMMSchedule, self._schedule[self._slm_mask_dmm])
```
EDIT: Nevermind, it was a different method.

pulser-core/pulser/sequence/_schedule.py
```diff
@@ -254,6 +253,10 @@ def __iter__(self) -> Iterator[_TimeSlot]:
 class _DMMSchedule(_ChannelSchedule):
     detuning_map: DetuningMap
 
+    def __post_init__(self) -> None:
+        super().__post_init__()
+        self._waiting_for_first_pulse: bool = True
```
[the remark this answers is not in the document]
a-corni: Ah I understood the name as "Is waiting for first pulse" so the answer would be yes at beginning, no after adding the slm pulse...
If you want it to be False at the beginning and True after the SLM pulse, I would suggest we switch for `_waited_for_first_pulse` or `_can_add_first_pulse` or `_can_be_modulated` ?

pulser-core/pulser/sequence/sequence.py
a-corni: No not tired haha I tried the casting but it was but as I feared by putting `_waiting_for_first_pulse` in only `_ChannelSchedule`, althought it makes sense I don't know how to handle modifying this attribute in `self._schedule[self._slm_mask_dmm]` (`cast` creates a new object from what I have observed)

pulser-core/pulser/sequence/_schedule.py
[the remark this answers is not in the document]
HGSilveri: No, it's just that this way it doesn't discriminate between a DMM used to make an SLM mask and a regular usage. My feeling is that you should have to deal with when using the DMM normally, hence the default being `False`. 
My suggestions is that it is only flipped to `True` when the SLM mask is configured and then back to `False` once the SLM pulse is added

pulser-core/pulser/sequence/sequence.py
HGSilveri: That's odd, `cast` should have no effect at runtime, much less create a new object. I tried the `cast`, `mypy` doesn't complain and no test failed. Should I be looking out for anything else?

pulser-core/pulser/sequence/sequence.py
HGSilveri: Sounds better! You can also make the name of the argument `first_pulse`, I think it'll help clarify a bit too

pulser-core/pulser/sequence/sequence.py
a-corni: I want to be sure that all the amplitudes are way smaller than the detuning of the DMM used for the SLM during the SLM mask time, that's why I found it necessary here. If I was to go only for the first pulse, I don't think I would use `find_slm_mask_times`

pulser-core/pulser/sequence/sequence.py
a-corni: Okay perhaps there was another issue at the time then 😅

pulser-core/pulser/sequence/sequence.py
HGSilveri: That's even better, I think

pulser-core/pulser/sequence/sequence.py
```diff
@@ -405,13 +464,48 @@ def set_magnetic_field(
...
 
+    def _set_slm_mask_dmm(self, dmm_id: str, targets: set[QubitId]) -> None:
+        ntargets = len(targets)
+        detuning_map = self.register.define_detuning_map(
+            {
+                qubit: (1 / ntargets if qubit in targets else 0)
+                for qubit in self.register.qubit_ids
+            }
+        )
+        self.config_detuning_map(detuning_map, dmm_id)
```
HGSilveri, line 475: Beware: this call can't be stored, it will result in the same detuning map being configured twice (because there will be the `config_slm_mask` call there too). We have to separate the "storing" part from the rest, like it's done in other methods

a-corni: I am not sure I get it... I get the fact that I will end up with an additional call to config_detuning_map that will result in an error, but at the moment I am not storing the calls to config_detuning_map. Is calling the decorator `@seq_decorators.store` prior to configuring the detuning map the solution ?

tests/test_simulation.py
```diff
@@ -1141,11 +1106,43 @@ def test_mask_local_channel():
...
+    assert np.all(sim.samples["Global"]["ground-rydberg"]["phase"] == 0.0)
+    masked_qubits = ["q0", "q3"]
+    for q in masked_qubits:
+        assert np.all(sim.samples["Local"]["ground-rydberg"][q]["amp"] == 0.0)
+        assert np.array_equal(
+            sim.samples["Local"]["ground-rydberg"][q]["det"],
+            np.concatenate(
+                (-10 / len(masked_qubits) * pulse.amplitude.samples, [0])
+            ),
+        )
```
HGSilveri, line 1122: You should also check that there is no detuning on the non-masked qubits

tests/test_simulation.py
```diff
@@ -1197,9 +1194,33 @@ def test_effective_size_disjoint(channel_type):
...
-    )
+    if channel_type == "mw_global":
+        assert sim.get_hamiltonian(0) == 0.5 * amp * sim.build_operator(
+            [(qutip.sigmax(), ["atom3"])]
+        )
+    else:
+        basis = (
+            "ground-rydberg" if channel_type == "rydberg_global" else "digital"
+        )
+        print(sim.samples["Local"])
```
HGSilveri, line 1200: You forgot this `print`

HGSilveri (commented):
Here's a review of the changes before the merge of #568, clearly nothing stands out

HGSilveri (approved):
🎉 🎉 🎉

PR #565, opened 2023-07-25
Special handling of samples from a DMM channel

HGSilveri:
**To Do:**

- [x] Use `Traps` in `WeightMap`
- [x] Write unit tests that don't depend on changes in `Sequence`

### DIFF
--- a/pulser-core/pulser/channels/base_channel.py
+++ b/pulser-core/pulser/channels/base_channel.py
@@ -360,8 +360,6 @@ class Channel(ABC):
 
         Args:
             pulse: The pulse to validate.
-            channel_id: The channel ID used to index the chosen channel
-                on this device.
         """
         if not isinstance(pulse, Pulse):
             raise TypeError(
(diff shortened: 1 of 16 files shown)


a-corni (commented):
I have not checked all the tests so far, but it looks good to me, I think it will solve the issues I have with to_nested_dict. I have changed my handling of slm mask in sequence such that when you first config an SLM mask and then declare a XY mode, it deletes the declared DMM from _schedule. This way, it should be compatible with your implementation

HGSilveri:
> I have not checked all the tests so far, but it looks good to me, I think it will solve the issues I have with to_nested_dict. I have changed my handling of slm mask in sequence such that when you first config an SLM mask and then declare a XY mode, it deletes the declared DMM from _schedule. This way, it should be compatible with your implementation

Ok, cool. I didn't write any new tests yet, I just put xfails on those we expect to fail for now. I'll try to work on the tests that are independent from your changes this afternoon

a-corni (approved):

PR #566, opened 2023-08-01
Give access to all EOM block parameters and allow for phase drift correction

### LINKED ISSUE #562: Provide public access to the settings of each EOM block
During EOM mode, though the Rabi frequency and the `detuning_on` are chosen by the user, the `detuning_off` is not, and there should be an easy way to find out what the chosen value was, which can then be used to correct for incurred phase changes with a phase shift, for example.

The way to implement this is straightforward for a non-parametrized sequence (simply access the chosen value and return it), but what if the sequence is parametrized? There are two scenarios here:

1. `amp_on` and `detuning_on` are not parametrized: In this case, we could still calculate the EOM parameters, with a method like `calculate_eom_parameters()`. This is more flexible that just accessing the chosen values because when the sequence becomes parametrized they stop being calculated.
2.  `amp_on` and/or `detuning_on` are parametrized: Here, we would have to somehow suspend the call to `calculate_eom_parameters()`, returning a parametrized object instead. Although possible, it sounds overly complex. I propose we leave this for later, in case it is requested.

**Built-in phase correction**

The usecase I currently have in mind could also be incorporated in `add_eom_pulse()` and it would be available under any scenario. I'm thinking we could call it `correct_phase: bool = False`. At build-time, it would have all the information it needs for the correction.

HGSilveri:
- [x]  Give access to the `detuning_off` value through `RydbergEOM.calculate_detuning_off()`
- [x] Add a `correct_phase_drift: bool` option to EOM operations to correct for the phase drift that occurs during detuned delays
- [x] Update the abstract representation JSON schema to support the new signature of the EOM operations

Fixes #562 .

### DIFF
--- a/pulser-core/pulser/json/abstract_repr/serializer.py
+++ b/pulser-core/pulser/json/abstract_repr/serializer.py
@@ -26,9 +26,9 @@ import numpy as np
 from pulser.json.abstract_repr.signatures import SIGNATURES
 from pulser.json.abstract_repr.validation import validate_abstract_repr
 from pulser.json.exceptions import AbstractReprError
-from pulser.register.base_register import QubitId
 
 if TYPE_CHECKING:
+    from pulser.register.base_register import QubitId
     from pulser.sequence import Sequence
     from pulser.sequence._call import _Call
 
--- a/pulser-core/pulser/json/abstract_repr/serializer.py
+++ b/pulser-core/pulser/json/abstract_repr/serializer.py
@@ -180,6 +180,15 @@ def serialize_abstract_sequence(
         }
         return {**default_values, **params}
 
+    def remove_kwarg_if_default(
+        data: dict[str, Any], call_name: str, kwarg_name: str
+    ) -> dict[str, Any]:
+        if data.get(kwarg_name, None) == get_kwarg_default(
+            call_name, kwarg_name
+        ):
+            data.pop(kwarg_name, None)
+        return data
+
     operations = res["operations"]
     for call in chain(seq._calls, seq._to_build_calls):
         if call.name == "__init__":
@@ -269,8 +278,17 @@ def serialize_abstract_sequence(
             res["slm_mask_targets"] = tuple(seq._slm_mask_targets)
         elif call.name == "enable_eom_mode":
             data = get_all_args(
-                ("channel", "amp_on", "detuning_on", "optimal_detuning_off"),
+                (
+                    "channel",
+                    "amp_on",
+                    "detuning_on",
+                    "optimal_detuning_off",
+                    "correct_phase_drift",
+                ),
                 call,
+            )
+            data = remove_kwarg_if_default(
+                data, call.name, "correct_phase_drift"
             )
             operations.append({"op": "enable_eom_mode", **data})
         elif call.name == "add_eom_pulse":
@@ -281,15 +299,20 @@ def serialize_abstract_sequence(
                     "phase",
                     "post_phase_shift",
                     "protocol",
+                    "correct_phase_drift",
                 ),
                 call,
             )
+            data = remove_kwarg_if_default(
+                data, call.name, "correct_phase_drift"
+            )
             operations.append({"op": "add_eom_pulse", **data})
         elif call.name == "disable_eom_mode":
-            data = get_all_args(("channel",), call)
-            operations.append(
-                {"op": "disable_eom_mode", "channel": data["channel"]}
+            data = get_all_args(("channel", "correct_phase_drift"), call)
+            data = remove_kwarg_if_default(
+                data, call.name, "correct_phase_drift"
             )
+            operations.append({"op": "disable_eom_mode", **data})
         else:
             raise AbstractReprError(f"Unknown call '{call.name}'.")
 
--- a/pulser-core/pulser/sequence/_schedule.py
+++ b/pulser-core/pulser/sequence/_schedule.py
@@ -24,6 +24,7 @@ import numpy as np
 from pulser.channels.base_channel import Channel
 from pulser.pulse import Pulse
 from pulser.register.base_register import QubitId
+from pulser.register.weight_maps import DetuningMap
 from pulser.sampler.samples import ChannelSamples, _PulseTargetSlot
 from pulser.waveforms import ConstantWaveform
 
@@ -245,6 +246,11 @@ class _ChannelSchedule:
     def __iter__(self) -> Iterator[_TimeSlot]:
         for slot in self.slots:
             yield slot
+
+
+@dataclass
+class _DMMSchedule(_ChannelSchedule):
+    detuning_map: DetuningMap
 
 
 class _Schedule(Dict[str, _ChannelSchedule]):
--- a/pulser-core/pulser/sequence/_schedule.py
+++ b/pulser-core/pulser/sequence/_schedule.py
@@ -48,6 +48,16 @@ class _EOMSettings:
 
 
 @dataclass
+class _PhaseDriftParams:
+    drift_rate: float  # rad/µs
+    ti: int  # ns
+
+    def calc_phase_drift(self, tf: int) -> float:
+        """Calculate the phase drift during the elapsed time."""
+        return self.drift_rate * (tf - self.ti) * 1e-3
+
+
+@dataclass
 class _ChannelSchedule:
     channel_id: str
     channel_obj: Channel
@@ -350,8 +360,16 @@ class _Schedule(Dict[str, _ChannelSchedule]):
         channel: str,
         phase_barrier_ts: list[int],
         protocol: str,
+        phase_drift_params: _PhaseDriftParams | None = None,
     ) -> None:
-        pass
+        def corrected_phase(tf: int) -> float:
+            phase_drift = (
+                phase_drift_params.calc_phase_drift(tf)
+                if phase_drift_params
+                else 0
+            )
+            return pulse.phase - phase_drift
+
         last = self[channel][-1]
         t0 = last.tf
         current_max_t = max(t0, *phase_barrier_ts)
@@ -368,7 +386,7 @@ class _Schedule(Dict[str, _ChannelSchedule]):
                 )
                 last_pulse = cast(Pulse, last_pulse_slot.type)
                 # Checks if the current pulse changes the phase
-                if last_pulse.phase != pulse.phase:
+                if last_pulse.phase != corrected_phase(current_max_t):
                     # Subtracts the time that has already elapsed since the
                     # last pulse from the phase_jump_time and adds the
                     # fall_time to let the last pulse ramp down
@@ -392,6 +410,14 @@ class _Schedule(Dict[str, _ChannelSchedule]):
         ti = t0 + delay_duration
         tf = ti + pulse.duration
         self._check_duration(tf)
+        # dataclasses.replace() does not work on Pulse (because init=False)
+        if phase_drift_params is not None:
+            pulse = Pulse(
+                amplitude=pulse.amplitude,
+                detuning=pulse.detuning,
+                phase=corrected_phase(ti),
+                post_phase_shift=pulse.post_phase_shift,
+            )
         self[channel].slots.append(_TimeSlot(pulse, ti, tf, last.targets))
 
     def add_delay(self, duration: int, channel: str) -> None:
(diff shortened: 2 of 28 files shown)


pulser-core/pulser/sequence/_schedule.py
```diff
@@ -392,6 +410,14 @@ def add_pulse(
         ti = t0 + delay_duration
         tf = ti + pulse.duration
         self._check_duration(tf)
+        # dataclasses.replace() does not work on Pulse (because init=False)
+        if phase_drift_params is not None:
+            pulse = Pulse(
+                amplitude=pulse.amplitude,
+                detuning=pulse.detuning,
+                phase=corrected_phase(ti),
```
lvignoli, line 418: If I get it correctly, this apply an additional phase that compensate the phase drift acquired by the non-zero detuning delays between EOM pulses.
That way there is no need for the user to track it.

pulser-core/pulser/sequence/sequence.py
```diff
@@ -913,6 +941,11 @@ def add_eom_pulse(
                 immediately after the end of the pulse.
             protocol: Stipulates how to deal with eventual conflicts with
                 other channels (see `Sequence.add()` for more details).
+            correct_phase_drift: Adjusts the phase to correct for the phase
+                drift that occured since the last pulse (or the start of the
+                EOM mode, if adding the first pulse). This effectively
+                changes the phase of the EOM pulse, so an extra delay might
+                be added to enforce the phase jump time.
```
lvignoli, line 948: Okay this makes things a bit clearer.

lvignoli (commented):
@HGSilveri thank you so much for working on this so quickly.

I just had a quick look without working out all the logic here since there is a lot of new code.

If I understand correctly, there is now a new option, off by default, to automatically correct the phase drift acquired by delays before new pulses when in EOM mode, so that the user get an effective pulse that has the intended effect.

The workaround we have with Lucas at the moment is to find the off detuning value of the sequence in EOM mode by some means, add a delay of known duration between pulses, and do the phase correction on the user side.

We have to make the delay long enough so that the phase can ramp up.
Is the ramp up for the correction of the phase taken into account when scheduling the pulses?

HGSilveri:
> @HGSilveri thank you so much for working on this so quickly.
> 
> I just had a quick look without working out all the logic here since there is a lot of new code.

Hey @lvignoli , thanks for the quick review! I just wanted to let you know about this feature to make sure it suits your needs, don't worry about reviewing everything in detail.

> If I understand correctly, there is now a new option, off by default, to automatically correct the phase drift acquired by delays before new pulses when in EOM mode, so that the user get an effective pulse that has the intended effect.

That is correct, yes.

> The workaround we have with Lucas at the moment is to find the off detuning value of the sequence in EOM mode by some means, add a delay of known duration between pulses, and do the phase correction on the user side.
> 
> We have to make the delay long enough so that the phase can ramp up. Is the ramp up for the correction of the phase taken into account when scheduling the pulses?

Yes, whenever you modify the phase the so called "phase jump time" is enforced and this is no exception. In fact, if the delay causing the phase drift you're trying to correct for is too short to change the phase, it is extend by the necessary amount and then the phase drift correction takes into account the full delay, not just the original one.

pulser-core/pulser/sequence/sequence.py
```diff
@@ -822,29 +830,35 @@ def enable_eom_mode(
...
+                    amp_on, detuning_on, optimal_detuning_off
+                )
                 off_pulse = Pulse.ConstantPulse(
                     channel_obj.min_duration, 0.0, detuning_off, 0.0
                 )
                 channel_obj.validate_pulse(off_pulse)
 
             if not self.is_parametrized():
+                phase_drift_params = _PhaseDriftParams(
+                    drift_rate=-detuning_off, ti=self.get_duration(channel)
```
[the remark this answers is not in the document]
HGSilveri: It's the time immediately before you enable EOM mode (and thus add the start buffer). In this case, we are trying to correct only for the phase drift incurred during the EOM start buffer

pulser-core/pulser/sequence/sequence.py
```diff
@@ -822,29 +830,35 @@ def enable_eom_mode(
...
+                    buffer_slot = self._last(channel)
+                    drift = phase_drift_params.calc_phase_drift(buffer_slot.tf)
+                    self._phase_shift(
+                        -drift, *buffer_slot.targets, basis=channel_obj.basis
+                    )
 
     @seq_decorators.store
     @seq_decorators.block_if_measured
-    def disable_eom_mode(self, channel: str) -> None:
+    def disable_eom_mode(
```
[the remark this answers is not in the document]
HGSilveri: It's only to correct for phase drift that has occured since the end of the last pulse and the moment you switch off the EOM mode. I don't except it to be used often, but it covers the case where the last thing you do before disabling the EOM mode is a delay.

pulser-core/pulser/sequence/sequence.py
```diff
@@ -774,6 +779,7 @@ def enable_eom_mode(
         amp_on: Union[float, Parametrized],
         detuning_on: Union[float, Parametrized],
         optimal_detuning_off: Union[float, Parametrized] = 0.0,
+        correct_phase_drift: bool = False,
```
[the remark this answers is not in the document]
HGSilveri: Because it will change the default behaviour and thus break backwards compatibility with previously written code. Plus, I always err on the side of letting the user choose the add-ons, even if I believe they will be generally beneficial

pulser-core/pulser/sequence/_schedule.py
HGSilveri: Exactly. Furthermore, since the compensation is only computed at runtime, it will also work for parametrized sequences.

pulser-core/pulser/sequence/sequence.py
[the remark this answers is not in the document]
lvignoli: Okay, make sense then. As long as it clearly documented and recommended somewhere in the docs, it's fine.

pulser-core/pulser/sequence/_schedule.py
lvignoli: Nice! =)

lvignoli:
Okay it makes things clear on my side. Also I am glad to know the phase ramp up I always taken into account (Pulser is doing so much behind the scenes during pulse scheduling, it's sometimes hard to know what is guaranteed and what is not).

pulser-core/pulser/json/abstract_repr/serializer.py
```diff
@@ -180,6 +180,15 @@ def get_all_args(
...
 
+    def remove_kwarg_if_default(
+        data: dict[str, Any], call_name: str, kwarg_name: str
+    ) -> dict[str, Any]:
+        if data.get(kwarg_name, None) == get_kwarg_default(
+            call_name, kwarg_name
+        ):
+            data.pop(kwarg_name, None)
+        return data
+
```
lvignoli, line 191: The name is so scary, I don't want to understand what's happening here 😅

lvignoli (approved):

pulser-core/pulser/sequence/sequence.py
```diff
@@ -822,29 +830,35 @@ def enable_eom_mode(
...
+                    drift_rate=-detuning_off, ti=self.get_duration(channel)
+                )
                 self._schedule.enable_eom(
                     channel, amp_on, detuning_on, detuning_off
                 )
+                if correct_phase_drift:
+                    buffer_slot = self._last(channel)
+                    drift = phase_drift_params.calc_phase_drift(buffer_slot.tf)
+                    self._phase_shift(
+                        -drift, *buffer_slot.targets, basis=channel_obj.basis
```
a-corni, line 854: Why is there a minus in front of drift ? I thought you were taking `-detuning_off` in `phase_drift_params` because that's this is the correcting phase drift... (at the moment I have the feeling you could have a + in front of drift here and in front of detuning_off in _PhaseDriftParams and the results would be the same)

a-corni: As it is the same in disable_eom_mode, I definitely have the feeling there is something I don't understand 😅

tests/test_sequence.py
```diff
@@ -1449,22 +1452,39 @@ def test_eom_mode(reg, mod_device, custom_buffer_time, patch_plt_show):
...
+        "ch0",
+        pulse_duration,
+        phase=0.0,
+        correct_phase_drift=correct_phase_drift,
+    )
     first_pulse_slot = seq._schedule["ch0"].last_pulse_slot()
     assert first_pulse_slot.ti == delay_slot.tf
     assert first_pulse_slot.tf == first_pulse_slot.ti + pulse_duration
-    eom_pulse = Pulse.ConstantPulse(pulse_duration, amp_on, detuning_on, 0.0)
+    phase = detuning_off * first_pulse_slot.ti * 1e-3 * correct_phase_drift
```
a-corni, line 1464: Should be 1e3 here as well I think

tests/test_sequence.py
```diff
@@ -1449,22 +1452,39 @@ def test_eom_mode(reg, mod_device, custom_buffer_time, patch_plt_show):
...
+        correct_phase_drift=correct_phase_drift,
+    )
     second_pulse_slot = seq._schedule["ch0"].last_pulse_slot()
     phase_buffer = (
         eom_pulse.fall_time(ch0_obj, in_eom_mode=True)
         + seq.declared_channels["ch0"].phase_jump_time
     )
     assert second_pulse_slot.ti == first_pulse_slot.tf + phase_buffer
+    # Corrects the phase acquired during the phase buffer
+    phase_ += detuning_off * phase_buffer * 1e-3 * correct_phase_drift
```
a-corni, line 1484: Coherent with the two minuses seen above... But shouldn't `detuning_off*phase_buffer*1e3` be substracted from the current phase ?

pulser-core/pulser/sequence/sequence.py
HGSilveri: Ah yes, I struggled with this a bit at first too. Remember that the Bloch sphere convention we follow has the z-component of the Bloch rotation vector be -∂(t). This means that, in free precession, the accumalated phase is indeed -∂*t (for constant ∂). Then, to correct for this, you apply a phase shift in the opposite direction. You are right when you say that we could just get rid of the minus signs, but I figured this would be more explicit.

bejito:
`correct_phase_drift` seems to be an operation applied in the frontend during the design of the Pulses. Is there a reason for it to be serialized ? Instead of having the final corrected pulse values serialized I mean.

HGSilveri:
> `correct_phase_drift` seems to be an operation applied in the frontend during the design of the Pulses. Is there a reason for it to be serialized ? Instead of having the final corrected pulse values serialized I mean.

It's because of the parametrized sequences, since you can't know before the sequence is built what the correction factor will be.

bejito:
Ah this makes sense indeed. So for non-parametrized sequences, do we directly apply the correction and serialize it or do we let the backend apply the correction ?

HGSilveri:
> Ah this makes sense indeed. So for non-parametrized sequences, do we directly apply the correction and serialize it or do we let the backend apply the correction ?

We let the backend apply the correction

pulser-core/pulser/sequence/sequence.py
```diff
@@ -806,6 +812,8 @@ def enable_eom_mode(
             optimal_detuning_off: The optimal value of detuning (in rad/µs)
                 when there is no pulse being played. It will choose the closest
                 value among the existing options.
+            correct_phase_drift: Performs a phase shift to correct for the
+                phase drift incurred while turning on the EOM mode.
```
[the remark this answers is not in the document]
HGSilveri: Isn't what's in the docstring of those methods enough in your opinion?

pulser-core/pulser/sequence/sequence.py
a-corni: Thanks for the explanation ! I had the -∂*t but I was mixing things up afterwards :) I think what puzzled me a bit was having a negative rate, but that's okay let's keep it this way.

pulser-core/pulser/sequence/sequence.py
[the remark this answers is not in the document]
a-corni: With the notebook it's definitely okay.

a-corni:
I am ready to approve, everything looks good to me. There seems to be an issue with the build of the docs, I wonder if it was not just a bug with the imports.

HGSilveri:
> I am ready to approve, everything looks good to me. There seems to be an issue with the build of the docs, I wonder if it was not just a bug with the imports.

Yes, this seems to be a general issue, I don't think it's related to these changes. I'll look into it

a-corni (approved):

PR #572, opened 2023-08-28
Update to jsonschema > 4.18, Sphinx < 7

HGSilveri:
I suspect this is what's causing problems in the docs build, so I'm switching to the reference system introduced in `jsonschema 4.18`.

### DIFF
--- a/pulser-core/pulser/json/abstract_repr/validation.py
+++ b/pulser-core/pulser/json/abstract_repr/validation.py
@@ -16,12 +16,12 @@ import json
 from typing import Literal
 
 import jsonschema
+from referencing import Registry, Resource
 
-from pulser.json.abstract_repr import SCHEMAS, SCHEMAS_PATH
+from pulser.json.abstract_repr import SCHEMAS
 
-RESOLVER = jsonschema.validators.RefResolver(
-    base_uri=f"{SCHEMAS_PATH.resolve().as_uri()}/",
-    referrer=SCHEMAS["sequence"],
+REGISTRY: Registry = Registry().with_resources(
+    [("device-schema.json", Resource.from_contents(SCHEMAS["device"]))]
 )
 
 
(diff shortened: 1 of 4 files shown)


pulser-core/pulser/json/abstract_repr/validation.py
```diff
@@ -16,12 +16,12 @@
...
+from referencing import Registry, Resource
 
-from pulser.json.abstract_repr import SCHEMAS, SCHEMAS_PATH
+from pulser.json.abstract_repr import SCHEMAS
 
-RESOLVER = jsonschema.validators.RefResolver(
-    base_uri=f"{SCHEMAS_PATH.resolve().as_uri()}/",
-    referrer=SCHEMAS["sequence"],
+REGISTRY: Registry = Registry().with_resources(
+    [("device-schema.json", Resource.from_contents(SCHEMAS["device"]))]
```
a-corni, line 24: Why is it device and not sequence ?

a-corni (commented):
I indeed don't think that jsonschema is responsible for our issues, since I made a commit that passed on August 8th whereas their latest change was on August 7th. Yet, it's good to make these changes, but I have a question regarding them: I see the resource taken for the registry is `SCHEMAS["device"]` whereas we used to take the `SCHEMAS["sequence"]` as referring document in the `RefResolver.

pulser-core/pulser/json/abstract_repr/validation.py
HGSilveri: Tbh, I think I never quite understood what the `referrer` argument did (in their docs, they have an example where they set it to `True` 🙃), I just went with something that worked. The schema being referred and the name (ie `device-schema.json`) was taken from the path in the `base_uri`.
Now, what we are doing is specifically building a `Registry` with the schemas that can be referred and the name they are referred by.

a-corni: Okay I just had a quick look at the doc but did not go further, I understood the role of the Registry and the Resource so let's go with this ;)

a-corni (approved):

PR #573, opened 2023-08-30
Clarify the Conventions in Pulser

### LINKED ISSUE #571: Write a "Conventions" page
Write a "Conventions" page to add under the "Fundamental Concepts" section of the Pulser docs. It should mention:

- [x] The different Hamiltonians and respective bases  
- [x] The single-qutrit state and the projections into the single-qubit states in the different bases 
- [x] Multi-partite state order
- [x] Measurement convention: what's 0 and what's 1 (depending on basis), order in a measurement sample

HGSilveri:
- Adds a "Conventions" page to the docs, establishing all the conventions in Pulser
- Modifies the simulation in XY mode to comply with this convention

Closes #571 .


tutorials/quantum_simulation/Spin chain of 3 atoms in XY mode.ipynb
```diff
@@ -14,7 +14,7 @@
     "We introduce here a new mode of interaction between the atoms driven by the dipole-dipole interaction. This mode is called the XY mode. The interaction hamiltonian of a system of $N$ qubits is now given by \n",
     "\n",
     "$$\n",
-    "H_{XY} = \\frac{1}{2}\\sum_{i<j} \\frac{C_3}{R_{ij}^3}(\\sigma^+_i \\sigma^-_j + \\sigma^-_i \\sigma^+_j) \n",
+    "H_{XY} = \\sum_{i=1}^N\\sum_{j<i} \\frac{C_3}{R_{ij}^3}(\\sigma^+_i \\sigma^-_j + \\sigma^-_i \\sigma^+_j) \n",
     "$$\n",
     "\n",
     "where $\\sigma^{±} = (\\sigma^x ± i\\sigma^y)/2$, $R_{ij}$ is the distance between atom $i$ and atom $j$, and $C_3$ is a coefficient dependent on the device and taken as  $\\frac{C_3}{\\hbar} = 3700 \\frac{\\mu m^3}{\\mu s}$ for the `MockDevice`.\n",
```
a-corni, line 20: Won't it be clearer to have the units not as a fraction ?

docs/source/conventions.rst
```diff
@@ -0,0 +1,239 @@
+****************************************
+Conventions
+****************************************
+
+States and Bases
+####################################
+
+Bases
+*******
+Essentially, a basis refers to a set of two eigenstates. The transition between
```
a-corni, line 10: Nit: Just wondering if the "Essentially" is needed "A basis refers to a set of two eigenstates. The transition between" or "A basis essentially refers to a set of two eigenstates. The transition between"

a-corni (commented):
Great documentation, as far as I am concerned it is very clear and it will for sure help users :) Just found a minor typo and have two suggestions

tutorials/quantum_simulation/Spin chain of 3 atoms in XY mode.ipynb
HGSilveri: To me it doesn't make a big difference, I'll change it

docs/source/conventions.rst
HGSilveri: I agree with you, I'll remove it

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -402,7 +402,7 @@ def set_initial_state(
         if isinstance(state, str) and state == "all-ground":
             self._initial_state = qutip.tensor(
                 [
-                    self.basis["d" if self._interaction == "XY" else "g"]
+                    self.basis["u" if self._interaction == "XY" else "g"]
```
a-corni, line 405: Thinking about it, I find it weird to have |0> -> "u" (for up ?) and |1> -> "d" (for down I guess) in the code. Won't it make more sense to invert the order of the "u" and "d" when building self.basis in _build_basis_and_op_matrices ?

a-corni (approved):

madagra:
@HGSilveri I have only a couple of small comments but it all looks good to me:

* I would add a note about the endianness to which the chosen basis state ordering corresponds when expressing bitstrings. I think it is little endian, if I am not mistaken. The "Measurements samples ordering" is probably the right place for it.
* In the XY basis, however, the state ordering looks flipped to me, now you get big endian when you sample. I guess the explanation given to @a-corni is valid also in this case but I want to point out this since it can be pretty dangerous in some cases.

madagra (approved):

HGSilveri:
> @HGSilveri I have only a couple of small comments but it all looks good to me:
> 
> * I would add a note about the endianness to which the chosen basis state ordering corresponds when expressing bitstrings. I think it is little endian, if I am not mistaken. The "Measurements samples ordering" is probably the right place for it.
> * In the XY basis, however, the state ordering looks flipped to me, now you get big endian when you sample. I guess the explanation given to @a-corni is valid also in this case but I want to point out this since it can be pretty dangerous in some cases.

Thanks for the review @madagra ! I'm not sure I understand your comments. From my understanding, endianness refers to how a binary number is stored in a register and displayed, so it is only relevant when you are reading your samples as a binary number. I think it's preferable to leave this up to the user and stick solely to matching the samples to the order of the qubit IDs. Let me give you an example:

Take a register with qubit IDs `("fish", "cat", "dog")`. The state vector is going to be `|fish, cat, dog>` and let's say that upon measurement in the ground-rydberg basis it is projected onto `|rgg>`. Then, we will register a count in sample `100`. That's all. The user can then decide whether "dog" or "fish" is the smallest-memory site in their Register and this will determine whether they read the sample as "4" or "1".

Also, can you point me to where ordering is flipped in the XY basis? This shouldn't be the case.

lvignoli:
I had a quick read without checking the validity of the maths.
From a user that comes back to these computations regularly, I would say most of the useful information is there. Nice job!

docs/source/conventions.rst
```diff
@@ -0,0 +1,239 @@
...
+and :math:`j`. Note that, if multiple basis are addressed, there will be a 
+corresponding driving Hamiltonian for each transition.
+
+
+Driving Hamiltonian
+*********************
+
+The driving Hamiltonian describes the coherent excitation of an individual atom
+between two energies levels, :math:`|a\rangle` and :math:`|b\rangle`, with
+Rabi frequency :math:`\Omega(t)`, detuning :math:`\delta(t)` and phase :math:`\phi(t)`.
```
lvignoli, line 140: I would put in the body text that b is the state of higher energy, not only in the warning below.
Also, a figure of the energy levels annotated with Ω, δ etc would make everything super clear.

```suggestion
The driving Hamiltonian describes the coherent excitation of an individual atom
between two energies levels, :math:`|a\rangle` and :math:`|b\rangle`, :math:`E_b > E_a`, with
Rabi frequency :math:`\Omega(t)`, detuning :math:`\delta(t)` and phase :math:`\phi(t)`.
```

lvignoli (approved):
I in fact had this small comment

docs/source/conventions.rst
HGSilveri: If I add the picture, do you think it's worth it to repeat in the text that the energy of b is higher than a? I feel like the picture and the warning should be enough

lvignoli: I'd rather yes, I add to re-read three time to make sure about the convention between an and b. The warning block had the unintended effect that, being a block of its own, I expected to find only _additional_ information, not the required ones to understand the body text 😅
In fact the warning is really here to insist on the ordering despite the vector space representation, so the energy ordering should be stated in the body text I think.

lvignoli: (Long comment for a nit, just merge <3 )

a-corni:
Nit: There is a reference to conventions page in the notebook about SLM 

> according to Pulser's conventions for XY basis states

Perhaps we could link both pages ?

a-corni:
This looks good to me !

HGSilveri:
> This looks good to me !

I was still waiting for a reply from @madagra, but I guess I can merge and then make modifications later if necessary

PR #574, opened 2023-08-31
Handle parametrized sequence, switching device and DMM, and serialize config_det_map as an operation

a-corni:
The last three commits take into account last review from #568:
- [x] Adds slug description
- [x] Parametrized sequences are now taken into account. This was particularly annoying with the slm mask, that does not config the detuning map right away but configures it at the build.
- [x] Config_det_map is now stored as an operation in the schema, this makes deserialization easier.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,5 +1,5 @@
 # For generating documentation.
-Sphinx
+Sphinx < 7
 sphinx-rtd-theme # documentation theme
 sphinx_autodoc_typehints == 1.21.3
 nbsphinx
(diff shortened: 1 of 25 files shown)


HGSilveri:
I'll change the target branch of the PR

HGSilveri:
The commit history is a bit messy here because the changes that were previously merged were squashed so they don't appear here. I think it would be better to checkout a new branch off of `add_dmm_sequence` and add the three new commits there.

a-corni:
That's what I have tried to perform in #575... Is that clearer ?

HGSilveri:
> That's what I have tried to perform in #575... Is that clearer ?

No, the old commits are still appearing... Did you make a new branch from `add_dmm_sequence` and then add to it only the three new commits?

a-corni:
No I merged the branch that was the mistake... Let's try again !

PR #576, opened 2023-09-01
Handle parametrized sequence, switching device and DMM, and serialize config_det_map as an operation

a-corni:
The last three commits take into account last review from #568:
- [x] Adds slug description
- [x] Parametrized sequences are now taken into account. The SLM Mask and config detuning map are seen as operations. They are stored to be performed at build. This has an impact on the way declared channels work since we now go into the to_build_calls to find the declared dmms.
- [x] Config_det_map is now stored as an operation in the schema, this makes deserialization easier.
- [x] Tests for compatibility between mappable register and detuning maps


HGSilveri:
Great! We can close the other PR now

pulser-core/pulser/sequence/sequence.py
```diff
@@ -615,9 +620,7 @@ def _config_detuning_map(
         # Manually store the channel declaration as a regular call
         if to_store:
```
HGSilveri, line 621: Do you still need the `to_store` argument? Can't we just put the store in `config_detuning_map` and not in `_config_detuning_map`?

pulser-core/pulser/sequence/sequence.py
```diff
@@ -728,6 +755,8 @@ def check_retarget(ch_obj: Channel) -> bool:
                         continue
                 if not strict:
                     channel_match[old_ch_name] = new_ch_id
+                    if ch_match_err.startswith(base_msg):
+                        ch_match_err = ""
```
HGSilveri, line 759: I'm not sure I understand why this is needed

pulser-core/pulser/sequence/sequence.py
```diff
@@ -681,10 +692,26 @@ def check_retarget(ch_obj: Channel) -> bool:
...
-        for old_ch_name, old_ch_obj in self.declared_channels.items():
+        all_channels_new_device = {
+            **new_device.channels,
+            **new_device.dmm_channels,
+        }
+        all_declared_channels = self.declared_channels
+        if self.is_parametrized():
+            # SLM mask configures a DMM after all the DMMs
+            for call in self._to_build_calls:
+                if call.name == "config_slm_mask":
```
HGSilveri, line 703: In a `strict` switch, can't the `bottom_detuning` influence the value of the DMM pulse for the SLM mask? Shouldn't we check for that?

pulser-core/pulser/sequence/sequence.py
```diff
@@ -1938,7 +1977,19 @@ def _validate_channel(
...
+            and [channel]
+            == [
+                (
+                    call.args[1]
+                    if "dmm_id" not in call.kwargs
+                    else call.kwargs["dmm_id"]
+                )
+                for call in self._to_build_calls
+                if call.name == "config_slm_mask"
+            ]
```
HGSilveri, line 1991: This makes me wonder if it's truly necessary to suspend the declaration of the DMM channel when using `config_slm_mask()`... Do you see a reason why?

pulser-core/pulser/sequence/sequence.py
a-corni: Ah yes I forgot about this case.

pulser-core/pulser/sequence/sequence.py
a-corni: Well I had an issue that I only managed to solve this way: When switching from a device with 1 reusable dmm to a device with 1 non-reusable dmm, for a sequence configuring two dmms after having declared a global channel, the error message was "No match for channel dmm_0" whereas dmm_0 was matched, and the error message should be "No match for channel dmm_0_1" (which it is now, see test_switch_device_down)

pulser-core/pulser/sequence/sequence.py
a-corni: The reason why I suspended the configuration of the DMM channel when configuring the SLM mask is that otherwise at build we would end up with two configurations of the DMM

HGSilveri: Since `config_slm_mask` does not take parametrized arguments, I could see it be always executed as a `_call` (like we do for `declare_channel`). 
The issue I see with this is that it will change the order of the instructions and this now matters (ie what is identified as the "first pulse" will depend on when `config_slm_mask()` is called, _I think_).

pulser-core/pulser/sequence/sequence.py
```diff
@@ -73,6 +73,19 @@
...
+    return "_".join(dmm_name.split("_")[0:2])
+
+
+def _get_dmm_name(dmm_id: str, channels: list[str]) -> str:
+    dmm_count = len(
+        [key for key in channels if _dmm_id_from_name(key) == dmm_id]
+    )
+    if dmm_count == 0:
+        return dmm_id
+    return dmm_id + f"_{dmm_count}"
```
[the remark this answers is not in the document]
a-corni: Eventually in the same file we define `DMM` ?

pulser-core/pulser/sequence/sequence.py
```diff
@@ -235,7 +248,29 @@ def get_register(
...
+                call.name == "config_slm_mask"
+                or call.name == "config_detuning_map"
+            ):
+                dmm_id: str
+                if "dmm_id" in call.kwargs:
+                    dmm_id = call.kwargs["dmm_id"]
+                elif len(call.args) > 1:
+                    dmm_id = call.args[1]
+                else:
+                    dmm_id = "dmm_0"
```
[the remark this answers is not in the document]
a-corni: There will be tests that will fail if the default value changes (there is at least one test with an SLM that does not define the dmm_name)

pulser-core/pulser/sequence/sequence.py
a-corni: So I tried with moving the base message above, there was still the old error because the issue does not come from the base message but from the fact that an error message is defined while a corresponding channel is found. I can spend more time figuring out what is really happening, but I find it normal to have an error message because the DMM channels don't have the same type as the Rydberg/Raman/XY channels - however I don't fully understand why we did not witness this issue before, personally I have never tried to switch a device with a Rydberg and two Raman channels for instance.

pulser-core/pulser/sequence/sequence.py
```diff
@@ -728,27 +770,38 @@ def check_retarget(ch_obj: Channel) -> bool:
...
                         old_ch_obj, param_
                     ):
-                        strict_error_message = strict_error_message or (
-                            base_msg + f" with the same {param_}."
+                        new_message = base_msg + f" with the same {param_}."
+                        new_message += (
+                            "Note: This might be too strict."
+                            if param_ == "bottom_detuning"
+                            else ""
+                        )
```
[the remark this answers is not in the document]
a-corni: Aah ok I will modify this haha

a-corni:
About the if: They are needed, it is actually an issue that was not reported before (and it is independent of the dmm, I can solve it in another PR if your want). The issue is basically that the wrong error message is displayed, see below with a raman channel declared twice, what is actually failing is the matching of `raman_1`,  not `raman`.
image

pulser-core/pulser/sequence/sequence.py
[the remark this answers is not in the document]
HGSilveri: Sure, at the end of the file for example

pulser-core/pulser/sequence/sequence.py
a-corni: I also want to add a test for the Mappable Register !
I will first solve the PR on the "Fix Error Messages for switch device" separately.
I will also change the name of `modulate_det_map` to` add_to_dmm`in `add_dmm_sequence`

a-corni:
@HGSilveri I think we are done with this PR, it does everything we wanted and beyond.

tests/test_sequence.py
```diff
@@ -1584,27 +1616,50 @@ def test_mappable_register(patch_plt_show):
...
-    with pytest.warns(UserWarning, match=warn_message_global):
+        + f"Target qubits of channel {ch} will be defined in build."
+        for ch in global_channels
+    ]
+    with pytest.warns(UserWarning) as records:
         seq.__str__()
+    assert len(records) == len(global_channels)
+    assert [
+        str(records[i].message) for i in range(len(global_channels))
+    ] == warn_message_rydberg
```
HGSilveri, line 1633: Very thorough 👍

HGSilveri (approved):
Very nice!

PR #583, opened 2023-09-20
Use jobs order provided by the cloud

### LINKED ISSUE #570: Adopt built-in job ordering from `pasqal-cloud 0.3.5`
`pasqal-cloud 0.3.5` added job ordering to the batch, so now we can get rid of how we are conserving the job order in the `PasqalCloud` and rely on `Batch.ordered_jobs` instead.

HGSilveri:
Closes #570 .

### DIFF
--- a/tests/test_pasqal.py
+++ b/tests/test_pasqal.py
@@ -85,7 +85,7 @@ def mock_batch(mock_job, seq):
     class MockBatch:
         id = "abcd"
         status = "DONE"
-        jobs = {mock_job.id: mock_job}
+        ordered_jobs = [mock_job]
         sequence_builder = seq_.to_abstract_repr()
 
     return MockBatch()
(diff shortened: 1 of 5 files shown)


tests/test_pasqal.py
```diff
@@ -208,16 +208,6 @@ def test_submit(fixt, parametrized, emulator, seq, mock_job):
...
-    job_params[0]["runs"] = 1
-    with pytest.raises(RuntimeError, match="Failed to find job ID"):
-        # Job runs don't match MockJob
-        fixt.pasqal_cloud.submit(
-            seq,
-            job_params=job_params,
-            emulator=emulator,
-            config=config,
-        )
-
```
a-corni, line 220: Why is this test no longer relevant ?

HGSilveri: Because we no longer try to determine the jobs order (the code where this was being raised was deleted)

a-corni: Okay, that was my guess... So it's no longer an issue that the number of runs in job_params is different than number of runs in "mock_job" ? The two issues seemed independent to me, but I am not I understood the test in the first place...

HGSilveri: That manual change was just a way for the job params to not match anymore and thus trigger the fail to find the job ID.

PR #585, opened 2023-09-20
Deprecate legacy serializer + Improve error messages

HGSilveri:
- Improves serialization error messages following user feedback
- Deprecates `Sequence.serialize()` and `Sequence.deserializer()`. These will still be supported has private method because other packages need them, but the public methods will be removed in v1
- Updates the serialization tutorial to show only the abstract representation


tutorials/advanced_features/Serialization.ipynb
```diff
@@ -1,182 +1,162 @@
 {
```
a-corni, line 1: <p>I think something is missing in the last sentence "as we'll <strong>see" </strong></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pasqal-io/Pulser/pull/585/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

HGSilveri: <p>Good catch!</p>

a-corni (approved):
It needs updating but then I approve !

PR #586, opened 2023-09-22
Adding register_is_from_calibrated_layout and is_calibrated_layout to Device

### LINKED ISSUE #577: Add function to validate that a given sequence's register matches the pre calibrated layout of a device
Given a `Sequence` and a `Device` object, I would like to be able to assert that the register of the sequence derives from the pre calibrated layouts of the device (if any). If no pre-calibrated layouts on the device then the function would return always True.

We initially thought about implementing our own logic but I think it would be better to have a properly tested function in the pulser repo directly that can be reused in any project and would be kept in sync with future developments of Pulser. For reference, here's the logic we initially went for:

```python
if device.pre_calibrated_layouts:
    if (
        sequence.register.layout is None
        or sequence.register.layout.static_hash()
        not in [
            layout.static_hash() for layout in device.pre_calibrated_layouts
        ]
    ):
        raise ValueError(
            "The sequence's register should derive from one of the device pre-calibrated layouts."
        )
```

a-corni:
As suggested by @HGSilveri:
Add two new methods to `Device`:
- [ ] `Device.is_calibrated_layout(layout: RegisterLayout)`: Checks whether a layout is within the calibrated layouts.
- [ ] `Device.register_is_from_calibrated_layout(register: BaseRegister|MappableRegister)`: Checks whether the register was constructed from a pre-calibrated layout (ie checks if it has a layout AND `is_calibrated_layout(layout)`)
As discussed, returns a boolean and not an error.
register_is_from_calibrated_layout works with BaseRegister (hence Register and Register3D) and MappableRegister (always has a layout in this case) because the type of the register in a sequence can be of type one of these two types. 

Closes #577

### DIFF
--- a/pulser-core/pulser/devices/_device_datacls.py
+++ b/pulser-core/pulser/devices/_device_datacls.py
@@ -538,6 +538,47 @@ class Device(BaseDevice):
         """Register layouts already calibrated on this device."""
         return {str(layout): layout for layout in self.pre_calibrated_layouts}
 
+    def is_calibrated_layout(self, register_layout: RegisterLayout) -> bool:
+        """Checks whether a layout is within the calibrated layouts.
+
+        Args:
+            register_layout: the RegisterLayout to check.
+
+        Returns:
+            True if register_layout is found among calibrated_register_layouts,
+            False otherwise.
+        """
+        for (
+            name,
+            calibrated_layout,
+        ) in self.calibrated_register_layouts.items():
+            if register_layout == calibrated_layout:
+                return True
+        return False
+
+    def register_is_from_calibrated_layout(
+        self, register: BaseRegister | MappableRegister
+    ) -> bool:
+        """Checks whether a register was constructed from a calibrated layout.
+
+        If the register is a BaseRegister, checks that it has a layout. If so,
+        or if it is a MappableRegister, check that its layout is within the
+        calibrated layouts.
+
+        Args:
+            register_layout: the Register or MappableRegister to check.
+
+        Returns:
+            True if register has a layout and it is found among
+            calibrated_register_layouts, False otherwise.
+        """
+        if (
+            isinstance(register, BaseRegister)
+            and register._layout_info is None
+        ):
+            return False
+        return self.is_calibrated_layout(cast(RegisterLayout, register.layout))
+
     def to_virtual(self) -> VirtualDevice:
         """Converts the Device into a VirtualDevice."""
         params = self._params()
--- a/pulser-core/pulser/devices/_device_datacls.py
+++ b/pulser-core/pulser/devices/_device_datacls.py
@@ -542,16 +542,13 @@ class Device(BaseDevice):
         """Checks whether a layout is within the calibrated layouts.
 
         Args:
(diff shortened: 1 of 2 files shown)


pulser-core/pulser/devices/_device_datacls.py
```diff
@@ -538,6 +538,47 @@ def calibrated_register_layouts(self) -> dict[str, RegisterLayout]:
         """Register layouts already calibrated on this device."""
         return {str(layout): layout for layout in self.pre_calibrated_layouts}
 
+    def is_calibrated_layout(self, register_layout: RegisterLayout) -> bool:
+        """Checks whether a layout is within the calibrated layouts.
+
+        Args:
+            register_layout: the RegisterLayout to check.
```
HGSilveri, line 545: nit: 
```suggestion
            register_layout: The RegisterLayout to check.
```

pulser-core/pulser/devices/_device_datacls.py
```diff
@@ -538,6 +538,47 @@ def calibrated_register_layouts(self) -> dict[str, RegisterLayout]:
...
+            True if register has a layout and it is found among
+            calibrated_register_layouts, False otherwise.
+        """
+        if (
+            isinstance(register, BaseRegister)
+            and register._layout_info is None
+        ):
+            return False
+        return self.is_calibrated_layout(cast(RegisterLayout, register.layout))
+
```
HGSilveri, line 581: This will raise an `AttributeError` when `register_layout` is not a `MappableRegister` nor a `BaseRegister`.  I think a `TypeError` would be more appropriate in that case, so perhaps we can check the type of `register_layout` first.

pulser-core/pulser/devices/_device_datacls.py
```diff
@@ -538,6 +538,47 @@ def calibrated_register_layouts(self) -> dict[str, RegisterLayout]:
...
+        Args:
+            register_layout: the Register or MappableRegister to check.
+
+        Returns:
+            True if register has a layout and it is found among
+            calibrated_register_layouts, False otherwise.
+        """
+        if (
+            isinstance(register, BaseRegister)
+            and register._layout_info is None
```
HGSilveri, line 577: You can check directly
```suggestion
            and register.layout is None
```

pulser-core/pulser/devices/_device_datacls.py
```diff
@@ -572,12 +569,16 @@ def register_is_from_calibrated_layout(
             True if register has a layout and it is found among
             calibrated_register_layouts, False otherwise.
         """
-        if (
-            isinstance(register, BaseRegister)
-            and register._layout_info is None
+        if not isinstance(register, BaseRegister) or isinstance(
+            register, MappableRegister
```
HGSilveri, line 573:
```suggestion
        if not isinstance(register, (BaseRegister, MappableRegister))
```

HGSilveri (approved):
LGTM, let's wait for @MatthieuMoreau0

MatthieuMoreau0 (approved):
LGTM! Thank you!

PR #587, opened 2023-09-27
Add DMM, methods of RegisterLayout inherited from Traps, Result

### LINKED ISSUE #584: API reference updates
- [ ] Add `Result` and its respective subclasses
- [ ] Add `DetuningMap` and `DMM`

a-corni:
- [x] Add `Result` and its respective subclasses
- [x] Add `DetuningMap` 
- [x] Add `DMM`
Closes #584

### DIFF
--- a/docs/source/apidoc/core.rst
+++ b/docs/source/apidoc/core.rst
@@ -52,6 +52,17 @@ Special cases
   :members:
   :show-inheritance:
 
+DetuningMap
+-------------------
+
+A ``DetuningMap`` is associated to a ``DMM`` in a ``Sequence``. It links a set
+of weights to a set of trap coordinates. As such, it is intended to be defined
+by the user from a ``RegisterLayout``, a ``Register`` or a ``MappableRegister``
+using ``define_detuning_map``.
+
+.. autoclass:: pulser.register.detuning_map
+  :members:
+  :inherited-members:
 
 Pulse
 -------------------
--- a/docs/source/apidoc/core.rst
+++ b/docs/source/apidoc/core.rst
@@ -38,6 +38,7 @@ register can be defined at build time.
 
 .. autoclass:: pulser.register.register_layout.RegisterLayout
   :members:
+  :inherited-members:
 
 .. autoclass:: pulser.register.mappable_reg.MappableRegister
   :members:
@@ -114,6 +115,9 @@ Available Channels
    :members:
    :show-inheritance:
 
+.. automodule:: pulser.channels.dmm
+   :members:
+   :show-inheritance:
 
 EOM Mode Configuration
 ^^^^^^^^^^^^^^^^^^^^^^^^
@@ -129,3 +133,7 @@ Sampler
 .. automodule:: pulser.sampler.samples
    :members:
 
+Result
+------------------
+.. automodule:: pulser.result
+   :members:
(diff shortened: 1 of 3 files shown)


docs/source/apidoc/core.rst
```diff
@@ -38,6 +38,7 @@ register can be defined at build time.
```
HGSilveri, line 38: I noticed there's a typo in the paragraph above (hardware **provider** instead of "provided"). Can you change it in this PR?

docs/source/apidoc/core.rst
```diff
@@ -114,6 +126,9 @@ Available Channels
    :members:
    :show-inheritance:
 
+.. automodule:: pulser.channels.dmm
```
HGSilveri, line 129: Maybe here we could make a subsection for the DMM, even change the docstring at the top of the file to be more informative. As it is, it appears a bit out of the blue.

docs/source/apidoc/core.rst
```diff
@@ -129,3 +144,8 @@ Sampler
 .. automodule:: pulser.sampler.samples
    :members:
 
+Result
+------------------
+.. automodule:: pulser.result
+   :members:
+   :show-inheritance:
```
HGSilveri, line 151: Here we can also include `QutipResult`, maybe. It's a bit of an awkward fit under "Core Features" but maybe it's not so bad, what do you think?

docs/source/apidoc/core.rst
a-corni: Sure !

docs/source/apidoc/core.rst
a-corni: Ah okay I see the issue. I still think that the dmm has its place here in the available channels. I suggest to have an autoclass instead an automodule. What do you think of this ?

docs/source/apidoc/core.rst
a-corni: Won't `QutipResult` be best in `simulation.rst` ?

docs/source/apidoc/core.rst
HGSilveri: Should be fine, yes. Just make sure the description in the class makes it clear it's not exactly a channel like the others

docs/source/apidoc/core.rst
a-corni: As we are presenting there `CoherentResults` and `NoisyResults`

HGSilveri: Yeah, it's a toss up to me, since it is a `Result` subclass. I'll let you choose :)

HGSilveri:
Do you need help figuring out what's up with the RTD build?

a-corni:
> Do you need help figuring out what's up with the RTD build?

Well I have a Warning raised when asking to display QutipResult. At the moment I don't find a way to circumvent it.
image

a-corni:
I think it's coming from this attribute:
image

HGSilveri:
> > Do you need help figuring out what's up with the RTD build?
> 
> Well I have a Warning raised when asking to display QutipResult. At the moment I don't find a way to circumvent it. image

`QubitId` is just a type alias, try to import it in `qutip_result.py` (and add an ignore to the flake8 error):
```python
from pulser.register import QubitId   # noqa: F401
```

HGSilveri (approved):
Nice!

PR #588, opened 2023-09-27
Override `optimal_detuning_off` on stored calls

HGSilveri:
In non-parametrized calls to `Sequence.enable_eom_mode()`, overrides the value of `optimal_detuning_off` with the chosen `detuning_off` in the stored call. 
This allows us to include the `detuning_off` information in the abstract representation without changing the schema and aids `Sequence.switch_device()` is more closely resembling the original sequence.

### DIFF
--- a/pulser-core/pulser/sequence/sequence.py
+++ b/pulser-core/pulser/sequence/sequence.py
@@ -648,7 +648,13 @@ class Sequence(Generic[DeviceType]):
     def switch_device(
         self, new_device: DeviceType, strict: bool = False
     ) -> Sequence:
-        """Switch the device of a sequence.
+        """Replicate the sequence with a different device.
+
+        This method is designed to replicate the sequence with as few changes
+        to the original contents as possible.
+        If the `strict` option is chosen, the device switch will fail whenever
+        it cannot guarantee that the new sequence's contents will not be
+        modified in the process.
 
         Args:
             new_device: The target device instance.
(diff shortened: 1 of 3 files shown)


pulser-core/pulser/sequence/sequence.py
```diff
@@ -648,7 +648,13 @@ def _config_detuning_map(
...
         self, new_device: DeviceType, strict: bool = False
     ) -> Sequence:
-        """Switch the device of a sequence.
+        """Replicate the sequence with a different device.
+
+        This method is designed to replicate the sequence with as few changes
+        to the original contents as possible.
+        If the `strict` option is chosen, the device switch will fail whenever
+        it cannot guarantee that the new sequence's contents will not be
+        modified in the process.
```
a-corni, line 657: Super nice !

tests/test_abstract_repr.py
```diff
@@ -1602,29 +1612,52 @@ def test_deserialize_eom_ops(self, correct_phase_drift):
...
                 del op["correct_phase_drift"]
-        _check_roundtrip(s)
+
         seq = Sequence.from_abstract_repr(json.dumps(s))
-        # init + declare_channel + enable_eom_mode
-        assert len(seq._calls) == 3
-        # add_eom_pulse + disable_eom
-        assert len(seq._to_build_calls) == 2
+        # init + declare_channel + enable_eom_mode (if not var_detuning_on)
+        assert len(seq._calls) == 3 - var_detuning_on
```
a-corni, line 1628: Why is there anything inside _calls if var_detuning_on ? Since the first operation (enable_eom_mode) is parametrized, it is stored in _to_build_calls, but afterwards the next operation should be added to _to_build_calls instead of _calls ?

a-corni: Thanks ! I had not noticed ! All good on my side 👍

a-corni (approved):

PR #595, opened 2023-10-09
Abstract representation fixes

HGSilveri:
- Fixes serialization of `phase_shift()` calls with custom a `basis`
- Adds support for calls to `Sequence.target()` with multiple qubits


pulser-core/pulser/json/abstract_repr/serializer.py
```diff
@@ -230,10 +243,19 @@ def remove_kwarg_if_default(
...
                 target = convert_targets(data["qubits"])
             elif call.name == "target_index":
-                target = data["qubits"]
+                if isinstance(
+                    data["qubits"], pulser.parametrized.Parametrized
+                ):
+                    # The qubits are given as a variable
+                    target = data["qubits"]
+                else:
+                    target = cast(
```
madagra, line 256: @HGSilveri can you add a comment about the possible target types (like done above)?

tests/test_abstract_repr.py
```diff
@@ -260,7 +260,10 @@ def sequence(self, request):
         reg = Register(qubits)
         device = request.param
         seq = Sequence(reg, device)
-        seq.declare_channel("digital", "raman_local", initial_target="control")
+
+        seq.declare_channel(
+            "digital", "raman_local", initial_target=("control",)
```
madagra, line 265: If you need use the `tuple` here why not below as well in the "rydberg_local"?

madagra (approved):
Looks overall good to me, I have just a couple of minor comments.

pulser-core/pulser/json/abstract_repr/deserializer.py
```diff
@@ -213,6 +213,7 @@ def _deserialize_operation(seq: Sequence, op: dict, vars: dict) -> None:
         seq.phase_shift_index(
             _deserialize_parameter(op["phi"], vars),
             *[_deserialize_parameter(t, vars) for t in op["targets"]],
+            basis=op["basis"],
```
madagra, line 223: @HGSilveri What exactly were the implications of this bug for affected runs?

tests/test_abstract_repr.py
HGSilveri: I don't have to, but I can (the target can be a sequence of qubit IDs). I just made it a `tuple` to have both cases covered (ie when it's a single ID or a sequence of IDs - even though there is only one element here).

madagra: Sounds good to me.

pulser-core/pulser/json/abstract_repr/deserializer.py
HGSilveri: Basically the basis was defaulting to `digital` every time. What happened generally was that an error was raised since there was no `Raman` declared channel

awennersteen:
Do I understand that this is a backwards incompatible change and we need to cast `int to [int]`?

No objections, just want to be clear

HGSilveri:
> Do I understand that this is a backwards incompatible change and we need to cast `int to [int]`?
> 
> No objections, just want to be clear

No, `int` is still valid, we just support [int] now too (which was already the case in Pulser, just not in the abstract repr).

awennersteen:
> No, `int` is still valid, we just support `[int]` now too (which was already the case in Pulser, just not in the abstract repr).

So now the abstract representation can come with both `int` and `List[int]` types?

HGSilveri:
> > No, `int` is still valid, we just support `[int]` now too (which was already the case in Pulser, just not in the abstract repr).
> 
> So now the abstract representation can come with both `int` and `List[int]` types?

For a `"target"` operation, the "target" field will either be an `int` or `list[int]` (of size >= 1). It can also be a variable or a variable item, in case you are dealing with parametrized sequences.

HGSilveri:
@awennersteen Thinking about it some more, it would be less disruptive if we continue serializing `list[int]` into an `int` whenever there is only one entry. `list[int]` will still need to be handled when there are multiple targets but that's a rarer case. Wdyt?

awennersteen:
>it would be less disruptive if we continue serializing `list[int]` into an `int` whenever there is only one entry.

Indeed it would be less disruptive, but I guess a question is also what do we think is the best thing going forward?
should we in the near future make it always a list, to have just a single type, instead of two?

Regardless, anyone implementing the abstract implementation will have to implement support for the `[int]` case

HGSilveri:
> > it would be less disruptive if we continue serializing `list[int]` into an `int` whenever there is only one entry.
> 
> Indeed it would be less disruptive, but I guess a question is also what do we think is the best thing going forward? should we in the near future make it always a list, to have just a single type, instead of two?
> 
> Regardless, anyone implementing the abstract implementation will have to implement support for the `[int]` case

I think having it be just a list would indeed be the way to go at some point but it's a breaking change we're not ready for atm. By using `int` whenever possible we still retain compatibility with the largest possible set of sequences.

awennersteen:
Agree. Let's go with the `int` wherever possible to have as few sequences fail because of not sufficiently recent version of pulser running in a service.

awennersteen (approved):
lgtm

PR #597, opened 2023-10-13
Use trusted publishing on PyPI

HGSilveri:
- Changes the `publish.yml` workflow to use a [Trusted Publisher](https://docs.pypi.org/trusted-publishers/using-a-publisher/) instead of a token (thus avoiding storing tokens as secrets)
- Replaces `twine` with the official [PyPI Publish Github Action](https://github.com/pypa/gh-action-pypi-publish)
- Adds a `release` Github environment to better manage the release process

The changes were tested through a dummy deployment to TestPyPI and then replicated for the PyPI case (without a dummy deployment so as to not pollute the official repo).


madagra (approved):
Looks good to me, @HGSilveri.

PR #598, opened 2023-10-13
Add `wait` option to remote backends

### LINKED ISSUE #596: Wait option for Backend.run( ) function
Hello ! In the Cloud SDK, there is an option wait in sdk.create_batch. Setting wait = True keeps your code running until the results are available on the Cloud platform (i.e. batch is marked as Done). It is very useful to build a continuous communication loop between a local computer and the distant Emulator/QPU. It is possible to make the code artificially wait but since the option is already in the Cloud SDK and that the Backend objects are wrappers around that, I feel like it would be useful and easy to keep that argument. Thank you a lot !

HGSilveri:
Fixes #596 .

### DIFF
--- a/pulser-core/pulser/backend/qpu.py
+++ b/pulser-core/pulser/backend/qpu.py
@@ -36,7 +36,10 @@ class QPUBackend(RemoteBackend):
                 is parametrized, the values for all the variables necessary
                 to build the sequence must be given in it's own mapping, for
                 each job, under the 'variables' field.
-            wait: Whether to wait until the results become available.
+            wait: Whether to wait until the results of the jobs become
+                available.  If set to False, the call is non-blocking and the
+                obtained results' status can be checked using their `status`
+                property.
 
         Returns:
             The results, which can be accessed once all sequences have been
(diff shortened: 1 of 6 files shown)


pulser-core/pulser/backend/qpu.py
```diff
@@ -34,6 +36,7 @@ def run(self, job_params: list[JobParams] | None = None) -> RemoteResults:
                 is parametrized, the values for all the variables necessary
                 to build the sequence must be given in it's own mapping, for
                 each job, under the 'variables' field.
+            wait: Whether to wait until the results become available.
```
madagra, line 39: Maybe add a bit more information: "Whether wait until the results of the jobs become available. If set to False, the call is non-blocking and the results must be checked using the `results()` property."

madagra (approved):
Looks good to me, just a minor comment.

pulser-core/pulser/backend/qpu.py
a-corni: Discussing with @LucasGitQ, we propose "Whether wait until the results of the jobs become available. If set to False, the call is non-blocking and the obtained results must be checked using their `status` property."

HGSilveri: Thanks for the suggestions!

a-corni (approved):

PR #599, opened 2023-10-18
[FIX] Allow parametrized phase in EOM mode

HGSilveri:
I stumbled upon this bug when trying to make EOM pulses with a parametrized phase.


a-corni (approved):
Nice find ! Looks good to me :)

PR #602, opened 2023-10-26
Refactoring `QutipEmulator` class

### LINKED ISSUE #600: Refactor the noise implementation in `pulser_simulation`

a-corni:
Add a `Hamiltonian` class to generate an hamiltonian using the samples and the noise model each time a new noise config is added or each time `generate_hamiltonian` is called.

Closes #600


pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -134,26 +130,10 @@ def __init__(
...
-        self.dim: int
-        self._eval_times_array: np.ndarray
-        self._bad_atoms: dict[Union[str, int], bool] = {}
-        self._doppler_detune: dict[Union[str, int], float] = {}
-
-        # Initializing sampling and evalutaion times
+        # Testing sampling
         if not (0 < sampling_rate <= 1.0):
             raise ValueError(
                 "The sampling rate (`sampling_rate` = "
```
a-corni, line 139: I have hesitated in moving these tests to the init of Hamiltonian. What do you think of this ?

tests/test_simulation.py
```diff
@@ -663,12 +663,12 @@ def test_config():
 
 
 def test_noise(seq, matrices):
-    np.random.seed(3)
+    np.random.seed(2)
     sim2 = QutipEmulator.from_sequence(
         seq, sampling_rate=0.01, config=SimConfig(noise=("SPAM"), eta=0.9)
     )
     assert sim2.run().sample_final_state() == Counter(
-        {"000": 857, "110": 73, "100": 70}
```
a-corni, line 671: I decided to modify the test to pass it.
- I changed the seed because the result I obtained with a seed of 3 did not reflect the presence of noise, it was `{"000": 1000}`.
- I can't figure out what is different now vs before, it seems that the number of times the noise is updated is the same and the number of time the code is run as well. Yet to me the difference in the result of the counter means that the number of calls to random functions is different.
- To me, the statistics seem the same.

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -1156,6 +585,9 @@ def from_sequence(
             evaluation_times,
         )
 
+    def __getattr__(self, name: str) -> Any:
+        return getattr(self.hamiltonian, name)
```
HGSilveri, line 589: I'm not a big fan of how much we are relying on this, it's somewhat hacky. This is giving full access to `Hamiltonian` even though it should be an internal class. Furthermore, it makes public methods not appear in the docs anymore. 
I think some public methods still belong in ` QutipEmulator` (namely those related to the config), For those who don't (eg ` build_operator`), we can make special cases here, for example:
 ```suggestion
    def __getattr__(self, name: str) -> Any:
        if name in exposed_methods:
            return getattr(self.hamiltonian, name)
        else:
            raise AttributeError(...)
```
I know this will probably break a few tests that access private attributes but this is fine imo.

File: pulser-simulation/pulser_simulation/simulation.py [hunk 4]
```diff
@@ -164,22 +144,18 @@ def __init__(
             raise ValueError(
                 "`sampling_rate` is too small, less than 4 data points."
             )
-        self._sampling_rate = sampling_rate
-        self.sampling_times = self._adapt_to_sampling_rate(
-            # Include extra time step for final instruction from samples:
-            np.arange(self._tot_duration + 1, dtype=np.double)
-            / 1000
+        # Sets the config as well as builds the hamiltonian
+        self.hamiltonian = Hamiltonian(
```
HGSilveri, line 148: Are we sure about ` hamiltonian` being a public attribute? To me, it makes more sense to hide it

tests/test_simulation.py
HGSilveri: I wouldn't stress about it, this change looks fine to me

pulser-simulation/pulser_simulation/hamiltonian.py
```diff
@@ -0,0 +1,640 @@
...
+    r"""Generates Hamiltonian from a sampled sequence and noise.
+
+    Args:
+        samples_obj: A sampled sequence whose ChannelSamples have same
+            duration.
+        qdict: A dictionary associating coordinates to qubit ids.
+        device: The device specifications.
+        sampling_rate: The fraction of samples that we wish to extract from
+            the samples to simulate. Has to be a value between 0.05 and 1.0.
+        config: Configuration to be used for this simulation.
```
HGSilveri, line 44: Ideally for me, ` Hamiltonian` wouldn't deal with `SimConfig` that has irrelevant info like `runs`, `samples_per_run` and `solver_options`. Instead, it would deal only with `NoiseModel`, which can be easily obtained from a `SimConfig` using `SimConfig.to_noise_model()`.

This is something that can also be tackled in a second PR though

pulser-simulation/pulser_simulation/hamiltonian.py
```diff
@@ -0,0 +1,640 @@
...
+        indices = np.linspace(
+            0,
+            len(full_array) - 1,
+            int(self._sampling_rate * self.duration),
+            dtype=int,
+        )
+        return cast(np.ndarray, full_array[indices])
+
+    @property
+    def config(self) -> SimConfig:
```
HGSilveri, line 104: I'd actually keep all `config` related methods and properties in `QutipEmulator`

HGSilveri (commented):
I have a few suggestions but it looks good overall!

pulser-simulation/pulser_simulation/hamiltonian.py
```diff
@@ -0,0 +1,640 @@
...
+            )
+        prev_config = self.config if hasattr(self, "_config") else SimConfig()
+        self._config = cfg
+        if not ("SPAM" in self.config.noise and self.config.eta > 0):
+            self._bad_atoms = {qid: False for qid in self._qid_index}
+        if "doppler" not in self.config.noise:
+            self._doppler_detune = {qid: 0.0 for qid in self._qid_index}
+        # Noise, samples and Hamiltonian update routine
+        self._construct_hamiltonian()
+        self._build_collapse_operators(prev_config)
```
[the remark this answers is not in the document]
a-corni: Moving it to 226 and getting rid of `prev_config` sounds definitely like a good idea !

pulser-simulation/pulser_simulation/simulation.py
a-corni: Ah I only focused on the issue of keeping backwards compatibility without making too many changes, but I agree it's hacky and I don't feel it is stable ground to keep developing QutipEmulator and Hamiltonian in the future.
I will move the properties and methods llinked with config to QutipEmulator, as well as modify the config type of Hamiltonian to be NoiseModel.
I am thinking of getting rid of the get_attr in QutipEmulator, and repeat all the public methods of Hamiltonian in QutipEmulator - will be perhaps more heavy but surely more explicit

pulser-simulation/pulser_simulation/simulation.py
a-corni: I am sure of nothing haha I made it public because I tought it could be useful to access this attribute, as well as I wanted to distinguish between _hamiltonian (inner property of hamiltonian) and hamilotonian, but since I am not testing it and since `_hamiltonian` was a private property here, let's keep it private.

pulser-simulation/pulser_simulation/hamiltonian.py
```diff
@@ -0,0 +1,638 @@
...
+        if "dephasing" in diff_noise_set:
+            param_dict["dephasing_prob"] = config.dephasing_prob
+        if "depolarizing" in diff_noise_set:
+            param_dict["depolarizing_prob"] = config.depolarizing_prob
+        if "eff_noise" in diff_noise_set:
+            param_dict["eff_noise_opers"] = config.eff_noise_opers
+            param_dict["eff_noise_probs"] = config.eff_noise_probs
+        # update runs:
+        param_dict["runs"] = config.runs
+        param_dict["samples_per_run"] = config.samples_per_run
```
HGSilveri, line 283: I don't see the point of having this complexity here, I think it should stay in `QutipEmulator` where it originally was. Plus, 'runs' and 'samples_per_run' are not part of the noise model

Since Hamiltonian is an internal class, I think we can just skip this method altogether and set a new config from `QutipEmulator` instead

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -334,55 +232,35 @@ def add_config(self, config: SimConfig) -> None:
...
 
     def show_config(self, solver_options: bool = False) -> None:
         """Shows current configuration."""
-        print(self._config.__str__(solver_options))
+        print(self.config.__str__(solver_options))
 
     def reset_config(self) -> None:
         """Resets configuration to default."""
-        self.set_config(SimConfig())
+        self._hamiltonian.set_config(NoiseModel())
```
HGSilveri, line 253: Just to be safe, I would rather do 
```suggestion
        self._hamiltonian.set_config(SimConfig().to_noise_model())
```

tests/test_simulation.py
```diff
@@ -822,18 +838,20 @@ def test_add_config(matrices):
...
+        noise=(
+            "SPAM",
+            "doppler",
+            "eff_noise",
+        ),
+        eff_noise_opers=[matrices["I"], matrices["X"]],
+        eff_noise_probs=[0.4, 0.6],
+        temperature=20000,
     )
+    print("init", config)
```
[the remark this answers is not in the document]
a-corni: Very sorry about these

pulser-simulation/pulser_simulation/hamiltonian.py
a-corni: I agree I was not a fan of having add_config in Hamiltonian

a-corni: Hum, however, I see `runs` and `samples_per_run` as being attributes of the NoiseModel class... https://github.com/pasqal-io/Pulser/blob/develop/pulser-core/pulser/backend/noise_model.py#L79

HGSilveri: Yes, I forgot they were there, I probably confused them with `quitp_options`, sorry

HGSilveri (approved):
LGTM, great job! I think we can merge and then change the target of PR #616 , what do you think?

PR #605, opened 2023-11-08
Fix use of spacing with a device of min_atom_distance=0 in Register.max_connectivity

### LINKED ISSUE #603: Register.max_connectivity: Allow to use device with minimal distance=0 if spacing>0
When building a register using `Register.max_connectivity`, we first check that minimal distance is greater than 0 before taking actions if spacing is None or checking the validity of spacing with the minimal distance.

**Error**: Due to this, `Register.max_connectivity(3, device=MockDevice, spacing=4)` raises an error because the minimal distance is 0, whereas the code is perfectly able to run and propose a result.

**Proposition**: Do in this order:
- if spacing is None -> replace it by min_atom_distance
- otherwise check that it is strictly greater than the min_atom_distance
- if spacing is 0 -> raise an Error (as you say it does not make sense)

a-corni:
Checks that spacing is greater than zero after the definition of spacing.
Fixes #603

### DIFF
--- a/pulser-core/pulser/register/register.py
+++ b/pulser-core/pulser/register/register.py
@@ -253,11 +253,6 @@ class Register(BaseRegister, RegDrawer):
                 f" ({device.max_atom_num})."
             )
 
-        if not device.min_atom_distance > 0.0:
-            raise NotImplementedError(
-                "Maximum connectivity layouts are not well defined for a "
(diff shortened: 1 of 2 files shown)


Louis-PaulHenry (approved):
Yes! This should do the trick!
Thx!

PR #610, opened 2023-11-10
Delete summing constraint in DetuningMap

### LINKED ISSUE #609: Delete constraint on sum of weights in DetuningMap
The constraint on the sum of weights in DetuningMap (sum of weights = 1) was introduced because detuning sent via the DMM was thought to be redistributed between the atoms. 

In fact, as underlined by @Lassabliere, some other physical effects are at stake and limit the minimum detuning of each atom more than the minimum detuning that can be sent via the laser.

Moreover, deleting this constraint eases the use of DMM and DetuningMap.

a-corni:
Deleting the condition "sum of weights = 1" in DetuningMap.
Condition is replaced by "weights are between 0 and 1".
Impacts the creation of the DetuningMap associated with an SLM (used to be with weights = 1/(#number of masked qubits), is now created with weights = 1 for masked qubits, 0 otherwise).
Notebooks are updated.
Is proposed to be a hotfix to advertise about this change among the developers.

**Possible Improvement to this PR**: To really make people aware that the summing constraint has been lifted, we could show a DeprecationWarning if the sum of the weights is equal to 1 ?

Fixes #609


madagra:
@a-corni thank you for this PR. Can I ask why exactly this constraint has been lifted? Is it not valid anymore?

a-corni:
> @a-corni thank you for this PR. Can I ask why exactly this constraint has been lifted? Is it not valid anymore?

It has in fact never been valid, @Lassabliere just realized that there had been a misunderstanding while discussing with a user who was telling him about this constraint.

madagra (commented):
@a-corni I am not in favor of fully removing the constraint on the weights. The alternative solution discussed offline might be more appropriate. For what concerns the actual changes in this PR, they look correct to me.

a-corni:
> @a-corni I am not in favor of fully removing the constraint on the weights. The alternative solution discussed offline might be more appropriate. For what concerns the actual changes in this PR, they look correct to me.

@madagra everything should be now tackled. @Lassabliere, bottom_detuning in notebook and devices is updated to -20 * 2 * pi rad/µs and global bottom detuning is -2000 * 2 * pi rad/µs.
 
In the end we don't have to delete the condition of the sum of the weights being equal to 1, we only need a constraint on the local detuning and a constraint on the global detuning. However, I think this implementation will help the user since it will be closer to how Local channels used to work (with weights equal to 1 or 0, you can define directly the detuning you want to apply on each qubit), even though this is making us drive away from the initial concept of a Global channel (where the detuning defined by the user should be between `global_bottom_detuning` and 0).

Lassabliere:
It won't be very user-friendly to work with weights.

For example, if you're working with 30 atoms and you have a DMM with a total available light shift of -2pi x 2000 MHz. In general, we want to work with detunings of the order of - 2pi x 1 MHz on the atoms, so you need to set very small weights [0.0005, ....]. And the algorithm developers will think in terms of local light-shift and not in term of weights (except for ramps maybe). Or, you can create maybe a function that gives weights for a desired light-shift.

Why don't you want to work with an additional term $-\hbar^2/2 \sum_i \Delta_i \sigma^z_i$ instead of what we have currently with a condition $\sum_i \Delta_i \le \Delta_{tot}$ ?

a-corni:
Hey @Lassabliere,

There might have been a misunderstanding with my previous message. I definitely agree with your point, and your example highlights it very well 😄 

Indeed the former implementation (defining the total lightshift applied on the qubits, applying on it weights whose sum is equal to 1 to obtain the detuning applied on each qubit) has the issue that the weights will be small if the number of atoms that have to receive light is large (to apply a detuning of -2pi rad/µs on each atom, you need to define a waveform with detuning -2pi30 rad/µs and a DetuningMap with weights $`1/30 \approx 0.033`$ on each atom).

Now the implementation under scrutiny here defines weights indeed, but that will generally be close to 1. To perform the operation you describe, we now send a waveform of detuning -2pi rad/µs on a DetuningMap with weights 1 on each atom, and we now internally check that $`\sum_{i=1}^30 -2pi = -60pi < -2 pi 2000`$. I find this very user-friendly. Another advantage of this solution is that whatever user did before with sum = 1 will still work.

Lassabliere:
My bad, I didn't read everything carefully. It looks ok for me!

pulser-core/pulser/channels/dmm.py
```diff
@@ -31,16 +32,20 @@ class DMM(Channel):
...
 
     Note:
         The protocol to add pulses to the DMM Channel is by default
         "no-delay".
 
     Args:
-        bottom_detuning: Minimum possible detuning (in rad/µs), must be below
-            zero.
+        bottom_detuning: Minimum possible detuning for each atom (in rad/µs),
+            must be below zero.
```
madagra, line 46: Maybe call it `local_bottom_detuning`? Furthermore, this is now just a single value for all qubits, right? So we only have one local bottom detuning and one global bottom detuning and that's it.

pulser-core/pulser/channels/dmm.py
```diff
@@ -72,26 +89,47 @@ def basis(self) -> Literal["ground-rydberg"]:
...
+            < self.bottom_detuning
         ):
             raise ValueError(
-                "The detuning goes below the bottom detuning "
-                f"of the DMM ({self.bottom_detuning} rad/µs)."
+                "The detunings on some atoms go below the local bottom "
+                f"detuning of the DMM ({self.bottom_detuning} rad/µs)."
+            )
+        # Check that distributed detuning is above global_bottom_detuning
+        if self.global_bottom_detuning is not None and np.any(
```
madagra, line 126: Why do you need the `np.any` here? I thought the sum would yield just a number.

pulser-core/pulser/json/abstract_repr/schemas/device-schema.json
```diff
@@ -39,6 +39,13 @@
           "description": "Time taken to change the target (in ns).",
           "type": "null"
         },
+        "global_bottom_detuning": {
+          "description": "Minimum possible detuning of the whole channel (in rad/µs), must be below zero.",
```
madagra, line 43:
```suggestion
          "description": "Minimum possible detuning of the whole DMM channel (in rad/µs), must be below zero.",
```

pulser-core/pulser/json/abstract_repr/schemas/device-schema.json
```diff
@@ -1435,7 +1443,7 @@
           "type": "string"
         },
         "bottom_detuning": {
-          "description": "Minimum possible detuning (in rad/µs), must be below zero.",
+          "description": "Minimum possible detuning per trap (in rad/µs), must be below zero.",
```
madagra, line 1446:
```suggestion
          "description": "Minimum possible detuning per trap (in rad/µs) for DMM channels, must be below zero.",
```

pulser-core/pulser/channels/dmm.py
a-corni: I thought of keeping `bottom_detuning` for backwards compatibility and because I was afraid `local_bottom_detuning` would be too long, but I guess the first reason does not really matter since it is a patch of the version that introduced that variable name - better to change it now than later. So I am fine with that change - will just imply modifying the schema

a-corni: I will also modify the comment, there has always been only one value for bottom_detuning and global_bottom_detuning

pulser-core/pulser/channels/dmm.py
a-corni: There is indeed no need of this np.any ! Thanks for noticing this :)

a-corni: Ah I misread round_detuning is still an array, but we can circumvent the use of an any by taking the minimum value of round_detuning.

pulser-core/pulser/channels/dmm.py
```diff
@@ -72,26 +89,50 @@ def basis(self) -> Literal["ground-rydberg"]:
...
+                f"detuning of the DMM ({self.bottom_detuning} rad/µs)."
+            )
+        # Check that distributed detuning is above total_bottom_detuning
+        if (
+            self.total_bottom_detuning is not None
+            and np.sum(detuning_map.weights) * min_round_detuning
+            < self.total_bottom_detuning
+        ):
+            raise ValueError(
+                "The applied detuning goes below the global bottom detuning "
```
HGSilveri, line 134:
```suggestion
                "The applied detuning goes below the total bottom detuning "
```

pulser-core/pulser/channels/dmm.py
```diff
@@ -72,26 +89,50 @@ def basis(self) -> Literal["ground-rydberg"]:
     def _undefined_fields(self) -> list[str]:
         optional = [
             "bottom_detuning",
+            "total_bottom_detuning",
```
HGSilveri, line 92: What do you think of allowing `total_bottom_detuning` do be left as `None` in physical devices by modifying `DMM.is_virtual()`? I would be in favor of doing so temporarily and raising a `DeprecationWarning` telling the user to define it.  This would allow us to keep backwards compatibility from 0.15 to 0.16. 
It would also involve not defining it for now in Chadoq2 and making the serializer skip this parameter when it is not `None`.
What do you think?

HGSilveri (commented):
I still need to review part of the changes but I wanted to get these comments out of the way first so we can discuss them

pulser-core/pulser/channels/dmm.py
```diff
@@ -31,16 +32,20 @@ class DMM(Channel):
     (of zero amplitude and phase). These Pulses are locally modulated by the
     weights of a `DetuningMap`, thus providing a local control over the
     detuning. The detuning of the pulses added to a DMM has to be negative,
-    between 0 and `bottom_detuning`. Channel targeting the transition between
-    the ground and rydberg states, thus encoding the 'ground-rydberg' basis.
+    between 0 and `bottom_detuning`, and the sum of the weights multiplied by
+    that detuning has to be blow `global_bottom_detuning`. Channel targeting
```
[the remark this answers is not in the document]
HGSilveri: I made this comment before your latest commit but perhaps the typo is still there?

pulser-core/pulser/sequence/sequence.py
```diff
@@ -598,8 +613,8 @@ def config_detuning_map(
             ``MockDevice`` DMM can be repeatedly declared if needed.
 
         Args:
-            detuning_map: A DetuningMap defining atoms to act on and bottom
-                detuning to modulate.
+            detuning_map: A DetuningMap defining the amont of detuning each
```
HGSilveri, line 616:
```suggestion
            detuning_map: A DetuningMap defining the amount of detuning each
```

pulser-core/pulser/sequence/sequence.py
```diff
@@ -2172,12 +2196,42 @@ def _validate_and_adjust_pulse(
...
+                    (
+                        current_dmm,
+                        current_det_map,
+                    ) = self._get_dmm_id_detuning_map(call)
+                    if current_dmm == dmm_id:
+                        dmm_idx += 1
+                        current_dmm_name = (
+                            current_dmm
+                            if dmm_idx == 0
+                            else current_dmm + f"_{dmm_idx}"
```
HGSilveri, line 2226: As we've done before, it would be better if we don't spread these arbitrary conventions on naming throughout the code but rather keep them in the same place. If we ever have to change them, we're more likely to miss something if they are scattered around

pulser-core/pulser/sequence/sequence.py
```diff
@@ -2172,12 +2196,42 @@ def _validate_and_adjust_pulse(
...
-        channel_obj.validate_pulse(pulse)
+            dmm_id = _dmm_id_from_name(channel)
+            channel_obj = self.device.dmm_channels[dmm_id]
+            dmm_idx = -1
+            for call in self._calls[1:] + self._to_build_calls:
+                if (
+                    call.name == "config_detuning_map"
+                    or call.name == "config_slm_mask"
+                ):
+                    # Check whether dmm_name matches with channel
```
HGSilveri, line 2216: It seems that nothing happens if there is no match though, is that intended?

pulser-core/pulser/channels/dmm.py
madagra: Yes, I am in favor of this (especially because we saw nasty errors in the past due to changes in the serialization).

pulser-core/pulser/sequence/sequence.py
```diff
@@ -2172,12 +2196,42 @@ def _validate_and_adjust_pulse(
...
+                    if current_dmm == dmm_id:
+                        dmm_idx += 1
+                        current_dmm_name = (
+                            current_dmm
+                            if dmm_idx == 0
+                            else current_dmm + f"_{dmm_idx}"
+                        )
+                        if current_dmm_name == channel:
+                            detuning_map = current_det_map
+                            break
```
HGSilveri, line 2230: I think this block needs some more comments

madagra (commented):
@a-corni it looks good to me now, I will leave @HGSilveri to give the final approval.

pulser-core/pulser/channels/dmm.py
madagra: I don't think the length is an issue given that you have another variable called `global_bottom_detuning` which is equally long :)

pulser-core/pulser/channels/dmm.py
madagra: It would make things a bit clearer. Or you can just add a comment about it.

pulser-core/pulser/channels/dmm.py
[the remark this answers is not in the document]
a-corni: Oops sorry I did not see it :'(

HGSilveri (commented):
As far as I can tell we still need to be more thorough with the abstract representation tests (especially if we are to try and maintain backwards compatibility)

pulser-core/pulser/channels/dmm.py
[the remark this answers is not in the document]
HGSilveri: No no, I hadn't published it yet, I left it on standby. I made this remark because I noticed `global_bottom_detuning` and figured that could be confusing

tutorials/advanced_features/Local addressability with DMM.ipynb
```diff
@@ -13,8 +13,12 @@
    "metadata": {},
```
HGSilveri, line 13: <p>Seems to be missing a space before "weights" here</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pasqal-io/Pulser/pull/610/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

HGSilveri, line 13: <p>A forgotten <code>global_bottom_detuning</code> here </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pasqal-io/Pulser/pull/610/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

pulser-core/pulser/sequence/sequence.py
a-corni: Yes sorry I will try to explain it better !

tutorials/advanced_features/Local addressability with DMM.ipynb
a-corni: <p>Weird, it seems to be a bug in review nb...</p>

a-corni:
@HGSilveri my modifications are ready for review :)

tests/test_abstract_repr.py
```diff
@@ -1050,13 +1075,27 @@ def _get_expression(op: dict) -> Any:
 
 
 class TestDeserialization:
```
HGSilveri, line 1077: I see a lot of warning catches in this class that I guess we don't really care for. What if we use something like this to avoid them? https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-mark-filterwarnings

pulser-core/pulser/sequence/sequence.py
HGSilveri: Is it really okay that we are failing silently?

HGSilveri (commented):
Final remarks :)

HGSilveri:
> **Possible Improvement to this PR**: To really make people aware that the summing constraint has been lifted, we could show a DeprecationWarning if the sum of the weights is equal to 1 ?

I forgot to comment on this: I think it's too noisy, I'd be more in favor of adding a note in a docstring signaling this change.

pulser-core/pulser/devices/_devices.py
```diff
@@ -12,60 +12,64 @@
...
-            max_abs_detuning=2 * np.pi * 20,
-            max_amp=2 * np.pi * 10,
-            min_retarget_interval=220,
-            fixed_retarget_t=0,
-            max_targets=1,
-            clock_period=4,
-            min_duration=16,
-            max_duration=2**26,
+with warnings.catch_warnings():
+    warnings.simplefilter("ignore")
```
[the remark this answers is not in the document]
a-corni: Sure ! I think this will work !

tests/test_abstract_repr.py
a-corni: Okay I hesitated with using this, in the mean time I found it nice to know which function raises a warning and which does not, but now that we know nothing is weird I guess I can use filterwarnings

pulser-core/pulser/sequence/sequence.py
a-corni: It's would say it's not supposed to fail: the channel has been tested by `validate_channel` before being as being in `declared_channels`, and `declared_channels` is exactly built by looping over the calls.

a-corni:
> > **Possible Improvement to this PR**: To really make people aware that the summing constraint has been lifted, we could show a DeprecationWarning if the sum of the weights is equal to 1 ?
> 
> I forgot to comment on this: I think it's too noisy, I'd be more in favor of adding a note in a docstring signaling this change.

Yes in the end I think it's irrelevant, let's just keep it this way isn't it ?

pulser-core/pulser/sequence/sequence.py
HGSilveri: Ok, good enough for me

HGSilveri: Actually, let's add an assert in that case to ensure the detuning map is always found

HGSilveri (approved):
Great, LGTM!

PR #617, opened 2023-11-24
JSON serialization support for numpy integer types

HGSilveri:
- Adds JSON serialization support for numpy integer types
- Updates the configuration of some development tools

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,6 +1,6 @@
 [flake8]
 docstring-convention = google
-exclude = ./build, ./docs
+exclude = ./build, ./docs, ./__venv__
 extend-ignore =
     # D105 Missing docstring in magic method
     D105,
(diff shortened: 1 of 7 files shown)


a-corni (approved):
Looks good to me ! Just out ouf curiosity, why were the changes in pre-commit-config.yaml and the exclusion of __venv__ necessary ?

HGSilveri:
> Looks good to me ! Just out ouf curiosity, why were the changes in pre-commit-config.yaml and the exclusion of **venv** necessary ?

They are unrelated, it's just because I started working in a new setup where I'm setting a local `__venv__` folder to store my virtual environment. The pre-commit probably needed updating for a while but it only became apparent once I tried to install it from scratch. I guess that these changes would ideally be in a separate PR but I was lazy...

PR #619, opened 2023-11-29
FIX: Detuning modulation for custom EOM buffer times

### LINKED ISSUE #614: EOM buffer's detuning is drawn with a high value
In pulser 15.2 when a pulse with some positive detuning is just before an EOM sequence, the initial buffer time detuning has a weird shape (see figure below).

```
from pulser import Sequence
from pulser import Register
from pulser import Pulse

from pulser.waveforms import ConstantWaveform
from pulser.devices import AnalogDevice

# make the sequence
seq = Sequence(Register.square(3, spacing=6), AnalogDevice)

#-------------------------------#
ch0 = "ryd_global"
seq.declare_channel(ch0, "rydberg_global")
# pulse 1
a1 = ConstantWaveform(200,5.0)
d1 = ConstantWaveform(200,5.0)
p1 = Pulse(a1, d1, phase=0)
seq.add(p1, ch0)
# EOM
seq.enable_eom_mode(ch0, amp_on=5.0, detuning_on=0.0, optimal_detuning_off=-5.0)
seq.add_eom_pulse(ch0, duration=100, phase=0.0)
seq.disable_eom_mode(ch0)
#-------------------------------#

seq.draw()

```
image

HGSilveri:
Fixes #614 .

The example outlined in #614 now renders like this:

image

### DIFF
--- a/pulser-core/pulser/sampler/samples.py
+++ b/pulser-core/pulser/sampler/samples.py
@@ -209,11 +209,41 @@ class ChannelSamples:
             The modulated channel samples.
         """
 
-        def masked(samples: np.ndarray, mask: np.ndarray) -> np.ndarray:
+        def masked(
+            samples: np.ndarray,
+            mask: np.ndarray,
+            keep_end_values: bool = False,
+        ) -> np.ndarray:
             new_samples = samples.copy()
             # Extend the mask to fit the size of the samples
             mask = np.pad(mask, (0, len(new_samples) - len(mask)), mode="edge")
-            new_samples[~mask] = 0
+            if keep_end_values:
+                # Extracts the contiguous masked regions as [ti, tf] pairs
+                masked_regions = (
+                    np.flatnonzero(
+                        np.diff(
+                            np.r_[
+                                np.int8(0), (~mask).view(np.int8), np.int8(0)
+                            ]
+                        )
+                    )
+                    .reshape(-1, 2)
+                    .tolist()
+                )
+                for reg in masked_regions:
+                    if not (delta := reg[1] - reg[0]):
+                        # Should never happen, added as a precaution
+                        continue  # pragma: no cover
+                    # Set the masked region to the final sample value
+                    new_samples[reg[0] : reg[1]] = samples[reg[1] - 1]
+                    if reg[0] > 0:
+                        # If not starting from 0, set the first half of
+                        # the region to the first sample value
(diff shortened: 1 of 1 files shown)


a-corni (approved):
So the error was that the detuning during the EOM buffers was not constant ? Now the detuning during the EOM buffers is taken as detuning off right ? I am not sure I have the full motivation with only the comments line 304-305, but other than the rest is clear and fixes the issue so I approve :)

HGSilveri:
> So the error was that the detuning during the EOM buffers was not constant ? Now the detuning during the EOM buffers is taken as detuning off right ? I am not sure I have the full motivation with only the comments line 304-305, but other than the rest is clear and fixes the issue so I approve :)

Sorry, it's a bit complex so I didn't want to go into a lot of detail. In simple terms, what we do in these modulations is to modulate the full samples with the AOM and EOM bandwidths separately and then combine them to make the final modulated samples. Plus, there are also the EOM buffers which take a different modulation bandwidth that is calculated based of `custom_buffer_time` (if defined). The problem here was coming from the modulation bandwidth in the buffer being lower than the AOM bandwidth (when they are the same there is no issue). The best solution I could find was to modify the samples that are modulated with the "buffer bandwidth" so that we avoid showing a delayed transient regime.

a-corni:
> > So the error was that the detuning during the EOM buffers was not constant ? Now the detuning during the EOM buffers is taken as detuning off right ? I am not sure I have the full motivation with only the comments line 304-305, but other than the rest is clear and fixes the issue so I approve :)
> 
> Sorry, it's a bit complex so I didn't want to go into a lot of detail. In simple terms, what we do in these modulations is to modulate the full samples with the AOM and EOM bandwidths separately and then combine them to make the final modulated samples. Plus, there are also the EOM buffers which take a different modulation bandwidth that is calculated based of `custom_buffer_time` (if defined). The problem here was coming from the modulation bandwidth in the buffer being lower than the AOM bandwidth (when they are the same there is no issue). The best solution I could find was to modify the samples that are modulated with the "buffer bandwidth" so that we avoid showing a delayed transient regime.

Ah I think I get it, now when the modulated detuning in the buffer is computed, the detuning he sees from 0 to the starting time of the buffer is 0 and not 5 from 0 to 200 and 0 from 200 to the starting time of the buffer (we used to see the modulation of the constant detuning from 0 to 200 with the modulation bandwidth of the buffer)

PR #620, opened 2023-12-01
Deprecate legacy devices

### LINKED ISSUE #607: Remove Chadoq2 device
The `Chadoq2` device is currently deprecated and it should be removed.

HGSilveri:
Closes #607 .


pulser-core/pulser/devices/__init__.py
```diff
@@ -27,4 +32,5 @@
     Chadoq2,
     IroiseMVP,
     AnalogDevice,
+    DigitalAnalogDevice,
```
a-corni, line 35: What is the use of _valid_devices ? I guess it's okay to keep Chadoq2 and IroiseMVP there, otherwise we could have put them in a list `_deprecated_devices` ?

pulser-core/pulser/sequence/sequence.py
```diff
@@ -121,6 +122,25 @@ def __init__(
                 f"'device' must be of type 'BaseDevice', not {type(device)}."
             )
 
+        with warnings.catch_warnings():
+            warnings.simplefilter("always")
```
a-corni, line 126: Why do you go into catch_warnings to set the filter on always ? (what could happen if you only had the warnings.warn ?)

a-corni (commented):
Looking good ! Congrats on the huge work, I am reviewing the rtd now

pulser-core/pulser/devices/__init__.py
HGSilveri: It's an internal parameter so I didn't bother. It's only used here: https://github.com/pasqal-io/Pulser/blob/b18e43428e9aa5f7c79010a3bacfc9dd3f5263ef/pulser-core/pulser/json/supported.py#L75

a-corni (approved):
Looks good to me !

PR #625, opened 2023-12-13
Bump Jsonschema version to v4.17.3

a-corni:


a-corni:
Moving condition in requirements.txt to `jsonschema >= 4.17.3` since it is working for `jsonschema == 4.17.3`

pulser-core/requirements.txt
```diff
@@ -1,4 +1,5 @@
-jsonschema >= 4.18
+jsonschema >= 4.17.3
```
HGSilveri, line 1: Technically, we are supporting only 
```suggestion
jsonschema >= 4.18, ==4.17.3
```
It may seem equivalent, but `4.17.4` could be released if there was a bug in 4.17.3.

tests/test_abstract_repr.py
```diff
@@ -278,7 +282,15 @@ def validate_schema(instance):
...
     ) as f:
         schema = json.load(f)
-    jsonschema.validate(instance=instance, schema=schema, registry=REGISTRY)
+    if DEPRECATED_JSONSCHEMA:
+        assert RESOLVER is not None
+        jsonschema.validate(
+            instance=instance, schema=schema, resolver=RESOLVER
+        )
+    else:
+        jsonschema.validate(
```
HGSilveri, line 291: nit: Maybe you can add here 
```suggestion
        assert RESOLVER is None
        jsonschema.validate(
```

pulser-core/pulser/json/abstract_repr/validation.py
```diff
@@ -13,13 +13,23 @@
...
+from importlib.metadata import version
 from typing import Literal
 
 import jsonschema
 from referencing import Registry, Resource
 
-from pulser.json.abstract_repr import SCHEMAS
+from pulser.json.abstract_repr import SCHEMAS, SCHEMAS_PATH
 
+DEPRECATED_JSONSCHEMA = version("jsonschema") == "4.17.3"
```
HGSilveri, line 24: nit: I would call it `LEGACY_JSONSCHEMA` but this is fine too

HGSilveri (commented):
LGTM but we should probably do an hotfix instead, otherwise the guys will be blocked until we release v0.17. Wdyt?

tests/test_abstract_repr.py
a-corni: Yes I hesitated to add an assert here, I will add it ;)

pulser-core/requirements.txt
a-corni: Ah I did not think about that, thanks for noticing !

a-corni:
> LGTM but we should probably do an hotfix instead, otherwise the guys will be blocked until we release v0.17. Wdyt?

That's True, I also have someone asking for #616 to be released in a hotfix, should we first merge this PR and make a hotfix with #616 containing them both ?

HGSilveri:
> That's True, I also have someone asking for #616 to be released in a hotfix, should we first merge this PR and make a hotfix with #616 containing them both ?
> 
> Edit: I guess we can merge this PR into master first

I would make a new `hotfix/v0.16.1` branch, merge both PRs into it and then merge that to `master`. Every merge to master expects a bump in the version, so merging separately would normally require two version bumps.

pulser-core/requirements.txt
HGSilveri: It seems the `,` works as a logical `and`, not `or`. This will work in the event `4.17.4` is released but not `4.17.5`:
```
jsonschema >= 4.17.3, != 4.17.4
```
Perhaps it's a safe assumption to make? The alternative would be to change `LEGACY_JSONSCHEMA`, I'll post a comment with a suggestion.

pulser-core/pulser/json/abstract_repr/validation.py
```diff
@@ -13,13 +13,23 @@
...
+from importlib.metadata import version
 from typing import Literal
 
 import jsonschema
 from referencing import Registry, Resource
 
-from pulser.json.abstract_repr import SCHEMAS
+from pulser.json.abstract_repr import SCHEMAS, SCHEMAS_PATH
 
+LEGACY_JSONSCHEMA = version("jsonschema") == "4.17.3"
```
HGSilveri, line 24: This should work too:
```suggestion
LEGACY_JSONSCHEMA = ("4.18"  > version("jsonschema") >= "4.17.3")
```
Then we can simply the dependency

HGSilveri (approved):

PR #630, opened 2024-01-08
Fix pyright import errors

### LINKED ISSUE #551: Pulser symbols exports not friendly to pyright
Many symbols that are effectively _exported_ by pulser are _not discovered_ by pyright.
This is the case for `pulser.Sequence` or `pulser.Register` for instance.

Capture d’écran 2023-07-07 à 18 33 10

I suspected that other static analyser may experience the same problem. This is inconvenient when developing, as one looses autocompletions and suggestions.

I am unsure on the reason of this, but 
- pyright reports some `reportPrivateImportUsage` on `pulser-core/pulser/__init__.py`. I guess this is related to ruff F401s
- There are no dunder `__all__` in any of pulser `__init__.py`, apparently unrelated but who knows.

HGSilveri:
@lvignoli I'm not going to add `pyright` to our CI, please let me know if this works for you.

Fixes #551 .

### DIFF
--- a/pulser-core/pulser/__init__.py
+++ b/pulser-core/pulser/__init__.py
@@ -14,9 +14,12 @@
 
 """A pulse-level composer for neutral-atom quantum devices."""
 
-from pulser._version import __version__
-from pulser.pulse import Pulse
-from pulser.register import Register, Register3D
-from pulser.sequence import Sequence
+# Redundant imports are necessary to avoid errors with pyright
 
-from pulser.backend import QPUBackend  # isort: skip
+from pulser._version import __version__ as __version__
+from pulser.pulse import Pulse as Pulse
+from pulser.register import Register as Register
+from pulser.register import Register3D as Register3D
+from pulser.sequence import Sequence as Sequence
+
+from pulser.backend import QPUBackend as QPUBackend  # isort: skip


lvignoli:
This works, but only for the 5 top-level symbols that are `Sequence`, `Register`, etc.
To properly close #551, it needs to fix imports for the whole pulser package.

pulser-core/pulser/__init__.py
```diff
@@ -14,9 +14,58 @@
...
+    NoiseModel,
+    QPUBackend,
+)
 
-from pulser.backend import QPUBackend  # isort: skip
+# Exposing relevant submodules
+from pulser import (
+    waveforms as waveforms,
+    channels as channels,
+    register as register,
```
a-corni, line 41: Should we consider register.special_layouts as relevant ? Since people should define their Register from a RegisterLayout for it to work on a real QPU, perhaps having `pulser.TriangularLatticeLayout` and `pulser.SquareLatticeLayout` would make sense ?

HGSilveri: I always hesitate on this one because, in the envisioned usage, the user should not be creating their own register layouts but rather taking choosing from the ones that are pre-calibrated. In practice, though, this is not what we observe (at least for now) so I do see value in moving the layout classes at least one level up (to the `register` module level).

As for the `pulser` namespace, I am making this sort of divide where I only expose  what I think "standard" users should be accessing. Perhaps we want to start encouraging the creating of registers from layouts, in which case it does make sense to move these all the way there?

a-corni: That's true, I figured this out after writing my question... I think it could make sense to move it one level up, just such that the user does not have to look to deep in the module to fetch it (makes sense to get them with pulser.register.AnyLayout)

I think history goes toward encouraging the creation of registers from layouts, but until we have a device where the user can define custom layouts, users should pick the layout in the calibrated_layouts of the device and there is no need to advertise about them too much (I fear it might just induce wrong layouts being used).

HGSilveri: Alright, let's be cautious and go only one level up for now

a-corni (approved):

HGSilveri:
Thanks for the review @a-corni !
@lvignoli : Any remarks?

lvignoli:
Oh god it's happening 🥁

lvignoli:
@HGSilveri I trust you to decide what's best to expose, I do not have anything to add 😀

PR #636, opened 2024-02-01
Reduce imports on main tutorials

### LINKED ISSUE #633: Update tutorials to use the classes exposed in the `pulser` namespace
Now that we properly expose the user-facing classes in the `pulser` namespace, we can update the tutorials to take advantage of this.

In this way, we can get rid of imports like:

```python
from pulser.waveforms import BlackmanWaveform, ConstantWaveform, ...
```

and just use `pulser.BlackmanWaveform` instead.

HGSilveri:
I went with only the main tutorials, I think it's overkill to change it everywhere.

Closes #633 .


HGSilveri:
CI should pass after #637 is merged

a-corni (approved):
Feels... Different than the way I used to write 😅 I will try to reduce imports in the next notebooks and python file I will write :)

PR #637, opened 2024-02-01
[FIX] Restore compatibility with pytest 8.0.0

HGSilveri:
It seems that the new pytest 8.0.0 uncovered some mistakes in our tests... These are the fixes.


a-corni (approved):
Thanks for tackling this issue !

PR #639, opened 2024-02-01
Replace `Register.rotate()` with `Register.rotated()`

### LINKED ISSUE #628: Deprecate `Register.rotate()`
`Register.rotate()` is the only method that changes the register instance in-place. which is a potentially problematic behaviour. I propose we deprecate it and create a new method (`Register.rotated()`) that instead creates a new register object with the rotated coordinates.

HGSilveri:
- Introduces `Register.rotated()`, which returns a new instance instead of rotating in place
- Deprecates `Register.rotate()` 

Closes #628 .

### DIFF
--- a/pulser-core/pulser/register/register.py
+++ b/pulser-core/pulser/register/register.py
@@ -16,6 +16,7 @@
 from __future__ import annotations
 
 import json
+import warnings
 from collections.abc import Mapping
 from typing import Any, Optional, Union, cast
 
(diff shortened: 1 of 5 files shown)


pulser-core/pulser/register/register.py
```diff
@@ -297,6 +310,33 @@ def rotate(self, degrees: float) -> None:
...
+            angle.
+        """
+        theta = np.deg2rad(degrees)
+        rot = np.array(
+            [[np.cos(theta), -np.sin(theta)], [np.sin(theta), np.cos(theta)]]
+        )
+        if self.layout is not None:
+            warnings.warn(
+                "The rotated register won't have an associated "
+                "'RegisterLayout'.",
```
a-corni, line 332: Perhaps specify the category ? It's a `Userwarning` by default and it might never change

a-corni: I also feel like there is a way to define the rotation for both layout and register, by defining a method `get_rotated_coords(self, angle)` in CoordsCollection (that would only work for 2D coords/rotate around phi=0 in 3D). But I guess it's not very interesting to rotate a register layout since it is obtained from calibration ?

HGSilveri: I doubt `warnings.warn()` will ever change the default warning category in Python 3.x, but I can still specify it. 
As you suggested, register layouts should give a notion of being static so having an easy option to rotate them is not something we want to be giving the user imo

a-corni: Yes let's go with this PR !

a-corni (approved):

PR #647, opened 2024-02-13
Enabling definition of multiple noise channels and noise channels in XY

### LINKED ISSUE #646: Allow running a simulation with Dephasing, Depolarizing and Effective Noise channels
Since we use collapse operators for the simulation in Qutip, the use of multiple noise during the same simulation is now implemented, we should stop raising an error
@dehond @MauroMendizabal

### LINKED ISSUE #648: Enable Simulation in XY mode using SPAM and collapse operators
- SPAM errors are already implemented, but a bug needs to be fixed for simulation to be **run** with qutip.
- Everything is in place for the user to be able to perform simulations in XY using collapse operators.
It does not make sense to develop other noise types since it uses microwave and if this is not implemented with a laser then:
- amplitude: no gaussian fluctuation in amplitude since since model works for a laser.
- doppler: speed will have an impact, but is it the same as for a laser ?
A noise on the magnetic field might be more interesting ?
@vdrouint @dehond

a-corni:
- [x] Enables use of multiple noise channels in one simulation: Fixes #646. The constraint is simply deleted, and the collapse operators are now added - this is tested in a test.
- [x] Collapse operators in XY mode is enabled. Fixes #648
- [x] Minor changes in the documentation: it's weird and confusing that the page referencing collapse operators is called `kraus_ops.html`


pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -514,7 +514,9 @@ def run(
             }
             if self.config.eta > 0 and self.initial_state != qutip.tensor(
                 [
-                    self._hamiltonian.basis["g"]
+                    self._hamiltonian.basis[
+                        "u" if self._hamiltonian._interaction == "XY" else "g"
+                    ]
```
HGSilveri, line 519: This might qualify this change as an hotfix, don't you think?

tests/test_simulation.py
```diff
@@ -934,17 +911,58 @@ def test_run_xy():
     assert sim.samples_obj._measurement == "XY"
 
 
-def test_noisy_xy():
+@pytest.mark.parametrize(
+    "masked_qubit",
```
HGSilveri, line 915: It would make more sense to split this into two variables:
```suggestion
    "masked_qubit, result",
```

HGSilveri:
This is starting to mix a fix and a feature :sweat_smile:

pulser-simulation/pulser_simulation/simulation.py
a-corni: I think so as well :)

tests/test_simulation.py
a-corni: Thanks for reminding me of this 😅

a-corni:
> This is starting to mix a fix and a feature 😅

I am stopping here ;) In a way it's a fix since everything was already there 😇 But yes I could make a separate PR with only the fix (SPAM in XY), and later merge this one with the rest that is more development

a-corni:
@HGSilveri should I still handle kraus_ops.html (but redirecting it to the same notebook as effective_noise.html) ?

a-corni:
> > @HGSilveri should I still handle kraus_ops.html (but redirecting it to the same notebook as effective_noise.html) ?
> 
> Isn't that done here already?

No I have only changed the name of the file 🙄 I will implement this then

a-corni:
> > > > @HGSilveri should I still handle kraus_ops.html (but redirecting it to the same notebook as effective_noise.html) ?
> > > 
> > > 
> > > Isn't that done here already?
> > 
> > 
> > No I have only changed the name of the file 🙄 I will implement this then
> 
> Then I think I don't get what you are referring to because I already see all the changes I thought were required

Now I understand the confusion 😅 All done then

tests/test_simulation.py
```diff
@@ -978,34 +957,22 @@ def test_noisy_xy(matrices, masked_qubit, result):
...
+            eff_noise_rates=[0.025],
+        )
+    )
     assert sim._hamiltonian._bad_atoms == {
         "atom0": True,
         "atom1": False,
         "atom2": True,
         "atom3": False,
     }
+    assert len(sim._hamiltonian._collapse_ops) // 4 == n_collapse_ops
```
HGSilveri, line 975: Maybe divide by `len(reg.qubits)` to be clearer?

HGSilveri (approved):
LGTM!

PR #649, opened 2024-02-14
FIX: SPAM noise in XY, Slope of RampWaveform

a-corni:
SPAM in XY mode is fixed and thoroughly tested this time. The issue has been here in many versions of Pulser and comes from the fact no tests had been developed to test QutipEmulator.run() for a QutipEmulator using a Sequence in XY mode. 

Partially fixes https://github.com/pasqal-io/Pulser/issues/648


HGSilveri (approved):
The changes LGTM, let's change the title to include a reference to the RampWaveform slop fix too

PR #653, opened 2024-03-01
FIX: Deserialization of sized variable items

HGSilveri:
We were only supporting the deserialization of unsized variable items (e.g. single values obtained through `var[0]`), although we also supported sized variable items through slices (e.g. `var[1:5:2]`). 

The abstract representation allowed for serialization of these variable items but then deserialization failed. The easiest way to fix this was actually to add support for indexing variables with a list of indices (e.g. `var[[0, 1, 3]]`), so this hotfix actually introduces a small enhancement.

- Fixes deserialization from the abstract representation of sized variable items
- Adds option to get a sized `VariableItem` by indexing a `Variable` with a list of indices
- Supports `__len__` for sized `VariableItem` instances


a-corni (approved):
Looks good to me !

PR #654, opened 2024-03-13
Add parameter to specify the numerical precision for matching coords to layout traps 

lvignoli:
Adds the new keyword parameter `decimals` to `get_traps_from_coordinates`.
Also, renames the associated constant to make it clear it specifies a number of decimals.

### DIFF
--- a/pulser-core/pulser/devices/_device_datacls.py
+++ b/pulser-core/pulser/devices/_device_datacls.py
@@ -32,7 +32,7 @@ from pulser.json.utils import get_dataclass_defaults, obj_to_dict
 from pulser.register.base_register import BaseRegister, QubitId
 from pulser.register.mappable_reg import MappableRegister
 from pulser.register.register_layout import RegisterLayout
-from pulser.register.traps import COORD_PRECISION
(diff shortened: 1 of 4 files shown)


lvignoli:
Actually, this is not implementing the wanted behavior.
Standby for a fix!

pulser-core/pulser/devices/_device_datacls.py
```diff
@@ -32,7 +32,7 @@
 from pulser.register.base_register import BaseRegister, QubitId
 from pulser.register.mappable_reg import MappableRegister
 from pulser.register.register_layout import RegisterLayout
-from pulser.register.traps import COORD_PRECISION
+from pulser.register.traps import COORD_DECIMAL_PRECISION
```
HGSilveri, line 35: I'm fine with the renaming but I'm not 100% this constant is not used outside of `pulser`. 
To be safe, we would need to keep an alias `COORD_PRECSION = COORD_DECIMAL_PRECSION`, which kinda defeats the purpose

lvignoli:
@HGSilveri I am closing this, feel free to reopen if the fix is not already addressed some other way :)

PR #660, opened 2024-03-20
Add error channel for simulation with noise

a-corni:
In this work I:
- [x] Associate a basis of two eigenvectors to each Channel via a property, and use it to clarify the eigenbasis associated to a SequenceSamples.
- [x] Refactor the definition of basis and op_matrix from SequenceSamples.eigenbasis
- [x] Add a RydbergError Channel and a new possible state to transition from g, the error state "x"
- [x] Enable the definition of collapse operators in the case of a qutrit states.
- [x] Test that all previous CI tests still passes (the changes are non destructive)
- [ ] Develop CI tests for the RydbergError Channel
- [ ] Add the new basis in the convention notebook
- [ ] Modify the notebook on effective noise to show how to define noise on a qutrit/qudit basis


a-corni:
@dehond This PR can be of interest for you and @LucasGitQ. I welcome your feedback on this idea of adding a RydbergError Channel that, if declared in a Sequence, enables you to simulate the interaction of your Rydberg/Hyperfine states with a garbage state via collapse operators.

dehond:
@a-corni Interesting idea! In the end I assume we can use this to define a branching ratio with which spontaneous emission brings population to this garbage state?

Would it also be possible to (stochastically) initialize part of the atoms in this state so as to model inefficient state preparation?

PR #662, opened 2024-03-22
Upgrade black version to 24.3.0 to fix  vulnerability

a-corni:
Fixes security warning https://github.com/pasqal-io/Pulser/security/dependabot/2

### DIFF
--- a/dev_requirements.txt
+++ b/dev_requirements.txt
@@ -1,5 +1,5 @@
 # tests
-black[jupyter] ~= 23.1
+black[jupyter] >= 24.3.0
 flake8
 flake8-docstrings
 isort


HGSilveri:
I can see you are treating this as an hotfix but `black` is not a pulser dependency... It should be enough to merge this to `develop`

VERSION.txt
```diff
@@ -1 +1 @@
-0.17.3
+0.17.4
```
HGSilveri, line 1: You'll need to undo this to get rid of the merge conflict

a-corni:
Indeed, it seemed more critical than a simple MR on develop at first sight, I have merged develop into it to fix it.

dev_requirements.txt
```diff
@@ -1,5 +1,5 @@
 # tests
-black[jupyter] ~= 23.1
+black[jupyter] >= 24.3.0
```
HGSilveri, line 2: Actually, I think we should go with `~=24.3` to avoid `25.0` breaking our CI unnecessarily

HGSilveri (approved):
LGTM, thanks!

PR #663, opened 2024-03-25
Improve access to output modulation durations

### LINKED ISSUE #643: Improvements to Sequence creation with modulation
Following user feedback, there are a few improvements we can make to improve the user experience when creating sequences on physical devices.

- [x] Add the option to start a `delay()` command after the last pulse has finished. We can allow this by adding a new `at_rest: bool = False` argument to the `Sequence.delay()` signature
- [x] Give access to the full duration of pulse (including modulation). We can add the method `Pulse.get_full_duration()` to achieve this; it should take into account the channel where the Pulse is to be executed and whether it will be in EOM mode (exclusively for square pulses).

HGSilveri:
- [x] Add the `at_rest: bool` option to `Sequence.delay()` and `Sequence.align()` to wait (or not) for a pulse to truly finish
- [x] Silently deal with applied delays of 0 (simple skips adding the slot to the schedule)
- [x] Add `Pulse.get_full_duration()` to access the full duration of a pulse after modulation, as shown in the sequence drawing.
- [x] Update the JSON schema to support the new `at_rest` argument  

Closes #643 .


pulser-core/pulser/sequence/sequence.py
```diff
@@ -2118,10 +2128,19 @@ def _check_qubits_give_ids(
...
+        channel: str,
+        at_rest: bool = False,
+    ) -> None:
         self._validate_channel(channel, block_if_slm=True)
         if self.is_parametrized():
             return
+        if at_rest:
+            self._schedule.wait_for_fall(channel)
+        if not duration:
+            return
```
a-corni, line 2143: So one can ask for a delay of duration 0 but with at_rest=True and it will create a delay until last pulse reaches 0 amplitude and detuning ?

a-corni: I am fine with that behaviour :)

tutorials/advanced_features/Output Modulation and EOM Mode.ipynb
```diff
@@ -198,7 +200,7 @@
...
     "3. Similarly, changing the target in a local channel will also wait for the output to ramp down before starting the retargeting.\n",
-    "4. For consecutive pulses in the same channel, there is no automatically imposed delay between them to allow one pulse to finish before the next one starts. As such, whenever the interval between two pulses is too short, they will be \"merged\" together, as is illustrated in the `raman` channel."
+    "4. For consecutive pulses in the same channel, there is no automatically imposed delay between them to allow one pulse to finish before the next one starts. As such, whenever the interval between two pulses is too short, they will be \"merged\" together, as is illustrated in the `raman` channel. To avoid this merging, we can set `at_rest=True` in the `delay()` call. As shown above, when this option is combined with an applied delay time of 0 the channel will be delayed by the exact ammount necessary for the the previous pulse to finish."
```
a-corni, line 203:
```suggestion
    "4. For consecutive pulses in the same channel, there is no automatically imposed delay between them to allow one pulse to finish before the next one starts. As such, whenever the interval between two pulses is too short, they will be \"merged\" together, as is illustrated in the `raman` channel. To avoid this merging, we can set `at_rest=True` in the `delay()` call. As shown above, when this option is combined with an applied delay time of 0 the channel will be delayed by the exact ammount necessary for the previous pulse to finish."
```

pulser-core/pulser/sequence/sequence.py
a-corni: Tested in test_sequence.py

PR #665, opened 2024-03-27
Rectangular lattice register and layout

soufianekaghad98:
Addition of classes to the special layouts :
RectangularLatticeLayout to be able to play with x and y spacing. Now SquareLatticeLayout inherits from it. Don't know how to deal with the print change, we can keep rectangular for a square. (one test fails that compares the strings)
TriangularLatticeLayoutRectShape to have triangular lattice in a rectangular layout
RandomLayout that generates a random layout for a given number of traps (limited number of iterations)

mypy failed because of no type defintion of an empty list in RandomLayout.

### DIFF
--- a/pulser-core/pulser/json/supported.py
+++ b/pulser-core/pulser/json/supported.py
@@ -66,6 +66,7 @@ SUPPORTED_MODULES = {
     "pulser.register.register3d": ("Register3D",),
     "pulser.register.register_layout": ("RegisterLayout",),
     "pulser.register.special_layouts": (
+        "RectangularLatticeLayout",
         "SquareLatticeLayout",
         "TriangularLatticeLayout",
     ),
--- a/pulser-core/pulser/json/supported.py
+++ b/pulser-core/pulser/json/supported.py
@@ -69,6 +69,7 @@ SUPPORTED_MODULES = {
         "RectangularLatticeLayout",
         "SquareLatticeLayout",
         "TriangularLatticeLayout",
+        # "RectangularLatticeLayout",
     ),
     "pulser.register.mappable_reg": ("MappableRegister",),
     "pulser.register.weight_maps": ("DetuningMap",),
--- a/pulser-core/pulser/json/supported.py
+++ b/pulser-core/pulser/json/supported.py
@@ -69,7 +69,6 @@ SUPPORTED_MODULES = {
         "RectangularLatticeLayout",
         "SquareLatticeLayout",
         "TriangularLatticeLayout",
-        # "RectangularLatticeLayout",
     ),
     "pulser.register.mappable_reg": ("MappableRegister",),
     "pulser.register.weight_maps": ("DetuningMap",),
--- a/pulser-core/pulser/register/__init__.py
+++ b/pulser-core/pulser/register/__init__.py
@@ -18,8 +18,11 @@ from pulser.register.register import Register
 from pulser.register.register3d import Register3D
 from pulser.register.register_layout import RegisterLayout
 from pulser.register.special_layouts import (
+    RectangularLatticeLayout,
     SquareLatticeLayout,
     TriangularLatticeLayout,
+    TriangularLatticeLayoutRectShape,
+    RandomLayout,
 )
 
 __all__ = [
@@ -29,4 +32,6 @@ __all__ = [
     "RegisterLayout",
     "SquareLatticeLayout",
     "TriangularLatticeLayout",
+    "TriangularLatticeLayoutRectShape",
+    "RandomLayout",
 ]
--- a/pulser-core/pulser/register/__init__.py
+++ b/pulser-core/pulser/register/__init__.py
@@ -18,11 +18,9 @@ from pulser.register.register import Register
 from pulser.register.register3d import Register3D
 from pulser.register.register_layout import RegisterLayout
 from pulser.register.special_layouts import (
-    RectangularLatticeLayout,
     SquareLatticeLayout,
     TriangularLatticeLayout,
-    TriangularLatticeLayoutRectShape,
-    RandomLayout,
+    RectangularLatticeLayout,
 )
 
 __all__ = [
@@ -32,6 +30,5 @@ __all__ = [
     "RegisterLayout",
     "SquareLatticeLayout",
     "TriangularLatticeLayout",
-    "TriangularLatticeLayoutRectShape",
-    "RandomLayout",
+    "RectangularLatticeLayout",
 ]
--- a/pulser-core/pulser/register/register.py
+++ b/pulser-core/pulser/register/register.py
@@ -89,7 +89,7 @@ class Register(BaseRegister, RegDrawer):
         spacing: float = 4.0,
         prefix: Optional[str] = None,
     ) -> Register:
-        """Initializes register, qubits in a rectangular array, square lattice.
+        """Creates a rectangular array of qubits on a square lattice.
 
         Args:
             rows: Number of rows.
@@ -102,33 +102,10 @@ class Register(BaseRegister, RegDrawer):
         Returns:
             A register with qubits placed in a rectangular array.
         """
-        # Check rows
-        if rows < 1:
-            raise ValueError(
-                f"The number of rows (`rows` = {rows})"
-                " must be greater than or equal to 1."
-            )
-
-        # Check columns
-        if columns < 1:
-            raise ValueError(
-                f"The number of columns (`columns` = {columns})"
-                " must be greater than or equal to 1."
-            )
-
-        # Check spacing
-        if spacing <= 0.0:
-            raise ValueError(
-                f"Spacing between atoms (`spacing` = {spacing})"
-                " must be greater than 0."
-            )
-
-        coords = patterns.square_rect(rows, columns) * spacing
-
-        return cls.from_coordinates(coords, center=True, prefix=prefix)
+        return cls.rectangular_lattice(rows, columns, spacing, spacing, prefix)
 
     @classmethod
-    def rectangle_rect_lattice(
+    def rectangular_lattice(
         cls,
         rows: int,
         columns: int,
@@ -136,7 +113,7 @@ class Register(BaseRegister, RegDrawer):
         col_spacing: float = 2.0,
         prefix: Optional[str] = None,
     ) -> Register:
-        """Initializes register, qubits in rectangle array, rectangle lattice.
+        """Creates a rectangular array of qubits on a rectangular lattice.
 
         Args:
             rows: Number of rows.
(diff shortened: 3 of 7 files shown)


HGSilveri:
Hi @soufianekaghad98 , thank you for your contribution! I have a few preliminary comments and questions:

1. I agree with the addition of `RectangularLatticeLayout` but I'm not sure about the others. 
2. In your opinion, what is the advantage of having a triangular lattice in a rectangular shape? Note that you can still define rectangular-shaped registers in the standard triangular lattice
3. `RandomLayout` is problematic because layouts rely on reproducibility, meaning that the same call should yield identical results. This will naturally not be the case for a random layout. Would it be so bad for the user to randomly generate the set of coordinates given to `RegisterLayout` instead?

pulser-core/pulser/register/special_layouts.py
```diff
@@ -84,16 +95,42 @@ def rectangular_register(
...
+
+    def __init__(self, rows: int, columns: int, spacing: float):
+        """Initializes a SquareLatticeLayout."""
+        self._rows = int(rows)
+        self._columns = int(columns)
+        self._spacing = float(spacing)
+        super().__init__(
+            self._rows, self._columns, self._spacing, self._spacing
+        )
+
```
HGSilveri, line 133: Here you just have to override `self.slug` with the specifc square lattice one

pulser-core/pulser/register/special_layouts.py
```diff
@@ -25,27 +25,38 @@
...
+class RectangularLatticeLayout(RegisterLayout):
+    """A RegisterLayout with a rectangular lattice pattern in a rectangular
+    shape.
 
     Args:
         rows: The number of rows of traps.
         columns: The number of columns of traps.
-        spacing: The distance between neighbouring traps (in µm).
+        x_spacing: Horizontal distance between neighbouring traps (in µm).
+        y_spacing: Vertical distance between neighbouring traps (in µm)
```
HGSilveri, line 40: I think `col_spacing` and `row_spacing` would be preferable

pulser-core/pulser/register/special_layouts.py
```diff
@@ -25,27 +25,38 @@
...
         self._rows = int(rows)
         self._columns = int(columns)
-        self._spacing = float(spacing)
+        self._x_spacing = float(x_spacing)
+        self._y_spacing = float(y_spacing)
         slug = (
-            f"SquareLatticeLayout({self._rows}x{self._columns}, "
-            f"{self._spacing}µm)"
+            f"RectangularLatticeLayout(Shape : {self._rows}x{self._columns}, "
+            f"Lattice pattern : {self._x_spacing}x{self._y_spacing}µm)"
```
HGSilveri, line 53: This slug is a bit too long, how about just this:
```suggestion
            f"RectangularLatticeLayout({self._rows}x{self._columns},  "
            f"{self._row_spacing}x{self._col_spacing}µm)"
```

tests/test_register_layout.py
```diff
@@ -190,6 +193,25 @@ def test_square_lattice_layout():
         square.rectangular_register(10, 3)
 
 
+def test_rectangular_lattice_layout():
+    rectangle = RectangularLatticeLayout(9, 7, 5, 5)
```
HGSilveri, line 197: You should write a test where you use different spacings

HGSilveri (commented):
Here goes a first batch of comments

soufianekaghad98:
Hi @HGSilveri, thanks for your feedback!
To give you some context, I am currently working on the equalization of the traps. So my goal is to generate a whole bunch of different layouts and test them to see how well we can get equalized traps.
This is why I want to use special_layouts which gives different kinds of patterns ready to use, and triangular lattice in rectangular shape is one among the easy layouts to try. I agree that in the end for a whole sequence it will suffice to just have the rectangular register, but for now I'm only working with layouts).
I feel like RandomLayout could be useful for other people who like me need to generate a lot of random layouts with given constraints (mostly for benchmarking and testing). Maybe it could be placed somewhere else then?

HGSilveri:
> This is why I want to use special_layouts which gives different kinds of patterns ready to use, and triangular lattice in rectangular shape is one among the easy layouts to try. I agree that in the end for a whole sequence it will suffice to just have the rectangular register, but for now I'm only working with layouts).
> I feel like RandomLayout could be useful for other people who like me need to generate a lot of random layouts with given constraints (mostly for benchmarking and testing). Maybe it could be placed somewhere else then?

Yes, I think `RandomLayout` should be placed at a higher-level package (which could be wherever you are using Pulser). `TriangularLatticeRectShape` could be included here but since you are already defining `RandomLayout` somewhere else, maybe it could go there as well? 
I'm not strongly opposed to the inclusion of `TriangularLatticeRectShape`, I just think its utility will be very limited and it can easily end up being something we support but nobody uses after a while.

pulser-core/pulser/json/supported.py
```diff
@@ -66,6 +66,7 @@
     "pulser.register.register3d": ("Register3D",),
     "pulser.register.register_layout": ("RegisterLayout",),
     "pulser.register.special_layouts": (
+        "RectangularLatticeLayout",
```
HGSilveri, line 69: You can delete this now

pulser-core/pulser/register/__init__.py
```diff
@@ -18,8 +18,11 @@
 from pulser.register.register3d import Register3D
 from pulser.register.register_layout import RegisterLayout
 from pulser.register.special_layouts import (
+    RectangularLatticeLayout,
     SquareLatticeLayout,
     TriangularLatticeLayout,
+    TriangularLatticeLayoutRectShape,
+    RandomLayout,
```
HGSilveri, line 25: These imports should be deleted as well

pulser-core/pulser/register/register.py
```diff
@@ -89,7 +89,7 @@ def rectangle(
         spacing: float = 4.0,
         prefix: Optional[str] = None,
     ) -> Register:
-        """Initializes the register with the qubits in a rectangular array.
+        """Initializes register, qubits in a rectangular array, square lattice.
```
HGSilveri, line 92: How about this?
```suggestion
        """Creates a rectangular array of qubits on a square lattice.
```

pulser-core/pulser/register/register.py
```diff
@@ -127,6 +127,55 @@ def rectangle(
 
         return cls.from_coordinates(coords, center=True, prefix=prefix)
 
+    @classmethod
+    def rectangle_rect_lattice(
```
HGSilveri, line 131: I think it would be fine to call this just `rectangular_lattice`

pulser-core/pulser/register/register.py
```diff
@@ -127,6 +127,55 @@ def rectangle(
...
+        cls,
+        rows: int,
+        columns: int,
+        row_spacing: float = 4.0,
+        col_spacing: float = 2.0,
+        prefix: Optional[str] = None,
+    ) -> Register:
+        """Initializes register, qubits in rectangle array, rectangle lattice.
```
HGSilveri, line 139: Similar to above
```suggestion
        """Creates a rectangular array of qubits on a rectangular lattice.
```

pulser-core/pulser/register/register.py
```diff
@@ -127,6 +127,55 @@ def rectangle(
...
+            rows: Number of rows.
+            columns: Number of columns.
+            row_spacing: The distance between rows in μm.
+            col_spacing: The distance between columns in μm.
+            prefix: The prefix for the qubit ids. If defined, each qubit
+                id starts with the prefix, followed by an int from 0 to N-1
+                (e.g. prefix='q' -> IDs: 'q0', 'q1', 'q2', ...)
+
+        Returns:
+            Register, qubits placed in a rectangular array, rectangle lattice.
```
HGSilveri, line 151: nit:
```suggestion
            Register with qubits placed in a rectangular array on a rectangular lattice.
```

pulser-core/pulser/register/register.py
HGSilveri, line 131: I think you can use this method inside `rectangle()` now

pulser-core/pulser/register/special_layouts.py
```diff
@@ -84,16 +94,50 @@ def rectangular_register(
...
         return cast(
             pulser.Register,
             self.define_register(*trap_ids, qubit_ids=qubit_ids),
         )
 
+    def _to_dict(self) -> dict[str, Any]:
+        return obj_to_dict(
+            self, self._rows, self._columns, self._col_spacing,
+            self._row_spacing
+        )
```
HGSilveri, line 113: You'll have to add this new class here now: https://github.com/pasqal-io/Pulser/blob/71b8e0321371d0cfd311d90e5972292cc33977eb/pulser-core/pulser/json/supported.py#L70

pulser-core/pulser/register/register.py
soufianekaghad98: I'm not sure I understand what you mean.

HGSilveri: The contents of the method just above (`Register.rectangle()`)  can be rewritten to use this one since 
`Register.rectangle(rows, columns, spacing) == Register.rectangular_lattice(rows, columns, spacing, spacing)`

pulser-core/pulser/register/__init__.py
soufianekaghad98: Don't I need it to be able to use the new class ? Same as for the supported.py file. I am not able to import the class otherwise.

HGSilveri: But we agreed you would remove these classes, no?

soufianekaghad98: Oh yes my bad, I thought the command was also about the `RectangulatLatticeLayout` class.

soufianekaghad98:
Feels like I messed up when trying to push again (apparently there was some issue with my local branch not being up to date with the remote... I tried different commands to pull then push which ended with me pushing stuff I didn't know about...). Sorry for the mess.

HGSilveri:
> Feels like I messed up when trying to push again (apparently there was some issue with my local branch not being up to date with the remote... I tried different commands to pull then push which ended with me pushing stuff I didn't know about...). Sorry for the mess.

No worries, it's nothing we can't fix. I'll give you a hand when I can

pulser-core/pulser/json/supported.py
```diff
@@ -66,8 +66,10 @@
...
         "SquareLatticeLayout",
         "TriangularLatticeLayout",
+        # "RectangularLatticeLayout",
```
HGSilveri, line 72: Let's delete instead of commenting

pulser-core/pulser/register/register.py
```diff
@@ -102,6 +102,32 @@ def rectangle(
...
+
+    @classmethod
+    def rectangular_lattice(
+        cls,
+        rows: int,
+        columns: int,
+        row_spacing: float = 4.0,
+        col_spacing: float = 2.0,
+        prefix: Optional[str] = None,
+    ) -> Register:
```
HGSilveri, line 115: I think we are still missing a dedicated test to this method

HGSilveri (commented):
Almost there, these should be the final comments before I approve :)

HGSilveri (approved):
LGTM, thanks a lot @soufianekaghad98 !

PR #666, opened 2024-03-28
Define max_step from minimum variations of samples

a-corni:
Currently, the max_step is defined automotically from the Sequence if the user don't provide it.
This definition was determined in this PR three years ago https://github.com/pasqal-io/Pulser/pull/225
However, it fails if a user provides one Pulse composed of a waveform (Composite) that would be equal to 0 during a long period and then has a short variation before going back to 0. This is a typical issue of qutip (and adams solver) as can be seen in the issues related to the previous PR and on this forum https://groups.google.com/g/qutip/c/slz4MCnkolw
I think it's best to define the max_step from the shortest variation of the waveforms that compose a Pulse.

example of a code that fails:
```
import pulser
import pulser_simulation
seq = pulser.Sequence(pulser.Register.square(2, 5), pulser.AnalogDevice)
seq.declare_channel("ryd_glob", "rydberg_global")
seq.add(
    pulser.Pulse.ConstantDetuning(
        pulser.CompositeWaveform(
            pulser.ConstantWaveform(2500, 0.0),
            pulser.BlackmanWaveform(1000, np.pi),
            pulser.ConstantWaveform(500, 0.0),
            ),
        0,
        0,
        ),
    channel="ryd_glob",
)
seq.draw()
sim = pulser_simulation.QutipEmulator.from_sequence(seq)
sim.run().get_final_state()
```
image

To be discussed: With the new implementation, this will surely increase the number of time steps for any InterpolatedWaveform, CustomWaveform, BlackmanWaveform. I think it will make the computation more robust, which is what we want for this package.


pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -492,19 +492,35 @@ def run(
...
+            min_variations: list[int] = []
+            for sample in (ch_sample.amp, ch_sample.det):
+                # Find the indices at which the sample varies
+                variations_indices: np.ndarray = np.concatenate(
+                    [
+                        np.array([-1]),
+                        np.nonzero(np.diff(sample)),
+                        np.array([end_point]),
+                    ],
+                    axis=None,
```
HGSilveri, line 507: I think you can use the `prepend` and `append` arguments in the `np.diff()` down below to avoid this concatenation

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -492,19 +492,35 @@ def run(
...
                 )
-            ]
-            if pulse_durations:
-                options["max_step"] = 0.5 * min(pulse_durations) / 1000
+                # Find the differences between two indices at which sample varies
+                min_variations.append(int(np.min(np.diff(variations_indices))))
+
+            return min(min_variations)
```
HGSilveri, line 512: I have a feeling we are going to get `max_step=0.001` whenever we have something other than constant waveforms in the sequence. Hopefully that doesn't slow us down too much

tests/test_simresults.py
```diff
@@ -244,7 +244,7 @@ def test_expect(results, pi_pulse, reg):
     config = SimConfig(noise="SPAM", eta=0)
     sim_single.set_config(config)
     sim_single.set_evaluation_times("Minimal")
-    results_single = sim_single.run()
+    results_single = sim_single.run(nsteps=2000)
```
HGSilveri, line 247: Is this because you were getting an ODE error now?

a-corni: I was getting a Warning followed by an error:
UserWarning: "zvode: Excess work done on this call (Perhaps wrong MF.)": 
Exception: ODE integration error: Try to increase the allowed number of substeps by increasing the nsteps parameter in the Options class.

At least the error is explicit

a-corni: I guess we can try to suggest a nsteps from the samples of the Sequence as well

pulser-simulation/pulser_simulation/simulation.py
a-corni: I think so too ;)

tests/test_simresults.py
a-corni: That was not an hard issue to solve ^^'

pulser-simulation/pulser_simulation/simulation.py
HGSilveri: I agree that robustness is preferable and speed can still be recovered by manually setting `max_step`. 
But I also think it's a good idea to have a notion of how much this impacts speed at the moment, if things become significantly slower the users won't like it.

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -521,6 +521,10 @@ def get_min_variation(ch_sample: ChannelSamples) -> int:
                 )
                 / 1000
             )
+        if "nsteps" not in options:
+            options["nsteps"] = max(
+                1000, self._tot_duration // options["max_step"]
+            )
```
HGSilveri, line 527: I'm not sure `nsteps` refers to the total number of steps over the entire sequence. From the [Qutip docs](https://qutip.org/docs/latest/guide/dynamics/dynamics-options.html#setting-options-for-the-dynamics-solvers) and the define it as: "Max. number of steps to take for each interval"

What is an interval though?

a-corni: Currently, this would be failing (with the error asking to raise nsteps) if the user was computing a sequence with one Pulse composed of a customwaveform of duration > 1000 changing value over large distance at each ns (for instance oscillating between 0 and 1) (untested, but I think this is the case)

a-corni: Tested: Fails
```
import pulser
import pulser_simulation
seq = pulser.Sequence(pulser.Register.square(2, 5), pulser.AnalogDevice)
seq.declare_channel("ryd_glob", "rydberg_global")
oscillations = [0, 1] * 1000
seq.add(
    pulser.Pulse.ConstantDetuning(
        pulser.CustomWaveform([0, 1] * 1000),
        0,
        0,
    ),
    channel="ryd_glob",
)
seq.draw()
sim = pulser_simulation.QutipEmulator.from_sequence(seq, evaluation_times="Minimal")
sim.run().get_final_state()
```
Exception: ODE integration error: Try to increase the allowed number of substeps by increasing the nsteps parameter in the Options class.

HGSilveri: Ok, but this is an unrealistic edge case so I wouldn't worry about. Plus, the user can still manually increase `nsteps` to fix it.

pulser-simulation/pulser_simulation/simulation.py
a-corni: I was considering looking at the slots and checking what is the type of the waveform used in each Pulse slot. If the slot is a CompositeWaveform/CustomWaveform/InterpolatedWaveform then it's best to define it using diff, otherwise we can say that max_step should be half this slot as it was before.
However, this loop might take more time than the simulation in itself...

pulser-simulation/pulser_simulation/simulation.py
a-corni: But it's still a nice feature, right ? It does not take too much effort on our side to introduce this

a-corni:
[test_perf_sim.pdf](https://github.com/pasqal-io/Pulser/files/14806196/test_perf_sim.pdf)
Conclusion: no major changes on the computation time.
I think that the RampWaveform is the worst case scenario for this condition on the max_step

pulser-simulation/pulser_simulation/simulation.py
HGSilveri: Yes, let's keep it :)

pulser-simulation/pulser_simulation/simulation.py
HGSilveri: This sounds like overkill to me

a-corni: To me as well haha

a-corni:
Should I add the "false positive" simulation in the tests ? In `tests/test_simresults.py` ?

HGSilveri (approved):
I approve the changes, but I was wondering if you could include a test with the case that was failing before

HGSilveri:
> Should I add the "false positive" simulation in the tests ? In `tests/test_simresults.py` ? Basically just checking that the final state is not equal to the ground state.

Hahaha you read my mind

a-corni:
> > Should I add the "false positive" simulation in the tests ? In `tests/test_simresults.py` ? Basically just checking that the final state is not equal to the ground state.
> 
> Hahaha you read my mind

mind-blown

tests/test_simresults.py
```diff
@@ -386,3 +390,24 @@ def test_results_xy(reg, pi_pulse):
...
+                ConstantWaveform(2500, 0.0),
+                BlackmanWaveform(1000, np.pi),
+                ConstantWaveform(500, 0.0),
+            ),
+            0,
+            0,
+        ),
+        channel="ryd_glob",
+    )
+    seq.draw()
```
HGSilveri, line 411: You forgot to delete this

a-corni: Daamn

HGSilveri (approved):

PR #669, opened 2024-04-03
Bump version to v0.18dev4

a-corni:
Following release of v0.17.4


HGSilveri:
I'm being very picky, but would it be possible to avoid the last merge commit?

a-corni:
Sorry I did it in a rush and branched from master instead of develop, I opened a new PR  #670, I think it is better :)

PR #672, opened 2024-04-15
Block add() on DMM channels

HGSilveri:
- Blocks `Sequence.add()` on DMM channels
- Blocks `Sequence.add_dmm_detuning()` on non-DMM channels

I think having one unique way of adding Pulses to DMM and non-DMM channels is preferable but let me know if you think otherwise

### DIFF
--- a/pulser-core/pulser/sequence/sequence.py
+++ b/pulser-core/pulser/sequence/sequence.py
@@ -1318,6 +1318,11 @@ class Sequence(Generic[DeviceType]):
             block_eom_mode=True,
             block_if_slm=channel.startswith("dmm_"),
         )
+        if isinstance(self.declared_channels[channel], DMM):
+            raise ValueError(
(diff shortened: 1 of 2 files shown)


pulser-core/pulser/sequence/sequence.py
```diff
@@ -1348,6 +1353,8 @@ def add_dmm_detuning(
                   latest pulse.
         """
         self._validate_channel(dmm_name, block_if_slm=True)
+        if not isinstance(self.declared_channels[dmm_name], DMM):
+            raise ValueError(f"'{dmm_name}' is not the name of a DMM channel.")
```
a-corni, line 1357: Daaamn I had not thought about this

a-corni (approved):
It is indeed important to enforce that `add_dmm_detuning` can be only used on a `DMM`. I have always hesitated in prohibiting the use of `add` on DMM, but I guess it will be easier to maintain and clearer for users.

PR #677, opened 2024-04-23
Fix CI tests on "macos" for python < 3.10

HGSilveri:
`macos-latest` no longer supports Python 3.8 and 3.9 so I'm using `macos-13` for these versions instead.


a-corni (approved):
👍

PR #678, opened 2024-04-23
Centralize all backend imports from a single `pulser.backends` module

HGSilveri:
This features adds a new `pulser.backends` module that should serve as the single source for all backends. Ideally, the user can now simple do eg

```python
import pulser

qutip_backend = pulser.backends.QutipBackend(...)
```
If `pulser_simulation` is not installed, an error message will appear suggesting its installation. 

```
AttributeError: 'QutipBackend' requires the 'pulser_simulation' package. To install it, run `pip install pulser_simulation`.
```

The same is true for the `pulser_pasqal` backends.

Imports like 

```pyhton
from pulser.backends import QutipBackend, EmuTNBackend
```
also work, but in this case the error message is not informative if the appropriate extension packages are not installed.


a-corni (approved):
Looks good to me !

Issue #661, opened 2024-03-20
Replicate remote sequence validation locally


HGSilveri:
When a user submits a sequence for remote execution, there are some checks that are executed only remotely. This makes it possible to have a sequence that's valid locally fail execution. 
To minimize the chance of this happening, we can centralize all checks within a method (in `Sequence`, perhaps?) that can be used to run the whole checks locally.  Furthermore, this would allow us to have different backends enforce the same checks as a QPU, thus increasing the similarity of the user experience with that of remote QPU.

PR #682, opened 2024-04-26
Add missing QPU validation fields to the Device class

HGSilveri:
These fields will be necessary to replicate QPU validation locally, as proposed in #661. I want to ship them already in the v0.18 release so that we can then change them in `AnalogDevice` on v0.19, where we will also tackle #661.  

I'm including these fields in the JSON schema update of #676 , so that PR has to be merged before this one.

### DIFF
--- a/pulser-core/pulser/devices/_device_datacls.py
+++ b/pulser-core/pulser/devices/_device_datacls.py
@@ -37,11 +37,15 @@ from pulser.register.traps import COORD_PRECISION
 
 DIMENSIONS = Literal[2, 3]
 
-ALWAYS_OPTIONAL_PARAMS = (
-    "max_sequence_duration",
-    "max_runs",
-    "dmm_objects",
-    "default_noise_model",
+ALWAYS_OPTIONAL_PARAMS = ("max_sequence_duration", "max_runs")
+OPTIONAL_IN_ABSTR_REPR = tuple(
+    list(ALWAYS_OPTIONAL_PARAMS)
+    + [
+        "dmm_objects",
+        "default_noise_model",
+        "requires_layout",
+        "accepts_new_layouts",
+    ]
 )
 PARAMS_WITH_ABSTR_REPR = ("channel_objects", "channel_ids", "dmm_objects")
 
@@ -83,6 +87,8 @@ class BaseDevice(ABC):
         default_noise_model: An optional noise model characterizing the default
             noise of the device. Can be used by emulator backends that support
             noise.
+        requires_layout: Whether the register used in the sequence must be
+            created from a register layout. Only enforced in QPU execution.
     """
 
     name: str
@@ -96,6 +102,7 @@ class BaseDevice(ABC):
     max_layout_filling: float = 0.5
     max_sequence_duration: int | None = None
     max_runs: int | None = None
+    requires_layout: bool = False
     reusable_channels: bool = field(default=False, init=False)
     channel_ids: tuple[str, ...] | None = None
     channel_objects: tuple[Channel, ...] = field(default_factory=tuple)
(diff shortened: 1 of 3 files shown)


pulser-core/pulser/devices/_device_datacls.py
```diff
@@ -37,11 +37,15 @@
...
+OPTIONAL_IN_ABSTR_REPR = tuple(
+    list(ALWAYS_OPTIONAL_PARAMS)
+    + [
+        "dmm_objects",
+        "default_noise_model",
+        "requires_layout",
+        "accepts_new_layouts",
+    ]
 )
 PARAMS_WITH_ABSTR_REPR = ("channel_objects", "channel_ids", "dmm_objects")
```
a-corni, line 50: I don't know if it has an impact: Does not "Noise Model" have a an abstract repr as well ?

HGSilveri: Yeah but the name is misleading. These are actually the parameters with a special format in the abstract representation (eg channel IDs and objects get grouped together in the Device abstract repr)

pulser-core/pulser/devices/_device_datacls.py
```diff
@@ -464,8 +471,8 @@ def _to_dict(self) -> dict[str, Any]:
     def _to_abstract_repr(self) -> dict[str, Any]:
         defaults = get_dataclass_defaults(fields(self))
         params = self._params()
-        for p in ALWAYS_OPTIONAL_PARAMS:
-            if params[p] == defaults[p]:
+        for p in OPTIONAL_IN_ABSTR_REPR:
+            if p in params and params[p] == defaults[p]:
```
a-corni, line 475: What value is covered by "p in params" ?

pulser-core/pulser/devices/_device_datacls.py
```diff
@@ -500,7 +507,15 @@ class Device(BaseDevice):
...
-        rybderg_level : The value of the principal quantum number :math:`n`
+        channel_objects: The Channel subclass instances specifying each
+            channel in the device.
+        channel_ids: Custom IDs for each channel object. When defined,
+            an ID must be given for each channel. If not defined, the IDs are
+            generated internally based on the channels' names and addressing.
+        dmm_objects: The DMM subclass instances specifying each channel in the
+            device. They are referenced by their order in the list, with the ID
+            "dmm_[index in dmm_objects]".
+        rybderg_level: The value of the principal quantum number :math:`n`
```
a-corni, line 518: Nit: There is a typo there "rydberg_level"

pulser-core/pulser/devices/_device_datacls.py
```diff
@@ -701,7 +723,15 @@ class VirtualDevice(BaseDevice):
...
```
a-corni, line 734: Nit:same here

pulser-core/pulser/devices/_device_datacls.py
```diff
@@ -522,15 +537,22 @@ class Device(BaseDevice):
...
+            QPU execution.
     """
 
     max_atom_num: int
     max_radial_distance: int
+    requires_layout: bool = True
     pre_calibrated_layouts: tuple[RegisterLayout, ...] = field(
         default_factory=tuple
     )
+    accepts_new_layouts: bool = True
```
a-corni, line 555: Very Nit: Should this be True or False by default ? Current QPUs would be False...

a-corni: What happens if you don't require a layout but don't accept new layouts ? Should it be prohibited here ?

pulser-core/pulser/devices/_device_datacls.py
HGSilveri: `accepts_new_layouts`

pulser-core/pulser/devices/_device_datacls.py
a-corni: Okay let's go with the least prohibitive default ! I was wondering if we should go with the least prohibitive or the most prohibitive (since it real device). Sounds good to me !

a-corni (approved):
Looks good to me ! I approve and let you fix the nits :)

PR #684, opened 2024-05-14
Add option to switch the Sequence register

HGSilveri:
This has been requested more than once, so I figured we could add a proper way to do it.

### DIFF
--- a/pulser-core/pulser/sequence/sequence.py
+++ b/pulser-core/pulser/sequence/sequence.py
@@ -689,6 +689,19 @@ class Sequence(Generic[DeviceType]):
     ) -> Sequence:
         """Replicate the sequence with a different register.
 
+        The new sequence is reconstructed with the provided register by
+        replicating all the instructions used to build the original sequence.
+        This means that operations referecing specific qubits IDs
+        (eg. `Sequence.target()`) expect to find the same qubit IDs in the new
+        register. By the same token, switching from a register to a mappable
+        register might fail if one of the instructions does not work with
+        mappable registers (e.g. `Sequence.configure_slm_mask()`).
+
+        Warns:
+            UserWarning: If the sequence is configuring a detuning map, a
+            warning is raised to remind the user that the detuning map is
+            unchanged and might no longer be aligend with the new register.
+
         Args:
             new_register: The new register to give the sequence.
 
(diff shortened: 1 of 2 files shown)


tests/test_sequence.py
```diff
@@ -533,6 +533,64 @@ def test_ising_mode(
...
+    pulse = Pulse.ConstantPulse(1000, 1, -1, 2)
+    seq = init_seq(
+        reg,
+        DigitalAnalogDevice,
+        "raman",
+        "raman_local",
+        [pulse],
+        initial_target="q0",
+        parametrized=parametrized,
+    )
```
a-corni, line 548: Do I get it right that you are not testing the conversion of a Sequence with a Mappable Register into a Sequence with a Register/another Mappable Register ? Is it worth testing ? Should we check what happens when DMM are configured in the Sequence as well ?

pulser-core/pulser/sequence/sequence.py
```diff
@@ -684,6 +684,24 @@ def _config_detuning_map(
...
+            new_register: The new register to give the sequence.
+
+        Returns:
+            The sequence with the new register.
+        """
+        new_seq = type(self)(register=new_register, device=self._device)
+        # Copy the variables to the new sequence
+        new_seq._variables = self.declared_variables
+        for call in self._calls[1:] + self._to_build_calls:
+            getattr(new_seq, call.name)(*call.args, **call.kwargs)
```
a-corni, line 702: So the only condition on the new Register/Mappable Register is that the labels in the register match the labels used in local operations ? Perhaps we should explain this in the docstring ?

tests/test_sequence.py
HGSilveri: You are correct. It doesn't hurt to check these cases, so I'll just add them

pulser-core/pulser/sequence/sequence.py
HGSilveri: I put some more info in the docstring, please take a look :)

pulser-core/pulser/sequence/sequence.py
```diff
@@ -684,6 +684,44 @@ def _config_detuning_map(
...
+        This means that operations referecing specific qubits IDs
+        (eg. `Sequence.target()`) expect to find the same qubit IDs in the new
+        register. By the same token, switching from a register to a mappable
+        register might fail if one of the instructions does not work with
+        mappable registers (e.g. `Sequence.configure_slm_mask()`).
+
+        Warns:
+            UserWarning: If the sequence is configuring a detuning map, a
+            warning is raised to remind the user that the detuning map is
+            unchanged and might no longer be aligend with the new register.
```
a-corni, line 703: Perfect ! I was expecting an issue of this genre :)

a-corni (approved):
Lgtm ! Thanks for clarifying the docstring and adding the tests, it's way clearer now :)

PR #686, opened 2024-05-21
Add support for pulser-diff backend

vytautas-a:
Modify `pulser-core` to support all current features of differentiable `torch` backend.

### DIFF
--- a/pulser-core/pulser/channels/dmm.py
+++ b/pulser-core/pulser/channels/dmm.py
@@ -20,6 +20,7 @@ from typing import Any, Literal, Optional
 
 import numpy as np
 
+import pulser.math as pm
 from pulser.channels.base_channel import Channel
 from pulser.json.utils import get_dataclass_defaults
 from pulser.pulse import Pulse
@@ -124,7 +125,9 @@ class DMM(Channel):
                 (defaults to a detuning map with weight 1.0).
         """
         super().validate_pulse(pulse)
-        round_detuning = round(pulse.detuning.samples, 6).as_array(detach=True)
+        round_detuning = pm.round(pulse.detuning.samples, 6).as_array(
+            detach=True
+        )
         # Check that detuning is negative
         if np.any(round_detuning > 0):
             raise ValueError("The detuning in a DMM must not be positive.")
--- a/pulser-core/pulser/math/abstract_array.py
+++ b/pulser-core/pulser/math/abstract_array.py
@@ -37,7 +37,12 @@ class AbstractArray:
         array: The array to store.
     """
 
-    def __init__(self, array: AbstractArrayLike, dtype: DTypeLike = None):
+    def __init__(
+        self,
+        array: AbstractArrayLike,
+        dtype: DTypeLike = None,
+        force_array: bool = False,
+    ):
         """Initializes a new AbstractArray."""
         self._array: np.ndarray | torch.Tensor
         if isinstance(array, AbstractArray):
@@ -49,6 +54,9 @@ class AbstractArray:
             )
         else:
             self._array = np.asarray(array, dtype=dtype)
+
+        if force_array:
+            self._array = self._array[None]
 
     @staticmethod
     @functools.lru_cache
--- a/pulser-core/pulser/math/abstract_array.py
+++ b/pulser-core/pulser/math/abstract_array.py
@@ -17,7 +17,7 @@ from __future__ import annotations
 import functools
 import importlib.util
 import operator
-from typing import Any, cast
+from typing import Any, Generator, cast
 
 import numpy as np
 from numpy.typing import ArrayLike, DTypeLike
@@ -42,7 +42,7 @@ class AbstractArray:
         self._array: np.ndarray | torch.Tensor
         if isinstance(array, AbstractArray):
             self._array = array._array
-        elif self.has_torch() and type(array) is torch.Tensor:
+        elif self.has_torch() and isinstance(array, torch.Tensor):
             self._array = torch.as_tensor(
                 array,
                 dtype=dtype,  # type: ignore[arg-type]
@@ -59,7 +59,7 @@ class AbstractArray:
     @functools.cached_property
     def is_tensor(self) -> bool:
         """Whether the stored array is a tensor."""
-        return self.has_torch() and type(self._array) is torch.Tensor
+        return self.has_torch() and isinstance(self._array, torch.Tensor)
 
     def astype(self, dtype: DTypeLike) -> AbstractArray:
         """Casts the data type of the array contents."""
@@ -106,6 +106,11 @@ class AbstractArray:
     def ndim(self) -> int:
         """The number of dimensions in the array."""
         return self._array.ndim
+
+    @property
+    def shape(self) -> tuple[int, ...]:
+        """Shape of the array."""
+        return self._array.shape
 
     @property
     def real(self) -> AbstractArray:
@@ -235,6 +240,10 @@ class AbstractArray:
         array[indices] = values  # type: ignore[assignment]
         self._array = array
         del self.is_tensor  # Clears cache
+
+    def __iter__(self) -> Generator:
+        for i in self._array:
+            yield i
 
     def __len__(self) -> int:
         return len(self._array)
--- a/pulser-core/pulser/math/abstract_array.py
+++ b/pulser-core/pulser/math/abstract_array.py
@@ -89,6 +89,10 @@ class AbstractArray:
             )
         return np.asarray(self._array)
 
+    def as_list(self) -> list:
+        """Converts the stored array to a Python list."""
+        return self._array.tolist()
+
     def copy(self) -> AbstractArray:
         """Makes a copy itself."""
         return AbstractArray(
@@ -241,9 +245,9 @@ class AbstractArray:
         self._array = array
         del self.is_tensor  # Clears cache
 
-    def __iter__(self) -> Generator:
-        for i in self._array:
-            yield i
+    def __iter__(self) -> Generator[AbstractArray, None, None]:
+        for i in range(self.__len__()):
+            yield self.__getitem__(i)
 
     def __len__(self) -> int:
         return len(self._array)
--- a/pulser-core/pulser/math/abstract_array.py
+++ b/pulser-core/pulser/math/abstract_array.py
@@ -77,7 +77,7 @@ class AbstractArray:
             raise RuntimeError("`torch` is not installed.")
         return torch.as_tensor(self._array)
 
-    def as_array(self, detach: bool = False) -> np.ndarray:
+    def as_array(self, detach: bool = True) -> np.ndarray:
         """Converts the stored array to a Numpy array.
 
         Args:
--- a/pulser-core/pulser/math/abstract_array.py
+++ b/pulser-core/pulser/math/abstract_array.py
@@ -77,7 +77,7 @@ class AbstractArray:
             raise RuntimeError("`torch` is not installed.")
         return torch.as_tensor(self._array)
 
-    def as_array(self, detach: bool = True) -> np.ndarray:
+    def as_array(self, detach: bool = False) -> np.ndarray:
         """Converts the stored array to a Numpy array.
 
         Args:
--- a/pulser-core/pulser/math/abstract_array.py
+++ b/pulser-core/pulser/math/abstract_array.py
@@ -89,7 +89,7 @@ class AbstractArray:
             )
         return np.asarray(self._array)
 
-    def as_list(self) -> list:
+    def tolist(self) -> list:
         """Converts the stored array to a Python list."""
         return self._array.tolist()
 
--- a/pulser-core/pulser/math/abstract_array.py
+++ b/pulser-core/pulser/math/abstract_array.py
@@ -55,7 +55,7 @@ class AbstractArray:
         else:
             self._array = np.asarray(array, dtype=dtype)
 
-        if force_array:
+        if force_array and self._array.ndim == 0:
             self._array = self._array[None]
 
     @staticmethod
--- a/pulser-core/pulser/parametrized/variable.py
+++ b/pulser-core/pulser/parametrized/variable.py
@@ -75,8 +75,7 @@ class Variable(Parametrized, OpSupport):
     def _validate_value(
         self, value: Union[AbstractArrayLike, float, int]
     ) -> pm.AbstractArray:
-        val = pm.AbstractArray(value)
-        # val = np.array(value, dtype=self.dtype, ndmin=1)
+        val = pm.AbstractArray(value, dtype=self.dtype, force_array=True)
         if val.size != self.size:
             raise ValueError(
                 f"Can't assign array of size {val.size} to "
--- a/pulser-core/pulser/parametrized/variable.py
+++ b/pulser-core/pulser/parametrized/variable.py
@@ -22,7 +22,9 @@ from typing import Any, Iterator, Optional, Union, cast
 import numpy as np
 from numpy.typing import ArrayLike
 
+import pulser.math as pm
 from pulser.json.utils import obj_to_dict
+from pulser.math import AbstractArrayLike
 from pulser.parametrized import Parametrized
 from pulser.parametrized.paramobj import OpSupport
 
@@ -71,9 +73,10 @@ class Variable(Parametrized, OpSupport):
         object.__setattr__(self, "_count", self._count + 1)
 
     def _validate_value(
-        self, value: Union[ArrayLike, float, int]
-    ) -> np.ndarray:
-        val = np.array(value, dtype=self.dtype, ndmin=1)
+        self, value: Union[AbstractArrayLike, float, int]
+    ) -> pm.AbstractArray:
+        val = pm.AbstractArray(value)
+        # val = np.array(value, dtype=self.dtype, ndmin=1)
         if val.size != self.size:
             raise ValueError(
                 f"Can't assign array of size {val.size} to "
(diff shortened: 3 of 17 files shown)


pulser-core/pulser/channels/dmm.py
```diff
@@ -124,7 +125,9 @@ def validate_pulse(
                 (defaults to a detuning map with weight 1.0).
         """
         super().validate_pulse(pulse)
-        round_detuning = round(pulse.detuning.samples, 6).as_array(detach=True)
+        round_detuning = pm.round(pulse.detuning.samples, 6).as_array(
+            detach=True
+        )
```
HGSilveri, line 130: What was the issue with the previous version?

pulser-core/pulser/math/abstract_array.py
```diff
@@ -77,7 +77,7 @@ def as_tensor(self) -> torch.Tensor:
             raise RuntimeError("`torch` is not installed.")
         return torch.as_tensor(self._array)
 
-    def as_array(self, detach: bool = False) -> np.ndarray:
+    def as_array(self, detach: bool = True) -> np.ndarray:
```
HGSilveri, line 80: I thought we decided that we wouldn't detach by default

pulser-core/pulser/pulse.py
```diff
@@ -103,10 +103,7 @@ def __init__(
             )
         object.__setattr__(self, "amplitude", amplitude)
         object.__setattr__(self, "detuning", detuning)
-        assert not isinstance(phase, Parametrized)
-        if (phase_ := pm.AbstractArray(phase, dtype=float)).ndim != 0:
-            raise ValueError(f"'phase' must be a single float, not {phase!r}.")
-        object.__setattr__(self, "phase", phase_ % (2 * np.pi))
+        object.__setattr__(self, "phase", cast(float, phase) % (2 * np.pi))
```
HGSilveri, line 106: I think what I had before is what we want here

pulser-core/pulser/waveforms.py
```diff
@@ -509,7 +509,9 @@ def _samples(self) -> pm.AbstractArray:
         Returns:
             A numpy array with a value for each time step.
         """
-        return self._value * np.ones(self.duration)
+        res = self._value * np.ones(self.duration)
+        # print(res)
+        return res
```
HGSilveri, line 514: I guess you don't need this anymore

HGSilveri (commented):
I think you had some changes that I ended up doing but they got applied anyway. I'll let you decided which ones actually make sense to keep

pulser-core/pulser/channels/dmm.py
vytautas-a: It gave me a mypy error related to incompatibility of types.

pulser-core/pulser/pulse.py
vytautas-a: I think phase should be treated on the same footing as amplitude and detuning. From the perspective of pulser-diff, user might do something like `phase=torch.rand(1, requires_grad=True)` and pass this tensor to pulse constructor. However, the phase here will be a 1D tensor and your implementation will fail. I suggest then changing your version to `phase_ := pm.AbstractArray(phase, dtype=float)).size != 1`. This will handle my case.

vytautas-a: Yes, your example has `ndim=0` because you create a tensor from a single float. Mine in the comment above has actually `ndim=1` because that's the way `rand` is implemented. At least in my local version of `torch` :)

pulser-core/pulser/math/abstract_array.py
```diff
@@ -42,7 +42,7 @@ def __init__(self, array: AbstractArrayLike, dtype: DTypeLike = None):
         self._array: np.ndarray | torch.Tensor
         if isinstance(array, AbstractArray):
             self._array = array._array
-        elif self.has_torch() and type(array) is torch.Tensor:
+        elif self.has_torch() and isinstance(array, torch.Tensor):
```
HGSilveri, line 45: So torch Tensors have subclasses?

pulser-core/pulser/math/abstract_array.py
```diff
@@ -236,6 +241,10 @@ def __setitem__(self, indices: Any, values: AbstractArrayLike) -> None:
         self._array = array
         del self.is_tensor  # Clears cache
 
+    def __iter__(self) -> Generator:
+        for i in self._array:
+            yield i
+
```
HGSilveri, line 247: I think that a class with `__getitem__` and `__len__` is automatically iterable. What's the usecase for defining `__iter__` explicitly?

pulser-core/pulser/register/_coordinates.py
```diff
@@ -24,41 +28,44 @@ class CoordsCollection:
         _coords: The coordinates.
     """
 
-    _coords: np.ndarray | list
+    _coords: pm.AbstractArray | list
```
HGSilveri, line 31: I would use just `AbstractArrayLike` instead

pulser-core/pulser/register/_coordinates.py
```diff
@@ -24,41 +28,44 @@ class CoordsCollection:
...
 
     @cached_property  # Acts as an attribute in a frozen dataclass
-    def _sorted_coords(self) -> np.ndarray:
-        coords = np.array(self._coords, dtype=float)
-        rounded_coords = np.round(coords, decimals=COORD_PRECISION)
-        sorting = self._calc_sorting_order()
-        return cast(np.ndarray, rounded_coords[sorting])
+    def _sorted_coords(self) -> pm.AbstractArray:
+        coords = pm.vstack(cast(Sequence[AbstractArrayLike], self._coords))
+        rounded_coords = pm.round(coords, decimals=COORD_PRECISION)
```
HGSilveri, line 47: Since we are replicating this in `WeightMap`, how about putting these two lines in a `CoordsCollection._rounded_cords` property? You could even cache it

pulser-core/pulser/register/weight_maps.py
```diff
@@ -159,7 +166,8 @@ def _to_abstract_repr(self) -> dict[str, Any]:
             traps=[
                 {"weight": weight, "x": x, "y": y}
                 for weight, (x, y) in zip(
-                    self.sorted_weights, self.sorted_coords
+                    self.sorted_weights,
+                    self.sorted_coords.as_array(detach=True),
```
HGSilveri, line 170: I think these sort of changes shouldn't be necessary. I guess if we `detach=True` in `AbstractArray._to_abstract_repr()` it will be fine

HGSilveri (commented):
New changes are looking :ok_hand:

pulser-core/pulser/math/abstract_array.py
vytautas-a: I got some `mypy` errors related to `AbstractClass` not having `__iter__` . For example, when passing this object to `zip`: `for name, (x, y) in zip(names, self._coords)`.

pulser-core/pulser/register/_coordinates.py
vytautas-a: Yes, good idea, I can do that.

pulser-core/pulser/register/weight_maps.py
vytautas-a: Correct, the change in this particular place is not necessary - it's a remnant of previous solution. Now `mypy` doesn't complain about it.

pulser-core/pulser/math/abstract_array.py
HGSilveri: Ok... Well, if we are going to define `__iter__` we have to do it properly. I propose something like this: 
```suggestion
    def __iter__(self) -> Generator[AbstractArray, None, None]]:
        for i in range(self.__len__()):
            yield self.__getitem__(i)

```

pulser-core/pulser/math/abstract_array.py
vytautas-a: What do you mean? I made this change since I think it's more pythonic to check for class with `isinstance` - at least in `qadence` we do that. :)

HGSilveri: It was a guess of your motivation to do the change, I'm fine with `isinstance` :)

pulser-core/pulser/register/_coordinates.py
vytautas-a: If I make such change - a lot of `mypy` errors pop up saying that some types of `AbstractArrayLike` have no needed attributes, e.g., `Item "int" of "Union[Union[Union[_SupportsArray[dtype[Any]], _NestedSequence[_SupportsArray[dtype[Any]]], bool, int, float, complex, str, bytes, _NestedSequence[Union[bool, int, float, complex, str, bytes]]], AbstractArray], List[Any]]" has no attribute "__iter__" (not iterable)  [union-attr]`.

I think it's easier to leave `pm.AbstractArray` here.

pulser-core/pulser/math/abstract_array.py
vytautas-a: If I include the specification `[AbstractArray, None, None]` for `Generator` type I get these `mypy` errors:
```
pulser-core/pulser/register/base_register.py:91: error: Argument 2 to "zip" has incompatible type "Union[AbstractArray, List[Any]]"; expected "Iterable[ndarray[Any, Any]]"  [arg-type]
pulser-core/pulser/register/base_register.py:261: error: Argument 1 to "tolist" of "_ArrayOrScalarCommon" has incompatible type "Union[AbstractArray, Any]"; expected "_ArrayOrScalarCommon"  [arg-type]
```

HGSilveri: These errors look sensible to me, the type hints in `BaseRegister` need to be updated or the methods changed

HGSilveri: For example, I would change the `qubits` property to:

```python
    @property
    def qubits(self) -> dict[QubitId, pm.AbstractArray]:
        """Dictionary of the qubit names and their position coordinates."""
        return dict(zip(self._ids, pm.vstack(self._coords)))
```

pulser-core/pulser/math/abstract_array.py
```diff
@@ -89,6 +89,10 @@ def as_array(self, detach: bool = False) -> np.ndarray:
             )
         return np.asarray(self._array)
 
+    def as_list(self) -> list:
```
HGSilveri, line 92: We should probably call it `tolist()` instead

pulser-core/pulser/register/_coordinates.py
```diff
@@ -41,12 +41,15 @@ def sorted_coords(self) -> pm.AbstractArray:
         # Copies to prevent direct access to self._sorted_coords
         return self._sorted_coords.copy()
 
+    @cached_property
+    def rounded_coords(self) -> pm.AbstractArray:
```
HGSilveri, line 45: I would prefer if this was a private property, unless you thought of a case where it makes sense to have it public

pulser-core/pulser/register/_coordinates.py
```diff
@@ -41,12 +41,15 @@ def sorted_coords(self) -> pm.AbstractArray:
...
+        coords = pm.vstack(cast(Sequence[AbstractArrayLike], self._coords))
+        return pm.round(coords, decimals=COORD_PRECISION)
+
     @cached_property  # Acts as an attribute in a frozen dataclass
     def _sorted_coords(self) -> pm.AbstractArray:
-        coords = pm.vstack(cast(Sequence[AbstractArrayLike], self._coords))
-        rounded_coords = pm.round(coords, decimals=COORD_PRECISION)
-        sorting = self._calc_sorting_order(rounded_coords)
-        return rounded_coords[sorting]
+        sorting = self._calc_sorting_order(self.rounded_coords)
```
HGSilveri, line 51: I guess `_calc_sorting_order()` no longer needs to receive `rounded_coords`, right?

pulser-core/pulser/register/base_register.py
```diff
@@ -258,7 +265,7 @@ def _to_dict(self) -> dict[str, Any]:
         return obj_to_dict(
             self,
             cls_dict,
-            [np.ndarray.tolist(qubit_coords) for qubit_coords in self._coords],
+            [qubit_coords.as_list() for qubit_coords in self._coords],
```
HGSilveri, line 268: Does this fail for tensors that require grad?

pulser-simulation/pulser_simulation/hamiltonian.py
```diff
@@ -46,14 +47,14 @@ class Hamiltonian:
...
-        qdict: dict[QubitId, np.ndarray],
+        qdict: dict[QubitId, pm.AbstractArray],
         device: BaseDevice,
         sampling_rate: float,
         config: NoiseModel,
     ) -> None:
         """Instantiates a Hamiltonian object."""
         self.samples_obj = samples_obj
-        self._qdict = qdict
+        self._qdict = {k: v.as_array() for k, v in qdict.items()}
```
HGSilveri, line 57: Should you detach here?

pulser-core/pulser/register/base_register.py
vytautas-a: I changed this line not due to functionality issues but `mypy` error.

pulser-simulation/pulser_simulation/hamiltonian.py
vytautas-a: correct

pulser-core/pulser/register/_coordinates.py
vytautas-a: correct

pulser-core/pulser/register/base_register.py
HGSilveri: I get that, but I was just wondering if `tolist()` failed for tensors that required grad. I just checked that it just doesn't care

pulser-core/pulser/parametrized/variable.py
```diff
@@ -71,9 +73,10 @@ def _assign(self, value: Union[ArrayLike, float, int]) -> None:
         object.__setattr__(self, "_count", self._count + 1)
 
     def _validate_value(
-        self, value: Union[ArrayLike, float, int]
-    ) -> np.ndarray:
-        val = np.array(value, dtype=self.dtype, ndmin=1)
+        self, value: Union[AbstractArrayLike, float, int]
+    ) -> pm.AbstractArray:
+        val = pm.AbstractArray(value)
+        # val = np.array(value, dtype=self.dtype, ndmin=1)
```
HGSilveri, line 79: The casting to `self.dtype` and the `ndmin=1` are both important. `AbstractArray` already has a `dtype` argument, we should add `ndmin` too

pulser-core/pulser/register/_coordinates.py
```diff
@@ -24,41 +28,45 @@ class CoordsCollection:
...
     def _calc_sorting_order(self) -> np.ndarray:
         """Calculates the unique order that sorts the coordinates."""
-        coords = np.array(self._coords, dtype=float)
         # Sorting the coordinates 1st left to right, 2nd bottom to top
-        rounded_coords = np.round(coords, decimals=COORD_PRECISION)
-        dims = rounded_coords.shape[1]
-        sorter = [rounded_coords[:, i] for i in range(dims - 1, -1, -1)]
+        dims = self._rounded_coords.shape[1]
+        sorter = [
+            self._rounded_coords.as_array(detach=True)[:, i]
```
HGSilveri, line 59: This is detaching for every `i`, it would be better to detach only once and store the array in a variable

HGSilveri (commented):
Last comment and we can merge, great job!

pulser-core/pulser/parametrized/variable.py
vytautas-a: Ok, I can do that. But what's the actual reason for this?

HGSilveri: A variable's value is always a 1D array, even when it has only one entry. We accept 0D values but then internally turn them into a single-entry 1D array through this `ndmin=1` option

vytautas-a: yes, makes sense

vytautas-a: What if we skip this argument completely and simply make sure that any tensor/array gets reshaped to 1D shape if it is 0D? I noticed that `ndmin` is used only in `variable.py` and with value 1, so I think there's no need to fully replicate the behavior of `ndmin` of numpy here.

HGSilveri: > What if we skip this argument completely and simply make sure that any tensor/array gets reshaped to 1D shape if it is 0D?

I think that is a strong and potentially problematic assumption that I would prefer to avoid making in general.

> I noticed that `ndmin` is used only in `variable.py` and with value 1, so I think there's no need to fully replicate the behavior of `ndmin` of numpy here.

I agree that there's no need to fully replicate the behaviour of `ndmin`. You can make a restricted `force_array: bool` argument for this purpose, how about that?

pulser-core/pulser/math/abstract_array.py
```diff
@@ -37,19 +37,27 @@ class AbstractArray:
...
+        elif self.has_torch() and isinstance(array, torch.Tensor):
             self._array = torch.as_tensor(
                 array,
                 dtype=dtype,  # type: ignore[arg-type]
             )
         else:
             self._array = np.asarray(array, dtype=dtype)
 
+        if force_array:
+            self._array = self._array[None]
```
HGSilveri, line 59: We only want to do this for scalars:
```suggestion
        if force_array and self._array.ndim == 0:
            self._array = self._array[None]
```

HGSilveri (approved):
Perfect, let's merge! Thanks a lot for your efforts

PR #687, opened 2024-05-24
Add lightshift coefficients to RydbergEOM

HGSilveri:
- [x] Adds `blue_shift_coeff` and `red_shift_coeff` to accommodate for each beam's contribution to lightshift
- [x] Records which beams are switching in a given EOM block to `_EOMSettings`
- [x] Supports the new parameters in the abstract repr JSON schema

### DIFF
--- a/pulser-core/pulser/channels/eom.py
+++ b/pulser-core/pulser/channels/eom.py
@@ -311,11 +311,10 @@ class RydbergEOM(_RydbergEOMDefaults, BaseEOM, _RydbergEOM):
         # limit_rabi_freq is the maximum effective rabi frequency value
         # below which the lightshift can be zero
         if rabi_frequency <= limit_rabi_freq:
-            base_amp = np.sqrt(2 * rabi_frequency * self.intermediate_detuning)
-            sqrt_shift_factor = np.sqrt(shift_factor)
+            base_amp_squared = 2 * rabi_frequency * self.intermediate_detuning
             return {
-                self.limiting_beam: base_amp / sqrt_shift_factor,
-                ~self.limiting_beam: base_amp * sqrt_shift_factor,
+                self.limiting_beam: np.sqrt(base_amp_squared / shift_factor),
+                ~self.limiting_beam: np.sqrt(base_amp_squared * shift_factor),
             }
 
         # The limiting beam is at its maximum amplitude while the other
--- a/pulser-core/pulser/channels/eom.py
+++ b/pulser-core/pulser/channels/eom.py
@@ -29,6 +29,7 @@ MODBW_TO_TR = 0.48
 OPTIONAL_ABSTR_EOM_FIELDS = (
     "multiple_beam_control",
     "custom_buffer_time",
+    "blue_shift_coeff",
     "red_shift_coeff",
 )
 
@@ -136,6 +137,7 @@ class _RydbergEOM:
 @dataclass(frozen=True)
 class _RydbergEOMDefaults:
     multiple_beam_control: bool = True
+    blue_shift_coeff: float = 1.0
     red_shift_coeff: float = 1.0
 
 
@@ -154,6 +156,8 @@ class RydbergEOM(_RydbergEOMDefaults, BaseEOM, _RydbergEOM):
         custom_buffer_time: A custom wait time to enforce during EOM buffers.
         multiple_beam_control: Whether both EOMs can be used simultaneously.
             Ignored when only one beam can be controlled.
+        blue_shift_coeff: The weight coefficient of the blue beam's
+            contribution to the lightshift.
         red_shift_coeff: The weight coefficient of the red beam's contribution
             to the lightshift.
     """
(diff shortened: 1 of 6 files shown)


pulser-core/pulser/channels/eom.py
```diff
@@ -250,16 +330,26 @@ def _lightshift(
...
         if rabi_frequency <= limit_rabi_freq:
-            # Both beams the same rabi_freq
-            beam_amp = np.sqrt(2 * rabi_frequency * self.intermediate_detuning)
-            return {beam: beam_amp for beam in RydbergBeam}
+            base_amp = np.sqrt(2 * rabi_frequency * self.intermediate_detuning)
+            sqrt_shift_factor = np.sqrt(shift_factor)
+            return {
+                self.limiting_beam: base_amp / sqrt_shift_factor,
+                ~self.limiting_beam: base_amp * sqrt_shift_factor,
+            }
```
a-corni, line 352: I don't think that kills the lighshift... It seems to me the expression is easier:
image

pulser-core/pulser/sequence/sequence.py
```diff
@@ -1149,8 +1149,14 @@ def enable_eom_mode(
...
-                detuning_off = eom_config.calculate_detuning_off(
-                    amp_on, detuning_on, optimal_detuning_off
+                (
+                    detuning_off,
+                    switching_beams,
+                ) = eom_config.calculate_detuning_off(
+                    amp_on,
+                    detuning_on,
+                    optimal_detuning_off,
+                    return_switching_beams=True,
```
a-corni, line 1159: I really find it that it would make sense to have switching beams a property, it would be also way closer to the fact that you are defining it as an attribute in EOMSettings

HGSilveri: I really don't get what class you are referring to... The switching beams returned by `calculate_detuning_off()` actually depend on the inputs (when there is more than one controlled beam). They say: "for this combination of `amp_on` and `detuning_on`, what beam(s) should I be switching so that my `detuning_off` is as close as possible to `optimal_detuning_off`?"

HGSilveri (commented):
Thanks for the review @a-corni ! I must say I'm a bit confused by some of your comments

pulser-core/pulser/channels/eom.py
```diff
@@ -180,9 +196,33 @@ def __post_init__(self) -> None:
...
                 )
 
+    @overload
     def calculate_detuning_off(
-        self, amp_on: float, detuning_on: float, optimal_detuning_off: float
+        self,
+        amp_on: float,
+        detuning_on: float,
+        optimal_detuning_off: float,
+        return_switching_beams: Literal[False],
```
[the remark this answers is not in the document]
a-corni: Ah yes I agree ! I was missing the fact that they are being switched on and off.

pulser-core/pulser/channels/eom.py
a-corni: Daaamn that was stupid

pulser-core/pulser/sequence/sequence.py
a-corni: ~Yes, I get this, but you have two switching_beams: the one from calculate_detuning_off and the one from detuning_off_options. I propose to place what you use in detuning_off_options in a separate property that you can use here by doing switching_beams[det_off]~

Okay now I see why you want to return switching_beams in calculate_detuning_off and I agree it's good to have it there (I thought the user had access to closest_option in which case eom_config.switching_beams[closest_option] would be easy). I am just afraid we are doing something overly complex/hard to maintain or build above with these functions having two return types, that's why I am still suggesting having a property for switching_beams in detuning_off_options ?

pulser-core/pulser/channels/eom.py
a-corni, line 352:
```suggestion
            base_amp_squared = 2 * rabi_frequency * self.intermediate_detuning
            return {
                self.limiting_beam: np.sqrt(base_amp / shift_factor),
                ~self.limiting_beam: np.sqrt(base_amp * shift_factor),
            }
```
or surely even better
```suggestion
            base_amp_lim = np.sqrt(2 * rabi_frequency * self.intermediate_detuning / shift_factor)
            return {
                self.limiting_beam: base_amp_lim,
                ~self.limiting_beam: base_amp_lim * shift_factor,
            }
```

a-corni: And the lightshift is actually 0 with these values 🙃 Good job and sorry !

HGSilveri: The first suggestion looks better to me, though it's missing the `_squared` in the last two `base_amp`s
```suggestion
            base_amp_squared = 2 * rabi_frequency * self.intermediate_detuning
            return {
                self.limiting_beam: np.sqrt(base_amp_squared / shift_factor),
                ~self.limiting_beam: np.sqrt(base_amp_squared * shift_factor),
            }
```

a-corni: Oops indeed ;) I think second one is "faster" but eventually less clear and we are definitely not optimising down to the number of sqrt ^^'

pulser-core/pulser/sequence/sequence.py
HGSilveri: I just pushed the changes. Not a big fan of having the switching beams combinations defined independently but I guess it's okay, it did allow to do some nice simplifications

a-corni (approved):
Great ! Thanks for the modifications

PR #690, opened 2024-06-06
Fixes in `correct_phase_drift` behaviour + more robust Pulse comparison + Python 3.12 support

HGSilveri:
- Corrects two bugs with `correct_phase_drift`:
  1. The phase shift correction in `enable_eom(..., correct_phase_drift=True)` was including the fall time before the buffer, while it should have considered just the buffer. 
  2. The correction in `add_eom_pulse()` was changing only the phase of the added pulse while it should have also shifted the phase reference by this amount
- Defines a custom `Pulse.__eq__()`: This was needed for the UTs of `correct_phase_drift`. The custom method introduces a minimal precision when comparing phase values and it is insensitive to the `% (2 *np.pi)` wrapping.
- Removes the fall time from the record of the last time a qubit was used. The fall time was added only so that the `phase_shifts` were drawn after the modulated pulse was over; in hindsight, this was excessive as it prevents two consecutive pulses (with a phase shift between them) to be placed back-to-back even under the `no-delay` protocol. 
- Supports python 3.12


pulser-core/pulser/pulse.py
```diff
@@ -263,6 +265,26 @@ def __repr__(self) -> str:
...
+            return False
+
+        def check_phase_eq(phase1: float, phase2: float) -> np.bool_:
+            # Comparing with an offset ensures we don't fail just because
+            # we are very close to the wraping point
+            return np.isclose(phase1, phase2, atol=1e-6) or np.isclose(
+                (phase1 + 1) % (2 * np.pi),
+                (phase2 + 1) % (2 * np.pi),
+                atol=PHASE_PRECISION,
+            )
```
a-corni, line 279: Well this only shifts the phase value at which the issue arrises no ? from $2\pi$  to $2\pi -1$

pulser-core/pulser/pulse.py
```diff
@@ -263,6 +265,26 @@ def __repr__(self) -> str:
...
+
+        return bool(
+            self.amplitude == other.amplitude
+            and self.detuning == other.detuning
```
a-corni, line 283: Should they be equal or isclose with an absolute precision ?

a-corni (approved):

pulser-core/pulser/pulse.py
HGSilveri: The waveform comparison already has some tolerance built-in

a-corni: Ah yes thanks for reminding me :)

pulser-core/pulser/pulse.py
a-corni: I am not sure I understand... If you have phase1 at 2pi-1 - 0.0000001 and phase2 at 2pi-1+0.0000001  they will not be considered as close right ?

HGSilveri: Yes, but they will be considered close on the first `np.isclose()` (note that all phases are already stored mod 2pi)

a-corni: Aaah damn now I get it sorry I didn't see the "or" 🙃 Thanks !

PR #693, opened 2024-06-13
Optionally skip JSON schema validation after serialization

### LINKED ISSUE #689: Optionally skip validation after `Sequence.to_abstract_repr()`
After serialization to the abstract repr, we validate the outcomes against the JSON schema to ensure its valid before returning. This is a defensive practice since, in the absence of bugs, the serialized sequence should always be valid. It is also this validation that takes the vast majority of the time. 

As such, adding a `skip_validation: bool` flag to `Sequence.to_abstract_repr()` is sensible and would allow the serialization to run significantly faster.

HGSilveri:
Closes #689 .

### DIFF
--- a/pulser-core/pulser/json/abstract_repr/serializer.py
+++ b/pulser-core/pulser/json/abstract_repr/serializer.py
@@ -117,6 +117,11 @@ def serialize_abstract_sequence(
             be "cls").
         skip_validation: Whether to skip the validation of the serialized
             sequence against the abstract representation's JSON schema.
+            Skipping the validation is useful to cut down on execution
+            time, as this step takes significantly longer than the
+            serialization itself; it is also low risk, as the validation
+            is only defensively checking that there are no bugs in the
+            serialized sequence.
         defaults: The default values for all the variables declared in this
             Sequence instance, indexed by the name given upon declaration.
             Check ``Sequence.declared_variables`` to see all the variables.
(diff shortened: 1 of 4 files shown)


MatthieuMoreau0 (approved):
LGTM. My only suggetion would be to clarify in the docstring when it can be fine to skip the validation and maybe also mention the associated performance gains

HGSilveri:
> LGTM. My only suggetion would be to clarify in the docstring when it can be fine to skip the validation and maybe also mention the associated performance gains

Thank you @MatthieuMoreau0 ! What do you think of this for the docstring?
```
            skip_validation: Whether to skip the validation of the serialized
                sequence against the abstract representation's JSON schema.
                Skipping the validation is useful to cut down on execution
                time, as this step takes significantly longer than the
                serialization itself; it is also low risk, as the validation
                is only defensively checking that there are no bugs in the
                serialized sequence.
```

MatthieuMoreau0:
> > LGTM. My only suggetion would be to clarify in the docstring when it can be fine to skip the validation and maybe also mention the associated performance gains
> 
> Thank you @MatthieuMoreau0 ! What do you think of this for the docstring?
> 
> ```
>             skip_validation: Whether to skip the validation of the serialized
>                 sequence against the abstract representation's JSON schema.
>                 Skipping the validation is useful to cut down on execution
>                 time, as this step takes significantly longer than the
>                 serialization itself; it is also low risk, as the validation
>                 is only defensively checking that there are no bugs in the
>                 serialized sequence.
> ```

Looks good!

PR #696, opened 2024-06-19
Support for 3D registers and layouts in the abstract representation

### LINKED ISSUE #695: Support 3D registers in the abstract representation

HGSilveri:
Closes #695 .

### DIFF
--- a/pulser-core/pulser/json/abstract_repr/deserializer.py
+++ b/pulser-core/pulser/json/abstract_repr/deserializer.py
@@ -16,7 +16,7 @@ from __future__ import annotations
 
 import dataclasses
 import json
-from typing import TYPE_CHECKING, Any, Type, Union, cast, overload
+from typing import TYPE_CHECKING, Any, Literal, Type, Union, cast, overload
 
 import jsonschema
 import jsonschema.exceptions
@@ -57,7 +57,7 @@ from pulser.waveforms import (
 
 if TYPE_CHECKING:
     from pulser.noise_model import NoiseModel
-    from pulser.register.base_register import BaseRegister
+    from pulser.register import Register, Register3D
     from pulser.sequence import Sequence
 
 
@@ -384,7 +384,7 @@ def _deserialize_layout(layout_obj: dict[str, Any]) -> RegisterLayout:
 
 def _deserialize_register(
     qubits: list[dict[str, Any]], layout: RegisterLayout | None
-) -> BaseRegister:
+) -> Register:
     coords = [(q["x"], q["y"]) for q in qubits]
     qubit_ids = [q["name"] for q in qubits]
     if layout:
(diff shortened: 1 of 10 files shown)


pulser-core/pulser/json/abstract_repr/deserializer.py
```diff
@@ -589,20 +605,59 @@ def deserialize_abstract_layout(obj_str: str) -> RegisterLayout:
     return _deserialize_layout(json.loads(obj_str))
 
 
-def deserialize_abstract_register(obj_str: str) -> BaseRegister:
+@overload
+def deserialize_abstract_register(
```
awennersteen, line 609: All of these overload: pass confuse me.
Is that just the way to do it, or is there a reason for doing it if not required?

pulser-core/pulser/json/abstract_repr/deserializer.py
```diff
@@ -589,20 +605,59 @@ def deserialize_abstract_layout(obj_str: str) -> RegisterLayout:
...
+    obj_str: str, required_dim: Literal[None, 2, 3] = None
+) -> Register | Register3D:
     """Deserialize a register from an abstract JSON object.
 
     Args:
-        obj_str: the JSON string representing the register encoded
+        obj_str: The JSON string representing the register encoded
             in the abstract JSON format.
+        required_dim: If defined, ensures the register is of the
```
awennersteen, line 635: Honestly, calling it required_dim to me is a bit strange.

Is just dim, dimension, dimensionality or similar in some way  better?

At the same time this is not something externals will care much about so if the naming makes more sense to you I'm OK with it

pulser-core/pulser/json/abstract_repr/deserializer.py
HGSilveri: They are only for `mypy`, they allow it to restrict the return type based on the type (or in this case the actual value) of the input. Without these, I'd have to `cast` the outputs to `Register` or `Register3D` on the `.from_abstract_repr()`

pulser-core/pulser/json/abstract_repr/deserializer.py
```diff
@@ -589,20 +605,59 @@ def deserialize_abstract_layout(obj_str: str) -> RegisterLayout:
...
+            specified dimensionality.
 
     Returns:
         The Register instance.
     """
+    if required_dim not in (None, 2, 3):
+        raise ValueError(
+            "When specified, 'required_dim' must be 2 or 3, "
+            f"not {required_dim!s}."
```
awennersteen, line 644: Is the !s a typo, or syntax I'm not familiar with?

pulser-core/pulser/json/abstract_repr/deserializer.py
HGSilveri: How about `expected_dim`?

pulser-core/pulser/json/abstract_repr/deserializer.py
awennersteen: Thanks!

pulser-core/pulser/json/abstract_repr/deserializer.py
awennersteen: Huh

Had no idea. But I see this is not exactly new syntax. Wonder how I've never seen it

pulser-core/pulser/json/abstract_repr/deserializer.py
awennersteen: Sounds better to me

awennersteen (approved):
Lgtm

Thanks!

PR #700, opened 2024-06-26
Ignore the phase reference on the DMM channel's pulses

HGSilveri:
Although we can't change the phase of the pulses added to a DMM channel, the phase was still being adjusted according to the changes to the phase reference. Since the phase of the DMM channel's pulses is irrelevant, we should just keep it at zero instead.


a-corni (approved):
LGTM :)

PR #701, opened 2024-06-28
Add open batches to pulser-pasqal

oliver-gordon:
On the PCS we support open batches, to continuously add new jobs and there is support for this in the SDK but this has not been translated over to pulser-pasqal yet.

Due to the abstraction of batches in this lib I've attempted to keep away the term batch and just flow with open_submission, since our known reference is a submission_id.

When creating a new submission, a user is to either pass the submission_id and we can append new jobs to an existing batch or a boolean flag to indicate whether or not we want the submission being created to be left as open


pulser-core/pulser/backend/qpu.py
```diff
@@ -40,11 +49,26 @@ def run(
                 available.  If set to False, the call is non-blocking and the
                 obtained results' status can be checked using their `status`
                 property.
+            open_submission: A flag indicating whether or not the submission
+                should be for a single job or accept future jobs before
+                closing.
+            submission_id: If you have a preexisting submission and wish to add
+                more jobs to it then you can provide the ID here and 'run'
+                will attempt allocate jobs.
```
HGSilveri, line 57: I feel this option is misplaced here... To me, it would make more sense to have a different method for adding jobs to an open submission. 

Another thought: would it make sense if a backend kept track of an open submission, disallowing you to make a new submission before the former is closed? What's the harm if a user forgets to close an open submission?

pulser-core/pulser/backend/qpu.py
```diff
@@ -40,11 +49,26 @@ def run(
...
 
         Returns:
             The results, which can be accessed once all sequences have been
             successfully executed.
         """
+
+        # if submission is already open, we don't need an ID for a preexisting
+        # one
+        if open_submission and submission_id:
```
HGSilveri, line 66: I think this check should be done inside `submit()`

pulser-pasqal/pulser_pasqal/backends.py
```diff
@@ -72,7 +76,7 @@ def run(
                 available.  If set to False, the call is non-blocking and the
                 obtained results' status can be checked using their `status`
                 property.
-
+            open: Flag to determine whether or not
```
HGSilveri, line 79: I guess you forgot to copy the full thing here?

HGSilveri (commented):
Thanks for this first draft @oliver-gordon ! I have a left few comments, but let me focus on a design question here.

Does it make sense to use the same `run()` method for both starting a submission and adding jobs to one? ~~I suggested below that it made more sense as a separate backend method, but it would also work as a method of the `RemoteResults` class you get after you call `run()`.~~

EDIT: That was a bad idea, let's please disregard it.

MatthieuMoreau0:
>Due to the abstraction of batches in this lib I've attempted to keep away the term batch and just flow with open_submission, since our known reference is a submission_id.

Should we not introduce the notion of batch at all in this lib? If it becomes the main entrypoint to submit jobs, shouldn't we use the same language as on other components of our platform - where the notion of batch is central. I am worried it would create some confusion for users when switching from pulser-pasqal to the user portal. Curious what you guys think on that front

MatthieuMoreau0:
I wanted to suggest using a context manager to handle the batch lifecycle after reading Henrique's comments. Let me know what you guys think 

Starting the context manager would create a batch and store the associated `batch_id` as an attribute of the backend instance. Whenever calling the run method within the context manager, it would add jobs automatically to that batch without the user having to pass the `submission_id` explicitly. Exiting the context manager closes the batch automatically.

If the run method is used outside the context manager, it creates a closed batch with the jobs passed as argument.

e.g the user interface would look something like

Open batch example:
```python
"""Open batch example"""
# opening the context manager creates a batch. 
# Here I used the wording "session" but it could be called "batch"
with QPUBackend(seq, connection).start_session() as backend:
     # these jobs are automatically added to the batch
     results = backend.run(job_params, wait=True)
     # some more jobs added to the same batch
     results_2 = backend.run(job_params_2, wait=True)
#we exited the context manager so the batch is now closed
```

Closed batch
```python
backend = QPUBackend(seq, connection)
# Create a closed batch with some jobs
results = backend.run(job_params, wait=True)
# Create another batch with some jobs
results_2 = session.run(job_params_2, wait=True)
```

I didn't think through the implementation details of the `start_session` but I think it should be manageable. Thoughts?

oliver-gordon:
@MatthieuMoreau0 `Should we not introduce the notion of batch at all in this lib?` That would not be for this MR, where the intention is to introduce functionality. To introduce the concept of a batch would be to modify all the existing interfaces

pulser-core/pulser/backend/qpu.py
oliver-gordon: Yeah I think so too now. My initial thought was `run()` is a generic term for executing something, so reusing it wouldn't be outrageous but, I can move the features into another method

HGSilveri:
> > Due to the abstraction of batches in this lib I've attempted to keep away the term batch and just flow with open_submission, since our known reference is a submission_id.
> 
> Should we not introduce the notion of batch at all in this lib? If it becomes the main entrypoint to submit jobs, shouldn't we use the same language as on other components of our platform - where the notion of batch is central. I am worried it would create some confusion for users when switching from pulser-pasqal to the user portal. Curious what you guys think on that front

Indeed, I have avoided referring to "batches" so far, but in reality "submission" is still only an internal parameter too, so we could start calling it "batch" from the get go. However, I'm not sure about this because it seems to me that the way the user will interact with the `Batch` is different - they won't be taking the batch, adding jobs to it and closing, but rather using the backend to do all these things. So on the one hand, calling a submission a "batch" will provide a useful link to a concept of `pasqal-cloud`, but on the other hand it might induce confusion.

HGSilveri:
> Open batch example:
> 
> ```python
> backend = QPUBackend(seq, connection)
> # opening the context manager creates a batch. 
> # Here I used the wording "start session" but it could be called "open_batch"
> with backend.start_session():
>      # these jobs are automatically added to the batch
>      results = backend.run(job_params, wait=True)
>      # some more jobs added to the same batch
>      results_2 = backend.run(job_params_2, wait=True)
> #we exited the context manager so the batch is now closed
> ```
> 
> Closed batch (no change)
> 
> ```python
> backend = QPUBackend(seq, connection)
> # Create a closed batch with some jobs
> results = backend.run(job_params, wait=True)
> # Create another batch with some jobs
> results_2 = session.run(job_params_2, wait=True)
> ```
> 
> Thoughts about this general interface? What about the context manager? It's simply syntactic sugar but it can help prevent users forgetting to close their batch

I quite like this, I find it elegant, relatively simple and it meets the criteria of having the actions performed by the backend. One question though: when you call `run()` for the second time in an open batch, does the `results_2` object give a view of only the results of the new jobs, or also include the results of previously submitted jobs? In other others, does it complement `results` or does it replace it? I suspect you were going for the first option and it is also the one that makes most sense to me

awennersteen:
I personally would welcome having the concept of Batch in Pulser. However it would be in a way that I can re-use in downstreams libraries, and ideally we'd have it consistent with the cloud backend and emulation backends (even if that means usually wrapping the results of the emulator as a single job in a batch).

Due to that, I think I Agree with Oliver that it might be out of scope here.

MatthieuMoreau0:
I was not suggesting to introduce batches in this MR specifically either; and definitely not as an object the user should manipulate. I was wondering mostly about the wording because as Henrique mentioned `submission_id` is currently an internal implementation detail the user is not aware of so we could rename it "batch" freely if we want to share the cloud naming conventions

MatthieuMoreau0:
>when you call run() for the second time in an open batch, does the results_2 object give a view of only the results of the new jobs, or also include the results of previously submitted jobs?

I don't know tbh. Whatever makes most sense given the current implementation of backends in Pulser

awennersteen:
In Henriques example with open batches what if we do:

```python
backend = QPUBackend(seq, connection)
# opening the context manager creates a batch. 
with batch as backend.start_session():
     # we can (optionally?) pass the context to run
     results = backend.run(job_params, wait=True, batch=batch)
     # some more jobs added to the same batch, and the results
     # are just the results from that single run
     results_2 = backend.run(job_params_2, wait=True, batch)
```

HGSilveri:
> In Henriques example with open batches what if we do:
> 

 *Matthieu's example, I don't want to take away his credit :slightly_smiling_face: 

> ```python
> backend = QPUBackend(seq, connection)
> # opening the context manager creates a batch. 
> with batch as backend.start_session():
>      # we can (optionally?) pass the context to run
>      results = backend.run(job_params, wait=True, batch=batch)
>      # some more jobs added to the same batch, and the results
>      # are just the results from that single run
>      results_2 = backend.run(job_params_2, wait=True, batch)
> ```
> 
> With the actual batch implementation could come in a followup, so that we can think of that interface a bit/not scope creep this mr

I'm not convinced with exposing a batch object here, I feel like it strays away from the backend interface philosophy and it defeats the purpose of trying to simplify things for the more "junior" Pulser users. Let's remember that `pasqal-cloud` is not going away, so the option to work with batches will remain available for whoever needs them.

For the purposes of this PR which is just allowing an open batch/submission, I feel like we can get away with the backend storing internally the open batch ID and just adding jobs to it on every `run()` call that happens within the context (instead of starting a new batch).

HGSilveri:
> I suppose there are some assumptions that the runtime doesn't end for us to retain an ID as a class property? What if a user is running incredibly short lived scripts? or executing something again to add another job after already creating and submitting a batch? We would be missing the submission_id property
> 
> I can't speak for the regular usecases of pulser-pasqal users but, this one seems rather plausible?

I would go back to the argument that using `pasqal-cloud` directly is still an option. This implementation provides a restricted access to the feature and that's okay. 
With `pulser-pasqal`, if a user wants to have use an open batch they have to stay within the context and the moment the exit it, the batch is closed. In exchange for forcing the user to work within these bounds, I think this approach provides a simpler interface (eg not having to deal with batch IDs) and it is foolproof (eg. the user doesn't have to know/remember to close the batch). In my opinion, this is fair tradeoff.

oliver-gordon:
I can live with this. I'll push my change to the draft today to see if we're aligned on how it will look

tests/test_backend.py
```diff
@@ -281,3 +290,26 @@ def test_qpu_backend(sequence):
...
+    unique_test_id = "unique-id-string"
+
+    with pytest.warns(DeprecationWarning, match="From v0.18"):
+        seq = sequence.switch_device(replace(DigitalAnalogDevice, max_runs=10))
+
+    with QPUBackend(seq, connection).open_submission(unique_test_id) as qpub:
+        remote_results = qpub.run(job_params=[{"runs": 10}])
+        assert isinstance(remote_results, RemoteResults)
+        # confirms context manager assigns ID
+        assert qpub.submission_id == unique_test_id
```
MatthieuMoreau0, line 306: nit: you could assert the submission_id is reset when you exit the context manager

MatthieuMoreau0 (commented):
Looks good to me.

I would like to reiterate the question about naming things `submission` vs `batch` especially now that these are public attributes/methods using these names. imo calling things `batch` makes the most sense.

pulser-core/pulser/backend/qpu.py
```diff
@@ -61,10 +69,51 @@ def run(
...
+    def __exit__(
+        self,
+        exc_type: Type[BaseException] | None,
+        exc_value: BaseException | None,
+        traceback: TracebackType | None,
+    ) -> None:
+        # exiting the context of an open submission will
+        # make a remote call to close it.
+        if self.submission_id:
+            self.close_submission(self.submission_id)
```
HGSilveri, line 115: Following up on the idea that `close_submission()` should not be exposed as a public method, this should used the connection directly
```suggestion
            self._connection.close_submission(submission_id)
```

pulser-core/pulser/backend/remote.py
```diff
@@ -70,6 +70,17 @@ def results(self) -> tuple[Result, ...]:
         """The actual results, obtained after execution is done."""
         return self._results
 
+    @property
+    def submission_id(self) -> str:
```
HGSilveri, line 74: Reminder that we still need to decide whether we want to use "batch" instead of "submission"

pulser-pasqal/pulser_pasqal/pasqal_cloud.py
```diff
@@ -137,14 +141,24 @@ def submit(
...
-            configuration=configuration,
-            wait=wait,
-        )
+        batch = None
+        if submission_id:
+            submit_jobs_fn = backoff_decorator(self._sdk_connection.add_jobs)
+            batch = submit_jobs_fn(
+                batch_id=submission_id,
+                jobs=job_params or [],  # type: ignore[arg-type]
+            )
```
HGSilveri, line 150: Another thing here: what makes more sense within the backend interface is that the `RemoteResults()` returned when adding jobs to a batch contains only the added jobs, not the full batch. This will require changing `RemoteResults` to accept the  job IDs for the subset of jobs in a batch it should care about.

HGSilveri (commented):
Ok, this still needs some work. 
Apart from the comments I just left, I would suggest that the open submission functionality is moved to `RemoteBackend` instead of being defined just in `QPUBackend`. Also, the ability to use open submissions should depend on whether the `RemoteConnection` supports it

pulser-core/pulser/backend/qpu.py
```diff
@@ -61,10 +69,51 @@ def run(
...
+            self reference for current object
+        """
+        if submission_id:
+            self.submission_id = submission_id
+        else:
+            submission = cast(
+                RemoteResults, self._connection.submit(self._sequence)
+            )
+            self.submission_id = submission.submission_id
+        return self
```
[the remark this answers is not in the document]
oliver-gordon: Regarding point 1, I wouldn't see this as an outrageous feature. Most methods you typically bind with a context manager are functional without. I wouldn't imagine there is an issue with a user writing their own control flow, which is why close_submission was left public

pulser-core/pulser/backend/qpu.py
```diff
@@ -61,10 +69,51 @@ def run(
...
+        # open returns an instance of self to use open_submission within a
+        # context manager
+        return self
+
+    def __exit__(
+        self,
+        exc_type: Type[BaseException] | None,
+        exc_value: BaseException | None,
+        traceback: TracebackType | None,
+    ) -> None:
```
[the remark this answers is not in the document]
oliver-gordon: Not entirely clear what you mean by `which is returned by open_submission()`. A class instance?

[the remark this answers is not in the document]
HGSilveri: Apologies, it was indeed not very clear. I meant that `open_submission()` should return an instance of a class with these `__enter__()` and `__exit__()` methods defined, and it will be there that the batch is opened and closed.

pulser-core/pulser/backend/qpu.py
```diff
@@ -42,7 +42,9 @@ def __init__(
         super().__init__(sequence, connection, mimic_qpu=True)
 
     def run(
-        self, job_params: list[JobParams] | None = None, wait: bool = False
+        self,
+        job_params: list[JobParams] | None = None,
+        wait: bool = False,
```
HGSilveri, line 47: nit: Given there's no actual change here anymore, maybe we can revert the formatting modification?

pulser-core/pulser/backend/remote.py
```diff
@@ -17,7 +17,10 @@
 import typing
 from abc import ABC, abstractmethod
 from enum import Enum, auto
-from typing import Any, TypedDict
+from types import TracebackType
+from typing import Any, Type, TypedDict, cast
+
+from typing_extensions import Self
```
HGSilveri, line 23: Hum, this might be problematic, `typing_extensions` is not in our requirements

pulser-core/pulser/backend/remote.py
```diff
@@ -96,7 +99,12 @@ class RemoteConnection(ABC):
 
     @abstractmethod
     def submit(
-        self, sequence: Sequence, wait: bool = False, **kwargs: Any
+        self,
+        sequence: Sequence,
+        wait: bool = False,
+        complete: bool = False,
+        batch_id: str | None = None,
```
HGSilveri, line 106: Since we are renaming `submission_id` with `batch_id`, we should do it everywhere on the methods below too

pulser-core/pulser/backend/remote.py
```diff
@@ -148,6 +159,7 @@ def __init__(
                 "'connection' must be a valid RemoteConnection instance."
             )
         self._connection = connection
+        self.batch_id = ""
```
HGSilveri, line 162: I think it'd be best to keep this private ie `self._batch_id`

pulser-core/pulser/backend/remote.py
```diff
@@ -161,3 +173,34 @@ def _type_check_job_params(job_params: list[JobParams] | None) -> None:
...
+
+    def __exit__(
+        self,
+        exc_type: Type[BaseException] | None,
+        exc_value: BaseException | None,
+        traceback: TracebackType | None,
+    ) -> None:
+        # On context exit, we make a remote call to close the open batch
+        self._connection._close_batch(self.batch_id)
+        self.batch_id = ""
```
HGSilveri, line 206: Here's my issue with defining these methods here: 
```
backend = QPUBackend(...)
with backend:
   backend.run(...)  # Fails because the batch was not opened and there's no batch ID
``` 
or also
```
backend.open_batch() # This still opens the batch
backend.run(...)  # A batch_id is stored so this will add to the open batch
```

This is why I suggested that, instead of returning `Self`, `open_batch()` could return an auxiliary class specifically designed to handle open batches. Also, I guess the batch should
only be opened on `__enter__()`.

pulser-pasqal/pulser_pasqal/backends.py
HGSilveri: This can be deleted now

pulser-pasqal/pulser_pasqal/pasqal_cloud.py
```diff
@@ -159,15 +164,28 @@ def submit(
...
-            jobs=job_params or [],  # type: ignore[arg-type]
-            emulator=emulator,
-            configuration=configuration,
-            wait=wait,
-        )
 
+        # If batch_id is not empty, thedn we can submit new jobs to a
+        # batch we just created otherwise, create a new one with
+        #  _sdk_connection.create_batch()
+        if batch_id:
```
oliver-gordon, line 171: So this condition exists here because it's the only thing called via `run()` but if we expand run into another method for `append` jobs or something, we can dump the `if` and just have another method that calls `sdk_connection.add_jobs()`

pulser-core/pulser/backend/remote.py
```diff
@@ -161,3 +173,34 @@ def _type_check_job_params(job_params: list[JobParams] | None) -> None:
...
                 )
+
+    def open_batch(self) -> Self:
+        """Create an open batch that can continue recieve new job.
+
+        As long as jobs are submitted within an open context
+        manager.
+
+        Returns:
+            A class instance with an associated batch_id property
```
HGSilveri, line 184:
```suggestion
        """Creates an open batch that can receive new jobs.

        Note:
            Jobs must be submitted within an open context manager.

        Returns:
            A context manager within which jobs can be submitted to 
            the open batch. 
```

pulser-core/pulser/backend/remote.py
```diff
@@ -161,3 +173,34 @@ def _type_check_job_params(job_params: list[JobParams] | None) -> None:
...
+        """
+        # Create batch and receive submission id
+        submission = cast(
+            RemoteResults, self._connection.submit(self._sequence, open=True)
+        )
```
HGSilveri, line 189: I guess this creates an empty batch because there are no `job_params`, right? I'm assuming that's ok for the SDK?

pulser-pasqal/pulser_pasqal/pasqal_cloud.py
```diff
@@ -100,7 +100,12 @@ def __init__(
         )
 
     def submit(
-        self, sequence: Sequence, wait: bool = False, **kwargs: Any
+        self,
+        sequence: Sequence,
+        wait: bool = False,
+        complete: bool = True,
```
HGSilveri, line 106: You are calling this parameter `open` elsewhere

pulser-pasqal/pulser_pasqal/pasqal_cloud.py
```diff
@@ -234,3 +252,10 @@ def _convert_configuration(
 
         pasqal_config_kwargs["strict_validation"] = strict_validation
         return emu_cls(**pasqal_config_kwargs)
+
+    def _close_submission(self, submission_id: str) -> SubmissionStatus | None:
+        """Mark a submission as closed so no more jobs can be submitted."""
+        if not submission_id:
+            return None
+        batch = self._sdk_connection.close_batch(submission_id)
+        return SubmissionStatus(batch.id)
```
HGSilveri, line 261: - You called this `_close_batch()` on the parent class
- Why do you allow `submisssion_id` to not be defined?

pulser-core/pulser/backend/remote.py
oliver-gordon: ah so would it be an issue for me to introduce the dependency?
Linting fails if I don't type `Self` and I can't type `Self` for 3.8 without it

pulser-pasqal/pulser_pasqal/pasqal_cloud.py
oliver-gordon: Ah yeah I'll clean this up. We're discussing to swap from complete -> open elsewhere offline my mind slipped on this

pulser-core/pulser/backend/remote.py
HGSilveri: I would say this doesn't warrant the extra dependency since we have alternatives:

1. We might not even be returning `Self` in the end (see comments below)
2. You can use a `TypeVar` instead, as mentioned [here](https://docs.python.org/3/library/typing.html#typing.Self)

pulser-core/pulser/backend/remote.py
oliver-gordon: So I know it's ok for the API. Will confirm the SDK has no issue with it

pulser-core/pulser/backend/remote.py
oliver-gordon: I am not sure I have a mental image of what you are expecting when you are thinking we return a class these are bound to. Unless you share provide a simple example?

I wrote these in an inherited parent class initially which I kept getting a method order problem for, so I just removed the inherited class and brought them one step lower.

HGSilveri: Alright, here's an example:

```
class _OpenBatchManager:
    def __init__(self, backend: RemoteBackend):
        self.backend = RemoteBackend
    def __enter__(self) -> _OpenBatchManager:
        self.backend._open_batch = True
    def __exit__(self, ...) -> None:
        self.backend._open_batch = False
        self.backend._batch_id = ""
```

Then, in `RemoteBackend` you'll have:
```
   def open_batch(self) -> _OpenBatchManager:
       return _OpenBatchManager(self)
    
    def run(
        self, job_params: list[JobParams] | None = None, wait: bool = False
    ) -> RemoteResults:
        results = self._connection.submit(
            self._sequence, job_params=job_params, wait=wait, open=self._open_batch, batch_id=self._batch_id
        )
        if self._open_batch:
            self._batch_id = results._submission_id
            self._open_batch = False
        return results
```

pulser-core/pulser/backend/remote.py
oliver-gordon: Just to close this thread, this will be fine if the branch is `open`. Then we can add jobs after

pulser-core/pulser/backend/qpu.py
```diff
@@ -66,7 +68,10 @@ def run(
             job_params or [], self._sequence.device.max_runs
         )
         results = self._connection.submit(
-            self._sequence, job_params=job_params, wait=wait
+            self._sequence,
+            job_params=job_params,
+            wait=wait,
+            batch_id=self.batch_id,
```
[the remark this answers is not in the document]
MatthieuMoreau0: nit suggestion: default it to None instead of empty string

pulser-core/pulser/backend/remote.py
```diff
@@ -148,6 +157,7 @@ def __init__(
                 "'connection' must be a valid RemoteConnection instance."
             )
         self._connection = connection
+        self._batch_id = ""
```
MatthieuMoreau0, line 160: nit: I think the default should be None to be consistent with the submit method args

tests/test_backend.py
```diff
@@ -169,3 +183,16 @@ def test_qpu_backend(sequence):
 
     results = remote_results.results
     assert results[0].sampling_dist == {"00": 1.0}
+
+    # Test create a batch and submitting jobs via a context manager
+    # behaves as expected.
+    qpu = QPUBackend(seq, connection)
+    with qpu.open_batch() as ob:
+        print(ob)
```
MatthieuMoreau0, line 191: nit: rm

MatthieuMoreau0 (commented):
The implementation with the context manager looks good. I pointed out 2 typos.

On a general note, in my opinion the context manager should be provided simply for convenience to the user and documented as the "quick start" usage. The user may open their batch and close it on their own without the context manager if they want to as suggested by @oliver-gordon in another comment. @HGSilveri you seem to think otherwise so we can discuss this further. We can start with this implementation and provide the non-context manager method down the road if we agree on it.

What is the process for documentation in this repo; should it be part of this MR or another one? In any case we need to document this feature before it gets released.


I'll let Henrique approve this MR as he has more knowledge of the codebase but I have no blocking comments at this point.

pulser-core/pulser/json/abstract_repr/deserializer.py
```diff
@@ -409,7 +409,6 @@ def _deserialize_register3d(
 
 
 def _deserialize_noise_model(noise_model_obj: dict[str, Any]) -> NoiseModel:
-
```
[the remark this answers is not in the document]
oliver-gordon: Can't say it was intentional or noticed

pulser-core/pulser/backend/remote.py
HGSilveri: I agree

pulser-core/pulser/backend/remote.py
```diff
@@ -161,3 +171,29 @@ def _type_check_job_params(job_params: list[JobParams] | None) -> None:
                     "All elements of 'job_params' must be dictionaries; "
                     f"got {type(d)} instead."
                 )
+
+    def open_batch(self) -> _OpenBatchContextManager:
+        """Create an open batch within a context manager object"""
```
HGSilveri, line 176: nit: Not sure if the linter caught it, but the style guide we follow needs a couple modifications:
```suggestion
        """Creates an open batch within a context manager object."""
```

pulser-core/pulser/backend/remote.py
HGSilveri: Ok, so `pasqal-cloud` supports opening an empty batch after all, or is that features being added?

pulser-pasqal/pulser_pasqal/pasqal_cloud.py
HGSilveri: So, we'll go with `open` in the end, right?

pulser-pasqal/pulser_pasqal/pasqal_cloud.py
HGSilveri: This still needs to be addressed. I can take over that part if you wish

pulser-core/pulser/backend/remote.py
```diff
@@ -122,6 +128,9 @@ def fetch_available_devices(self) -> dict[str, Device]:
             "remote connection."
         )
 
+    def _close_batch(self, batch_id: str) -> SubmissionStatus | None:
+        pass
```
HGSilveri, line 132: - You still have to fill this in on the `PasqalCloud` class. 
- Since this is not an `abstractmethod`, let's raise a `NotImplementedError` here.
- Why are you returning `SubmissionStatus`? I think returning `None` should be fine

pulser-pasqal/pulser_pasqal/pasqal_cloud.py
HGSilveri: Ok, but in that case you have to change it in the `__enter__()` because there it's `open` atm. Also, I think `complete` warrants a small comment because to me the meaning is not obvious from the name (especially if I was unfamiliar with the Cloud-SDK)

pulser-core/pulser/backend/remote.py
oliver-gordon: Mypy thinks otherwise

pulser-core/pulser/backend/remote.py
```diff
@@ -161,3 +179,34 @@ def _type_check_job_params(job_params: list[JobParams] | None) -> None:
                     "All elements of 'job_params' must be dictionaries; "
                     f"got {type(d)} instead."
                 )
+
+    def open_batch(self) -> _OpenBatchContextManager:
+        """Creates an open batch within a context manager object."""
+        if not self._connection.supports_open_batch():
+            raise NotImplementedError(  # pragma: no cover
```
HGSilveri, line 186: This seems like something we can test, so I'd get rid of the `no cover`

pulser-core/pulser/backend/remote.py
```diff
@@ -161,3 +179,34 @@ def _type_check_job_params(job_params: list[JobParams] | None) -> None:
...
+
+
+class _OpenBatchContextManager:
+    def __init__(self, backend: RemoteBackend) -> None:
+        self.backend = backend
+
+    def __enter__(self) -> _OpenBatchContextManager:
+        batch = cast(
+            RemoteResults,
+            self.backend._connection.submit(self.backend._sequence, open=True),
```
HGSilveri, line 199:
```suggestion
            self.backend._connection.submit(self.backend._sequence, complete=False),
```

pulser-pasqal/pulser_pasqal/pasqal_cloud.py
oliver-gordon: Yeah, I'm aware it's not clear. Poor naming conventions is popular cloud side, cc rebatch

I'll add a comment. We have changes down the line to rename this so once the SDK has that change we can make an update here in the future

MatthieuMoreau0: We have an open MR to rename `complete` to `open` in the SDK:  https://github.com/pasqal-io/pasqal-cloud/pull/131

So we can name this `open` in pulser-pasqal right away

MatthieuMoreau0: I was sure I left a comment on this yesterday but I can't find it again, apologies if this is a duplicate.

There is an open [MR](https://github.com/pasqal-io/pasqal-cloud/pull/131) to rename `complete` to `open` in the SDK so it should be named `open` right away in pulser pasqal

oliver-gordon: Yeah, I just replied to your comment? We'll need to wait until the MR goes in and I can bump the SDK version otherwise it's only going to be referred to as `open` on the pulser-pasqal side

HGSilveri:
Still needed:

- [ ] Incorporate changes in #707 once that is merged
- [ ] Test everything is working with an actual connection

HGSilveri:
@oliver-gordon @MatthieuMoreau0 I managed to replace Submission -> Batch in a backwards compatible way everywhere except in `RemoteResults` (basically, the notion of "submission" is contained in the `RemoteResults` class). There are still links between submission and batch in `RemoteResults` though (eg the `RemoteResults.batch_id` property).
I had to replicate `SubmissionStatus` as `BatchStatus` and keep both because there are users that might have scripts where they call `RemoteResults.get_status()` and they expect a `SubmissionStatus` instance as the return type.
Is this acceptable for you?

pulser-core/pulser/backend/remote.py
```diff
@@ -81,14 +99,14 @@ def __init__(
         job_ids: list[str] | None = None,
```
MatthieuMoreau0, line 99: why not call the submission_id argument batch_id ?

pulser-core/pulser/backend/remote.py
```diff
@@ -81,14 +99,14 @@ def __init__(
...
+        self._batch_id = submission_id
         self._connection = connection
         if job_ids is not None and not set(job_ids).issubset(
-            all_job_ids := self._connection._get_job_ids(self._submission_id)
+            all_job_ids := self._connection._get_job_ids(self._batch_id)
         ):
             unknown_ids = [id_ for id_ in job_ids if id_ not in all_job_ids]
             raise RuntimeError(
-                f"Submission {self._submission_id!r} does not contain jobs "
+                f"Submission {self._batch_id!r} does not contain jobs "
```
MatthieuMoreau0, line 109: The error should mention batch instead of submission, no ?

MatthieuMoreau0 (approved):
A few suggestions to deprecate "submission" but otherwise lgtm

pulser-core/pulser/backend/remote.py
HGSilveri: Because I'm afraid someone is calling `RemoteResults(submission_id=...)` and I don't want to break it

pulser-core/pulser/backend/remote.py
HGSilveri: Good question, I didn't change it to avoid mixing up "batch" and "submission" within the same class, but I guess here it wouldn't be so bad

pulser-core/pulser/backend/remote.py
```diff
@@ -98,23 +116,30 @@ def results(self) -> tuple[Result, ...]:
...
     @property
     def job_ids(self) -> list[str]:
-        """The IDs of the jobs within this results submission."""
+        """The IDs of the jobs within these results' submission."""
         if self._job_ids is None:
-            return self._connection._get_job_ids(self._submission_id)
+            return self._connection._get_job_ids(self._batch_id)
         return self._job_ids
 
     def get_status(self) -> SubmissionStatus:
```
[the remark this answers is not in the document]
HGSilveri: Yeah, I guess that will have to happen at some point. I just feel like this will be a mess to deprecate properly so I was avoiding doing that in this PR but if you tell me it should happen right away, I'll do it

MatthieuMoreau0 (approved):
Thanks for the changes!

PR #703, opened 2024-07-01
Support differentiability through Torch tensors

HGSilveri:
This PR adds native support for `torch.Tensor` on select quantities, so that the final Hamiltonian can be differentiated wrt to these parameters. A few important design considerations were taken here:

1. `torch` will be an optional dependency: Users that don't require this feature should be able to keep installing and using pulser as before. For those who do want the `torch` support, they can have it by adding `[torch]` to their `pip install` commands:
```bash
# Both options work
pip install pulser[torch]
pip install pulser-core[torch]
```
2. To handle both `np.ndarray`s and `torch.Tensor`s indiscriminately, the wrapper class `AbstractArray` is created. This class contains an array-like object, which is stored either as `torch.Tensor` or a `np.ndarray`. To the largest extent possible, instances of this class can be handled like array, allowing operations, indexing, etc... However, their use should be only internal, so they contain only the necessary functionality for the existent codebase - whenever extra functionality is needed, it can always be added.
3. Operations on `AbstractArray` must use custom functions to preserve the `AbstractArray` type in the output. As such, analgous definitions of the necessary `numpy/torch` functions were defined in the new `pulser.math` module. This new module is, by convention, imported with  `import pulser.math as pm` and can often be used as a drop-in replacement for `np`.
4. Parameters supporting ~~differentiability~~tensor-like behaviour are marked with the new ~~`pm.Differentiable`~~`pm.TensorLike` type-hint. When this is present, a `torch.Tensor` (with `requires_grad=True` or not) can be provided. Whenever a parameter is not marked as `pm.TensorLike`, providing a `torch.tensor` is not guaranteed to work as intended.


pulser-core/pulser/math/__init__.py
```diff
@@ -0,0 +1,238 @@
...
+
+
+# Custom function definitions
+
+
+def exp(a: AbstractArrayLike, /) -> AbstractArray:
+    a = AbstractArray(a)
+    if a.is_tensor:
+        return AbstractArray(torch.exp(a.as_tensor()))
+    return AbstractArray(np.exp(a.as_array()))
```
a-corni, line 49: I see a common pattern to all these functions... Perhaps we could have the pattern in a function `_func_abstr_array`

```suggestion
def _fun_abstr_array(a: AbstractArrayLike, torch_fun: Callable, np_fun:Callable) -> AbstractArray
    a = AbstractArray(a)
    if a.is_tensor:
        return AbstractArray(torch_fun(a.as_tensor()))
    return AbstractArray(np_fun(a.as_array()))

def exp(a: AbstractArrayLike, /) -> AbstractArray:
    return _fun_abstr_array(a, torch.exp, np.exp)

```

a-corni (commented):
Thanks for tackling this ! Here are some comments about the integration of this new feature in github workflow, as well as some comments on the new pulser.math module :)

pulser-core/pulser/pulse.py
```diff
@@ -103,15 +104,17 @@ def __init__(
...
+        if np.any(amplitude.samples.as_array(detach=True) < 0):
             raise ValueError(
                 "All samples of an amplitude waveform must be "
                 "greater than or equal to zero."
             )
         object.__setattr__(self, "amplitude", amplitude)
         object.__setattr__(self, "detuning", detuning)
-        phase = cast(float, phase)
-        object.__setattr__(self, "phase", float(phase) % (2 * np.pi))
+        assert not isinstance(phase, Parametrized)
```
a-corni, line 114: So phase can't be Parametrized ? If so, we should remove Parametrized from the possible types of phase.

pulser-core/pulser/waveforms.py
```diff
@@ -69,14 +82,9 @@ def __init__(self, duration: Union[int, Parametrized]):
...
         """
-        duration = cast(int, duration)
-        try:
-            _duration = int(duration)
-        except (TypeError, ValueError):
-            raise TypeError(
-                "duration needs to be castable to an int but "
-                f"type {type(duration)} was provided."
-            )
+        assert not isinstance(duration, Parametrized)
```
a-corni, line 85: Can't the duration be Parametrized ? I am puzzled with the type of `duration` that is `Union[int, Parametrized]`...

pulser-core/pulser/waveforms.py
```diff
@@ -181,11 +189,22 @@ def modulated_samples(
         Returns:
             The array of samples after modulation.
         """
+        detach = True  # We detach unless...
+        if self.samples.is_tensor and self.samples.as_tensor().requires_grad:
+            # ... the samples require grad. In this case, we clear the cache
+            # so that the modulation is recalculated with the current samples
+            self._modulated_samples.cache_clear()
```
a-corni, line 196: The modulated buffers don't have to be recomputed because in differentiation, the duration is not expected to a variable in the gradient (only the values are used), do I get it right ? 😄

pulser-core/pulser/waveforms.py
```diff
@@ -295,28 +314,36 @@ def _check_slice(self, s: slice) -> slice:
         return slice(start, stop)
 
     @abstractmethod
-    def __mul__(self, other: float) -> Waveform:
+    def __mul__(self, other: float | ArrayLike) -> Waveform:
```
a-corni, line 317: ArrayLike or AbstractArrayLike ? :)

pulser-core/pulser/waveforms.py
```diff
@@ -295,28 +314,36 @@ def _check_slice(self, s: slice) -> slice:
...
         pass
 
     def __neg__(self) -> Waveform:
         return self.__mul__(-1.0)
 
-    def __truediv__(self, other: float) -> Waveform:
-        if other == 0:
+    def __truediv__(self, other: float | ArrayLike) -> Waveform:
```
a-corni, line 323: - Isn't AbstractArrayLike better than ArrayLike ?
- Should not we either check for the dimension of the provided AbstractArrayLike, or, if the dimension can be higher than 0, that the length of the other is the same as self, and that it does not contain any 0 value ?

pulser-core/pulser/waveforms.py
```diff
@@ -466,8 +492,10 @@ def __str__(self) -> str:
     def __repr__(self) -> str:
         return f"CustomWaveform({self.duration} ns, {self.samples!r})"
 
-    def __mul__(self, other: float) -> CustomWaveform:
-        return CustomWaveform(self._samples * float(other))
+    def __mul__(self, other: float | ArrayLike) -> CustomWaveform:
+        return CustomWaveform(
+            self._samples * pm.AbstractArray(other, dtype=float)
```
a-corni, line 497: Are we sure this dtype work with torch tensors ? 😮

pulser-core/pulser/waveforms.py
```diff
@@ -481,26 +509,27 @@ class ConstantWaveform(Waveform):
...
         self,
         duration: Union[int, Parametrized],
-        value: Union[float, Parametrized],
+        value: Union[float, pm.Differentiable, Parametrized],
     ):
         """Initializes a constant waveform."""
         super().__init__(duration)
-        value = cast(float, value)
-        self._value = float(value)
+        assert not isinstance(value, Parametrized)
```
a-corni, line 516: I am still puzzled by these assertions

pulser-core/pulser/waveforms.py
```diff
@@ -734,13 +772,18 @@ def _to_abstract_repr(self) -> dict[str, Any]:
...
 
-    def __mul__(self, other: float) -> BlackmanWaveform:
-        return BlackmanWaveform(self._duration, self._area * float(other))
+    def __mul__(self, other: float | ArrayLike) -> BlackmanWaveform:
+        return BlackmanWaveform(
+            self._duration, self._area * pm.AbstractArray(other, dtype=float)
+        )
 
 
 class InterpolatedWaveform(Waveform):
```
a-corni, line 789: We don't touch the type of values in InterpolatedWaveform because we don't have the interpolator in torch ?

a-corni (commented):
Some comments about waveform, Pulse, Channels and Devices

pulser-core/pulser/register/register.py
```diff
@@ -189,20 +201,26 @@ def triangular_lattice(
...
+        if spacing_ <= 0.0:
             raise ValueError(
                 f"Spacing between atoms (`spacing` = {spacing})"
                 " must be greater than 0."
             )
 
-        coords = patterns.triangular_rect(rows, atoms_per_row) * spacing
-
+        coords = (
+            pm.AbstractArray(patterns.triangular_rect(rows, atoms_per_row))
```
a-corni, line 213: Could there be some value in returning the patterns directly as AbstractArray ?

pulser-core/pulser/register/_coordinates.py
```diff
@@ -24,7 +27,7 @@ class CoordsCollection:
         _coords: The coordinates.
     """
 
-    _coords: np.ndarray | list
+    _coords: pm.AbstractArray | list
```
a-corni, line 30: If I am right, in Traps, we still define _coords from trap_coordinates, that is a `np.ndarray`. Either we go for 
```suggestion
    _coords: AbstractArrayLike | list
```
Or we transform trap_coordinates in traps.py `object.__setattr__(self, "_coords", pm.AbstractArray(trap_coordinates))`. By the way, do I get it right that trap_coordinates in Traps is not expected to be differentiable because it is related with a physical object and because it must be match some calibrated layouts that are not AbstractArray ?

pulser-core/pulser/register/weight_maps.py
```diff
@@ -63,7 +65,7 @@ def __init__(
     @property
```
a-corni, line 65: Can't there be use cases where we wan't to modify the weights of the detuning map ? Such that weights could be typing.Sequence[float | pm.Differentiable] ?

pulser-core/pulser/register/weight_maps.py
```diff
@@ -72,7 +74,7 @@ def sorted_weights(self) -> np.ndarray:
         return cast(np.ndarray, np.array(self.weights)[sorting])
 
     def get_qubit_weight_map(
-        self, qubits: Mapping[QubitId, np.ndarray]
+        self, qubits: Mapping[QubitId, pm.AbstractArray]
```
a-corni, line 77: From a user perspective, wouldn't `AbstractArrayLike` be better suited (the current implementation is introducing a breaking change, we can't provide np.ndarray anymore) ?

a-corni (commented):
This is regarding the register :)

a-corni (commented):
I agree with the handling of torch tensors in the more complex structures. I have only couple questions on the pulser-pasqal side - but it's more future work that work that should be done at the moment.

.github/workflows/ci.yml
```diff
@@ -67,8 +68,13 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
           extra-packages: pytest
+          with-torch: ${{ matrix.with-torch }}
       - name: Run the unit tests & generate coverage report
+        if: ${{ matrix.with-torch == 'with-torch' }}
         run: pytest --cov --cov-fail-under=100
+      - name: Run the unit tests without torch installed
+        if: ${{ matrix.with-torch != 'with-torch' }}
+        run: pytest --cov
```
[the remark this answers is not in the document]
HGSilveri: When we test "with-torch", we are testing both with torch _and_ with numpy, so we are covering everything. Without torch, we just care whether nothing fails because we already tested everything in principle. I'm not seeing the need to complicate things further but perhaps I'm missing something

pulser-core/pulser/math/abstract_array.py
```diff
@@ -0,0 +1,301 @@
...
+            self._array = np.asarray(array, dtype=dtype)
+
+        if force_array and self._array.ndim == 0:
+            self._array = self._array[None]
+
+    @staticmethod
+    @functools.lru_cache
+    def has_torch() -> bool:
+        """Checks whether torch is installed."""
+        return importlib.util.find_spec("torch") is not None
```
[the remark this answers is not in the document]
HGSilveri: `find_spec()` does not import the module, so we'd still need the `import torch`. We could make it conditional on `HAS_TORCH` but I don't see how that would be an improvement tbh.
I think this approach is more robust because the value of `has_torch()` is protected, unlike with `HAS_TORCH` being a constant.
Finally, I'm afraid the typing issues would not resolve because all typing checks are done statically ie the code is not executed. Therefore, the static type checkers won't know what `HAS_TORCH` is and they don't evaluate `if` statements. I suspect defining the types as you proposed will even throw an error.

pulser-core/pulser/math/abstract_array.py
```diff
@@ -0,0 +1,301 @@
...
+    """An abstract array containing an array or tensor.
+
+    Args:
+        array: The array to store.
+    """
+
+    def __init__(
+        self,
+        array: AbstractArrayLike,
+        dtype: DTypeLike = None,
```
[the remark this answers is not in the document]
HGSilveri: - `torch.Tensor` is actually `ArrayLike` so it is also `AbstractArrayLike`. 
- `DTypeLike` is a bit abusive but it works for the usual built-in types (eg `int`, `float`, `complex`, `bool`, ...). You'll see there is a `# type: ignore[arg-type]` further down because it doesn't fully work.

pulser-core/pulser/math/abstract_array.py
```diff
@@ -0,0 +1,301 @@
...
+        return torch.as_tensor(self._array)
+
+    def as_array(self, *, detach: bool = False) -> np.ndarray:
+        """Converts the stored array to a Numpy array.
+
+        Args:
+            detach: Whether to detach before converting.
+        """
+        if detach and self.is_tensor:
+            return cast(torch.Tensor, self._array).detach().numpy()
```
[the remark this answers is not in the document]
HGSilveri: Keep in mind that `AbstractArray` is not supposed to be a user facing object, we want the users to keep providing either numpy arrays or torch tensors. Rather, `AbstractArray` allows us to internally handle both input types adequately.

> Perhaps we could also have the `detach()` method implemented ?

I have not had the need to have a `AbstractArray.detach()` method so far, nor do I see how it could be useful tbh.

> I think you (almost) always have to detach() a torch tensor to be able to convert it to a numpy array. Additionnaly, having the `detach` would also enable the serializing of a torch tensor (detaching it from its graph) https://pytorch.org/docs/stable/generated/torch.Tensor.detach.html#torch.Tensor.detach

Not really, you only have to `detach()` when `requires_grad=True`. That is way this method allows you to choose whether to detach or not: if you don't care about the computational graph, you can set `detach=True`; otherwise it won't detach so as to preserve the computational graph and will eventually fail if you are not handling things properly. 

> More globally, I think I understand from this function that you don't want to implement all the methods a torch tensor and a numpy array would have in their syntax, but rather develop a new syntax to implement all these methods. It's fine by me, but I am wondering if it's good from a user point of view, if we want to think of a user programming in the torch framework with other purposes than differentiation (as suggested by @awennersteen) ?

Again, keeping in mind that `AbstractArray` should only be used internally, we only need to implement the methods and functions we need, as that need arises. In the few instances where a user is given an `AbstractArray`, we expect them to convert them to the type of their choice (`array` or `tensor`) before doing whatever computations they want. Nonetheless, you should be able to treat an `AbstractArray` as a `np.darray` the majority of the time.

pulser-core/pulser/math/abstract_array.py
```diff
@@ -0,0 +1,301 @@
...
+
+    def _to_abstract_repr(self) -> Any:
+        try:
+            return self.as_array().tolist()
+        except RuntimeError as e:
+            raise NotImplementedError(
+                "A tensor that requires grad can't be serialized without"
+                " losing the computational graph information."
+            ) from e
+
```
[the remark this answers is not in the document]
HGSilveri: This I can't really say. So far it was not necessary

pulser-core/pulser/math/__init__.py
```diff
@@ -0,0 +1,238 @@
...
+)
+
+try:
+    import torch
+except ImportError:  # pragma: no cover
+    pass
+
+
+class Differentiable(Protocol):
+    """A type hint to signal that a parameter may be differentiable."""
```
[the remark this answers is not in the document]
HGSilveri: Yeah, I just liked the idea of having the type hint signaling whether a parameter is differentiable or not. Especially for scalars, `TensorLike` does not convey the right idea to me

pulser-core/pulser/waveforms.py
HGSilveri: Yes :)

pulser-core/pulser/waveforms.py
HGSilveri: Yeah, pretty much

HGSilveri (commented):
Thanks a lot for the review @a-corni !

pulser-core/pulser/math/__init__.py
a-corni: As you like :) Perhaps some type errors might emerge with `Callable` indeed... I was also thinking about understandability, I am wondering if this solution does not decrease the understandability of the whole concept - even though I think it increases readability, since there are a lot of functions in these page...

pulser-core/pulser/channels/base_channel.py
```diff
@@ -450,10 +451,10 @@ def _modulation_padding(self) -> int:
 
     def modulate(
         self,
-        input_samples: np.ndarray,
+        input_samples: ArrayLike,
```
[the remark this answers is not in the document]
a-corni: Ok thanks for these explanations, I was solely focused on the union of types, not on what they contain.

pulser-core/pulser/devices/_device_datacls.py
```diff
@@ -452,10 +459,14 @@ def _params(self, init_only: bool = False) -> dict[str, Any]:
         }
 
     def _validate_coords(
-        self, coords_dict: dict[QubitId, np.ndarray], kind: str = "atoms"
+        self,
+        coords_dict: (
+            Mapping[QubitId, pm.AbstractArray] | Mapping[int, np.ndarray]
```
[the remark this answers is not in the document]
a-corni: 😢 Mypy rules

pulser-core/pulser/register/_coordinates.py
a-corni: If I got right what you suggested above, won't `ArrayLike` work ?
```suggestion
    _coords: ArrayLike | list
```

pulser-core/pulser/register/weight_maps.py
a-corni: I agree, let's keep it "simple" 😄

pulser-core/pulser/register/weight_maps.py
a-corni: From your comment above,  I guess you can even go for `ArrayLike` ?

pulser-core/pulser/sampler/samples.py
```diff
@@ -167,26 +166,26 @@ def extend_duration(self, new_duration: int) -> ChannelSamples:
         if extension < 0:
             raise ValueError("Can't extend samples to a lower duration.")
 
-        new_amp = np.pad(self.amp, (0, extension))
+        new_amp = pm.pad(self.amp, (0, extension))
         # When in EOM mode, we need to keep the detuning at detuning_off
         if self.eom_blocks and self.eom_blocks[-1].tf is None:
-            final_detuning = self.eom_blocks[-1].detuning_off
+            final_detuning = float(self.eom_blocks[-1].detuning_off)
```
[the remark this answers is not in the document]
a-corni: I see that you regularly do `float(AbstractArray)`. From what I get, it calls `AbstractArray.astype` with `dtype=float`. In the case where a torch tensor is in `AbstractArray`, it converts it using `Array.to(dtype=float)`, but I am wondering if that effectively works, since I think it should be  `Array.to(dtype=torch.float)` (see [here](https://pytorch.org/docs/stable/generated/torch.Tensor.to.html#torch.Tensor.to))

[Edit] That's silly, I just got that it calls AbstractArray.__float__ that only handles AbstractArray._array. So it's fine by me.

pulser-core/pulser/waveforms.py
a-corni: Okay, so line 66 of this file you keep the "Parametrized" in `def __init__(self, duration: Union[int, Parametrized]):` only for the description ?

pulser-core/pulser/waveforms.py
a-corni: Solved.

pulser-core/pulser/waveforms.py
a-corni: Then I flatly resolve this conservation.

pulser-core/pulser/waveforms.py
a-corni: But now, I understand them.

pulser-core/pulser/waveforms.py
a-corni: Resolved.

pulser-core/pulser/waveforms.py
a-corni: - From the discussion above: definitely ArrayLike.
- Ookay, I did not get that __mul__ (and, per extension, __truediv__) was for 1-dimensional elements. It was failing ungracefully before, so I guess let's keep it failing ungracefully. However, even though it's intended to work with scalars only, I think it now works with 1D arrays as well... Do you want to put a condition on the dimension of the arrays in __mul__ ?

pulser-core/pulser/pulse.py
a-corni: Resolved.

.github/workflows/ci.yml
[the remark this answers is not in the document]
HGSilveri: Well, they will see incomplete coverage, yes. But note that I included `torch` in the development install, so if you are a developer you need to have it.

pulser-core/pulser/register/_coordinates.py
HGSilveri: Here it's a bit tricky because it is both the type hint of both the input (which would be fine) but also the attribute itself - and there you run into issues.

pulser-core/pulser/register/weight_maps.py
HGSilveri: Yes, that's what I did :)

pulser-core/pulser/register/base_register.py
```diff
@@ -271,16 +284,14 @@ def __eq__(self, other: Any) -> bool:
...
-                for i, id in enumerate(self._ids)
-            )
+        return self._ids == other._ids and np.allclose(
+            self._coords_arr.as_array(detach=True),
+            other._coords_arr.as_array(detach=True),
         )
 
+    def __repr__(self) -> str:
+        return f"{self.__class__.__name__}({self.qubits})"
+
```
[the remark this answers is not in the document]
HGSilveri: I just realized that registers printed badly when I was debugging so I added it. I know it's out of scope but I figured it was harmless

pulser-pasqal/pulser_pasqal/pasqal_cloud.py
```diff
@@ -110,8 +109,12 @@ def submit(
...
                     "for a sequence not addressing a single basis."
                 )
-            # The copy prevents changing the input sequence
-            sequence = copy.deepcopy(sequence)
+            # This is equivalent to performing a deepcopy
+            # All tensors are converted to arrays but that's ok, it would
+            # have happened anyway later on
+            sequence = Sequence.from_abstract_repr(
+                sequence.to_abstract_repr(skip_validation=True)
+            )
```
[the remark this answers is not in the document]
HGSilveri: That's the thing, for now we don't have a way to serialize the full computational graph so we won't support it. If you try to submit to a torch-based backend, two things can happen:
- Your tensors don't `require_grad`, in which case they don't even have to be tensors and this will work fine
- Some tensors `require_grad` and the serialization will fail (an error will be raised because you didn't detach). This is on purpose, it tells the user that differentiability is not supported on remote backends (at least for now)

.github/workflows/ci.yml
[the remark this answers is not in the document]
a-corni: Ooh I did not get that :) Then I am fine !

a-corni (approved):
Looks good to me ! 👍

PR #705, opened 2024-07-02
Add states labels to Channels and SequenceSamples

a-corni:
- [x] I am associating states to the bases addressed by Channels
- [x] such that they can be combined to make addressed states in Device and Sequence
- [x] and they unify the handling of these quantities in hamiltonian.py
Conventions.rst is also modified to take into account these labels, but I am not sure it actually improves the understanding of the conventions since it introduces a different label for XY.


a-corni:
Is this good enough ? 
image
Otherwise I am ready for review

pulser-core/pulser/channels/base_channel.py
```diff
@@ -90,12 +100,45 @@ def basis(self) -> str:
         """The addressed basis name."""
         pass
 
+    @property
+    def eigenstates(self) -> list[States]:
+        r"""The eigenstates associated with the basis.
+
+        Returns a tuple of labels, ranked decreasingly according
+        to their associated eigenenergy, as such:
```
HGSilveri, line 108: nit:
```suggestion
        Returns a tuple of labels, ranked in decreasing order
        of their associated eigenenergy, as such:
```

pulser-core/pulser/channels/base_channel.py
```diff
@@ -35,6 +35,16 @@
...
 
+States = Literal["u", "d", "r", "g", "h"]  # TODO: add "x" for leakage
+
+STATES_RANK = get_args(States)
+
+EIGENSTATES: dict[str, list[States]] = {
+    "ground-rydberg": ["r", "g"],
+    "digital": ["g", "h"],
+    "XY": ["u", "d"],
+}
```
HGSilveri, line 46: It seems that we are assuming the states here will be always in decreasing order of eigenenergy. I would maybe add a check (an assert is fine) to avoid human error

pulser-core/pulser/sequence/sequence.py
```diff
@@ -460,6 +465,13 @@ def get_addressed_bases(self) -> tuple[str, ...]:
         """Returns the bases addressed by the declared channels."""
         return tuple(self._basis_ref)
 
+    def get_addressed_states(self) -> list[States]:
+        """Returns the states addressed by the declared channels."""
+        all_states = set().union(
+            *(set(EIGENSTATES[basis]) for basis in self.get_addressed_bases())
+        )
+        return [state for state in STATES_RANK if state in all_states]
```
HGSilveri, line 473: I feel like there is a bit of code replication in this block, with small variations. Do you think we could make a single function that takes in a list of bases and returns the ordered states? If it makes sense, we could even make it a static method

HGSilveri (commented):
LGTM overall, I just have a few suggestions :)

pulser-core/pulser/channels/base_channel.py
a-corni: I have added an assert directly in the file, I am not sure that's what you had in mind ?

pulser-core/pulser/sequence/sequence.py
a-corni: I have added a function to tackle this, let me know what you think (and what you think of the name of this function as well ^^')

pulser-core/pulser/channels/base_channel.py
```diff
@@ -18,7 +18,16 @@
 import warnings
 from abc import ABC, abstractmethod
 from dataclasses import MISSING, dataclass, field, fields
-from typing import Any, Literal, Optional, Type, TypeVar, cast, get_args
+from typing import (
+    Any,
+    Collection,
```
HGSilveri, line 23: nit: I think this should be imported from `collections.abc` instead

HGSilveri (approved):
LGTM, thanks!

pulser-core/pulser/channels/base_channel.py
a-corni: Indeed ! Did not know that :) https://stackoverflow.com/questions/75285711/difference-between-collections-abc-sequence-and-typing-sequence

PR #707, opened 2024-07-10
[FEAT] Handle batches with partial results

MatthieuMoreau0:
- Stop raising an Exception if one of the job of the batch has no results after fetching. 
- Stop raising an exception when fetching results if the status of the batch is a termination status.

Currently if one of the job of a batch fails, it is impossible to retrieve the results for any jobs of the batch - this raises an exception. In this PR, I suggest instead to add `None` to the list of results for that submission when a job has no results. I don't know if returning None is the most idiomatic way to do this in pulser; I also contemplated creating a new subclass of `Result` called `EmptyResult` but I wanted to gather some feedback before going for that because it would imply some refactoring of the `Result` class and I'm not entirely convinced it would bring much.

### DIFF
--- a/pulser-core/pulser/backend/remote.py
+++ b/pulser-core/pulser/backend/remote.py
@@ -121,10 +121,9 @@ class RemoteResults(Results):
         some jobs associated to the submission do not have results.
 
         Returns:
-            dict[str, Result]: A dictionary mapping the job ID to their results.
+            dict[str, Result]: A dictionary mapping the job ID to its results.
             Jobs with no result are omitted.
         """
-
         results = {
             k: v[1]
             for k, v in self._connection._query_job_progress(
@@ -183,7 +182,7 @@ class RemoteConnection(ABC):
         Unlike `_fetch_result`, this method does not raise an error if some
         jobs associated to the submission do not have results.
 
-        It returns a dictionnary mapping the job ID to their status and results.
+        It returns a dictionnary mapping the job ID to its status and results.
         """
         pass
 
--- a/pulser-core/pulser/backend/remote.py
+++ b/pulser-core/pulser/backend/remote.py
@@ -120,8 +120,9 @@ class RemoteResults(Results):
         Unlike the `results` property, this method does not raise an error if
         some jobs associated to the submission do not have results.
 
-        It returns a dictionnary mapping the job ID to their results. Jobs with
-        no result are omitted.
+        Returns:
+            dict[str, Result]: A dictionary mapping the job ID to their results. 
+            Jobs with no result are omitted.
         """
 
         results = {
@@ -176,7 +177,7 @@ class RemoteConnection(ABC):
     def _query_job_progress(
         self, submission_id: str
     ) -> Mapping[str, tuple[JobStatus, Result | None]]:
-        """Fetches the status and results of a submission.
+        """Fetches the status and results of all the jobs in a submission.
 
         Unlike `_fetch_result`, this method does not raise an error if some
         jobs associated to the submission do not have results.
--- a/pulser-core/pulser/backend/remote.py
+++ b/pulser-core/pulser/backend/remote.py
@@ -103,6 +103,13 @@ class RemoteResults(Results):
         """Gets the status of the remote submission."""
         return self._connection._get_submission_status(self._submission_id)
 
+    def _get_available_result(self, submission_id: str) -> dict[str, Result]:
+        """Return available results and ignore jobs with no results."""
+        return {
+            k: v
+            for k, v in self._connection._query_result(submission_id).items()
+        }
+
     def __getattr__(self, name: str) -> Any:
         if name == "_results":
             status = self.get_status()
--- a/pulser-core/pulser/backend/remote.py
+++ b/pulser-core/pulser/backend/remote.py
@@ -44,6 +44,17 @@ class SubmissionStatus(Enum):
     PAUSED = auto()
 
 
+class JobStatus(Enum):
+    """Status of a remote job."""
+
+    PENDING = auto()
+    RUNNING = auto()
+    DONE = auto()
+    CANCELED = auto()
+    ERROR = auto()
+    PAUSED = auto()
+
+
 class RemoteResultsError(Exception):
     """Error raised when fetching remote results fails."""
 
@@ -103,28 +114,42 @@ class RemoteResults(Results):
         """Gets the status of the remote submission."""
         return self._connection._get_submission_status(self._submission_id)
 
-    def _get_available_result(self, submission_id: str) -> dict[str, Result]:
-        """Return available results and ignore jobs with no results."""
-        return {
-            k: v
-            for k, v in self._connection._query_result(submission_id).items()
-            if v is not None
+    def get_available_results(self, submission_id: str) -> dict[str, Result]:
+        """Returns the available results of a submission.
+
+        Unlike the `results` property, this method does not raise an error if
+        some jobs associated to the submission do not have results.
+
+        It returns a dictionnary mapping the job ID to their results. Jobs with
+        no result are omitted.
+        """
+
+        results = {
+            k: v[1]
+            for k, v in self._connection._query_job_progress(
+                submission_id
+            ).items()
+            if v[1] is not None
         }
+
+        if self._job_ids:
+            return {k: v for k, v in results.items() if k in self._job_ids}
+        return results
 
     def __getattr__(self, name: str) -> Any:
         if name == "_results":
-            status = self.get_status()
-            if status == SubmissionStatus.DONE:
+            try:
                 self._results = tuple(
                     self._connection._fetch_result(
                         self._submission_id, self._job_ids
                     )
                 )
-                return self._results
-            raise RemoteResultsError(
-                "The results are not available. The submission's status is "
-                f"{str(status)}."
-            )
+            except RemoteResultsError as e:
+                raise RemoteResultsError(
+                    "Results are not available for all jobs. Use the "
+                    "`get_available_results` method to retrieve partial "
+                    "results."
+                ) from e
         raise AttributeError(
             f"'RemoteResults' object has no attribute '{name}'."
         )
@@ -148,7 +173,16 @@ class RemoteConnection(ABC):
         pass
 
     @abstractmethod
-    def _query_result(self, submission_id: str) -> Mapping[str, Result | None]:
+    def _query_job_progress(
+        self, submission_id: str
+    ) -> Mapping[str, tuple[JobStatus, Result | None]]:
+        """Fetches the status and results of a submission.
+
+        Unlike `_fetch_result`, this method does not raise an error if some
+        jobs associated to the submission do not have results.
+
+        It returns a dictionnary mapping the job ID to their status and results.
+        """
         pass
 
     @abstractmethod
--- a/pulser-core/pulser/backend/remote.py
+++ b/pulser-core/pulser/backend/remote.py
@@ -17,7 +17,7 @@ from __future__ import annotations
 import typing
 from abc import ABC, abstractmethod
 from enum import Enum, auto
-from typing import Any, TypedDict
+from typing import Any, Mapping, TypedDict
 
 from pulser.backend.abc import Backend
 from pulser.devices import Device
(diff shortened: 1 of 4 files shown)


HGSilveri:
> I don't know if returning None is the most idiomatic way to do this in pulser; I also contemplated creating a new subclass of `Result` called `EmptyResult` but I wanted to gather some feedback before going for that because it would imply some refactoring of the `Result` class and I'm not entirely convinced it would bring much.

I was thinking of something along those lines too, I think it would be better than `None`. I also considering having `PendingResult` for distinguishing pending jobs from failed ones.

HGSilveri:
I think we both agree that the current format that we use to return results is limiting us and that there doesn't seem to be a nice way to include partial results within it. 
As such, I would propose we add an alternative, more flexible method to `RemoteResults` instead. I was thinking of something along the lines of 
```
RemoteResults.get_available_results() -> dict[JobId, Result]:
```
This would give the user a workaround when something failed in the traditional method. We could also point the user to this alternative method through the error message.
What do you think about this @MatthieuMoreau0 ?

pulser-core/pulser/backend/remote.py
```diff
@@ -139,6 +147,10 @@ def _fetch_result(
         """Fetches the results of a completed submission."""
         pass
 
+    @abstractmethod
+    def _query_result(self, submission_id: str) -> Mapping[str, Result | None]:
```
HGSilveri, line 151: Could you add a short description to this?

pulser-core/pulser/backend/remote.py
```diff
@@ -103,6 +103,14 @@ def get_status(self) -> SubmissionStatus:
         """Gets the status of the remote submission."""
         return self._connection._get_submission_status(self._submission_id)
 
+    def _get_available_result(self, submission_id: str) -> dict[str, Result]:
```
HGSilveri, line 106: Shouldn't we make this public?

pulser-core/pulser/backend/remote.py
```diff
@@ -103,6 +103,14 @@ def get_status(self) -> SubmissionStatus:
...
+        """Return available results and ignore jobs with no results."""
```
HGSilveri, line 107: I think explaining the output format would be helpful

pulser-core/pulser/backend/remote.py
```diff
@@ -103,6 +103,14 @@ def get_status(self) -> SubmissionStatus:
...
+        return {
+            k: v
+            for k, v in self._connection._query_result(submission_id).items()
```
HGSilveri, line 110: You'll want to select the subset of job IDs withint this `RemoteResults` instance

pulser-pasqal/pulser_pasqal/pasqal_cloud.py
```diff
@@ -183,37 +183,44 @@ def _fetch_result(
...
     ) -> tuple[Result, ...]:
         # For now, the results are always sampled results
+        full_results = self._query_result(submission_id)
+
+        if job_ids is None:
+            job_ids = list(full_results.keys())
+
+        results = []
+        for id in job_ids:
+            assert full_results[id] is not None, "Failed to fetch the results."
```
HGSilveri, line 193: Let's turn this into a proper `raise` statement, informing the user they can still get the result of successful jobs via the new method

pulser-pasqal/pulser_pasqal/pasqal_cloud.py
```diff
@@ -183,37 +183,44 @@ def _fetch_result(
...
         sdk_jobs = batch.ordered_jobs
-        if job_ids is not None:
-            ind_job_pairs = [
-                (job_ids.index(job.id), job)
-                for job in sdk_jobs
-                if job.id in job_ids
-            ]
-            ind_job_pairs.sort()
-            sdk_jobs = [job for _, job in ind_job_pairs]
+
```
HGSilveri, line 209: nit: we don't need `sdk_jobs`, we can give `batch.ordered_jobs` directly to the for loop

HGSilveri (commented):
Thanks for your work on this!

pulser-pasqal/pulser_pasqal/pasqal_cloud.py
HGSilveri: Let's also make it conditional on the status?

pulser-core/pulser/backend/remote.py
HGSilveri:
```suggestion
    def get_available_results(self, submission_id: str) -> dict[str, Result]:
```

pulser-core/pulser/backend/remote.py
HGSilveri: nit: Just a bit clear to me
```suggestion
    def _query_job_results(self, submission_id: str) -> Mapping[str, Result | None]:
```

pulser-core/pulser/backend/remote.py
```diff
@@ -103,6 +103,14 @@ def get_status(self) -> SubmissionStatus:
...
+            if v is not None
+        }
+
     def __getattr__(self, name: str) -> Any:
```
HGSilveri, line 114: Let's stop blocking the call to `_fetch_result()` depending on the status

pulser-core/pulser/backend/remote.py
MatthieuMoreau0: Done.

fetch_result uses singular while query_job_results is now plural. Is that ok?

pulser-core/pulser/backend/remote.py
MatthieuMoreau0: Done

pulser-core/pulser/backend/remote.py
MatthieuMoreau0: Added some extra comment, lmk if that looks good to you

pulser-core/pulser/backend/remote.py
MatthieuMoreau0: Done

pulser-pasqal/pulser_pasqal/pasqal_cloud.py
MatthieuMoreau0: Should be good

pulser-core/pulser/backend/remote.py
HGSilveri: Yes, let's not overthink it :)

pulser-pasqal/pulser_pasqal/pasqal_cloud.py
```diff
@@ -183,19 +185,27 @@ def _fetch_result(
...
-            job_ids = list(full_results.keys())
+            job_ids = list(jobs.keys())
 
-        results = []
+        results: list[Result] = []
         for id in job_ids:
-            assert full_results[id] is not None, "Failed to fetch the results."
-            results.append(cast(Result, full_results[id]))
+            status, result = jobs[id]
+            if status != {JobStatus.PENDING, JobStatus.RUNNING}:
```
HGSilveri, line 196: Didn't you mean
```suggestion
            if status in {JobStatus.PENDING, JobStatus.RUNNING}:
```
?

pulser-core/pulser/backend/remote.py
```diff
@@ -148,7 +173,16 @@ def _fetch_result(
         pass
 
     @abstractmethod
-    def _query_result(self, submission_id: str) -> Mapping[str, Result | None]:
+    def _query_job_progress(
+        self, submission_id: str
+    ) -> Mapping[str, tuple[JobStatus, Result | None]]:
+        """Fetches the status and results of a submission.
```
HGSilveri, line 179: nit: 
```suggestion
        """Fetches the status and results of all the jobs in a submission.
```

pulser-pasqal/pulser_pasqal/pasqal_cloud.py
HGSilveri: Yup!

tests/test_pasqal.py
```diff
@@ -77,40 +82,38 @@ def __init__(
...
+            _MockJob(),
             _MockJob(result={"00": 10}),
             _MockJob(result={"11": 10}),
         ]
-        sequence_builder = seq_.to_abstract_repr()
+    )
+    seq_ = copy.deepcopy(_seq)
+    seq_.declare_channel("rydberg_global", "rydberg_global")
+    seq_.measure()
+    sequence_builder = seq_.to_abstract_repr()
```
HGSilveri, line 108: It's weird to have this within the class, why did you need it?

tests/test_pasqal.py
```diff
@@ -77,40 +82,38 @@ def __init__(
...
-        status = "DONE"
-        ordered_jobs = [
-            mock_job,
+@dataclasses.dataclass
+class MockBatch:
+    id = "abcd"
+    status: str = "DONE"
+    ordered_jobs: list[_MockJob] = dataclasses.field(
+        default_factory=lambda: [
+            _MockJob(),
```
HGSilveri, line 100: If I get it right, the changes were needed so that you gained access to take the `MockBatch` definition outside of the fixture?

tests/test_pasqal.py
```diff
@@ -190,15 +196,82 @@ def test_remote_results(fixt, mock_batch, with_job_id):
...
+
+    fixt = mock_pasqal_cloud_sdk(batch)
+
+    remote_results = RemoteResults(
+        batch.id,
+        fixt.pasqal_cloud,
+    )
+
+    fixt.mock_cloud_sdk.get_batch.reset_mock()
+    with pytest.raises(RemoteResultsError):
```
HGSilveri, line 228: Can you check the error message matches what you expect?

tests/test_pasqal.py
```diff
@@ -190,15 +196,82 @@ def test_remote_results(fixt, mock_batch, with_job_id):
...
+        ],
+    )
+
+    fixt = mock_pasqal_cloud_sdk(batch)
+    remote_results = RemoteResults(
+        batch.id,
+        fixt.pasqal_cloud,
+    )
+
+    with pytest.raises(RemoteResultsError):
```
HGSilveri, line 255: Same here, the message should be different right?

tests/test_pasqal.py
```diff
@@ -190,15 +196,82 @@ def test_remote_results(fixt, mock_batch, with_job_id):
...
+    assert available_results == {
+        job.id: SampledResult(
+            atom_order=("q0", "q1", "q2", "q3"),
+            meas_basis="ground-rydberg",
+            bitstring_counts=job.result,
+        )
+        for job in batch.ordered_jobs
+        if job.result is not None
+    }
+
```
HGSilveri, line 267: If you could, some checks that the expected methods in the mock are being called would be nice

HGSilveri (commented):
I think I got why you needed so many changes in the tests but let me know if I got it wrong

tests/test_pasqal.py
MatthieuMoreau0: Yeah the goal was to be able to customize the MockBatch, e.g. to set its status or results. This enables testing for the "partial results" case

tests/test_pasqal.py
MatthieuMoreau0: Ah yes I very much intended to do that, on it!

tests/test_pasqal.py
MatthieuMoreau0: Ah in fact the message we catch in the test is the same because we catch the error and re-raise it [here](https://github.com/pasqal-io/Pulser/blob/mm/partial-results/pulser-core/pulser/backend/remote.py#L150). I did this because it felt wrong to mention the `get_vailable_results` method of the `RemoteResults` object inside the `PasqalCloud` class.

The details of why the results are not available will be displayed in the stacktrace though. If you think having to check the stacktrace is unpractical, I can maybe concatenate the error messages instead.

HGSilveri: That sounds like overkill, no need to bother

tests/test_pasqal.py
MatthieuMoreau0: It was in the fixture before. I can move it to the top level if you prefer

Note: I'm not sure we even need a valid sequence builder, I'll remove it and see if tests still pass

tests/test_pasqal.py
MatthieuMoreau0: Done

tests/test_pasqal.py
MatthieuMoreau0: I moved it to another function, let me know if that's better for you

HGSilveri: Yeah, looks fine :)

PR #708, opened 2024-07-10
Allow modification of the EOM setpoint without disabling EOM mode

HGSilveri:
This turned out to be easier than I first expected. By reusing existing methods as much as possible, there was no need to modify internal logic and the modulation seems to work without need for changes too.

- [x] Implements `Sequence.modify_eom_setpoint()`
- [x] Updates the abstract representation schema to support this new instruction


a-corni (approved):
Looks good to me :) I let you add the schema !

a-corni (approved):

PR #710, opened 2024-07-17
Reworking the NoiseModel interface

### LINKED ISSUE #697: Rework the NoiseModel interface
There are few issues with the existing `NoiseModel` interface:

 1. **Cumbersome UX**: When activating a set of noises, they have to be given to the `noise_types` argument _and_ the corresponding parameters have to be modified from their defaults. For example, to include only a 10% rate of false positives, one needs to define `NoiseModel(noise_types=("SPAM",), p_false_pos=0.1, p_false_neg=0., state_prep_error=0.)`. Preferably, one should only need to define `NoiseModel(p_false_pos=0.1)`.
 
 2. **Error-prone**: In fact, the "preferred way" shown above seems to be something users are intuitively doing already and it currently results in no noise being actually included, which is admittedly misleading.
 
 3. **Arbitrary/misleading default values**: The default values for every parameter where inherited from the `SimConfig` class, which was created in a time where providing reasonable defaults was needed. Now that we can associate default noise models to devices, keeping some universal default values makes much less sense. Furthermore, having non-zero default values is an issue in itself, since it includes noise through parameters the user never explicitly defined.
 
 4. **Confusing `__repr__()`**: As a consequence of having all these default values, the `NoiseModel.__repr__()` (autogenerated because `NoiseModel` is a dataclass) is a confusing mess of default values. 
 
 **Proposed changes:**

1. **Deprecate the default values**: We can't just change them because it will break existing code (or worse, silently change the results). However, we can deprecate them and mark them for removal in v1.

2. **Not requiring the explicit definition of `noise_types`**: Instead of relying on the user "activating" the noises they want, we can move to an interface where they just specify the parameters that should be taken into account and the corresponding noise type is internally "activated". For now this will work well because each noise parameter is associated to a unique noise type. However, I can see at some point a parameter like `temperature` being associated to more than one noise type (eg. Doppler noise and thermal motion of the atoms). In this case, having the desired noise type(s) explicitly defined would be needed for finer control, so I don't think we can completely move away from explicit noise type definition. Nonetheless, we can still make it an optional parameter that is generally not required.

3. **Define a custom `__repr__()`**: We can make it so only the active noises and associated parameters appear.

HGSilveri:
- Allows a `NoiseModel` to be defined solely by its relevant parameters, assuming everything else to be 0.0/undefined
- Deprecates the explicit definition of `noise_types` and reliance on default values
- Updates the `NoiseModel.__repr__()` to show only relevant parameters
- Modifies `QutipEmulator` to avoid Hamiltonian resampling with `"amplitude"` noise when `amp_sigma=0.`
- Updates the JSON schema to accommodate these changes

Fixes #697 .


pulser-core/pulser/noise_model.py
```diff
@@ -113,115 +158,203 @@ class NoiseModel:
...
+            if true_noise_types != set(noise_types):
+                raise ValueError(
+                    "The explicit definition of noise types (deprecated) is"
+                    " not compatible with the modification of unrelated noise "
+                    "parameters. Defining only the relevant noise parameters "
+                    "(without specifying the noise types) is recommended."
                 )
+            run_params_ = [
+                p for p in relevant_params if p in ("runs", "samples_per_run")
+            ]
```
a-corni, line 272: What about:
```suggestion
            run_params_ =relevant_params.intersection(("runs", "samples_per_run"))
```

pulser-core/pulser/noise_model.py
```diff
@@ -113,115 +158,203 @@ class NoiseModel:
...
+            for p_ in run_params_:
+                param_vals[p_] = param_vals[p_] or _LEGACY_DEFAULTS[p_]
```
a-corni, line 274: I must say I got lost and can't really understand the need of this for loop - why do we end up using LEGACY_DEFAULTS ? I thought we got rid of None...

pulser-core/pulser/noise_model.py
```diff
@@ -237,13 +370,47 @@ def _check_eff_noise(self) -> None:
...
+            self.noise_types,
+            self.state_prep_error,
+            self.amp_sigma,
+            self.laser_waist,
+        )
+        relevant_params.add("noise_types")
+        params_list = []
+        for f in fields(self):
+            if f.name in relevant_params:
+                params_list.append(f"{f.name}={getattr(self, f.name)!r}")
```
a-corni, line 411: Why not do:
```suggestion
       if param in relevant_params:
            params_list.append(f"{param}={getattr(self, param)!r}")
```

pulser-simulation/pulser_simulation/hamiltonian.py
```diff
@@ -307,10 +310,9 @@ def _update_noise(self) -> None:
             )
             self._bad_atoms = dict(zip(self._qid_index, dist))
         if "doppler" in self.config.noise_types:
+            temp = cast(float, self.config.temperature) * 1e-6
```
a-corni, line 313: That can be solved by allowing the temperature to be zero (np.random.normal allows 0 value -> returns an array with zeros)

pulser-simulation/pulser_simulation/hamiltonian.py
```diff
@@ -351,6 +353,12 @@ def _construct_hamiltonian(self, update: bool = True) -> None:
         Also builds qutip.Qobjs related to the Sequence if not built already,
         and refreshes potential noise parameters by drawing new at random.
 
+        Warning:
+            The refreshed noise parameters (when update=True) are only those
+            that change from shot to shot (ie doppler and state preparation).
+            Amplitude fluctuations change from pulse to pulse and are always
```
a-corni, line 359: I don't find the expression "pulse to pulse" very clear, what do you mean by this ? Aren't amplitude fluctuations updated whatever the value of update ?

a-corni (commented):
Thanks for tackling this ! I made few comments, the most interesting one might be about temperature - I think we can allow it to be 0, it fits with the current code and it gives a nice physical intuition of Doppler noise - it's perfect if T=0K.

pulser-core/pulser/noise_model.py
```diff
@@ -37,51 +39,94 @@
...
+    "relaxation": ("relaxation_rate",),
+    "depolarizing": ("depolarizing_rate",),
+    "eff_noise": ("eff_noise_rates", "eff_noise_opers"),
+}
+
+_PARAM_TO_NOISE_TYPE = {
+    param: noise_type
+    for noise_type, params in _NOISE_TYPE_PARAMS.items()
+    for param in params
+}
```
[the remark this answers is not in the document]
HGSilveri: It's true, I can add a check or a test

pulser-core/pulser/noise_model.py
```diff
@@ -37,51 +39,94 @@
...
+      ``depolarizing_rate``. While it does not describe a
+      physical phenomenon, it is a commonly used tool to test
+      the system under a uniform combination of phase flip (Z) and
+      bit flip (X) errors.
+    - **eff_noise**: General effective noise channel defined by the
+      set of collapse operators ``eff_noise_opers`` and the
+      corresponding rates distribution ``eff_noise_rates``.
+    - **doppler**: Local atom detuning due to termal motion of the
+      atoms and Doppler effect with respect to laser frequency.
+      Parametrized by the ``temperature`` field.
```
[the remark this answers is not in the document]
HGSilveri: Won't that force me to explain what the other parameters in the formula are too?

pulser-core/pulser/noise_model.py
HGSilveri: Cool, didn't remember this one

pulser-core/pulser/noise_model.py
HGSilveri: - We got rid of `None` in quantities that are not strictly positive (which is not the case for the run parameters)
- Here we rely on the legacy default for the run parameters in case `noise_types` was explicitly given **and** the run parameters are relevant. It is done here because we can only know if the run parameters are relevant after we know the values of `state_prep_error` and `amp_sigma` (which we potentially filled with the legacy defaults above)

I'll add a comment

pulser-simulation/pulser_simulation/hamiltonian.py
HGSilveri: I mean that from one pulse to another, the amplitude fluctuation is always updated. How would it be clearer for you?

pulser-core/pulser/noise_model.py
```diff
@@ -37,51 +39,94 @@
...
+_POSITIVE = {
+    "dephasing_rate",
+    "hyperfine_dephasing_rate",
+    "relaxation_rate",
+    "depolarizing_rate",
+}
+_STRICT_POSITIVE = {
+    "runs",
+    "samples_per_run",
+    "temperature",
```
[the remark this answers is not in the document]
HGSilveri: I agree, I was thinking that physically a temperature of 0K doesn't make sense

pulser-simulation/pulser_simulation/hamiltonian.py
HGSilveri: I agree with you but I'd say that if the temperature is 0, `"doppler"` should not even be included in the noise types.

pulser-core/pulser/noise_model.py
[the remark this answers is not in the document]
a-corni: Yeah, I thought about that and I guess it would be too cumbersome... I thought of pointing to an article, but can't find an article with the `$k_{eff}$` https://opg.optica.org/oe/fulltext.cfm?uri=oe-30-14-24261&id=477110 or in the bible https://steck.us/alkalidata/rubidium87numbers.1.6.pdf

pulser-core/pulser/noise_model.py
a-corni: Well I think that if you insert noise_types at 0, then everything is in order, but I agree it's surely not great to rely on that :)

HGSilveri: Not really because `relevant_params` is a set, so it just scrambles everything

pulser-simulation/pulser_simulation/hamiltonian.py
a-corni: Yep, I think so as well :) You could have a matching "doppler" not activated in the noise types <-> temperature is zero (in my opinion, it would be more speaking to see that temperature is zero when the user did not define "doppler" in noise types rather than seeing None)

HGSilveri: Yes, that's how it is now after the latest changes

pulser-simulation/pulser_simulation/hamiltonian.py
a-corni: Hum I am puzzled, is it really between two Pulses ? Isn't it for a whole SequenceSamples ?

HGSilveri: Yes, check it out here: https://github.com/pasqal-io/Pulser/blob/e16257ad5f9c6a88cbf3ffffb7f36aa6b2db129d/pulser-simulation/pulser_simulation/hamiltonian.py#L201-L233

`add_noise()` is applied on each pulse time slot and every time it is called, the random distribution is sampled. Contrast this against the doppler detuning which is stored in `self._doppler_detune` and is only updated when `_update_noise()` is called, ie between runs

HGSilveri: The idea is that the atoms are slow compared to the timescale of a shot, so you can assume the atoms stay in the same place on a single run. On the other hand, the amplitude fluctuations are fast, so they change within a run

HGSilveri: The idea is that the atoms are slow compared to the timescale of a shot, so you can assume the atoms move in the same direction with constant speed over a single run (and hence the detuning from the doppler effect stays constant over the run). On the other hand, the amplitude fluctuations are fast, so they change within a run

a-corni (commented):
Thanks for tackling the last comments :) Just have few questions to increase my understanding, and then it will be good for me !

pulser-core/pulser/noise_model.py
```diff
@@ -267,9 +267,10 @@ def to_tuple(obj: tuple) -> tuple:
                     "parameters. Defining only the relevant noise parameters "
                     "(without specifying the noise types) is recommended."
                 )
-            run_params_ = [
-                p for p in relevant_params if p in ("runs", "samples_per_run")
-            ]
+            # Only now that we know the relevant_params can we determine if
+            # we need to use the legacy defaults for the run parameters (ie in
+            # case they were not provided by the user)
+            run_params_ = relevant_params & {"runs", "samples_per_run"}
```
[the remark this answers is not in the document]
HGSilveri: We only use the legacy default values if the user specified `noise_types` explicitly (ie the default option). Otherwise, `runs` and `samples_per_run` are left as `None` when not specified and later raise an error upon validation

pulser-simulation/pulser_simulation/hamiltonian.py
a-corni: Ok, I hadn't see that, thanks for the explanation ! Now that I see it, the implementation is perhaps a bit weird, since in the case of a Sequence containing only one super-long pulse, the value of noise_base that will be applied will be the same all along - there will be no "fluctuation" actually.
I would suggest to define a rate at which fluctuations should happen (and perhaps they should fluctuate at each ns ?), and sample them at each run (as it's done at the moment).

pulser-core/pulser/noise_model.py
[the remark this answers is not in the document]
a-corni: Ah damn now I see it. Sorry !

pulser-simulation/pulser_simulation/hamiltonian.py
HGSilveri: I agree but it's out of scope, could you open an issue please?

a-corni (approved):
I approve the PR ! I think my comment on the implementation of noise from amplitude fluctuations should be seen as possible improvements and tackled in a separate PR.

PR #711, opened 2024-07-18
Modifications needed for pulser-diff

vytautas-a:
Support torch tensors in noise configuration.

### DIFF
--- a/pulser-core/pulser/noise_model.py
+++ b/pulser-core/pulser/noise_model.py
@@ -24,7 +24,6 @@ import numpy as np
 from numpy.typing import ArrayLike
 
 import pulser.json.abstract_repr as pulser_abstract_repr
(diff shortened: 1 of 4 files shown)


HGSilveri:
Hum, this will conflict heavily with the changes in #710 ... I would say it would be better to merge those first because we'll just have to redo these anyway. Wdyt?

vytautas-a:
> Hum, this will conflict heavily with the changes in #710 ... I would say it would be better to merge those first because we'll just have to redo these anyway. Wdyt?

Sounds reasonable. I just restored some changes I made previously. I leave the requirements for noise rates to be floats and if we see that in some situations, when no averaging is required in pulser-diff and these rates must be differentiable params, the `SimConfig` and `NoiseModel` can be updated accordingly.

PR #714, opened 2024-07-25
Add leakage noise in NoiseModel

a-corni:
Add the option to include leakage in NoiseModel.
It's currently done by modifying the parameter `with_leakage` in NoiseModel.
`with_leakage` is a boolean. It is associated with the noise type "leakage". 
For backward compatibility reason, with_leakage is an option in the schema of noise_model.
I have also added it with_leakage in the default values (if someone uses noise_types). However, I am not sure if the behaviour is correct: 
- by default, the parameter with_leakage is False in the __init__, and to have leakage the user should set `with_leakage` to True.
- however, if the user asks for leakage in noise_types, then I guess he doesn't want to also set with_leakage to True, so the default value of with_leakage in _LEGACY_DEFAULTS is True.
- In SimConfig, the user only defines leakage in the noises, and not via with_leakage, that is only a property instead of an argument.

### DIFF
--- a/pulser-core/pulser/json/abstract_repr/schemas/noise-schema.json
+++ b/pulser-core/pulser/json/abstract_repr/schemas/noise-schema.json
@@ -103,6 +103,9 @@
         },
         "temperature": {
           "type": "number"
+        },
+        "with_leakage": {
+          "type": "boolean"
         }
       },
       "required": [
@@ -125,13 +128,13 @@
     },
     "NoiseType": {
       "enum": [
+        "leakage",
         "doppler",
         "amplitude",
         "SPAM",
         "relaxation",
         "dephasing",
         "depolarizing",
-        "leakage",
         "eff_noise"
       ],
       "type": "string"
--- a/pulser-core/pulser/json/abstract_repr/schemas/noise-schema.json
+++ b/pulser-core/pulser/json/abstract_repr/schemas/noise-schema.json
@@ -103,9 +103,6 @@
         },
         "temperature": {
           "type": "number"
-        },
-        "with_leakage": {
-          "type": "boolean"
         }
       },
       "required": [
@@ -128,13 +125,13 @@
     },
     "NoiseType": {
       "enum": [
-        "leakage",
         "doppler",
         "amplitude",
         "SPAM",
         "relaxation",
         "dephasing",
         "depolarizing",
+        "leakage",
         "eff_noise"
       ],
       "type": "string"
--- a/pulser-core/pulser/noise_model.py
+++ b/pulser-core/pulser/noise_model.py
@@ -337,13 +337,6 @@ class NoiseModel:
         # Can't define "dephasing", "depolarizing" with "leakage"
         if "leakage" not in noise_types:
             return
-        # TODO: Implement the depolarizing and dephasing operations with
-        # projectors will stop raising an error.
-        if "dephasing" in noise_types or "depolarizing" in noise_types:
-            raise NotImplementedError(
-                "Dephasing and depolarizing channels can't be defined "
-                "with a leakage noise."
-            )
         if "eff_noise" not in noise_types:
             raise ValueError(
                 "At least one effective noise operator must be defined to"
--- a/pulser-core/pulser/noise_model.py
+++ b/pulser-core/pulser/noise_model.py
@@ -30,6 +30,7 @@ from pulser.json.abstract_repr.validation import validate_abstract_repr
 __all__ = ["NoiseModel"]
 
 NoiseTypes = Literal[
+    "leakage",
     "doppler",
     "amplitude",
     "SPAM",
@@ -40,6 +41,7 @@ NoiseTypes = Literal[
 ]
 
 _NOISE_TYPE_PARAMS: dict[NoiseTypes, tuple[str, ...]] = {
+    "leakage": ("with_leakage",),
     "doppler": ("temperature",),
     "amplitude": ("laser_waist", "amp_sigma"),
     "SPAM": ("p_false_pos", "p_false_neg", "state_prep_error"),
@@ -76,7 +78,10 @@ _PROBABILITY_LIKE = {
     "amp_sigma",
 }
 
+_BOOLEAN = {"with_leakage"}
+
 _LEGACY_DEFAULTS = {
+    "with_leakage": False,
     "runs": 15,
     "samples_per_run": 5,
     "state_prep_error": 0.005,
@@ -98,6 +103,11 @@ class NoiseModel:
 
     Supported noise types:
 
+    - "leakage": Adds an error state 'x' to the computational
+        basis, that can interact with the other states via an
+        effective noise channel. Must be defined with an effective
+        noise channel, but is incompatible with dephasing and
+        depolarizing noise channels.
     - **relaxation**: Noise due to a decay from the Rydberg to
       the ground state (parametrized by ``relaxation_rate``),
       commonly characterized experimentally by the T1 time.
@@ -156,6 +166,8 @@ class NoiseModel:
         eff_noise_rates: The rate associated to each effective noise operator
             (in 1/µs).
         eff_noise_opers: The operators for the effective noise model.
+        with_leakage: Whether or not to include an error state in the
+            computations (default to False).
     """
 
     noise_types: tuple[NoiseTypes, ...]
@@ -173,6 +185,7 @@ class NoiseModel:
     depolarizing_rate: float
     eff_noise_rates: tuple[float, ...]
     eff_noise_opers: tuple[ArrayLike, ...]
+    with_leakage: bool
 
     def __init__(
         self,
@@ -191,6 +204,7 @@ class NoiseModel:
         depolarizing_rate: float | None = None,
         eff_noise_rates: tuple[float, ...] = (),
         eff_noise_opers: tuple[ArrayLike, ...] = (),
+        with_leakage: bool = False,
     ) -> None:
         """Initializes a noise model."""
 
@@ -214,6 +228,7 @@ class NoiseModel:
             depolarizing_rate=depolarizing_rate,
             eff_noise_rates=to_tuple(eff_noise_rates),
             eff_noise_opers=to_tuple(eff_noise_opers),
+            with_leakage=with_leakage,
         )
 
         if noise_types is not None:
@@ -241,11 +256,14 @@ class NoiseModel:
             for p_ in param_vals
             if param_vals[p_] and p_ in _PARAM_TO_NOISE_TYPE
         }
-
+        self._check_leakage_noise(
+            true_noise_types if noise_types is None else noise_types
+        )
         self._check_eff_noise(
             cast(tuple, param_vals["eff_noise_rates"]),
             cast(tuple, param_vals["eff_noise_opers"]),
             "eff_noise" in (noise_types or true_noise_types),
+            with_leakage=with_leakage,
         )
 
         # Get rid of unnecessary None's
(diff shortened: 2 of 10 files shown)


HGSilveri:
Thanks for this @a-corni ! I was wondering, isn't there some sort of rate for leakage?

pulser-core/pulser/noise_model.py
```diff
@@ -366,9 +416,11 @@ def _check_eff_noise(
...
                 raise ValueError(f"Operator '{op!r}' is not a 2D array.")
 
-            if operator.shape != (2, 2):
-                raise NotImplementedError(
-                    f"Operator's shape must be (2,2) not {operator.shape}."
+            if operator.shape not in possible_shapes:
+                raise ValueError(
+                    f"With{'' if with_leakage else 'out'} leakage, operator's "
+                    f"shape must be {possible_shapes[0]} or "
+                    f"{possible_shapes[1]}, not {operator.shape}."
```
HGSilveri, line 423: So the changes for the leakage will also allow qutrit operators without leakage, right? I guess this will make a weird in-between state where our NoiseModel supports effective noise on qutrits but our `QutipEmulator` doesn't

pulser-core/pulser/noise_model.py
```diff
@@ -312,8 +334,29 @@ def _find_relevant_params(
         # Disregard laser_waist when not defined
         if laser_waist is None:
             relevant_params.discard("laser_waist")
+        # Diregard leakage if with_leakage is False
+        if not with_leakage:
+            relevant_params.discard("with_leakage")
```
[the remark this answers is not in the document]
a-corni: That's not needed indeed, if with_leakage is false then leakage will not be in true_noise_types at its creation :)

pulser-core/pulser/noise_model.py
```diff
@@ -353,6 +397,12 @@ def _check_eff_noise(
 
         if np.any(np.array(eff_noise_rates) < 0):
             raise ValueError("The provided rates must be greater than 0.")
+        print(with_leakage)
```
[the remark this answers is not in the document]
a-corni: Yeah sorry I definitely was not focused in my last commit ^^'

pulser-core/pulser/noise_model.py
a-corni: You are right, I am adding the leakage in effective noise operators in a separate branch from this PR and I think I should modify this test there.

a-corni:
> Thanks for this @a-corni ! ~I was wondering, isn't there some sort of rate for leakage?~ EDIT: Nevermind, I just remembered that it has to be given by the user through the effective noise parameters

Indeed ! I would like to first implement it for effective noise operator, and then have a look at implementing it with dephasing and depolarizing noise. For dephasing, we might want to have a error_dephasing_rate (like the hyperfine_dephasing_rate), but let's see this when it will be needed :) 

One thing I am not sure of is the way to go from this PR:
- should I merge it ? But then users can only define effective_noise_opers of shape (2, 2) and won't perform simulations with noise. Should I then add a NotImplementedError in Hamiltonian/QutipEmulator saying that simulation with leakage is not implemented yet ?
- should I change this PR to also incorporate the implementation of leakage in pulser-simulation ? Should I make a separate PR starting from these changes ?
I think the best is going with NotImplementedError in Hamiltonian for the moment... What is your opinion on this ?

HGSilveri:
> One thing I am not sure of is the way to go from this PR:
> 
>     * should I merge it ? But then users can only define effective_noise_opers of shape (2, 2) and won't perform simulations with noise. Should I then add a NotImplementedError in Hamiltonian/QutipEmulator saying that simulation with leakage is not implemented yet ?
> 
>     * should I change this PR to also incorporate the implementation of leakage in pulser-simulation ? Should I make a separate PR starting from these changes ?
>       I think the best is going with NotImplementedError in Hamiltonian for the moment... What is your opinion on this ?

Let's go with the `NotImplementedError`

pulser-core/pulser/noise_model.py
```diff
@@ -246,26 +244,26 @@ def to_tuple(obj: tuple) -> tuple:
                 )
             self._check_noise_types(noise_types)
             for nt_ in noise_types:
+                if nt_ == "leakage":
+                    raise ValueError(
+                        "'Leakage' cannot be explicitely defined in the noise"
```
HGSilveri, line 249: nit:
```suggestion
                        "'leakage' cannot be explicitely defined in the noise"
```

pulser-core/pulser/noise_model.py
```diff
@@ -298,7 +296,11 @@ def to_tuple(obj: tuple) -> tuple:
         relevant_param_vals = {
             p: param_vals[p]
             for p in param_vals
-            if param_vals[p] is not None or (p in relevant_params)
+            if not (
+                param_vals[p] is None
+                or (isinstance(param_vals[p], bool) and not param_vals[p])
+            )
+            or p in relevant_params
```
HGSilveri, line 303: I'm not sure, but maybe we can get away with
```suggestion
            if param_vals[p] or p in relevant_params
```
Did you try this?

pulser-core/pulser/noise_model.py
```diff
@@ -416,11 +409,9 @@ def _check_eff_noise(
...
                 raise ValueError(f"Operator '{op!r}' is not a 2D array.")
 
-            if operator.shape not in possible_shapes:
-                raise ValueError(
-                    f"With{'' if with_leakage else 'out'} leakage, operator's "
-                    f"shape must be {possible_shapes[0]} or "
-                    f"{possible_shapes[1]}, not {operator.shape}."
+            if operator.shape != (2, 2):
+                raise NotImplementedError(
+                    f"Operator's shape must be (2,2) not {operator.shape}."
```
HGSilveri, line 414: I think you can still allow (3, 3) if `with_leakage=True` (it will then hit the `NotImplementedError` on the hamiltonian but that's a different story), just not (4,4)

pulser-core/pulser/noise_model.py
```diff
@@ -299,6 +322,7 @@ def _find_relevant_params(
         state_prep_error: float,
         amp_sigma: float,
         laser_waist: float | None,
+        with_leakage: bool,
```
HGSilveri, line 325: You should remove this now

pulser-core/pulser/noise_model.py
a-corni: I tried this, you have a misbehaviour with runs=0 (the value of run is not validated, and the NoiseModel is generated with runs=0)

pulser-core/pulser/noise_model.py
a-corni: Yeah I agree,  I will do it this way :)

pulser-core/pulser/noise_model.py
HGSilveri: Ok, in that case I'd say this special case is not worth it, I'd suggest reverting to the less strict validation that a `bool` is a boolean, not necessarily `True`. In reality, we also allow other parameters to be zero (when not strictly positive), so it's more consistent behaviour

a-corni: I agree, I had the feeling of duplicating the test.

pulser-core/pulser/noise_model.py
a-corni: I have decided to go for something a bit more clever: raise an error depending on if the shape might be implemented some day or not (assuming that we might never support shapes outside (2, 2), (3, 3) without leakage, and (3, 3), (4, 4) with leakage.

HGSilveri (commented):
I only have a few nits, so it's pretty much good for me. One thing though: wouldn't it perhaps make more sense to merge #715 first?

tests/test_noise_model.py
```diff
@@ -176,8 +176,28 @@ def matrices(self):
...
+            assert noise_model.with_leakage == value
+            return
+        with pytest.raises(
+            ValueError, match=f"'with_leakage' must be a boolean, not {value}"
+        ):
+            noise_model = NoiseModel(
+                eff_noise_rates=[0.1],
+                eff_noise_opers=[matrices["I3"] if value else matrices["I"]],
+                with_leakage=value,
+            )
```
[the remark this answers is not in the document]
a-corni: I can do this. Initially I inspired myself from the aforedefined test_init_prob_like: https://github.com/pasqal-io/Pulser/blob/add_leakage/tests/test_noise_model.py#L145

tests/test_noise_model.py
```diff
@@ -235,7 +289,12 @@ def test_relevant_params(self):
             "p_false_pos",
             "p_false_neg",
         }
-        assert NoiseModel._find_relevant_params({"SPAM"}, 0.1, 0.5, 100) == {
+        assert NoiseModel._find_relevant_params(
```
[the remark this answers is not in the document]
a-corni: Oops sorry, thanks for the remark :)

a-corni:
> I only have a few nits, so it's pretty much good for me. One thing though: wouldn't it perhaps make more sense to merge #715 first?

Thanks for the review ! I have addressed the nits. As you want, we can definitely merge #715 before merging this PR 😄

HGSilveri:
> > I only have a few nits, so it's pretty much good for me. One thing though: wouldn't it perhaps make more sense to merge #715 first?
> 
> Thanks for the review ! I have addressed the nits. As you want, we can definitely merge #715 before merging this PR 😄

I thought that there were some `TODOs` here that could be done already, that's why I suggested it

HGSilveri (approved):
LGTM!

a-corni:
> > > I only have a few nits, so it's pretty much good for me. One thing though: wouldn't it perhaps make more sense to merge #715 first?
> > 
> > 
> > Thanks for the review ! I have addressed the nits. As you want, we can definitely merge #715 before merging this PR 😄
> 
> I thought that there were some `TODOs` here that could be done already, that's why I suggested it

That's True, I will let you merge your PR, I will delete the TODOs in this PR afterwards.

HGSilveri (approved):
Great!

pulser-core/pulser/json/abstract_repr/schemas/noise-schema.json
```diff
@@ -125,13 +128,13 @@
     },
     "NoiseType": {
       "enum": [
+        "leakage",
```
HGSilveri, line 131: This was already there

pulser-core/pulser/json/abstract_repr/schemas/noise-schema.json
```diff
@@ -103,6 +103,9 @@
         },
         "temperature": {
           "type": "number"
+        },
+        "with_leakage": {
+          "type": "boolean"
```
HGSilveri, line 108: I guess this is not needed, we can just check if `"leakage"` is in the `noise_types`. That way we won't need to make any modifications to the schema

HGSilveri (changes requested):
Sorry, I forgot to check the schema

a-corni (commented):
Thanks for this review of the schema. I have followed your suggestions. Should I keep 'with_leakage' property in the __repr__ as well ?

HGSilveri:
> Thanks for this review of the schema. I have followed your suggestions. Should I keep 'with_leakage' property in the **repr** as well ?

I thought about it and I'm torn actually. It's obviously redundant but if someone sets `with_leakage=True` I imagine they might expect to see it appear in the repr. I guess leaving it is the safer option, at least it won't cause any confusion

HGSilveri (approved):
Now I think we're good!

a-corni:
> > Thanks for this review of the schema. I have followed your suggestions. Should I keep 'with_leakage' property in the **repr** as well ?
> 
> I thought about it and I'm torn actually. It's obviously redundant but if someone sets `with_leakage=True` I imagine they might expect to see it appear in the repr. I guess leaving it is the safer option, at least it won't cause any confusion

That's also my feeling, let's go with this :)

PR #715, opened 2024-07-26
Defining dephasing and depolarizing operators with projectors

### LINKED ISSUE #691: Implement depolarizing and dephasing operations with projectors
Currently, depolarizing and dephasing operations are defined in pulser_simulation/hamiltonian.py/Hamiltonian._build_collapse_operators using `qutip.sigmax/y/z()`. It would be better to define them with projectors `sigma_{a}{b}` with a, b in [r, g, h].

HGSilveri:
Fixes #691 .

### DIFF
--- a/pulser-simulation/pulser_simulation/hamiltonian.py
+++ b/pulser-simulation/pulser_simulation/hamiltonian.py
@@ -114,6 +114,17 @@ class Hamiltonian:
                     f"Cannot include {noise_type} noise in all-basis."
                 )
 
+        # NOTE: These operators only make sense when basis != "all"
+        b, a = self.eigenbasis[:2]
+        pauli_2d = {
+            "x": self.op_matrix[f"sigma_{a}{b}"]
+            + self.op_matrix[f"sigma_{b}{a}"],
+            "y": 1j * self.op_matrix[f"sigma_{a}{b}"]
+            - 1j * self.op_matrix[f"sigma_{b}{a}"],
+            "z": self.op_matrix[f"sigma_{b}{b}"]
+            - self.op_matrix[f"sigma_{a}{a}"],
+        }
+
         local_collapse_ops = []
         if "dephasing" in config.noise_types:
             basis_check("dephasing")
@@ -122,7 +133,7 @@ class Hamiltonian:
                 if self.basis_name == "digital"
                 else config.dephasing_rate
             )
-            local_collapse_ops.append(np.sqrt(rate / 2) * qutip.sigmaz())
+            local_collapse_ops.append(np.sqrt(rate / 2) * pauli_2d["z"])
 
         if "relaxation" in config.noise_types:
             coeff = np.sqrt(config.relaxation_rate)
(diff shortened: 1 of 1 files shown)


pulser-simulation/pulser_simulation/hamiltonian.py
```diff
@@ -122,7 +133,7 @@ def basis_check(noise_type: str) -> None:
                 if self.basis_name == "digital"
                 else config.dephasing_rate
             )
-            local_collapse_ops.append(np.sqrt(rate / 2) * qutip.sigmaz())
+            local_collapse_ops.append(np.sqrt(rate / 2) * pauli_2d["z"])
```
a-corni, line 136: I agree with the current implementation. For scalability, I think we once discussed it was possible to define the collapse operator for dephasing channel as: 
```suggestion
            state, rate = "h", config.hyperfine_dephasing_rate if self.basis_name =="digital" else "r", config.dephasing_rate
            local_collapse_ops.append(np.sqrt(rate) * self.op_matrix[f"sigma_{a}{a}"])
```
But that will imply changing the tests...

a-corni: We can also keep this idea for a future PR where we implement noise in 'all' channel (which I guess we should implement before tackling the implementation of "leakage").

HGSilveri: > I agree with the current implementation. For scalability, I think we once discussed it was possible to define the collapse operator for dephasing channel as:
> 
> But that will imply changing the tests...

I'm pretty sure this is true for Kraus operators but I don't rememeber checking it also applies for collapse operators.

> We can also keep this idea for a future PR where we implement noise in 'all' channel (which I guess we should implement before tackling the implementation of "leakage")

Yeah, we still have to figure out how we'll do that...

a-corni: Well if it works for kraus it can also be extended to collapse operators :) 
I think that was the core of what I had in mind when I created this issue (I had in mind to first implement effective noise operators with 'all' and then move on and dephasing and depolarizing with "all"). I propose to tackle the implementation of effective noise operators in the next PR.

a-corni: Perhaps we should resume this PR after having implemented the simulation with effective noise operators in 'all' basis (and having in mind the simulations with leakage loise), since at the moment the feature is nice but it does not add much value (as you commented in the issue). What do you think ?

HGSilveri: > Perhaps we should resume this PR after having implemented the simulation with effective noise operators in 'all' basis (and having in mind the simulations with leakage loise), since at the moment the feature is nice but it does not add much value (as you commented in the issue). What do you think ?

Well, this is already done so I guess we might as well merge it and then make the necessary modifications once "all" is supported

a-corni: I agree, I had not seen your comments in the other PR :)

a-corni (approved):
LGTM

PR #716, opened 2024-08-02
Enable definition of effective noise operators in all basis

a-corni:
- Enables the definition of effective noise operators in a Simulation of a Sequence with Rydberg and Raman channels.
- Redefines the collapse operators associated with dephasing noise: the collapse operators become `$ \sqrt{2\gamma} \left |r \right>\left <r \right |$` and `$ \sqrt{2\gamma_h} \left |h \right>\left <h \right |$`.
- Test the reduction to basis of a density matrix obtained after simulation using effective noise operators in all basis (not implemented error)


tests/test_simulation.py
```diff
@@ -865,6 +855,73 @@ def test_noises_digital(matrices, noise, result, n_collapse_ops, seq_digital):
         seq_digital.register.qubits
     )
     trace_2 = res.states[-1] ** 2
+    assert np.abs(np.trace(trace_2)) < 1 and not np.isclose(
```
HGSilveri, line 858: Was the `np.abs()` needed because there was an imaginary part?

HGSilveri (commented):
Looking good! Are you still updating the tutorial on this one?

HGSilveri:
I fixed the issue in the test on #717 , it is unrelated

HGSilveri (approved):
LGTM, great job!

tests/test_simulation.py
a-corni: It was not really needed, but just at some point I had some issues with this test and I was wondering if it was coming from there...

HGSilveri: Ok, let's revert it then!

PR #717, opened 2024-08-05
Hide Rabi frequency when drawing DMM channels

### LINKED ISSUE #651: Hide Rabi frequency when drawing DMM channels in Sequence
I think that the blank Rabi frequency section is just taking up space and could be omitted, much like we already do for other channels with no detuning.

HGSilveri:
Closes #651 .


a-corni (approved):
Looks good to me :)

PR #718, opened 2024-08-06
Allow specification of job IDs in RemoteResults

HGSilveri:
- Allows specification of a subset of job IDs in `RemoteResults`
- Exposes `RemoteResults.batch_id` and `RemoteResults.job_ids`
- Supports specification of `job_ids` in `PasqalCloud._fetch_result()`

These changes should be useful to PRs #701 and #707 .


pulser-core/pulser/backend/remote.py
```diff
@@ -58,18 +58,47 @@ class RemoteResults(Results):
...
+    ):
         """Instantiates a new collection of remote results."""
         self._submission_id = submission_id
         self._connection = connection
+        if job_ids is not None and not set(job_ids).issubset(
+            all_job_ids := self._connection._get_job_ids(self._submission_id)
+        ):
+            unknown_ids = [id_ for id_ in job_ids if id_ not in all_job_ids]
+            raise RuntimeError(
+                f"Submission {self._submission_id!r} does not contain jobs "
```
oliver-gordon, line 80: `!r` is there for inline quotes right? It's _not_ misplaced correct?

pulser-core/pulser/backend/remote.py
```diff
@@ -102,7 +133,9 @@ def submit(
         pass
 
     @abstractmethod
-    def _fetch_result(self, submission_id: str) -> typing.Sequence[Result]:
+    def _fetch_result(
+        self, submission_id: str, job_ids: list[str] | None
+    ) -> typing.Sequence[Result]:
```
oliver-gordon, line 138: Not a related change but, how have you defined which types are imported via `typing.Type` or imported at the top? It's just the abstract classes?

pulser-pasqal/pulser_pasqal/pasqal_cloud.py
```diff
@@ -189,7 +191,16 @@ def _fetch_result(self, submission_id: str) -> tuple[Result, ...]:
         meas_basis = seq_builder.get_measurement_basis()
 
         results = []
-        for job in batch.ordered_jobs:
+        sdk_jobs = batch.ordered_jobs
+        if job_ids is not None:
+            ind_job_pairs = [
```
oliver-gordon, line 196: Just to confirm, this is so we can retain the returned order that they came in? if we associate and order by the index?

oliver-gordon (approved):
Some basic Q's but looks fine with me.

pulser-core/pulser/backend/remote.py
HGSilveri: Yes, that's right

pulser-core/pulser/backend/remote.py
HGSilveri: It's basically everything except `typing.Sequence` so as to not confuse it with `pulser.Sequence` (I know, it's unfortunate naming but there's no going back now)

pulser-pasqal/pulser_pasqal/pasqal_cloud.py
HGSilveri: Yes, whenever `job_ids` are provided we want to return the results in the same order they were given. This is simultaneously ensuring that's the case and selecting only the relevant subset of jobs

PR #719, opened 2024-08-07
Fix failing pipeline for test of relaxation noise in qutrit state

a-corni:
Fixes merge of #716

### DIFF
--- a/tests/test_simulation.py
+++ b/tests/test_simulation.py
@@ -858,6 +858,16 @@ def test_noises_digital(matrices, noise, result, n_collapse_ops, seq_digital):
     assert np.trace(trace_2) < 1 and not np.isclose(np.trace(trace_2), 1)
 
 
+res_deph_relax = {
+    "000": 412,
+    "010": 230,
+    "001": 176,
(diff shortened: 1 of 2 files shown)


HGSilveri:
You should rebase on `develop` to include just the last commit. Also, let's temporarily enable the CI tests on `mac-os` to make sure it is indeed fixed

tests/test_simulation.py
```diff
@@ -887,7 +887,7 @@ def test_noises_all(matrices, noise, result, n_collapse_ops, seq):
             noise=noise,
             dephasing_rate=0.1,
             hyperfine_dephasing_rate=0.1,
-            relaxation_rate=1000,
+            relaxation_rate=2000,
```
HGSilveri, line 890: It's slightly ridiculous that we have to go this high in relaxation rate to see an effect. I would say it is preferable to change the sequence to something where relaxation has an effect, even if just for this one case - otherwise we're not really testing anything

HGSilveri (approved):
LGTM!

PR #724, opened 2024-08-22
Soften switching device with strict conditions

### LINKED ISSUE #721: Avoid failing to switch device prematurely
Currently, `Sequence.switch_device(strict=True)` is unnecessarily rigid in some cases:
- It requires the `RydbergEOM` configs match exactly between channels, which will become even more problematic now that the lightshift coefficients can be specified
- It enforces matching between channel parameters that might not be relevant for the particular sequence being switched, eg.
    - `bottom_detuning` and `total_bottom_detuning` are enforced during reconstruction so there is no need to see if they match
    - The parameters that are required to match in `eom_config` depend on whether the amplitude of the pulses is high enough to make the limiting beam reach the maximum amplitude
    
It would be preferable to not fail unnecessarily and instead check that channel samples all match between the old and the new sequence.

a-corni:
Conditions on total_bottom_detuning and bottom_detuning for DMM channels, and on EOM config for channels using EOM (eom config being the same) are deleted.
They are replaced by a condition on the building of the sequence being possible, and the samples of the Sequence being unaffected by the switching of the device.
To do this, we first loop over all the possible matchings between declared channels and devices in the new device. For the one having correct type, addressing, modulations, ... we try to build the associated sequence. If the strict condition is provided and the Sequence can be successfully built, we check that the samples are the same for the two Sequences. Finally, the Sequence built with the first compatible matching.
If no matching is found to be compatible, we go through the same error detection process as before to provide a good hint on the error.
Closes #721


pulser-core/pulser/sequence/sequence.py
```diff
@@ -785,10 +786,223 @@ def check_retarget(ch_obj: Channel) -> bool:
...
+                    elif ch_match_err_suffix != "":
+                        ch_match_err = (
+                            ch_match_err
+                            if (
+                                "type, basis and addressing"
+                                in ch_match_err_suffix
+                                and ch_match_err
+                            )
+                            else base_msg + ch_match_err_suffix
+                        )
```
HGSilveri, line 912: Why do you keep the existing message only in this case. If I remember correctly, before we had somehting like
```suggestion
                        ch_match_err = (
                            ch_match_err
                            or (base_msg + ch_match_err_suffix)
                        )
```
What's wrong with this?

pulser-core/pulser/sequence/sequence.py
```diff
@@ -785,10 +786,223 @@ def check_retarget(ch_obj: Channel) -> bool:
...
+                )
+
+            if strict:
+                for eom_channel in active_eom_channels:
+                    current_samples = self._schedule[eom_channel].get_samples()
+                    new_samples = new_seq._schedule[eom_channel].get_samples()
+                    if (
+                        np.any(current_samples.amp != new_samples.amp)
+                        or np.any(current_samples.det != new_samples.det)
+                        or np.any(current_samples.phase != new_samples.phase)
```
HGSilveri, line 996: How about if we use `isclose()` instead?

HGSilveri (commented):
Looking good overall! The `switch_device()` method is getting too bulky, how would you feel about moving it into a separate file?

pulser-core/pulser/sequence/sequence.py
a-corni: My bad, I think what you are suggesting is better than what I did at the moment, I forgot how it was before when writing this 😅

a-corni:
> Looking good overall! The `switch_device()` method is getting too bulky, how would you feel about moving it into a separate file?

I have created a "helpers" folder, and placed this function `switch_device` in it. I have also moved _seq_str.py to this file since it is also a helper function. The other files does not seem to contain only a helper function, so I have decided to let it like that.

HGSilveri:
> I have created a "helpers" folder, and placed this function `switch_device` in it. I have also moved _seq_str.py to this file since it is also a helper function. The other files does not seem to contain only a helper function, so I have decided to let it like that.

This is okay with me. I think other files could potentially go there but it's not out-of-scope for this PR.

pulser-core/pulser/sequence/sequence.py
```diff
@@ -747,180 +748,7 @@ def switch_device(
...
-                    # EOM configuration
-                    if new_ch_obj.eom_config is None:
-                        ch_match_err = base_msg + " with an EOM configuration."
-                        continue
-                    if (
-                        # TODO: Improvements to this check:
-                        # 1. multiple_beam_control doesn't matter when there
-                        # is only one beam
-                        # 2. custom_buffer_time doesn't have to match as long
-                        # as `Channel_eom_buffer_time`` does
```
HGSilveri, line 839: This is something we could check for even when the sequence is parametrized

pulser-core/pulser/sequence/helpers/_switch_device.py
```diff
@@ -0,0 +1,351 @@
...
+            return build_sequence_from_matching(
+                new_device, channel_match, active_eom_channels, strict
+            )
+        except ValueError as e:
+            err_channel_match[tuple(channel_match.items())] = e.args
+            continue
+    raise ValueError(
+        "No matching found between declared channels and channels in the "
+        "new device that does not modify the samples of the Sequence. "
+        "Here is a list of matching tested and their associated errors: "
```
HGSilveri, line 349: nit: 
```suggestion
        "Here is a list of matchings tested and their associated errors: "
```

pulser-core/pulser/sequence/helpers/_switch_device.py
```diff
@@ -16,6 +16,7 @@
 
 import itertools
 import warnings
+from dataclasses import asdict
```
HGSilveri, line 19: nit: `asdict` is a bit too vague, I would rather have an `import dataclasses` and then a call to `dataclasses.asdict()`

pulser-core/pulser/sequence/helpers/_switch_device.py
```diff
@@ -117,20 +119,41 @@ def check_channels_match(
...
+                    # However, multiple_beam_control only matters when
+                    # the two beams are controlled
+                    if len(old_ch_obj.eom_config.controlled_beams) == 1:
+                        new_eom_config.pop("multiple_beam_control")
+                        old_eom_config.pop("multiple_beam_control")
+                        # Controlled beams only matter when only one beam
+                        # is controlled by the new eom
+                        if len(new_ch_obj.eom_config.controlled_beams) > 1:
+                            new_eom_config.pop("controlled_beams")
+                            old_eom_config.pop("controlled_beams")
```
HGSilveri: This is unreachable, I don't think this is what you want to do

HGSilveri: Sorry, I missed that up top it is `old_ch_obj` and then it's `new_ch_obj`, this is fine

pulser-core/pulser/sequence/helpers/_switch_device.py
```diff
@@ -117,20 +119,41 @@ def check_channels_match(
...
+                            "",
+                            " with the same mod_bandwidth for the EOM.",
+                        )
+                else:
+                    # Eom configs have to match is Sequence is parametrized
+                    new_eom_config = asdict(new_ch_obj.eom_config)
+                    old_eom_config = asdict(old_ch_obj.eom_config)
+                    # However, multiple_beam_control only matters when
+                    # the two beams are controlled
+                    if len(old_ch_obj.eom_config.controlled_beams) == 1:
```
HGSilveri: Regarding `controlled_beams`, they should be compared as `set`s to avoid failing because of the order

tests/test_sequence.py
```diff
@@ -1074,16 +1074,33 @@ def test_switch_device_up(
...
+        eom = dataclasses.replace(
+            AnalogDevice.channels["rydberg_global"].eom_config,
+            controlled_beams=tuple(RydbergBeam),
+            multiple_beam_control=True,
+            custom_buffer_time=None,
+        )
+        channel = dataclasses.replace(
+            AnalogDevice.channels["rydberg_global"], eom_config=eom
+        )
+        device = dataclasses.replace(AnalogDevice, channel_objects=(channel,))
```
HGSilveri, line 1098: nit: you could do this outside the function and just give the `device` as an argument, no?

HGSilveri (approved):
Only one nit, nice job!

tests/test_sequence.py
a-corni: Sure :)

a-corni:
Ready @HGSilveri 😄

PR #727, opened 2024-08-29
Add from_abstract_repr to Device and VirtualDevice

### LINKED ISSUE #725: Have a method `from_abstract_repr` in Device to deserialize an abstract Device
At the moment, users wanting to deserialize the abstract representation of a Device have to do 
`from pulser.json.abstract_repr.deserializer import deserialize_device` followed by `deserialize_device(abstract_device)`. It would be simpler to have `Device.from_abstract_repr(abstract_device)`.

Initially suggested by @nataliejpg, useful for users of pulser_myqlm to build the device associated with the current state of the QPU (provided as a String, in the abstract format).

a-corni:
from_abstract_repr uses deserialize_device and checks the type of the returned device for it to match the class that generated it:
- Device class only allows a serialized Device object.
- VirtualDevice class allows both serialized Device and VirtualDevice -> If it's a serialized Device, it is converted into a VirtualDevice using to_virtual().
Should I keep this behaviour for VirtualDevice ? 
Closes #725

### DIFF
--- a/pulser-core/pulser/devices/_device_datacls.py
+++ b/pulser-core/pulser/devices/_device_datacls.py
@@ -23,6 +23,7 @@ from typing import Any, Literal, cast, get_args
 import numpy as np
 from scipy.spatial.distance import pdist, squareform
 
+import pulser.json.abstract_repr as pulser_abstract_repr
 from pulser.channels.base_channel import Channel, States, get_states_from_bases
 from pulser.channels.dmm import DMM
 from pulser.devices.interaction_coefficients import c6_dict
@@ -715,6 +716,29 @@ class Device(BaseDevice):
         d["is_virtual"] = False
         return d
 
+    @staticmethod
+    def from_abstract_repr(obj_str: str) -> Device:
+        """Deserialize a noise model from an abstract JSON object.
+
+        Args:
+            obj_str (str): the JSON string representing the noise model
+                encoded in the abstract JSON format.
+        """
+        if not isinstance(obj_str, str):
+            raise TypeError(
+                "The serialized Device must be given as a string. "
+                f"Instead, got object of type {type(obj_str)}."
+            )
+
+        # Avoids circular imports
+        device = pulser_abstract_repr.deserializer.deserialize_device(obj_str)
+        if not isinstance(device, Device):
+            raise TypeError(
+                "The given schema is not related to a Device, but to a"
+                f" {type(device).__name__}."
+            )
+        return device
+
 
 @dataclass(frozen=True)
 class VirtualDevice(BaseDevice):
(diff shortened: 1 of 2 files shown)


a-corni:
@nataliejpg does this close your issue ?

pulser-core/pulser/devices/_device_datacls.py
```diff
@@ -796,3 +820,23 @@ def _to_abstract_repr(self) -> dict[str, Any]:
...
+        if not isinstance(obj_str, str):
+            raise TypeError(
+                "The serialized VirtualDevice must be given as a string. "
+                f"Instead, got object of type {type(obj_str)}."
+            )
+
+        # Avoids circular imports
+        device = pulser_abstract_repr.deserializer.deserialize_device(obj_str)
+        if isinstance(device, Device):
+            return device.to_virtual()
```
HGSilveri, line 841: I think this is not a bad idea. Let's just add a `Warning:` section in the docstring to document this behaviour

tests/test_abstract_repr.py
```diff
@@ -228,27 +229,75 @@ def _roundtrip(abstract_device):
...
+                    VirtualDevice.from_abstract_repr,
+                ]
+            for func in func_to_test:
+                with pytest.raises(DeserializeDeviceError) as exc_info:
+                    func(obj_str)
+                err_raised.append(exc_info)
+            for exc_info in err_raised:
+                cause = exc_info.value.__cause__
+                assert isinstance(cause, original_err)
+                assert re.search(re.escape(err_msg), str(cause)) is not None
```
HGSilveri, line 251: This is looking unnecessarily convoluted. Why not just make the function to test an argument to the function (potentially with the default being deserialize_device) and then do the test for the from_abstract_repr() outside of the function?

a-corni:
@MatthieuMoreau0 that's of interest for you as well :)

pulser-core/pulser/devices/_device_datacls.py
```diff
@@ -715,6 +716,33 @@ def _to_abstract_repr(self) -> dict[str, Any]:
         d["is_virtual"] = False
         return d
 
+    @staticmethod
+    def from_abstract_repr(obj_str: str) -> Device:
```
MatthieuMoreau0, line 720: On the cloud side can we simply use the `VirtualDevice.from_abstract_repr` method to build our device object? I am not super up to date on the differences between Devices and VirtualDevices. 

We build the device object to 1) validate the layout 2) switch the device of the sequence for validation.

In our current code I see that we only validate the layout if the device is not virtual, is it correct virtual devices do not have pre calibrated layouts?

HGSilveri: It seems that what you are describing is QPU validation, so what you want is `Device.from_abstract_repr()`.

A `VirtualDevice` is not tied to any physical implementation, which is why it doesn't have certain parameters like `pre_calibrated_layouts`. I'd say you probably won't have a need for `VirtualDevice.from_abstract_repr()` unless you are deserializing a device for execution on an emulator at some point.

HGSilveri (approved):
LGTM!

PR #734, opened 2024-09-23
Remove legacy NoiseModel interface

HGSilveri:
Removes the legacy `NoiseModel` interface, where `noise_types` are explicitly given rather than inferred.


pulser-core/pulser/noise_model.py
```diff
@@ -102,7 +102,7 @@ class NoiseModel:
 
     Supported noise types:
 
-    - "leakage": Adds an error state 'x' to the computational
+    - **leakage**: Adds an error state 'x' to the computational
```
a-corni, line 105: Daamn, Nice catch !

a-corni (approved):

PR #736, opened 2024-09-23
Remove Simulation() and pulser.simulation

HGSilveri:
- Removes the deprecated `Simulation()` class
- Removes the legacy `pulser.simulation` module


pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -698,7 +698,7 @@ def phase_str(phi: Any) -> str:
                         alpha=0.3,
                         hatch="////",
                     )
-                else:
+                else:  # pragma: no cover
```
a-corni, line 701: Why is this needed ?

a-corni (commented):
Just one nit :)

pulser-core/pulser/sequence/_seq_drawer.py
HGSilveri: Basically this case is only touched when using `draw_sequence()` with modulation and a sampling rate. While `Simulation()` was doing this, `QutipEmulator()` uses `draw_samples()`. I could delete it but it might come in handy again in the future, so I just kept it

a-corni: Sure, let's keep it then ;)

a-corni (approved):

PR #744, opened 2024-10-03
Correct XY Hamiltonian interaction term

HGSilveri:
There was a wrong factor of `/2` in the XY interaction term: what should have been `(1 - 3*cos**2) ` was rather `(1 - 3*cos**2) / 2`.
This PR:
- Removes that factor
- Calculates the cosine with the quantization axis more generically
- Updates the tutorial using the XY mode to accomodate for the higher interaction strength


a-corni (approved):
LGTM :)

PR #747, opened 2024-10-09
Fix rounding error in RampWaveform samples

HGSilveri:
Rounding errors in the generation of a `RampWaveform`'s samples could lead the edge values to lie outside of the range of the `start`and `stop` values. This ensures they stay within said range.


a-corni (approved):
Definitely better than the "bricolage" I did ;)

PR #752, opened 2024-10-15
Improve the NoiseModel unused parameters warning message

### LINKED ISSUE #742: UserWarning: 'runs' is not used by any active noise type ('SPAM',) when measurement errors without state preparation errors
Dear team,

It seems Pulser requires `runs` and `samples_per_run` to NOT be specified in `NoiseModel` when
`p_false_pos > 0` and `state_prep_error == 0`.

To reproduce, execute the following script
```
import pulser

noise_model = pulser.noise_model.NoiseModel(
            runs=1,
            samples_per_run=1,
            p_false_pos=0.1,
        )
```

Output: 
```
UserWarning: 'runs' is not used by any active noise type ('SPAM',).
  noise_model = pulser.noise_model.NoiseModel(
UserWarning: 'samples_per_run' is not used by any active noise type ('SPAM',).
  noise_model = pulser.noise_model.NoiseModel(
```

However, as soon as `state_prep_error` is `>0` then the warning disappears.
This is unexpected, since then the active noise types are then exactly the same, so the warning has an issue, I believe.
I would expect no warning in either case, or dissociate state preparation (which would require `runs`) from measurement errors (which maybe wouldn't). Let me know if that makes sense.

Using `pulser-core` 0.20.1

HGSilveri:
The warning message indicating there are unused parameters in the `NoiseModel` now includes the non-zero noise parameters, e.g. for
```
pulser.NoiseModel(p_false_pos=0.1, runs=10)
```

Before:
```
UserWarning: 'runs' is not used by any active noise type ('SPAM',).
```

After:
```
UserWarning: 'runs' is not used by any active noise type in ('SPAM',) when the only defined parameters are ['p_false_pos'].
```

Fixes #742 .

### DIFF
--- a/pulser-core/pulser/noise_model.py
+++ b/pulser-core/pulser/noise_model.py
@@ -258,12 +258,16 @@ class NoiseModel:
         object.__setattr__(
             self, "noise_types", tuple(sorted(true_noise_types))
         )
+        non_zero_relevant_params = [
+            p for p in relevant_params if param_vals[p]
+        ]
         for param_, val_ in param_vals.items():
             object.__setattr__(self, param_, val_)
             if val_ and param_ not in relevant_params:
                 warnings.warn(
                     f"{param_!r} is not used by any active noise type "
-                    f"{self.noise_types}.",
+                    f"in {self.noise_types} when the only defined parameters "
+                    f"are {non_zero_relevant_params}.",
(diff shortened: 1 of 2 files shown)


a-corni (approved):
I approve, I also think it brings value to add this :)

PR #753, opened 2024-10-16
Incorporate automatic layout generation in `Register`

### LINKED ISSUE #748: Add a way to generate a layout from a register

HGSilveri:
- Adds an algorithm for defining the layout trap coordinates from a set of atomic coordinates
- Defines `Register.with_automatic_layout(device: Device)`, which allows the user to get an automatically generated layout that obeys the `device` constrains from any register

Closes #748 .

### DIFF
--- a/pulser-core/pulser/register/_layout_gen.py
+++ b/pulser-core/pulser/register/_layout_gen.py
@@ -16,22 +16,23 @@ from __future__ import annotations
 import numpy as np
 from scipy.spatial.distance import cdist
 
-from pulser.register import Register, RegisterLayout
+from pulser.register import RegisterLayout
 
 
 def generate_register_layout(
-    register: Register,
+    atom_coords: np.ndarray,
     min_trap_dist: float,
     max_radial_dist: int,
     max_layout_filling: float,
     optimal_layout_filling: float | None = None,
     mesh_resolution: float = 1.0,
+    min_traps: int = 1,
     slug: str | None = None,
 ) -> RegisterLayout:
-    """Generates a layout for a register.
+    """Generates a layout from a collection of coordinates.
 
     Args:
-        register: The register to generate the layout for.
+        atom_coords: The coordinates where atoms will be placed.
         device: The Pulser device on which the layout and register
             will be implemented.
         min_trap_dist: The minimum distance between traps.
@@ -41,6 +42,7 @@ def generate_register_layout(
             atoms to traps. If not given, takes max_layout_filling.
         mesh_resolution: The spacing between points in the mesh of candidate
             coordinates, in µm.
+        min_traps: The minimum number of traps in the resulting layout.
         slug: An optional slug for the resulting layout.
 
     Returns:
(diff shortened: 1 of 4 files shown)


pulser-core/pulser/register/_layout_gen.py
```diff
@@ -0,0 +1,96 @@
...
+        np.ceil(N_seeds / max_layout_filling).astype(int), min_traps
+    )
+
+    # Use max() in case min_traps is larger than the optimal number
+    target_traps = max(
+        np.ceil(N_seeds / optimal_layout_filling).astype(int),
+        min_traps,
+    )
+    if max_traps:
+        target_traps = min(target_traps, max_traps)
```
a-corni, line 74: Isn't it also possible that min_traps be bigger than max_traps ? Shouldn't we also have 
min_traps = min(min_traps, max_traps)

pulser-core/pulser/register/_layout_gen.py
```diff
@@ -0,0 +1,96 @@
...
+    all_dists = cdist(coords, seeds)
+
+    # Accounts for the case when the needed number is less than min_traps
+    min_traps = max(
+        np.ceil(N_seeds / max_layout_filling).astype(int), min_traps
+    )
+
+    # Use max() in case min_traps is larger than the optimal number
+    target_traps = max(
+        np.ceil(N_seeds / optimal_layout_filling).astype(int),
```
a-corni, line 70: Why is it ceil and not `round` for the optimal_layout_filling ?

pulser-core/pulser/register/_layout_gen.py
```diff
@@ -0,0 +1,96 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from __future__ import annotations
+
+import numpy as np
+from scipy.spatial.distance import cdist
+
+
+def generate_layout_trap_coordinates(
```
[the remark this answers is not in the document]
HGSilveri: Sounds good

pulser-core/pulser/register/_layout_gen.py
HGSilveri: I didn't think much about it but I do want `target_traps >= min_traps`, which won't always happen if I use `round()`.

pulser-core/pulser/register/_layout_gen.py
HGSilveri: When they come from `Device` it's not possible; `min_traps > max_traps` has no solution, after all

pulser-core/pulser/register/_layout_gen.py
```diff
@@ -0,0 +1,96 @@
...
+    min_trap_dist: float,
+    max_radial_dist: int,
+    max_layout_filling: float,
+    optimal_layout_filling: float | None = None,
+    mesh_resolution: float = 1.0,
+    min_traps: int = 1,
+    max_traps: int | None = None,
+) -> list[np.ndarray]:
+    """Generates layout traps for a collection of atomic coordinates.
+
```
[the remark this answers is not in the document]
HGSilveri: I'll take it!

pulser-core/pulser/register/_layout_gen.py
HGSilveri: However, it's true that with the `max()` that won't happen, so I can do it :)

pulser-core/pulser/register/_layout_gen.py
a-corni: Ah I think there was a little misunderstanding. I meant that if N_seeds is extremely big, you might have the new min_traps higher than max_traps after the np.ceil, so I guess it would be good to have 
```suggestion
        target_traps = min(target_traps, max_traps)
        min_traps = min(min_traps, max_traps)
```

HGSilveri: I see what you mean but if you do this, the program will successfully terminate when it reach `max_traps`, which will be lower than what you need (`min_traps` before your new line). We don't want this to happen because it won't respect the `max_layout_filling`.
What I'll do instead is always check at the end that `len(traps) >= min_traps` and raise the `RuntimeError` otherwise.

a-corni (approved):

Issue #756, opened 2024-10-21
Show all the specs of the `Device`


a-corni:
Current `_specs` of the Device is only implemented for part of the attributes of the Device. For instance it does not display `default_noise_model`, `requires_layout`, `accepts_new_layout`, ...

HaroldErbin:
I had a few questions related to this point:
1. Why not store also the properties in a `dict`, in case we want to pass all properties (for example, to define a new device, save as json for easy check of the specs, etc.)? I find having a string not always convenient (also it's supposed to be an internal method).
2. Why not implement `_specs` also for virtual devices? Following [this tutorial](https://pulser.readthedocs.io/en/stable/tutorials/virtual_devices.html), we could expect to be able to print the specs also for the derived virtual device.
3. Why implement `_specs` as an internal method and not as a public property?

HaroldErbin:
Note that I just found the `_params()` internal method, which does work for both physical and virtual devices. However, it does not contain all parameters (for example, `interaction_coefficient` is missing, while `interaction_coefficient_xy` is present). It could also better be a public property.

PR #763, opened 2024-11-19
Update displayed device specs

HaroldErbin:
Add more information to `_specs()` and move the method to `BaseDevice`.

A public property `specs` has been added.

See issue #756.


HaroldErbin:
I don't understand how to make the `typing` test pass. If I use a `try: ... except AttributeError` to see if `ch` has an attribute `bottom_detuning`, then I get an error saying that `ch` has no attribute `bottom_detuning` (which is what I am testing`. If I am using a different test with `hasattr`, then I get an error saying that there is a redundant cast to `float`. Any pointer to solve this would be appreciated.

pulser-core/pulser/devices/_device_datacls.py
```diff
@@ -579,6 +579,140 @@ def to_abstract_repr(self) -> str:
...
+                    max_abs_detuning = (
+                        f"{float(cast(float, ch.max_abs_detuning)):.4g} rad/µs"
+                    )
+
+                bottom_detuning = "None"
+                if hasattr(ch, "bottom_detuning"):
+                    if ch.bottom_detuning is not None:
+                        bottom_detuning = (
+                            f"{float(cast(float, ch.bottom_detuning)):.4g}"
+                            " rad/µs"
```
HGSilveri, line 677: Did you try this?
```suggestion
                if isinstance(ch, DMM) and ch.bottom_detuning is not None:
                    bottom_detuning = (
                        f"{float(ch.bottom_detuning):.4g} rad/µs"
```

HaroldErbin: I have modified as you suggested. I also removed the cast for  `max_abs_detuning` because mypy was also complaining locally. However, I did not remove it for `max_map` because then I get an error:
```
pulser-core/pulser/devices/_device_datacls.py:664: error: Argument 1 to "float" has incompatible type "float | None"; expected "str | Buffer | SupportsFloat | SupportsIndex"  [arg-type]
```

Let's see what mypy gives now.

HaroldErbin:
mypy continues to fail, but in a file I did not modify. Should I try to fix it nonetheless?

HGSilveri:
> mypy continues to fail, but in a file I did not modify. Should I try to fix it nonetheless?

Yeah, I saw this pop-up too in another PR. Please just copy this line and it goes away: https://github.com/pasqal-io/Pulser/blob/67197da4032f2dbc37bc40e14618111312f11883/pulser-pasqal/pulser_pasqal/pasqal_cloud.py#L241

HGSilveri:
> Concerning the test coverage, which is now needed for `specs` since it's not internal, I was wondering what's the best approach. The simplest way is to assert that the `specs` string of different devices matches what is expected. However, this means having many lines of text (since we have to split the lines to respect the code style). Do you see a simpler way?

I'm afraid I don't see a way around it, no. I would maybe use a string with some format fields so that you can change some parameters (using @pytest.mark.parametrize) and check that the string always matches what you expect.

HaroldErbin:
I have added a test for three devices. At the end, I did not use `pytest.mark.parametrize` but a function, let me know if this works for you.

HaroldErbin:
To give an example of what has been implemented in the PR, here is the output of `AnalogDevice._specs()` before:
```
Register parameters:
 - Dimensions: 2D
 - Rydberg level: 60
 - Maximum number of atoms: 25
 - Maximum distance from origin: 35 μm
 - Minimum distance between neighbouring atoms: 5 μm
 - Maximum layout filling fraction: 0.5
 - SLM Mask: No
 - Maximum sequence duration: 4000 ns

Channels:
 - 'rydberg_global': Rydberg.Global(Max Absolute Detuning: 125.66370614359172 rad/µs, Max Amplitude: 12.566370614359172 rad/µs, Clock period: 4 ns, Minimum pulse duration: 16 ns, Maximum pulse duration: 100000000 ns, Modulation Bandwidth: 8 MHz, Basis: 'ground-rydberg')
```
and after:
```
Register parameters:
 - Dimensions: 2D
 - Rydberg level: 60
 - Maximum number of atoms: 25
 - Maximum distance from origin: 35 μm
 - Minimum distance between neighbouring atoms: 5 μm
 - Maximum layout filling fraction: 0.5
 - SLM Mask: No
 - Maximum sequence duration: 4000 ns

Device parameters:
 - Maximum number of runs: 2000
 - Channels can be reused: No
 - Supported bases: ground-rydberg
 - Supported states: r, g
 - Ising interaction coefficient: 865723.02

Layout parameters:
 - Requires layout: Yes
 - Accepts new layout: No
 - Minimal number of traps: 1
 - Maximal number of traps: None

Channels:
 - 'rydberg_global': Rydberg.Global(Max Absolute Detuning: 125.66370614359172 rad/µs, Max Amplitude: 12.566370614359172 rad/µs, Clock period: 4 ns, Minimum pulse duration: 16 ns, Maximum pulse duration: 100000000 ns, Modulation Bandwidth: 8 MHz, Basis: 'ground-rydberg')
```

pulser-core/pulser/devices/_device_datacls.py
```diff
@@ -579,6 +579,137 @@ def to_abstract_repr(self) -> str:
...
+            ),
+            f" - Maximum layout filling fraction: {self.max_layout_filling}",
+            f" - SLM Mask: {yes_no_fn(self.supports_slm_mask)}",
+        ]
+
+        if self.max_sequence_duration is not None:
+            lines.append(
+                " - Maximum sequence duration: "
+                f"{self.max_sequence_duration} ns"
+            )
```
HGSilveri, line 616: Looks like this was already mistakenly placed under "Register parameters", can you maybe place it under "Device parameters" instead?

pulser-core/pulser/devices/_device_datacls.py
```diff
@@ -579,6 +579,137 @@ def to_abstract_repr(self) -> str:
...
+    @property
+    def specs(self) -> str:
+        """Text summarizing the specifications of the device."""
+        return self._specs(for_docs=False)
+
+    def _specs(self, for_docs: bool = False) -> str:
+        def yes_no_fn(cond: bool) -> str:
+            return "Yes" if cond else "No"
+
+        lines = [
```
HGSilveri, line 598: nit: I guess it makes more sense to call these `reg_lines` now

pulser-core/pulser/devices/_device_datacls.py
```diff
@@ -579,6 +579,137 @@ def to_abstract_repr(self) -> str:
...
+            "\nRegister parameters:",
+            f" - Dimensions: {self.dimensions}D",
+            f" - Rydberg level: {self.rydberg_level}",
+            f" - Maximum number of atoms: {self.max_atom_num}",
+            f" - Maximum distance from origin: {self.max_radial_distance} μm",
```
HGSilveri, line 603: These parameters can be `None` in a `VirtualDevice`. 
I think saying `"Maximum number of atoms: None"` is acceptable (albeit still potentially confusing), but `"None μm" should be avoided.

pulser-core/pulser/devices/_device_datacls.py
```diff
@@ -579,6 +579,137 @@ def to_abstract_repr(self) -> str:
...
+            (
+                " - Minimum distance between neighbouring atoms: "
+                f"{self.min_atom_distance} μm"
+            ),
+            f" - Maximum layout filling fraction: {self.max_layout_filling}",
```
HGSilveri, line 608: This should be moved to "Layout parameters", since we are making that distinction. Also, it would make sense to have the "Layout parameters" come right after "Register parameters", since they are closely related

pulser-core/pulser/devices/_device_datacls.py
```diff
@@ -579,6 +579,137 @@ def to_abstract_repr(self) -> str:
...
+
+        layout_lines = [
+            "\nLayout parameters:",
+            f" - Requires layout: {yes_no_fn(self.requires_layout)}",
+        ]
+
+        if hasattr(self, "accepts_new_layouts"):
+            layout_lines.append(
+                f" - Accepts new layout: {yes_no_fn(self.accepts_new_layouts)}"
+            )
```
HGSilveri, line 653: How would you feel about separating the different lines in private methods (ie `_reg_lines()`, `_device_lines()`, ...) and then have `_specs()` combine them? I think it would potentially improve readability and it would allow us to override `Device._layout_lines()` to add this extra parameter that is not part of `BaseDevice`.

tests/test_devices.py
```diff
@@ -257,6 +258,65 @@ def test_tuple_conversion(test_params):
...
+            + "\nChannels:\n"
+            + "\n".join(
+                f" - '{name}': {ch!r}"
+                for name, ch in {**dev.channels, **dev.dmm_channels}.items()
+            )
+        )
+
+    assert MockDevice.specs == specs(MockDevice)
+    assert AnalogDevice.specs == specs(AnalogDevice)
+    assert DigitalAnalogDevice.specs == specs(DigitalAnalogDevice)
```
HGSilveri, line 317: This looks fine but you can still easily parametrize this, which will allow us to detect if this matching fails for one device and not others. You could simply add something like this:
```
@pytest.mark.parametrize("device", [MockDevice, AnalogDevice, DigitalAnalogDevice])
def test_device_specs(device):
    ...
    assert device.specs == specs(device)
```

pulser-core/pulser/devices/_device_datacls.py
```diff
@@ -579,6 +579,137 @@ def to_abstract_repr(self) -> str:
...
+        """Prints the device specifications."""
+        title = f"{self.name} Specifications"
+        header = ["-" * len(title), title, "-" * len(title)]
+        print("\n".join(header))
+        print(self._specs())
+
+    @property
+    def specs(self) -> str:
+        """Text summarizing the specifications of the device."""
+        return self._specs(for_docs=False)
```
HGSilveri, line 592: I'm curious as to why you, as a user, would like access to this string... Are you perhaps parsing it? I would agree that this quite inconvenient and I would think unnecessary given that you can access the device instance attributes.

HGSilveri (commented):
Thank you for this @HaroldErbin , very nice work! I have some questions and suggestions, I'd be happy to hear your thoughts.

pulser-core/pulser/devices/_device_datacls.py
HaroldErbin: I need this string in two places:
1. I am using the `logging` module to log all events of each run. The specs of the device is one thing I want to log for future reference if needed.
2. In my script, I have a function which checks and displays many different properties of the device. This includes the specs, but various other things I am computing.

HGSilveri: Ok, that's clear, thanks for the explanation!

pulser-core/pulser/devices/_device_datacls.py
HaroldErbin: In fact there were only two sections: "Register" and "Channels", so it made sense to have it under register before. I did not want to change what was already present, but I agree that it makes more sense to move it.

pulser-core/pulser/devices/_device_datacls.py
HaroldErbin: I have been hesitating on this point, because it seems useful to indicate that there are no limits. For this reason, when I was working on the code, I did hesitate to still add a line for any field being `None`, to indicate that there is no constraint. On the other hand, this may be confusing as you say, so i decided to follow the current convention of removing the lines altogether. I did not know that these values could also be `None`, so I suggest (at least for now) to also remove the lines. But I think it would be worth having other opnions on the matter.

HaroldErbin: There was a slight inconsistency with what I said, since I had left `max_layout_traps`. Now I am also removing the line if it's `None`.

HaroldErbin:
I have updated the codes with your suggestions, leaving only two conversations open for further discussions.

pulser-core/pulser/devices/_device_datacls.py
HGSilveri: I agree that showing there is no constraint can be useful but let's stick with not showing it for now.

pulser-core/pulser/devices/_device_datacls.py
HGSilveri: I can see you are not too keen on doing this, no worries :)

pulser-core/pulser/devices/_device_datacls.py
```diff
@@ -595,67 +595,70 @@ def _specs(self, for_docs: bool = False) -> str:
...
+                return line.format(var)
+
+        reg_lines = (
+            "\nRegister parameters:\n"
+            + f" - Dimensions: {self.dimensions}D\n"
+            + f" - Rydberg level: {self.rydberg_level}\n"
+            + check_none(self.max_atom_num, " - Maximum number of atoms: {}\n")
+            + check_none(
+                self.max_radial_distance,
+                " - Maximum distance from origin: {} µm\n",
```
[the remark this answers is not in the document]
HaroldErbin: If you prefer, I can change back, but I can give some more arguments. What I did not like with list, is that there were three different syntaxes – `lines = [...]`, `lines.append()`, `lines += []` – and one had to go back and forth between each, which is quite verbose. With a string, there is a single syntax, except when I need to check if an attribute is present (but this is caused by layout, and this would be unnecessary if I implemented your suggestion of having `layout_lines` only for `Device`).

pulser-core/pulser/devices/_device_datacls.py
HaroldErbin: Let me implement it, after your comment on using a string vs a list, I see more the advantage of doing it now, such that the code can be simplified further if we don't have to check if an object has an attribute.

pulser-core/pulser/devices/_device_datacls.py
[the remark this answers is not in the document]
HaroldErbin: These are good arguments, I will revert back to lists.

pulser-core/pulser/devices/_device_datacls.py
HaroldErbin: What do you suggest for the `yes_no_fn` and `check_none` helpers? Should I redefine them in each method? Or just go back to the previous form, where the test was done explicitly in each case?

HGSilveri: I'd prefer reusing them over redefining them. It's not the end of the world to have them as private functions at the end of the file. Between reusing and just doing things explicitly, I'll leave that up to you

HaroldErbin:
I have implemented the changes suggested by @HGSilveri:
1. Individual private methods have been created for each section (register, layout, device, channels).
2. The layout section is defined only for `Device`.
3. I have reverted back to using lists instead of strings.
4. I have defined two new private methods which are used to improve handling of `None` and boolean parameters.

Here are the results:

`DigitalAnalogDevice`
```
Register parameters:
 - Dimensions: 2D
 - Rydberg level: 70
 - Maximum number of atoms: 100
 - Maximum distance from origin: 50 µm
 - Minimum distance between neighbouring atoms: 4 μm
 - SLM Mask: Yes

Layout parameters:
 - Requires layout: Yes
 - Accepts new layout: Yes
 - Minimal number of traps: 1
 - Maximum layout filling fraction: 0.5

Device parameters:
 - Channels can be reused: No
 - Supported bases: digital, ground-rydberg
 - Supported states: r, g, h
 - Ising interaction coefficient: 5420158.53

Channels:
 - 'rydberg_global': Rydberg.Global(Max Absolute Detuning: 125.66370614359172 rad/µs, Max Amplitude: 15.707963267948966 rad/µs, Clock period: 4 ns, Minimum pulse duration: 16 ns, Maximum pulse duration: 67108864 ns, Basis: 'ground-rydberg')
 - 'rydberg_local': Rydberg.Local(Max Absolute Detuning: 125.66370614359172 rad/µs, Max Amplitude: 62.83185307179586 rad/µs, Minimum retarget time: 220 ns, Fixed retarget time: 0 ns, Max targets: 1, Clock period: 4 ns, Minimum pulse duration: 16 ns, Maximum pulse duration: 67108864 ns, Basis: 'ground-rydberg')
 - 'raman_local': Raman.Local(Max Absolute Detuning: 125.66370614359172 rad/µs, Max Amplitude: 62.83185307179586 rad/µs, Minimum retarget time: 220 ns, Fixed retarget time: 0 ns, Max targets: 1, Clock period: 4 ns, Minimum pulse duration: 16 ns, Maximum pulse duration: 67108864 ns, Basis: 'digital')
 - 'dmm_0': DMM.Global(Max Absolute Detuning: None, Max Amplitude: 0, Clock period: 4 ns, Minimum pulse duration: 16 ns, Maximum pulse duration: 67108864 ns, Basis: 'ground-rydberg')
 ```

`MockDevice`
```
Register parameters:
 - Dimensions: 3D
 - Rydberg level: 70
 - Minimum distance between neighbouring atoms: 0.0 μm
 - SLM Mask: Yes

Device parameters:
 - Channels can be reused: Yes
 - Supported bases: digital, ground-rydberg, XY
 - Supported states: u, d, r, g, h
 - Ising interaction coefficient: 5420158.53
 - XY interaction coefficient: 3700.0

Channels:
 - 'rydberg_global': Rydberg.Global(Max Absolute Detuning: None, Max Amplitude: None, Clock period: 1 ns, Minimum pulse duration: 1 ns, Basis: 'ground-rydberg')
 - 'rydberg_local': Rydberg.Local(Max Absolute Detuning: None, Max Amplitude: None, Minimum retarget time: 0 ns, Fixed retarget time: 0 ns, Clock period: 1 ns, Minimum pulse duration: 1 ns, Basis: 'ground-rydberg')
 - 'raman_global': Raman.Global(Max Absolute Detuning: None, Max Amplitude: None, Clock period: 1 ns, Minimum pulse duration: 1 ns, Basis: 'digital')
 - 'raman_local': Raman.Local(Max Absolute Detuning: None, Max Amplitude: None, Minimum retarget time: 0 ns, Fixed retarget time: 0 ns, Clock period: 1 ns, Minimum pulse duration: 1 ns, Basis: 'digital')
 - 'mw_global': Microwave.Global(Max Absolute Detuning: None, Max Amplitude: None, Clock period: 1 ns, Minimum pulse duration: 1 ns, Basis: 'XY')
 - 'dmm_0': DMM.Global(Max Absolute Detuning: None, Max Amplitude: 0, Clock period: 1 ns, Minimum pulse duration: 1 ns, Maximum pulse duration: 100000000 ns, Basis: 'ground-rydberg')
```

File: pulser-core/pulser/devices/_device_datacls.py [hunk 10]
```diff
@@ -591,71 +591,65 @@ def specs(self) -> str:
...
+            + f"{self.min_atom_distance} μm",
+            f" - SLM Mask: {self._param_yes_no(self.supports_slm_mask)}",
         ]
 
-        if self.max_sequence_duration is not None:
-            lines.append(
-                " - Maximum sequence duration: "
-                f"{self.max_sequence_duration} ns"
-            )
+        return "\n".join(line for line in register_lines if line != "")
```
HGSilveri, line 623: Can we return `list[str]` instead? This (and the other similar methods) could change to
```suggestion
        return [line for line in register_lines if line != ""]
```

pulser-core/pulser/devices/_device_datacls.py
```diff
@@ -579,6 +579,141 @@ def to_abstract_repr(self) -> str:
...
+
+    def _specs(self, for_docs: bool = False) -> str:
+
+        return "\n".join(
+            [
+                self._register_lines(),
+                self._device_lines(),
+                self._channel_lines(for_docs=for_docs),
+            ]
+        )
```
HGSilveri, line 715: This would then become
```suggestion
        return "\n".join(
                self._register_lines()
                + self._device_lines()
                + self._channel_lines(for_docs=for_docs)
            )
```

pulser-core/pulser/devices/_device_datacls.py
```diff
@@ -828,6 +890,33 @@ def from_abstract_repr(obj_str: str) -> Device:
...
+            " - Accepts new layout: "
+            + self._param_yes_no(self.accepts_new_layouts),
+            f" - Minimal number of traps: {self.min_layout_traps}",
+            self._param_check_none(self.max_layout_traps)(
+                " - Maximal number of traps: {}"
+            ),
+            f" - Maximum layout filling fraction: {self.max_layout_filling}",
+        ]
+
+        return "\n".join(line for line in layout_lines if line != "")
```
HGSilveri, line 907: I had something slightly different in mind for the layout parameters. I think the ones that are in `VirtualDevice` should also appear in its specs, so this method could be moved to `BaseDevice` _expect_ for the `accepts_new_layout` line. That one, we could add it like this:
```suggestion
    def _layout_lines(self) -> list[str]:
        layout_lines = super()._layout_lines()
        layout_lines.insert(2, f" - Accepts new layout: {self._param_yes_no(self.accepts_new_layouts)}" 
        return layout_lines
```

pulser-core/pulser/devices/_device_datacls.py
```diff
@@ -828,6 +890,33 @@ def from_abstract_repr(obj_str: str) -> Device:
...
+    def _specs(self, for_docs: bool = False) -> str:
+
+        return "\n".join(
+            [
+                self._register_lines(),
+                self._layout_lines(),
+                self._device_lines(),
+                self._channel_lines(for_docs=for_docs),
+            ]
+        )
```
HGSilveri, line 918: This would then be in `BaseDevice` and we don't need to redefine it here

HGSilveri (commented):
Super nice work @HaroldErbin , thanks a lot! I have a final proposition for the return types of these new methods that I think can make them slightly more flexible. Consider all the comments below together, they are related

HaroldErbin:
@HGSilveri : Just a quick message to say that I did not forget, but I am very busy with other projects until next week, I will finish the PR afterwards.

HGSilveri:
> @HGSilveri : Just a quick message to say that I did not forget, but I am very busy with other projects until next week, I will finish the PR afterward.

Thanks for letting me know @HaroldErbin , no problem on my side :)

HaroldErbin:
Sorry for that, it was a long time I did not do it so I messed up somewhere. Let me have a look and come back to you if needed.

HGSilveri (approved):
Looks great, thanks a lot for your efforts @HaroldErbin !

PR #764, opened 2024-11-19
Defining the new Backend API classes

HGSilveri:
- Introduces the `Results` class and make `Result` a subclass of it in backwards compatible way
- Defines the `State` and `Operator` ABCs
- Updates `BackendConfig` and defines `EmulatorConfig`
- Defines the `EmulatorBackend` ABC
-  Defines the `Callback` and `Observable` ABCs
- Defines the default observables

**To do:**

- [x] Define `QutipState` to have an example
- [x] Idem for `QutipOperator`
- [x] Solve relevant TODOs
- [x] Write UTs


pulser-core/pulser/backend/config.py
```diff
@@ -14,28 +14,223 @@
...
+            specified, the state type needs to be compatible with the emulator
+            backend. If left undefined, defaults to starting with all qudits
+            in the ground state.
+        with_modulation: Whether to emulate the sequence with the programmed
+            input or the expected output.
+        interaction_matrix: An optional interaction matrix to replace the
+            interaction terms in the Hamiltonian. For an N-qudit system,
+            must be an NxN symmetric matrix where entry (i, j) dictates
+            the interaction coefficient between qudits i and j, ie it replaces
+            the C/r_{ij}^6.
```
a-quelle, line 111: c_n/r{ij}^n? There is also XY.

a-quelle: Pablo recently added a `log_file` to our `BackendConfig` it would cause the application to pipe all output into the log file. It seems like a thing we could generally support, though I'm also fine with keeping it an emu-mps specific config value.

pulser-core/pulser/backend/default_observables.py
```diff
@@ -0,0 +1,369 @@
...
+                type(hamiltonian),
+            ).expect(state)
+
+        return [
+            [
+                calc_expectation(frozenset((i, j)))
+                for j in range(state.n_qudits)
+            ]
+            for i in range(state.n_qudits)
+        ]
```
a-quelle, line 247: We have recently had to add the following to this class

```
if hasattr(state, "get_correlation_matrix") and callable(
            state.get_correlation_matrix
        ):
            return state.get_correlation_matrix(
```

to resolve a performance bottleneck. We managed to improve the runtime of this callback a factor of 30 for the case we benchmarked. I would very much like to have some mechanism that prevents us from having to write an emu-mps specific callback to avoid a regression.

pulser-core/pulser/backend/config.py
HGSilveri: It sounds a bit EMU-MPS specific at this moment. I'd keep it there for now, we can always extend it to `EmulationConfig` if it makes sense in the future

pulser-core/pulser/backend/default_observables.py
```diff
@@ -0,0 +1,369 @@
...
+    def name(self) -> str:
+        """Returns the name of the observable."""
+        return "qubit_density"
+
+    def apply(
+        self, *, state: State, hamiltonian: Operator, **kwargs: Any
+    ) -> list:
+        """Calculates the observable to store in the Results."""
+        return [
+            CorrelationMatrix._get_number_operator(
```
a-quelle, line 282: Same as with the CorrelationMatrix, except it's less critical here. Maybe we can figure something out

pulser-core/pulser/backend/default_observables.py
```diff
@@ -0,0 +1,369 @@
...
+    def name(self) -> str:
+        """Returns the name of the observable."""
+        return "energy_variance"
+
+    def apply(
+        self, *, state: State, hamiltonian: Operator, **kwargs: Any
+    ) -> Any:
+        """Calculates the observable to store in the Results."""
+        # This works for state vectors and density matrices and avoids
+        # squaring the hamiltonian
```
a-quelle, line 331: last week I found the following implementation to be significantly faster and more memory efficient

```
h_squared = H @ H
        return h_squared.expect(state).real - H.expect(state).real ** 2
```

pulser-core/pulser/backend/default_observables.py
a-quelle: I've done some thinking, and in the worst case, we can always override the Callback behaviour in the backend ourselves. This holds for all the comments I wrote in this file, so I don't consider any of it truly blocking, but some thought might deliver a more elegant solution.

pulser-core/pulser/backend/default_observables.py
HGSilveri: Interesting, I thought that matrix-matrix multiplication would always be more resource intensive, so I tried to avoid it.  Just goes to show how little I know about matrix product states, I guess

pulser-core/pulser/backend/default_observables.py
a-quelle: I think in the end it's best if you just take implementations that work well for state-vector, and we monkeypatch them in the emulators as needed.

HGSilveri: So, in your `MPSConfig` or `MPSConfig`, you would basically override `apply()` of select observables at runtime?

pulser-core/pulser/backend/default_observables.py
```diff
@@ -0,0 +1,369 @@
...
+        h_state = hamiltonian.apply_to(state)
+        identity = hamiltonian.from_operator_repr(
+            eigenstates=state.eigenstates,
+            n_qudits=state.n_qudits,
+            operations=[(1.0, [])],
+        )
+        return identity.expect(h_state)
```
a-quelle, line 369: same here with the overlap, I guess.

pulser-core/pulser/backend/state.py
```diff
@@ -0,0 +1,151 @@
...
+            eigenstates: The basis states (e.g., ('r', 'g')).
+            amplitudes: A mapping between basis state combinations and
+                complex amplitudes.
+
+        Returns:
+            The state constructed from the amplitudes.
+        """
+        pass
+
+    def get_correlation_matrix(
```
a-quelle, line 120: Based on my monkey patching idea, we could remove this. Having it is a bit of a slipperly slope, since why not also add a `get_magnetization` and whatever other items you could think of.

HGSilveri: Sure, I'm fine with that

pulser-core/pulser/backend/config.py
```diff
@@ -14,28 +14,223 @@
...
+        """Initializes the EmulationConfig."""
+        for obs in observables:
+            if not isinstance(obs, Observable):
+                raise TypeError(
+                    "All entries in 'observables' must be instances of "
+                    f"Observable. Instead, got instance of type {type(obs)}."
+                )
+
+        if default_evaluation_times != "Full":
+            eval_times_arr = np.array(default_evaluation_times, dtype=float)
```
a-quelle, line 152: why is this a `np.array` but below you use `AbstractArray`?

HGSilveri: You mean for `interaction_matrix`? I use `AbstractArray` because I expect people might want to differentiate against it

pulser-core/pulser/backend/results.py
```diff
@@ -0,0 +1,122 @@
...
+
+        Args:
+            name: Name of the result to get times of.
+
+        Returns:
+            List of relative times.
+        """
+        return list(getattr(self, name).keys())
+
+    def get_result(self, name: str, time: float) -> Any:
```
a-quelle, line 80: I think it'd be useful if this also accepted an instance of `Observable` which would then simply call `name` on itself. The same for `get_result_times` etc.

pulser-core/pulser/backend/state.py
```diff
@@ -0,0 +1,236 @@
...
+        Args:
+            other: The other state.
+
+        Returns:
+            The overlap between the two states.
+        """
+        pass
+
+    @abstractmethod
+    def probabilities(
```
a-quelle, line 101: This also does not scale well with qubit number. Say you create a 100 qubit hadamard state, which is quite trivial to do for an MPS at least.  This has the uniform distribution on the bitstrings, and these probabilities cannot be represented in any floating point currently in use since the number `2^{-100}` is well below machine precision. Conversely sample methods bases on a Markov chain (such as we currently use in emu-mps) would only deal with numbers of order unity and sample from the underlying distribution correctly.

pulser-core/pulser/backend/state.py
```diff
@@ -0,0 +1,236 @@
...
+            cutoff: The value below which a probability is considered to be
+                zero.
+
+        Returns:
+            A mapping between basis state combinations and their respective
+            probabilities.
+        """
+        pass
+
+    def bitstring_probabilities(
```
a-quelle, line 116: same issue as above

pulser-core/pulser/backend/state.py
```diff
@@ -0,0 +1,236 @@
...
+            eigenstates: The basis states (e.g., ('r', 'g')).
+            amplitudes: A mapping between basis state combinations and
+                complex amplitudes.
+
+        Returns:
+            The state constructed from the amplitudes.
+        """
+        pass
+
+    def infer_one_state(self) -> Eigenstate:
```
a-quelle, line 218: why do we not simply throw an error if the eigenbasis is not predefined? No backend would know how to simulate it anyway.

pulser-core/pulser/backend/state.py
HGSilveri: Ok, that's good to know. So bascially what you're saying is that we need to have the whole `sample()` method be the `@abstractmethod` and we can't have any shared logic in `State`, right?

pulser-core/pulser/backend/state.py
```diff
@@ -0,0 +1,236 @@
...
+
+        The order of the states should match the order in a numerical (ie
+        state vector or density matrix) representation.
+        For example, with eigenstates ("a", "b", ...),  "a" will be associated
+        to eigenvector (1, 0, ...), "b" to (0, 1, ...) and so on.
+        """
+        return tuple(self._eigenstates)
+
+    @property
+    def basis_states(self) -> tuple[str, ...]:
```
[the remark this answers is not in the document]
HGSilveri: I believe it's still important that we have the order of the basis states enforced in the base class so I replaced this with `State.get_basis_state_from_index()`. I think you will find this one acceptable

pulser-core/pulser/backend/state.py
HGSilveri: Moved out of `State`

pulser-core/pulser/backend/state.py
a-quelle: yes

HGSilveri: That sounds like something each subclass may chose to do, I don't want to enforce the supported eigenbasis in the ABC

pulser-core/pulser/backend/operator.py
```diff
@@ -0,0 +1,140 @@
...
+        pass
+
+    @classmethod
+    @abstractmethod
+    def from_operator_repr(
+        cls: Type[OperatorType],
+        *,
+        eigenstates: Sequence[Eigenstate],
+        n_qudits: int,
+        operations: FullOp,
```
a-quelle, line 105: in the notion page we had an extra argument here `operators: dict[str, QubitOp]` where you could define aliases for certain `QuditOp`s such as `{"X": {"rg": 1, "gr": 1}}`. I found this super useful and started depending on it in several places. Now I realize it's not in the method. Could we add it back in?

HGSilveri: Can't it be specfic to `MPO`? I'm not a big fan of it and chose not to have it in `QutipOperator` as I found it overcomplicated the implementation unnecessarily. 
I think that what you wanted could be achieved by defining an auxiliary variable instead, eg.
```
X = {"rg": 1, "gr": 1}
QutipOperator.from_operator_repr(
    eigenstates=("r", "g"),
    n_qudits=3,
    operations=[(1.0, [(X, {0, 2})]],
)
```

a-quelle: That is fair.

pulser-core/pulser/backend/state.py
a-quelle: I'm not sure. For example, in this implementation, you're considering that the bases in lines 153-160 are known. We only support `rg` and `01` currently, and that is fine. But what is the use-case for supporting `uv` and having the user supply that `u` measures as `1`? Perhaps in the qutip backend the state object is independent enough from the simulation that this works, even though no dynamics are really defined for it, but actually enforcing all backends to have a state so independent from the simulation is what seems excessive to me.

HGSilveri: The reason why I'm allowing this flexibility is because, already, we have a case where we can't infer the one state - notably, when we have the qutrit with eigenstates `rgh`, the `one_state` actually depends on the measurement basis.
I defined `infer_one_state()` to avoid having the user defining the one state as much as possible but sometimes it is unavoidable.
I think that if you want to block the definition of unknown eigenstates or custom one states, you can do so in `MPS`. As for the ABC, I think we should allow this flexibility. If we don't and at some point want to support some other basis in a particular backend, we would have to open a PR on `pulser-core` to change the base class, which doesn't make sense to me.

pulser-core/pulser/backend/config.py
```diff
@@ -14,28 +14,236 @@
...
+            default_evaluation_times = cast(Sequence[float], eval_times_arr)
+
+        if initial_state is not None and not isinstance(initial_state, State):
+            raise TypeError(
+                "When defined, 'initial_state' must be an instance of State;"
+                f" got object of type {type(initial_state)} instead."
+            )
+
+        if interaction_matrix is not None:
+            interaction_matrix = pm.AbstractArray(interaction_matrix)
```
a-quelle, line 178: should we symmetrize this? or at least check for symmetry?

HGSilveri: Yes, I'll add a check

pulser-core/pulser/backend/results.py
```diff
@@ -0,0 +1,157 @@
...
+        Args:
+            observable: The observable instance used to calculate the result
+                or its tag.
+
+        Returns:
+            List of relative times.
+        """
+        return list(self._results[self._find_uuid(observable)].keys())
+
+    def get_result(self, observable: Observable | str, time: float) -> Any:
```
pablolh, line 85: There is an [interesting feature request in emu-mps](https://github.com/pasqal-io/emulators/issues/23) that is related to this.

Also this morning we discussed with Kemal that the "dict" data structure indexed by time is maybe not the most appropriate:
- most people use the data to make plots, where they need a list of values in chronological order
- few people request value at a given time individually, except the latest time (with a list `[-1]`)
- dict has probably a lot of overhead compared to list

pablolh: Should we specify that the `_store` method is called in increasing order of time for a given callback? In which case the dict entries are always sorted by time [according to this](https://docs.python.org/3/library/stdtypes.html#dictionary-view-objects) 

> Keys and values are iterated over in insertion order.

HGSilveri: I agree with this, although we should still store the evaluation times of each observable in a separate list

pablolh: That sounds reasonable I think. Maybe allow sth like `plt.plot(*results["energy"])`, that is, return both times and values in unzipped format in the `__getitem__`?

There is also this thing that `evaluation_times` could just be a sort of observable dt/integer multiple of dt (or -1 for just the final time for instance), except if there is some strong use case for evaluating observables at non-regular times. In that case, it is not needed to store the times. At least all notebooks I saw of people using emu_mps API use a range of x multiples.
Sounds insignificant but when we get to serialize it for repeated usage in cloud, can save some traffic.

On the side, I think having a `plot` method for adequate observables would greatly help people working with simulation data. Throwing here some ideas:
```
results = MPSBackend().run(...)

plt.plot(*results["energy"])
# or
results["energy"].plot()
# or
results.plot("energy")
```
This plot method would be observable-specific and NotImplementedError by default in the base class

HGSilveri: > That sounds reasonable I think. Maybe allow sth like `plt.plot(*results["energy"])`, that is, return both times and values in unzipped format in the `__getitem__`?

I'm not sure about this, not all observables are plottable and I think it would make for a counter-intuitive return type for `__getitem__()`.
(Side note: you might have noticed that I went for `__getattr__` instead of `__getitem__` - so it would be `results.energy` rather than `results["energy"]`. This is to avoid confusion between `Results` and `Sequence[Results]`, which some backends return. )

> There is also this thing that `evaluation_times` could just be a sort of observable dt/integer multiple of dt (or -1 for just the final time for instance), except if there is some strong use case for evaluating observables at non-regular times. In that case, it is not needed to store the times. At least all notebooks I saw of people using emu_mps API use a range of x multiples.
> Sounds insignificant but when we get to serialize it for repeated usage in cloud, can save some traffic.

I don't feel comfortable enforcing this at the base class level. I'd rather take the (minor) hit of making the serialized object a bit larger, we won't be saving much there anyway.

> On the side, I think having a `plot` method for adequate observables would greatly help people working with simulation data. Throwing here some ideas:
> 
> ```
> results = MPSBackend().run(...)
> 
> plt.plot(*results["energy"])
> # or
> results["energy"].plot()
> # or
> results.plot("energy")
> ```
> 
> This plot method would be observable-specific and NotImplementedError by default in the base class

I'm not opposed to this but I think this is the sort of thing that would make more sense to implement first in a MPS specific `Results` subclass, for two main reasons: 
- All users are currently interacting with `Results` via EMU-MPS
- I suspect there will be some trial and error before the best format is found, so it's better for you to have the freedom and flexibility to make adjustments without relying on `pulser-core`
Once the format is well-established, we can potentially move it up to `Results` so it's shared by all backends.

pulser-core/pulser/backend/results.py
```diff
@@ -51,19 +53,19 @@ def _store(
...
-        if time in self._results[observable.uuid]:
+        _times = self._times.setdefault(observable.uuid, [])
+        if time in _times:
             raise RuntimeError(
                 f"A value is already stored for observable '{observable.tag}'"
                 f" at time {time}."
             )
         self._tagmap[observable.tag] = observable.uuid
-        self._results[observable.uuid][time] = value
+        _times.append(time)
```
pablolh, line 63: `assert time > _times[-1]` ?

pulser-core/pulser/backend/results.py
```diff
@@ -51,19 +53,19 @@ def _store(
...
+        self._results.setdefault(observable.uuid, []).append(value)
```
pablolh, line 64:
```
_results = self._results.setdefault(observable.uuid, [])
assert len(_times) == len(_results)
_results.append(value)
```
is that also an idea?

pulser-core/pulser/backend/results.py
HGSilveri: This is not something we can assert because we are actually not enforcing any ordering to the evaluation times. Perhaps we should, but for now we just leave it up to the user

pulser-core/pulser/backend/results.py
HGSilveri: This should always be true we can `assert` it, yes

HGSilveri:
@a-corni This is ready for review

pulser-core/pulser/backend/results.py
pablolh: > This is not something we can assert because we are actually not enforcing any ordering to the evaluation times. 

"we cannot do it because we don't do it" ?

people are plotting directly with that list of times as x axis
plotting with unsorted data gives bad plots

imagine a physicist having to sort the times in increasing order and needing to apply the same ordering to another list of values, not so difficult in Python but a clear programming challenge for some that will take them time

For me there is a clear expectation from this API to get sorted times, and in practice it is sorted in emumps. that's why I was suggesting an assert

An alternative to the assert is to maintain the ordering/sort the list inside the API code each time, with adaptive sorting algos should be pretty quick.

HGSilveri: You only assert for things you are sure will be true unless there are bugs - in this case, the `assert` can fail if the user gives disordered evaluation times and we are not prohibiting that.

> people are plotting directly with that list of times as x axis
> plotting with unsorted data gives bad plots

They shouldn't be providing unsorted evaluation times in the first place - if they are, I would argue they are actually going out of their way to do it and have a reason to do so.

pulser-core/pulser/backend/__init__.py
```diff
@@ -14,8 +14,44 @@
...
+from pulser.backend.observable import Callback, Observable
+from pulser.backend.default_observables import (
+    BitStrings,
+    CorrelationMatrix,
+    Energy,
+    EnergyVariance,
+    Expectation,
+    Fidelity,
+    Occupation,
+    SecondMomentOfEnergy,
```
a-corni, line 33: 1) I find this name funny 
2) Given that we have Energy, EnergyVariance, I would suggest to have EnergySecondMoment

pulser-core/pulser/backend/config.py
```diff
@@ -14,31 +14,245 @@
...
+                )
+            self._backend_options.update(backend_options["backend_options"])
+
+    def _expected_kwargs(self) -> set[str]:
+        return set()
+
+    def __getattr__(self, name: str) -> Any:
+        if (
+            # Needed to avoid recursion error
+            "_backend_options" in self.__dict__
```
a-corni, line 83: Is it possible that _backend_options is not among the attributes ?

a-corni (commented):
First two comments

pulser-core/pulser/backend/__init__.py
HGSilveri: It's what is being used in `EMU-MPS`. Personally, I would have called it `EnergySquared` but I don't have a big issue with it. 
What do you think @a-quelle , @pablolh ? Is it worth changing?

pulser-core/pulser/backend/config.py
HGSilveri: Yeah, it is only added in the middle of `__init__()` and I think Python was trying to access attributes before that, causing issues that forced me to add this here

pulser-core/pulser/backend/observable.py
```diff
@@ -191,4 +191,9 @@ def _validate_eval_times(
                 "Evaluation times must be unique but "
                 f"{evaluation_times!r} has repeated values."
             )
+        if not np.all(eval_times_arr[:-1] < eval_times_arr[1:]):
```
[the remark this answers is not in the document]
HGSilveri: You are correct, I'm assuming that the backend is respecting the order of the evaluation times. I'll make it explicit instead

pulser-core/pulser/backend/__init__.py
pablolh: So that's the thing, it is **not** the energy squared, it is the _second moment_ in statistics terms, which is the mean of the square of the Hamiltonian.
It is used for computing the energy variance with [_monte-carlo jump method_](https://www.phys.ens.psl.eu/~dalibard/publi3/osa_93.pdf). When you have n runs with monte carlo, you **cannot** average the `EnergyVariance` of all runs (which is an existing observable as well) to get the noisy `EnergyVariance`. Instead, you compute the average of the energy `<H>` for all runs and square it, then compute the average of the second moment, **then** substract:
`Var = <psi|H^2|psi> - <psi|H|psi>^2 = SecondMomentOfEnergy - Energy^2 = SecondMomentOfEnergy - EnergySquared`

Concerning the name, I tend to agree that `EnergySecondMoment` is a better name indeed. I don't mind changing, it will require minor refactoring in emu-mps that we can afford, but Anton is in charge of this integration atm so I cannot really decide for him. Regarding existing users of this feature it's a bit difficult to say how many there are, but there are people using `SecondMomentOfEnergy` already in their scripts for sure.

pulser-core/pulser/backend/results.py
```diff
@@ -60,6 +60,10 @@ def _store(
                 f" at time {time}."
             )
         self._tagmap[observable.tag] = observable.uuid
+        try:
```
pablolh, line 63: what about
`assert _times == [] or _times[-1] < time`
instead of the try-catch block?

pablolh (approved):

pulser-core/pulser/backend/__init__.py
a-quelle: I agree that we should probably have the name start with `Energy`. @pablolh, you mention

> So that's the thing, it is not the energy squared, it is the second moment in statistics terms, which is the mean of the square of the Hamiltonian.

but what's the difference between the two things for you? In my mind, it's kind of the same thing, and I suspect it's the same for @HGSilveri

HGSilveri: Let's go with `EnergySecondMoment`

pulser-core/pulser/backend/observable.py
```diff
@@ -0,0 +1,199 @@
...
+            return self._base_tag
+        return f"{self._base_tag}_{self._tag_suffix}"
+
+    def __call__(
+        self,
+        config: EmulationConfig,
+        t: float,
+        state: State,
+        hamiltonian: Operator,
+        result: Results,
```
a-corni, line 122: From a user perspective, I think I would have prefered to have call returning a value (like what apply does) and have a method that would use call and associate that value to a result object (apply could be the name for this, add otherwise ?)

pulser-core/pulser/backend/observable.py
```diff
@@ -0,0 +1,199 @@
...
+        """
+        time_tol = (
+            (0.5 / result.total_duration) if result.total_duration else 1e-6
+        )
+        if (
+            self.evaluation_times is not None
+            and config.is_time_in_evaluation_times(
+                t, self.evaluation_times, tol=time_tol
+            )
+        ) or config.is_evaluation_time(t, tol=time_tol):
```
a-corni, line 149: Why isn't this test in apply ?

pulser-core/pulser/backend/observable.py
```diff
@@ -0,0 +1,199 @@
...
+    from pulser.backend.config import EmulationConfig
+    from pulser.backend.results import Results
+
+
+class Callback(ABC):
+    """A general Callback that is called during the emulation."""
+
+    def __init__(self) -> None:
+        """Initializes a Callback."""
+        self._uuid: uuid.UUID = uuid.uuid4()
```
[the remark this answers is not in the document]
a-corni: Got it later, it's for the repr

pulser-core/pulser/backend/results.py
a-corni: Euh if there is no assertion then the append is not needed, right ? 
```suggestion
        self._results.setdefault(observable.uuid, [value])
```

pulser-core/pulser/backend/state.py
```diff
@@ -0,0 +1,173 @@
...
+from abc import ABC, abstractmethod
+from collections import Counter
+from collections.abc import Sequence
+from typing import Generic, Literal, Type, TypeVar, Union
+
+import numpy as np
+
+from pulser.channels.base_channel import States
+
+Eigenstate = Union[States, Literal["0", "1"]]
```
a-corni, line 26: What are "0" and "1" for ? Is it Microwave ?

pulser-core/pulser/result.py
```diff
@@ -136,12 +152,28 @@ class SampledResult(Result):
         meas_basis: The measurement basis.
         bitstring_counts: The number of times each bitstring was
             measured.
+        evaluation_time: Relative time at which the samples were
```
a-corni, line 155: nit: Relative times ?

pulser-core/pulser/result.py
```diff
@@ -136,12 +152,28 @@ class SampledResult(Result):
...
+            taken.
     """
 
     bitstring_counts: dict[str, int]
+    evaluation_time: float = 1.0
 
     def __post_init__(self) -> None:
+        super().__post_init__()
         self.n_samples = sum(self.bitstring_counts.values())
+        # TODO: Make sure this is not too hacky
```
a-corni, line 165: Definitely hacky, but valid 😆

pulser-simulation/pulser_simulation/qutip_op.py
```diff
@@ -0,0 +1,270 @@
...
+        Args:
+            state: The state to apply this operator to.
+
+        Returns:
+            The resulting state.
+        """
+        self._validate_other(state, QutipState, "QutipOperator.apply_to()")
+        out = self._operator * state._state
+        if state._state.isoper:
+            out = out * self._operator.dag()
```
a-corni, line 88: I am not sure I get it, if the state is represented by an operator (a density matrix then) $\rho$ then applying our operator $O$ on it is $O\rho \rho^\dagger$ ?

a-corni (commented):
Additionnal comments:
- I think we can now close https://github.com/pasqal-io/Pulser/pull/706
- Will it one day be relevant to be able to go from qutip objects/states to the abstract representation ? Do we create an issue for this ? I can be a nice thing to propose at unitary hack for instance...

pulser-simulation/pulser_simulation/qutip_state.py
```diff
@@ -0,0 +1,286 @@
...
+            )
+            if set(self.eigenstates) != set(other.eigenstates):
+                raise ValueError(msg)
+            raise NotImplementedError(msg)
+        overlap = self._state.overlap(other._state)
+        if self._state.isket and other._state.isket:
+            overlap = np.abs(overlap) ** 2
+        return float(overlap.real)
+
+    def probabilities(self, *, cutoff: float = 1e-10) -> dict[str, float]:
```
a-corni, line 111: Just to know, why do we choose a cutoff of 1e-10 ?
I see only two default tolerances:
- the absolute tolerance of the simulation (1e-8)
- the value below which to consider a number 0, which seems to be 1e-12 in qutip https://groups.google.com/g/qutip/c/Pe2zhHBzNkE

pulser-simulation/pulser_simulation/qutip_state.py
```diff
@@ -0,0 +1,286 @@
...
+        else:
+            probs = (np.abs(self._state.full()) ** 2).flatten().real
+        non_zero = np.argwhere(probs > cutoff).flatten()
+        # Renormalize to make the non-zero probablilites sum to 1.
+        probs = probs[non_zero]
+        probs = probs / np.sum(probs)
+        return dict(zip(map(self.get_basis_state_from_index, non_zero), probs))
+
+    def bitstring_probabilities(
+        self, *, one_state: Eigenstate | None = None, cutoff: float = 1e-10
```
a-corni, line 135: idem on cutoff

pulser-simulation/pulser_simulation/qutip_state.py
```diff
@@ -0,0 +1,286 @@
...
+            one_state: The eigenstate that measures to 1. Can be left undefined
+                if the state's eigenstates form a known eigenbasis with a
+                defined "one state".
+            cutoff: The value below which a probability is considered to be
+                zero.
+
+        Returns:
+            A mapping between bitstrings and their respective probabilities.
+        """
+        one_state = one_state or self.infer_one_state()
```
a-corni, line 149: Ah that's where it comes into play ! Nice !

pulser-core/pulser/backend/config.py
```diff
@@ -14,31 +14,245 @@
...
+            if (
+                initial_state is not None
+                and _shape[0] != initial_state.n_qudits
+            ):
+                raise ValueError(
+                    f"The received interaction matrix of shape {_shape} is "
+                    "incompatible with the received initial state of "
+                    f"{initial_state.n_qudits} qudits."
+                )
+            matrix_arr = interaction_matrix.as_array(detach=True)
```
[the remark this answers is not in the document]
kbidzhiev: Hi,
I did smth similar on our side in old base classes, some part might be useful

```
if not (
  int_mat.numel() != 0  # Empty input
  and torch.all(torch.isreal(int_mat))  # real values for having hermitean Hamiltonian
  and int_mat.dim() == 2  # to avoid 3nested lists list[list[list[float]]]
  and int_mat.shape[0] == int_mat.shape[1]  # no rectangular matrices
  and torch.allclose(int_mat, int_mat.T, atol=tol)  # symmetric matrix
  and torch.norm(torch.diag(int_mat)) < tol  # zero on diag
):
```

also I've added one check to avoid trapezoidal lists. I think it's not necessary, cause numpy and torch constructors error

```
if interaction_matrix is not None and not (
  isinstance(interaction_matrix, list)
  and isinstance(interaction_matrix[0], list)
  and isinstance(interaction_matrix[0][0], float) :   # this line checks depth of array ~int_mat.dim() == 2
```

a-corni (commented):
I am done with my comments :)

pulser-core/pulser/backend/config.py
```diff
@@ -14,31 +14,245 @@
...
+    initial_state: StateType | None
+    with_modulation: bool
+    interaction_matrix: pm.AbstractArray | None
+    prefer_device_noise_model: bool
+    noise_model: NoiseModel
+
+    def __init__(
+        self,
+        *,
+        observables: Sequence[Observable] = (),
```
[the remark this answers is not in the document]
HGSilveri: Good point, perhaps we should... But I'm not sure if that will affect the work already done with `MPSConfig`

pulser-core/pulser/backend/config.py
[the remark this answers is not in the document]
HGSilveri: We're just ignoring the diagonal, perhaps a `Warning` is enough, no?

pulser-core/pulser/backend/observable.py
HGSilveri: This is not supposed to be called by users though, it's the backends that call the observables

pulser-core/pulser/backend/observable.py
HGSilveri: Because `apply()` is solely for calculating the observable. It's the `abstractmethod` so we want it to be focused only on the observable specific part

pulser-core/pulser/backend/state.py
HGSilveri: It can be, yes.

pulser-simulation/pulser_simulation/qutip_op.py
HGSilveri: `out` is already $\hat{O}\rho$ , so here we are doing $\hat{O}\rho\hat{O}^{\dagger}$

pulser-simulation/pulser_simulation/qutip_state.py
HGSilveri: It was arbitrary tbh. Which do you prefer I change it to?

HGSilveri:
> Additionnal comments:
> 
>     * I think we can now close [Create tools to build operators #706](https://github.com/pasqal-io/Pulser/pull/706)
> 
>     * Will it one day be relevant to be able to go from qutip objects/states to the abstract representation ? Do we create an issue for this ? I can be a nice thing to propose at unitary hack for instance...

That will be hard to do in general. I think what we'll have instead is an `AbstractState` and `AbstractOperator` that allow you to start from simple things and sum/multiple/tensor into more complicated objects (a bit like in #706 :sweat_smile: )
This will also make it easier to switch Observables between backends

pulser-core/pulser/backend/config.py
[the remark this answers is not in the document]
a-corni: @pablolh @a-quelle @kbidzhiev What's your opinion ?

pulser-core/pulser/backend/config.py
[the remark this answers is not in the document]
a-corni: If it's really ignored by all backends, then a Warning is enough :)

pulser-core/pulser/backend/config.py
```diff
@@ -14,31 +14,245 @@
...
+                initial_state is not None
+                and _shape[0] != initial_state.n_qudits
+            ):
+                raise ValueError(
+                    f"The received interaction matrix of shape {_shape} is "
+                    "incompatible with the received initial state of "
+                    f"{initial_state.n_qudits} qudits."
+                )
+            matrix_arr = interaction_matrix.as_array(detach=True)
+            if not np.allclose(matrix_arr, matrix_arr.transpose()):
```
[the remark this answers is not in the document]
a-corni: Good, I am ok with that.

pulser-core/pulser/backend/observable.py
a-corni: Fine

a-corni (approved):
~Have to check the doc build I approve :)~

pulser-core/pulser/backend/config.py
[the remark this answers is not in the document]
a-quelle: I sometimes run sequences without any output for benchmarking purposes. I think it's useful to allow this. Maybe issue a warning instead?

pulser-core/pulser/backend/observable.py
a-quelle: In addition, it's perfectly valid to have functions that don't return anything. Why should it be expected that `call` returns something?

pulser-core/pulser/backend/observable.py
a-quelle: indeed, this solution means `apply` only contains logic appropriate to that individual observable, sharing logic as much as possible, and keeping the subclasses light-weight.

a-corni: Ah yes there is no to time involved in apply, my bad !

pulser-core/pulser/backend/state.py
a-corni: You are being mysterious haha Why do we have to add them, and not only stick with ("u", "d") ?

pulser-core/pulser/result.py
a-corni: My bad, went too fast

pulser-simulation/pulser_simulation/qutip_state.py
a-corni: 1e-12 seems to be the best for qutip :)

pulser-core/pulser/backend/config.py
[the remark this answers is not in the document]
HGSilveri: I'll go with a Warning

pulser-core/pulser/backend/state.py
HGSilveri: Sorry, more lazy than mysterious really :sweat_smile: 
The truth is that we are not enforcing the states have to be one of the options in `EIGENSTATES`, nor do we want to. Technically, a user can define a self-consistent set of arbitrary eigenstates - as long as they match in the state, the operator and the observables, it will work.
The point here is that `"0"` and `"1"` is a very reasonable choice of eigenstate names and it is trivial to tell which is the `one_state`, so I wanted to include that in `infer_one_state()`. 
It's an inconsequential decision anyway, as this is just a type hint alias.

a-corni: Thanks for the precisions !

a-corni (approved):

PR #765, opened 2024-11-19
Fix typo in docstring of Sequence.draw

HaroldErbin:
The dosctring of `Sequence.draw` contained the words "excepted curves" instead of "expected curves".


HaroldErbin:
Branch was started out of wrong branch, see PR #766 for corrected version.

PR #768, opened 2024-11-29
Fixes to the amplitude noise implementation

### LINKED ISSUE #712: Apply amplitude fluctutations at a fixed rate
Now that I see it, the implementation is perhaps a bit weird, since in the case of a Sequence containing only one super-long pulse, the value of noise_base that will be applied will be the same all along - there will be no "fluctuation" actually.
I would suggest to define a rate at which fluctuations should happen (and perhaps they should fluctuate at each ns ?), and sample them at each run (as it's done at the moment).

_Originally posted by @a-corni in https://github.com/pasqal-io/Pulser/pull/710#discussion_r1684308905_

HGSilveri:
- Adds information of the beam's propagation direction to `Channel` 
- Makes finite-waist effects depend on the position relative to the optical axis
- Makes amplitude fluctuations occur from run to run instead of pulse to pulse
- Fixes the transfer of `laser_waist` between `NoiseModel` and `SimConfig`
- Starts adding the `pulser_version` to serialized devices 

Note: To avoid disruptions for serialized devices between different versions, `propagation_dir` is defaulting to `None` in all channels and `QutipEmulator` is assuming using `propagation_dir=(0,1,0)` whenever it encounter `None`. In a future version, it would make more sense to change the default `propagation_dir` of `Global` channels to be `(0,1,0)` and remove this assumption from `QutipEmulator`. 

Fixes #712

### DIFF
--- a/pulser-core/pulser/channels/base_channel.py
+++ b/pulser-core/pulser/channels/base_channel.py
@@ -254,6 +254,10 @@ class Channel(ABC):
                     "'propagation_dir' must be given as a non-zero 3D vector;"
                     f" got {self.propagation_dir} instead."
                 )
+            # Make sure it's stored as a tuple
+            object.__setattr__(
+                self, "propagation_dir", tuple(self.propagation_dir)
+            )
 
     @property
     def rise_time(self) -> int:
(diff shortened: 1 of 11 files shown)


a-corni (approved):
LGTM

a-corni:
Ah I forgot about one question: should we raise a NotImplementedError if the propagation vector is implemented in a channel that is not Rydberg.Global ?

HGSilveri:
> Ah I forgot about one question: should we raise a NotImplementedError if the propagation vector is implemented in a channel that is not Rydberg.Global ?

I think that is overly restrictive, how about I do that only for `Local` channels?

a-corni (approved):
LGTM !

PR #771, opened 2024-12-02
Draw the phase curve by default in Sequence.draw()

### LINKED ISSUE #760: Draw the phase curve by default in Sequence.draw()
Right now, to see the phase curve one has to explicitly set `draw_phase_curve=True`. I think we can change the default to always show the phase curve whenever it is non-trivial (ie not zero during the whole sequence).

a-corni:
Sets the default value of the draw_phase_curve argument to True.
It's working: can be checked in the documentation on EOM.
It's perhaps a bit annoying for people working with draw_phase_area: since draw_phase_curve is True by default now, they have to put draw_phase_cruve=False if they want the value of the phase displayed.

Suggestion: What about always printing the phase when draw_phase_area is selected (no matter the value of draw_phase_curve) ?
 
Fixes #760


HGSilveri (approved):
LGTM!

PR #774, opened 2024-12-04
Deprecate usage of int in QubitIds

### LINKED ISSUE #769: Deprecate usage of `int` as qubit IDs and uniformly recommend usage of the "q" prefix
Currently, we allow `QubitId = str | int`. There are a few problems with this:

- Incompatible with the abstract repr: Whenever we serialize, all `QubitId`s are turned into `str`. Upon deserialization, they remain `str` because we can't know whether a `"1"` was originally `1` or `"1"`. This is also susceptible to clashes, eg if both `1` and `"1"` exist as separate indices in the same `Register`.
 - Harder to handle: we have to always account for two separate types

So, we should deprecate the usage of `int`s as `QubitId`s. Unfortunately, the classmethods of `Register` and `Register3D` currently use `int`s as the default qubit IDs. For consistency with what `RegisterLayout.define_register()` does, we should also recommend usage of the `prefix="q"` in the deprecation warning, as that will become the default behaviour for these methods once usage of `int` as qubit IDs is removed.

a-corni:
Fixes #769


pulser-core/pulser/register/base_register.py
```diff
@@ -77,6 +78,16 @@ def __init__(
             [pm.AbstractArray(v, dtype=float) for v in qubits.values()]
         )
         self._ids: tuple[QubitId, ...] = tuple(qubits.keys())
+        not_str = [id for id in self._ids if not isinstance(id, str)]
+        if not_str:
+            warnings.simplefilter("always")
+            warnings.warn(
+                "Usage of `int`s as `QubitId`s will be deprecated. Define "
+                "your `QubitId`s as `str`s, prefer setting `prefix='q'` "
+                " when using classmethods.",
```
HGSilveri, line 87: nit:
```suggestion
                "Usage of `int`s as `QubitId`s will be deprecated. Define "
                "your `QubitId`s as `str`s and favour setting `prefix='q'` "
                "when using classmethods, as that will become the new "
                "default once `int` qubit IDs become invalid. ",
```

pulser-core/pulser/register/base_register.py
```diff
@@ -77,6 +78,17 @@ def __init__(
             [pm.AbstractArray(v, dtype=float) for v in qubits.values()]
         )
         self._ids: tuple[QubitId, ...] = tuple(qubits.keys())
+        not_str = [id for id in self._ids if not isinstance(id, str)]
+        if not_str:
+            warnings.simplefilter("always")
+            warnings.warn(
+                "Usage of `int`s as `QubitId`s will be deprecated. Define "
+                "your `QubitId`s as `str`s, prefer setting `prefix='q'` "
```
HGSilveri, line 86: nit: 
```suggestion
                "your `QubitId`s as `str`s and prefer setting `prefix='q'` "
```

HGSilveri (commented):
Can we change the `QubitId` type hint to be just `str`?

pulser-core/pulser/register/base_register.py
```diff
@@ -77,6 +78,17 @@ def __init__(
             [pm.AbstractArray(v, dtype=float) for v in qubits.values()]
         )
         self._ids: tuple[QubitId, ...] = tuple(qubits.keys())
+        not_str = [id for id in self._ids if not isinstance(id, str)]
+        if not_str:
```
HGSilveri, line 82: nit: 
```suggestion
        if any(not isinstance(id, str) for id in self._ids):
```

pulser-core/pulser/register/base_register.py
```diff
@@ -77,6 +78,17 @@ def __init__(
...
+            warnings.simplefilter("always")
+            warnings.warn(
+                "Usage of `int`s as `QubitId`s will be deprecated. Define "
```
HGSilveri, line 85: We are not actually checking that they are `int`, just that they are not str... To be precise, we should say
```suggestion
                "Usage of `int`s or other non-`str` types as `QubitId`s will be deprecated. Define "
```

HGSilveri (approved):
LGTM!

PR #779, opened 2024-12-11
Allow definition of a custom phase jump time + Suppress output modulation for phase

### LINKED ISSUE #762: Allow for the definition of a custom phase jump time in Channel
Right now, `Channel.phase_jump_time` is always taken to be `2 * rise_time`. However, in some cases we might want to set a custom value for this parameter, so we should allow that via `custom_phase_jump_time: int | None`.

HGSilveri:
- Adds `Channel.custom_phase_jump_time`
- Suppresses modulation of phase samples as it is not physically accurate

Closes #762.

### DIFF
--- a/pulser-core/pulser/channels/base_channel.py
+++ b/pulser-core/pulser/channels/base_channel.py
@@ -34,7 +34,11 @@ warnings.filterwarnings("once", "A duration of")
 
 ChannelType = TypeVar("ChannelType", bound="Channel")
 
-OPTIONAL_ABSTR_CH_FIELDS = ("min_avg_amp", "propagation_dir")
+OPTIONAL_ABSTR_CH_FIELDS = (
+    "min_avg_amp",
+    "custom_phase_jump_time",
+    "propagation_dir",
+)
 
 # States ranked in decreasing order of their associated eigenenergy
 States = Literal["u", "d", "r", "g", "h", "x"]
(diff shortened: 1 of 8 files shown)


pulser-core/pulser/channels/base_channel.py
```diff
@@ -78,6 +82,9 @@ class Channel(ABC):
         min_avg_amp: The minimum average amplitude of a pulse (when not zero).
         mod_bandwidth: The modulation bandwidth at -3dB (50% reduction), in
             MHz.
+        custom_phase_jump_time: An optional custom value for the phase jump
```
a-corni, line 85: Perhaps should we mention that this is not used in EOM mode ?

tests/test_sequence.py
```diff
@@ -1729,10 +1729,16 @@ def test_sequence(reg, device, patch_plt_show):
...
-def test_estimate_added_delay(eom):
+def test_estimate_added_delay(eom, custom_phase_jump_time):
+    ryd_ch_obj = dataclasses.replace(
+        AnalogDevice.channels["rydberg_global"],
+        custom_phase_jump_time=custom_phase_jump_time,
+    )
+    device = dataclasses.replace(AnalogDevice, channel_objects=(ryd_ch_obj,))
     reg = Register.square(2, 5)
-    seq = Sequence(reg, AnalogDevice)
+    seq = Sequence(reg, device)
```
a-corni, line 1741: I am wondering if this is hyper user-friendly... I don't know if any users will modify the channel in a Device to define their custom phase_jump_time (it will be a good test to see if people understand the difference between a backend and a device though)

HGSilveri: This change is more for us, so that we can override the default phase jump time on our devices.

a-corni (approved):
LGTM

PR #783, opened 2024-12-16
Upgrade pulser-simulation to qutip 5

### LINKED ISSUE #730: Upgrade to Qutip 5
EDIT: On standby for eventual modification to the internals of `QutipEmulator` - for now, Qutip 5 is substantially worse than Qutip 4. 
The changes needed to go to Qutip 5 have been pushed to [this branch](https://github.com/pasqal-io/Pulser/tree/hs/qutip-5)

a-corni:
Bumps the requirements to qutip 5.
Handles the introduction of the new data type "Dense" in qutip (previously, there was only "CSR"): by default, all the operators used to be initialized as "CSR", but they were now stored as "Dense". This PR converts any Qobj to CSR, such that the performances are kept the same (notably, having CSR in collapse operators is very important, see https://github.com/qutip/qutip/issues/2328). 
- Here is the benchmark performed with qutip 4.7.6: [benchmark_qutip_4_7_6.pdf](https://github.com/user-attachments/files/18154270/benchmark_qutip_4_7_6.pdf)
- Here is the benchmark performed with all the QObj as CSR: 
[benchmark_qutip_5_all_CSR.pdf](https://github.com/user-attachments/files/18154298/benchmark_qutip_5_all_CSR.pdf)
- Here is the benchmark with the additional trick of https://github.com/qutip/qutip/issues/2406: 
[benchmark_qutip_5_all_CSR_no_norm.pdf](https://github.com/user-attachments/files/18154309/benchmark_qutip_5_all_CSR_no_norm.pdf)

Closes #730


pulser-simulation/pulser_simulation/hamiltonian.py
```diff
@@ -157,6 +157,7 @@ def _build_collapse_operators(
         if "eff_noise" in config.noise_types:
             for id, rate in enumerate(config.eff_noise_rates):
                 op = np.array(config.eff_noise_opers[id])
+                print(op)
```
awennersteen, line 160: leftover debug print

a-corni: Arg damn, I didn't even used it for debug 😢

pulser-simulation/pulser_simulation/hamiltonian.py
```diff
@@ -522,7 +527,8 @@ def make_interaction_term(masked: bool = False) -> qutip.Qobj:
                     return 0 * self.build_operator([("I", "global")])
 
             # make interaction term
-            dipole_interaction = cast(qutip.Qobj, 0)
+            with qutip.CoreOptions(default_dtype="CSR"):
+                dipole_interaction = cast(qutip.Qobj, 0)
```
HGSilveri, line 531: Is this doing anything?

pulser-simulation/pulser_simulation/simconfig.py
```diff
@@ -144,6 +144,10 @@ def from_noise_model(cls: Type[T], noise_model: NoiseModel) -> T:
         if "amplitude" in noise_model.noise_types:
             kwargs.setdefault("laser_waist", float("inf"))
         kwargs.pop("with_leakage", None)
+        if "eff_noise_opers" in kwargs:
+            kwargs["eff_noise_opers"] = list(
+                map(qutip.Qobj, kwargs["eff_noise_opers"])
+            )
```
HGSilveri, line 150: Here it doesn't matter if they are dense or CSR?

HGSilveri (approved):
I have a couple of question but it LGTM, very nice job!

pulser-simulation/pulser_simulation/hamiltonian.py
a-corni: Indeed, this is not doing anything, since `cast` is only used to check the typing

pulser-simulation/pulser_simulation/simconfig.py
a-corni: Normally not, I have handled the conversion in the Hamiltonian class. I let the possibility to define dense Qobj in the NoiseModel, so I think there can be dense Qobj in the SimConfig as well :)

a-corni:
In the end it seems that it is always better to have the initial state in CSR (but it's being picky for few tens of ms) [benchmark_qutip_5_operators_as_CSR_state_as_CSR.pdf](https://github.com/user-attachments/files/18166488/benchmark_qutip_5_operators_as_CSR_state_as_CSR.pdf)

awennersteen:
Based on general sparse linear algebra CSR x CSR is faster than CSR x CSR if the ket is sufficiently sparse.
IMHO it would be nice to have that user configurable

It's also the sort of thing that can depend a lot on your system

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -348,7 +348,7 @@ def set_initial_state(
                     "Incompatible shape of initial state."
                     + f"Expected {legal_shape}, got {shape}."
                 )
-            self._initial_state = qutip.Qobj(state, dims=legal_dims)
+            self._initial_state = qutip.Qobj(state, dims=legal_dims).to("CSR")
```
awennersteen, line 351: As mentioned in a standalone comment, it would be nice to be able to set this.
Also, doesn't this have to be done for the case above as well?

a-corni: About the case above, I am making the tensor product of objects that I know are CSR, so I know the output is CSR :) 

I suggest to go back to how it was before: 
- If no initial state is provided, we store the initial state in CSR (the initial state is the ground state |gg...g>, so very sparse). 
- If the user provides an initial state, then he is going to provide a numpy array in the EmulatorConfig, and the initial state will be stored as Dense. We could then have an optional argument in `run` of `QutipEmulator` and `QutipBackend` like `initial_state_type` that would be by default None (meaning, don't perform any conversion), but could be a str/a qutip.core.data type.

Yet, I am not sure how much this is relevant to care too much about, and how that would fit with the backend API @HGSilveri (notably, how that fits with the user defining the initial state in the new Config class)

awennersteen (approved):
lgtm

pulser-simulation/pulser_simulation/simulation.py
HGSilveri: The new backend API has room for this, but I definitely would leave this to a future PR

PR #784, opened 2024-12-16
Add shortcut to find out if an AbstractArray is differentiable

HGSilveri:
- Adds `AbstractArray.is_differentiable`, which checks if the stored array is a tensor with `requires_grad=True`

### DIFF
--- a/pulser-core/pulser/math/abstract_array.py
+++ b/pulser-core/pulser/math/abstract_array.py
@@ -71,6 +71,11 @@ class AbstractArray:
         """Whether the stored array is a tensor."""
         return self.has_torch() and isinstance(self._array, torch.Tensor)
 
+    @property
+    def is_differentiable(self) -> bool:
+        """Whether the stored array is a differentiable tensor."""
+        return self.is_tensor and cast(torch.Tensor, self._array).requires_grad
+
     def astype(self, dtype: DTypeLike) -> AbstractArray:
         """Casts the data type of the array contents."""
         if self.is_tensor:
@@ -271,10 +276,7 @@ class AbstractArray:
                 self._process_indices(indices)
             ] = values  # type: ignore[assignment]
         except RuntimeError as e:
-            if (
-                self.is_tensor
-                and cast(torch.Tensor, self._array).requires_grad
-            ):
+            if self.is_differentiable:
                 raise RuntimeError(
                     "Failed to modify a tensor that requires grad in place."
                 ) from e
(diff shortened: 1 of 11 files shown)


pulser-core/pulser/math/abstract_array.py
```diff
@@ -71,6 +71,11 @@ def is_tensor(self) -> bool:
         """Whether the stored array is a tensor."""
         return self.has_torch() and isinstance(self._array, torch.Tensor)
 
+    @property
+    def is_differentiable(self) -> bool:
```
a-corni, line 75: Wouldn't `requires_grad` or `has_grad` be a more generic name for this property ? I have the feeling it does not only apply to differentiability (differentiability requires_grad, but requires_grad can be used for other perspectives, right ?)

pulser-core/pulser/math/abstract_array.py
```diff
@@ -271,10 +276,7 @@ def __setitem__(self, indices: Any, values: AbstractArrayLike) -> None:
                 self._process_indices(indices)
             ] = values  # type: ignore[assignment]
         except RuntimeError as e:
-            if (
-                self.is_tensor
-                and cast(torch.Tensor, self._array).requires_grad
-            ):
+            if self.is_differentiable:
```
a-corni, line 279: Here for instance I have the feeling that self.requires_grad would be more suited...

a-corni (commented):
It really looks good to me, but I would really much prefer having this property be named `requires_grad`, because that's how we use it in all our tests. However, if having `is_differentiable` is needed to improve the user experience, I suggest we have two properties `requires_grad` and `is_differentiable`, with `is_differentiable` simply returning `requires_grad`.

pulser-core/pulser/math/abstract_array.py
HGSilveri: I used `is_differentiable` because I felt it was more explicit. `requires_grad` is clear for people familiar with torch but I find it quite opaque otherwise. 
Also, to me being differentiable = having a gradient. What do you have in mind that would fall outside of this scope?

a-corni: `is_differentiable` is definitely explicit. My only point is what this function does is what a function `requires_grad`  should do. We could imagine that at some point, `is_differentiable` can be different that `requires_grad` (might be conditions on the gradient, idk), I would prefer having two properties, one `requires_grad` and one `is_differentiable`, with `is_differentiable` currently returning the `requires_grad` property (I think that some people will still try to check the differentiability using `requires_grad`).

But it's very minor ;)

HGSilveri: Tbh I think it's overkill to bloat the source code with two ways to access the same property, at least at this stage. If you feel like `requires_grad` is better, I'll replace it

a-corni (approved):
LGTM ! Thanks :)

PR #796, opened 2025-01-31
Refactor methods in RemoteConnections and PasqalCloud

a-corni:
Some methods implemented in PasqalCloud are quite general (not specific to PasqalCloud, are checks that any connection should implement prior to submitting to a QPU) and therefore have their place directly in the RemoteConnection class.


pulser-core/pulser/backend/remote.py
```diff
@@ -294,6 +362,26 @@ def _type_check_job_params(job_params: list[JobParams] | None) -> None:
...
+            if "runs" not in j:
+                raise ValueError(
+                    "All elements of 'job_params' must specify 'runs'" + suffix
+                )
+            if max_runs is not None and j["runs"] > max_runs:
+                raise ValueError(
+                    "All 'runs' must be below the maximum allowed by the "
+                    f"device ({max_runs})" + suffix
+                )
+
```
HGSilveri, line 384: This level of validation is something we tend to enforce only for QPU submissions or when `mimic_qpu=True`. What's the rationale behind moving it here? 
My fear is that it might be measleading, ie it suggests that any job submit to a `RemoteBackend` should pass this validation.

pulser-core/pulser/backend/remote.py
```diff
@@ -221,6 +221,74 @@ def supports_open_batch(self) -> bool:
         """Flag to confirm this class can support creating an open batch."""
         pass
 
+    @staticmethod
+    def add_measurement_to_sequence(sequence: Sequence) -> Sequence:
```
HGSilveri, line 225: If you make it a public method like this, you should add a Warning in the docstring saying that it will convert all tensors to arrays (which is a very weird side-effect).
I think it would make sense to keep it private, ie
```suggestion
    def _add_measurement_to_sequence(sequence: Sequence) -> Sequence:
```

pulser-core/pulser/json/utils.py
```diff
@@ -79,6 +82,23 @@ def obj_to_dict(
...
+    class NumpyEncoder(json.JSONEncoder):
+        def default(self, o: Any) -> Any:
+            if isinstance(o, np.ndarray):
+                return o.tolist()
+            return json.JSONEncoder.default(self, o)
+
+    # Serializes with the custom encoder and then deserializes back
+    return json.loads(json.dumps(obj, cls=NumpyEncoder))
+
+
```
HGSilveri, line 101: This is a good place for it, indeed :)

pulser-core/pulser/backend/remote.py
a-corni: Yeah let's make it private

pulser-core/pulser/backend/remote.py
a-corni: In the end, I don't need to move this method, so I agree with you it can be put back in QPUBackend.

HGSilveri (approved):
LGTM! (After fixing the bug :wink: )

HGSilveri:
@a-corni Can we merge this?

a-corni:
Yes, let's merge this, it will be useful anyway.

PR #806, opened 2025-02-11
Modifications to the TOC tree

HGSilveri:
- Modifies the order of pages in the TOC
- Retires the AFM state preparation tutorial
- Slight improvements to the Backend Execution page


tutorials/advanced_features/Backends for Sequence Execution.ipynb
```diff
@@ -1,454 +1,471 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "a7601ae9",
+   "metadata": {},
+   "source": [
+    "## 1. Choosing the type of backend\n",
+    "\n",
+    "Although the backend interface nearly doesn't change between backends, some will unavoidably enforce more restrictions on the sequence being executed or require extra steps. In particular, there are two questions to answer:\n",
+    "\n",
+    "1. **Is it local or remote?** Execution on remote backends requires a working remote connection. For now, this is only available through `pulser_pasqal.PasqalCloud`.\n",
```
a-corni, line 39: [independet of this PR] Can we imagine having the possibility to list all the connections ? Just like we do for backends ?

tutorials/advanced_features/Backends for Sequence Execution.ipynb
```diff
@@ -1,454 +1,471 @@
...
+   "source": [
+    "If the appropriate packages are installed, all backends should be available via the `pulser.backends` module so we don't need to explicitly import them."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "365ed331",
+   "metadata": {},
+   "source": [
+    "Upon creation, all backends require the sequence they will execute. Emulator backends also accept, optionally, a configuration given as an instance of the `EmulationConfig` class. This class allows for setting all the parameters available in `QutipEmulator` and is forward looking, meaning that it envisions that these options will at some point be availabe on other emulator backends. This also means that trying to change parameters in the configuration of a backend that does not support them yet will raise an error.\n",
```
a-corni, line 186:
```suggestion
    "Upon creation, all backends require the sequence they will execute. Emulator backends also accept, optionally, a configuration given as an instance of the `EmulationConfig` class. This class allows for setting all the parameters available in `QutipEmulator` and is forward looking, meaning that it envisions that these options will at some point be available on other emulator backends. This also means that trying to change parameters in the configuration of a backend that does not support them yet will raise an error.\n",
```

a-corni (approved):
One typo, otherwise it looks good to me 👍

tutorials/advanced_features/Backends for Sequence Execution.ipynb
HGSilveri: Maybe... Depends on how public we want the other connections to be, I suppose

PR #808, opened 2025-02-13
Add show option to `RegisterLayout.draw()`

HGSilveri:


a-corni (approved):
Nice ! I think this can be used to solve #772 rapidly

PR #809, opened 2025-02-13
Fix drawing of interpolated waveform on DMM

HGSilveri:
The drawing of the interpolated points on a DMM channel failed because it was assuming the amplitude would always be plotted. This solution may be overkill but at least it does not make any assumptions anymore


a-corni (approved):
LGTM

PR #813, opened 2025-02-14
Improve handling of ParamObj in InterpolatedWaveform

### LINKED ISSUE #812: Improve error handling when working with Parametrized object in InterpolatedWaveform
# Code:

```
import pulser
seq = pulser.Sequence(pulser.Register.square(2, 5), device=pulser.DigitalAnalogDevice)

values = seq.declare_variable('values', size=5)
duration, *other_values = values

pulser.InterpolatedWaveform(duration, values=other_values)  # 1
pulser.InterpolatedWaveform(1000, values=values[1:])  # 2
pulser.InterpolatedWaveform(1000, values=values[1::2])  # 3
pulser.InterpolatedWaveform(1000, values=other_values)  # 4
```

# Current behaviour

`values` is a parametrized object
`other_values` is a list of parametrized objects

- `InterpolatedWaveform(duration, values=other_values)`:
     -  **works but it is not okay**
     - duration is parametrized -> we don't enter the init of the Interpolated Waveform and store all the arguments, waiting for a value to be assigned to the duration/values.
     - No matter the value assigned to duration, the building will always fail (as in expression 4)
- `InterpolatedWaveform(1000, values=values[1:])` and `InterpolatedWaveform(1000, values=values[1::2])` 
     - **work and it's okay**
     - duration is a float, but second arguments (values) is parametrized -> we don't enter the init of the Interpolated Waveform and store all the arguments, waiting for a value to be assigned to the values.
     - This is the intended behaviour, building will work. Duration could be parametrized as well, and building will always work.
- `InterpolatedWaveform(1000, values=other_values)` 
     - **doesn't work, it's normal but the error is perhaps a bit cryptic**
     - duration and values argument are non-parametrized -> we enter the init of InterpolatedWaveform and try to cast the list of values in other_values to float, but it fails because it is a list of Parametrized Objects
     - Error is `TypeError: float() argument must be a string or a real number, not 'VariableItem'`, originating from https://github.com/pasqal-io/Pulser/blob/develop/pulser-core/pulser/waveforms.py#L817. We could improve the warning and place a hint: if you want to work with a list of variable items, instantiate a variable item with a size.

# Proposal

- the error message could be more explicit for the case in which someone provides a list of Parametrized objects (see above)
- the description of InterpolatedWaveform could be more explicit to ask specifically for "either an array of elements castable to float, or a parametrized object". 
- InterpolatedWaveform(duration, values=other_values) should fail at creation, because it will always fail at building: this means reimplementing the __new__ class in InterpolatedWaveform and introduce a check on the values (either a parametrized object, or castable to float)

a-corni:
Fixes #812


a-corni:
This pattern can be reused to improve the instantiation of all the Waveforms, in order to avoid creating parametrized waveforms that would always fail at build

pulser-core/pulser/waveforms.py
```diff
@@ -804,6 +806,18 @@ class InterpolatedWaveform(Waveform):
             interpolator class.
     """
 
+    def __new__(cls, *args, **kwargs):  # type: ignore
```
HGSilveri, line 809: nit: I think that you can replicate the arguments in `__init__()` here, that way you don't have to do the tricks with args and kwargs below. Something like this should work:
```suggestion
    def __new__(
        cls,
        duration: Union[int, Parametrized],
        values: Union[ArrayLike, Parametrized],
        times: Optional[Union[ArrayLike, Parametrized]],
        *args,
        **kwargs,
    ):
```
With `__new__()` I'm never sure without testing though
If this is too much trouble, just leave it be

pulser-core/pulser/waveforms.py
```diff
@@ -865,6 +860,51 @@ def __init__(
...
+
+        if not isinstance(values, Parametrized):
+            try:
+                _values = np.array(values, dtype=float)
+            except TypeError as e:
+                raise TypeError(_err_message("values")) from e
+        if times is not None and not isinstance(times, Parametrized):
+            try:
+                times = cast(ArrayLike, times)
+                times_ = np.array(times, dtype=float)
```
HGSilveri, line 885: This will fail for torch tensors with requires_grad. You should use `AbstractArray` instead (and cover all possibilities in the tests)

tests/test_waveforms.py
```diff
@@ -266,6 +267,41 @@ def test_interpolated():
     )
     assert np.all((wf.samples >= 0).as_array())
 
+    # Init an InterpolatedWaveform that always fails at build fails
+    seq = pulser.Sequence(
+        pulser.Register.square(2, 5), device=pulser.DigitalAnalogDevice
+    )
+
+    values = seq.declare_variable("values", size=5)  # a VariableItem
```
HGSilveri, line 275:
```suggestion
    values = seq.declare_variable("values", size=5)  # a Variable
```

pulser-core/pulser/waveforms.py
HGSilveri: Nevermind, we actually don't support tensors on `InterpolatedWaveform`

HGSilveri (approved):

PR #817, opened 2025-02-20
Update programming.md

Louis-PaulHenry:
Time ordering operator missing in the integral definition of the final state


a-corni (approved):
Nice catch !

PR #825, opened 2025-03-11
Fix drawing of legends for local qubit quantities

a-corni:
We used to be saving the quantities per qubit instead of the legend of the qubit quantities in "myfile_per_qubit_legend.png". This is solved with this PR. No need to be a hotfix.


a-corni:
@victor-onofre

HGSilveri (approved):

PR #826, opened 2025-03-12
Support EmulationConfig serialization

sgrava:
PR only to serialize `EmulationConfig` and validate the serialization with the proposed schema.

To be done is separate PR
- Deserialization (str -> `EmulationConfig`) will be done in a separate PR.
- Emulation config should allow for an additional dict of backend specific options
- Operator._validate_operations()

### Changes
- `_to_abstract_repr()` on every config attribute
- `StateResult()` observable not allowed in remote backends.
- `EmulationConfig.to_abstract_repr(skip_validation=False)`
- added `pulser-core/pulser/json/abstract_repr/schemas/config-schema.json`
- `StateRepr(State)` subclass to store and serialize the arguments to make a state from repr (dict).
- `OperatorRepr(Operator)` subclass to store and serialize the arguments to make a operator from repr (dict).
- additional eigenstate validation, only ordered list (Sequence), e.g., ("r", "g") or ["0","1"] (it will probably break some test in emulators but easy to fix)

### To discuss
- vscode complains about the config-schema with:
```The schema uses meta-schema features ($dynamicRef) that are not yet supported by the validator.(769)```
Figure out why

### CODE REVIEW

**Comment by MatthieuMoreau0** [comment]
So my knowledge of this part of pulser is quite limited so I won't be able to give a proper review on the code. COuld you provide a quick example of how to use this? (e.g. serialisation on user side and deserialisation on server side)

File: pulser-core/pulser/backend/config.py [hunk 1]
```diff
@@ -256,6 +258,17 @@ def is_time_in_evaluation_times(
             np.any(np.abs(np.array(evaluation_times, dtype=float) - t) <= tol)
         )
 
+    def _to_abstract_repr(self) -> dict[str, Any]:
+        keys = self._expected_kwargs()
+        res: dict[str, Any] = {k: self.__getattr__(k) for k in keys}
+        return res
```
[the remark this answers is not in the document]
sgrava: You are right, I totally missed that serialization part in the base `BackendConfig`. I'll make the change.

pulser-core/pulser/backend/config.py
```diff
@@ -256,6 +258,17 @@ def is_time_in_evaluation_times(
...
+
+    def to_abstract_repr(self) -> str:
+        """Serializes the noise model into an abstract JSON object."""
```
HGSilveri, line 267:
```suggestion
        """Serializes the configuration into an abstract JSON object."""
```

pulser-core/pulser/backend/observable.py
```diff
@@ -99,6 +99,16 @@ def __init__(
...
 
+    def _expected_kwargs(self) -> set[str]:
+        return {"evaluation_times", "_tag_suffix"}
+
+    def _to_abstract_repr(self) -> dict[str, Any]:
+        obs_kwargs = {
+            key: self.__dict__[key] for key in self._expected_kwargs()
+        }
+        obs_tag = self._base_tag
+        return {obs_tag: obs_kwargs}
```
HGSilveri, line 110: I must say I find this unnecessarily convoluted and restrictive. I'd imagine in your schema you might be using `"tag_suffix"` instead of `"_tag_suffix"`, for instance. You could just do this:
```suggestion
    def _to_abstract_repr(self) -> dict[str, Any]:
        return {self._base_tag: {
            "evaluation_times": self.evaluation_times,
            "tag_suffix": self._tag_suffix,
        }
        }
```
This is simpler, more explicit and more flexible. For the subclasses, you can simply do e.g.
```
        repr = super()._to_abstract_repr()
        repr[self._base_tag]["one_state"]: self.one_state,
        return repr
```
I think this is not so bad

pulser-core/pulser/backend/operator.py
```diff
@@ -140,3 +140,29 @@ def from_operator_repr(
             The constructed operator.
         """
         pass
+
+
+class OperatorFromString(Operator):
+    """Operator subclass that supports serialization for remote backends."""
+
+    tag: str = "operator_from_repr"
```
HGSilveri, line 148: Is this tag just for the abstract repr?

pulser-core/pulser/backend/config.py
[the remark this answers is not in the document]
sgrava: done

pulser-core/pulser/backend/observable.py
sgrava: done

pulser-core/pulser/backend/operator.py
sgrava: Removed, the correct tag is already the "state" one

pulser-core/pulser/backend/config.py
sgrava: Changed to ```"""Serialize `EmulationConfig` to a JSON formatted str."""```. 
I think it better conveys what it does.

sgrava:
@a-quelle @HGSilveri I need some feedback on the `StateRepr(State)` class because I'm not sure if it is the best solution.
Since at this moment we don't want to send concrete states through the cloud connection, there is a need of a way to tell if a state has just been defined with its abstract representation. That is, a dict of eigenstates and amplitudes.

Alternative solutions could be:

1. `StateRepr` as a separate class, not inheriting from State. In this way, we don't need all the `NotImplementedError`  but the observables will need to accept it like `Fidelity(state: State | StateRepr,....)` and similarly the `EmulationConfig(..., initial_state: State | StateRepr, ...)`
2. We can force children of `State` to implement `_to_abstract_repr()`
3. Have in `State` class two methods: `_as_abtract_repr()` or an `__init__` to pack the dict and `_to_abstract_repr()` to unpack.

HGSilveri:
> @a-quelle @HGSilveri I need some feedback on the `StateRepr(State)` class because I'm not sure if it is the best solution. Since at this moment we don't want to send concrete states through the cloud connection, there is a need of a way to tell if a state has just been defined with its abstract representation. That is, a dict of eigenstates and amplitudes.
> 
> Alternative solutions could be:
> 
>     1. `StateRepr` as a separate class, not inheriting from State. In this way, we don't need all the `NotImplementedError`  but the observables will need to accept it like `Fidelity(state: State | StateRepr,....)` and similarly the `EmulationConfig(..., initial_state: State | StateRepr, ...)`
> 
>     2. We can force children of `State` to implement `_to_abstract_repr()`
> 
>     3. Have in `State` class two methods: `_as_abtract_repr()` or an `__init__` to pack the dict and `_to_abstract_repr()` to unpack.

Hi @sgrava , I have an alternative suggestion: how about we change `State` to internally store its "repr" when it can? Here's what I have in mind:

- By default, in `State.__init__()` we set `self._str_repr = None`.
- In `State.from_state_repr()` we finish by injecting the provided state in `_str_repr` before return.
- When serializing, we check whether `_str_repr` is None:
    - If not, we have our representation and can use it
    - If yes the state was not created with `from_state_repr()` and we raise an error informing the user that they must use it if they want to serialize it

pulser-core/pulser/backend/observable.py
```diff
@@ -99,6 +99,16 @@ def __init__(
...
 
+    def _to_abstract_repr(self) -> dict[str, Any]:
+        return {
+            self.tag: {
+                self._base_tag: {
+                    "evaluation_times": self.evaluation_times,
+                    "tag_suffix": self._tag_suffix,
+                }
+            }
+        }
```
HGSilveri, line 110: I understand why you did this but this kind of nesting is not very nice on the schema. In fact, I'd argue that an observable should be serialized as a flat object, ie
```suggestion
    def _to_abstract_repr(self) -> dict[str, Any]:
        return {
                "observable": self._base_tag,
                "evaluation_times": self.evaluation_times,
                "tag_suffix": self._tag_suffix,
        }
```

a-quelle: yes, I agree that it's preferable to just serialize the object as a flat dict, so you can deserialize it as
`Observable(**serialized_dict)`

a-quelle:
> > @a-quelle @HGSilveri I need some feedback on the `StateRepr(State)` class because I'm not sure if it is the best solution. Since at this moment we don't want to send concrete states through the cloud connection, there is a need of a way to tell if a state has just been defined with its abstract representation. That is, a dict of eigenstates and amplitudes.
> > Alternative solutions could be:
> > ```
> > 1. `StateRepr` as a separate class, not inheriting from State. In this way, we don't need all the `NotImplementedError`  but the observables will need to accept it like `Fidelity(state: State | StateRepr,....)` and similarly the `EmulationConfig(..., initial_state: State | StateRepr, ...)`
> > 
> > 2. We can force children of `State` to implement `_to_abstract_repr()`
> > 
> > 3. Have in `State` class two methods: `_as_abtract_repr()` or an `__init__` to pack the dict and `_to_abstract_repr()` to unpack.
> > ```
> 
> Hi @sgrava , I have an alternative suggestion: how about we change `State` to internally store its "repr" when it can? Here's what I have in mind:
> 
>     * By default, in `State.__init__()` we set `self._str_repr = None`.
> 
>     * In `State.from_state_repr()` we finish by injecting the provided state in `_str_repr` before return.
> 
>     * When serializing, we check whether `_str_repr` is None:
>       
>       * If not, we have our representation and can use it
>       * If yes the state was not created with `from_state_repr()` and we raise an error informing the user that they must use it if they want to serialize it

This would lead one to the situation that you can only define initial states for a remote backend by importing an emulator, and creating a state object usings it's specific subclass. We decided to have `EmulationConfig` be more lax about what arguments it takes to avoid forcing the user to import a potentially heavy emulator package, and this would make all that irrelevant.

HGSilveri:
> This would lead one to the situation that you can only define initial states for a remote backend by importing an emulator, and creating a state object usings it's specific subclass. We decided to have `EmulationConfig` be more lax about what arguments it takes to avoid forcing the user to import a potentially heavy emulator package, and this would make all that irrelevant.

I see this as a complementary solution, it would be behavior shared by all `State` subclasses. We can still have a light-weight `State` subclass that comes built in `pulser.backends` but it doesn't have to be THE one used for submission to a remote backend. When present, this state repr would also allow the backends to switch to the concrete state instance they support, allowing you to take a local submission to `QutipBackend` using `QutipState.from_abstract_repr()` and switch seamlessly to any other emulator backend, local or remote.

pulser-core/pulser/backend/config.py
```diff
@@ -256,6 +260,16 @@ def is_time_in_evaluation_times(
             np.any(np.abs(np.array(evaluation_times, dtype=float) - t) <= tol)
         )
 
+    def _to_abstract_repr(self) -> dict[str, Any]:
+        return self._backend_options
+
+    def to_abstract_repr(self, validate: bool = False) -> str:
```
HGSilveri, line 266: Normally we validate by default, I think that is preferable. Also, to mimic the signature of other cases, it would be better to use
```suggestion
    def to_abstract_repr(self, skip_validation: bool = False) -> str:
```

pulser-core/pulser/json/abstract_repr/schemas/config-schema.json
```diff
@@ -0,0 +1,410 @@
...
+    },
+    "operator": {
+        "description": "an operator, stored in a serialized format",
+        "type": "array",
+        "items": {
+            "description": "this array represents the state as a weighted sum of tensor product operators",
+            "type": "array",
+            "prefixItems": [
+                {
+                    "type": "number",
```
HGSilveri, line 144: This can be complex too

pulser-core/pulser/json/abstract_repr/schemas/config-schema.json
```diff
@@ -0,0 +1,410 @@
...
+                    "description": "tensor product operator, represented as a list of single qubit operators and the qubits upon which to apply it. Identity ops are implied",
+                    "items": {
+                        "type": "array",
+                        "description": "tuple of single qudit operator + support",
+                        "prefixItems": [
+                            {
+                                "type": "object",
+                                "description": "single qudit operator, the matrix elements are encoded using a string format",
+                                "additionalProperties": {
+                                    "type": "number",
```
HGSilveri, line 158: This can be complex too

HGSilveri (commented):
@a-quelle I'm halfway through reviewing the schema, here are some initial comments

pulser-core/pulser/json/abstract_repr/schemas/config-schema.json
```diff
@@ -0,0 +1,410 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "config-schema.json",
+    "$ref": "#/emulatorConfig",
+    "emulatorConfig": {
```
HGSilveri, line 5: nit: Might be better to call it `emulationConfig` to match the base class

pulser-core/pulser/json/abstract_repr/schemas/config-schema.json
```diff
@@ -0,0 +1,410 @@
...
+                "type": [
+                    "string",
+                    "null"
+                ]
+            }
+        },
+        "required": [
+            "observable",
+            "operator"
+        ]
```
HGSilveri, line 201: It's generally better to require all fields than to infer they are `null` when not given. This goes for all observables

HGSilveri (commented):
@a-quelle Here's the rest

HGSilveri:
@sgrava You can now merge the changes you need for the serialization

pulser-core/pulser/backend/operator.py
```diff
@@ -191,3 +191,58 @@ def _to_abstract_repr(self) -> dict[str, Any]:
             "n_qudits": self._n_qudits,
             "operations": self._operations,
         }
+
+
+class OperatorRepr(Operator):
```
HGSilveri, line 196: There's the issue that people won't know that they have to use `from_operator_repr()` to initialize. Let's at least write in the docstring, otherwise they will just be told by the error message once they try to serialize

pulser-core/pulser/backend/state.py
```diff
@@ -242,3 +242,53 @@ def _to_abstract_repr(self) -> dict[str, Any]:
             "eigenstates": tuple(self._eigenstates),
             "amplitudes": dict(self._amplitudes),
         }
+
+
+class StateRepr(State):
+    """State subclass that supports serialization for remote backends."""
```
HGSilveri, line 248: Same here, let's tell the user how to use this

pulser-core/pulser/backend/state.py
```diff
@@ -242,3 +242,53 @@ def _to_abstract_repr(self) -> dict[str, Any]:
...
+            )
+        return {
+            "eigenstates": tuple(self._eigenstates),
+            "amplitudes": dict(self._amplitudes),
+        }
+
+    @property
+    def n_qudits(self) -> int:
+        """``n_qudits`` not implemented in ``StateRepr``."""
+        raise NotImplementedError
```
HGSilveri, line 279: You can actually define this one through the amplitudes
```suggestion
        return self._validate_amplitudes(eigenstates=self._eigenstates, amplitudes=self._amplitudes)
```
Maybe you can just store this under `state._n_qudits` so you don't compute it every time

tests/test_backend_abstract_repr.py
```diff
@@ -0,0 +1,200 @@
...
+    )
+    fidelity = Fidelity(state, **kwargs)
+
+    fidelity_repr = fidelity._to_abstract_repr()
+
+    state_repr = fidelity_repr["state"]
+    assert state_repr is state
+
+
+def test_config_repr():
```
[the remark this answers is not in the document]
a-quelle: in the case that multiple types are allowed, I agree, but could you give example for when it would be relevant otherwise?

tests/test_backend_abstract_repr.py
```diff
@@ -0,0 +1,200 @@
...
+        "with_modulation": with_modulation,
+        "default_evaluation_times": default_evaluation_times,
+    }
+
+    config = EmulationConfig(**expected_kwargs)
+    # dump with AbstrctReprEncoder & validation
+    config_str = config.to_abstract_repr()
+    # load and redump but with default JSON encoder
+    # equivalent to go key by key and check single str repr
+    config_load_dump_str = json.dumps(json.loads(config_str))
```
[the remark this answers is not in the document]
a-quelle: If your serialization logic is not attuned to your deserialization logic. I'd say this is the most important test of all, assuming `config_str` is sufficiently complete, from an integration point of view.

tests/test_backend_abstract_repr.py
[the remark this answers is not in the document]
HGSilveri: No, I just care to check all the `None` or alternative types work fine, no need to check multiple values of the same type

tests/test_backend_abstract_repr.py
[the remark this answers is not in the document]
HGSilveri: But there is no custom deserialization happening here, you are just turning a JSON-formatted string into a dictionary (which will always work) and back to a string.

tests/test_backend_abstract_repr.py
```diff
@@ -0,0 +1,200 @@
...
+    EnergyVariance,
+    Fidelity,
+    Occupation,
+    StateResult,
+)
+from pulser.backend.operator import OperatorRepr
+from pulser.backend.state import StateRepr
+from pulser.json.exceptions import AbstractReprError
+
+# TODO: decide where to put these tests
```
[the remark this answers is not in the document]
HGSilveri: If you wish

pulser-core/pulser/backend/config.py
sgrava: Done

pulser-core/pulser/json/abstract_repr/schemas/config-schema.json
```diff
@@ -0,0 +1,410 @@
...
+                    }
+                ]
+            },
+            "prefer_device_noise_model": {
+                "description": "whether to prefer the noise model inferred from the sequence over the config one",
+                "type": "boolean"
+            },
+            "noise_model": {
+                "description": "the noise model to use",
+                "type": "object"
```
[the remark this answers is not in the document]
sgrava: Done. Now we point to `noise-schema.json` for the definition of the noise model and for the complex type.

pulser-core/pulser/json/abstract_repr/schemas/config-schema.json
```diff
@@ -0,0 +1,410 @@
...
+                        "g",
+                        "h",
+                        "x",
+                        "0",
+                        "1"
+                    ]
+                }
+            },
+            "amplitudes": {
+                "type": "object"
```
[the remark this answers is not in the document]
sgrava: Done.
Please, have another look. 
Maybe we can define some `{"$ref": "#/number_or_complex"}`

pulser-core/pulser/json/abstract_repr/schemas/config-schema.json
sgrava: Done, see previous comment

pulser-core/pulser/json/abstract_repr/schemas/config-schema.json
sgrava: Done, see previous comment

pulser-core/pulser/json/abstract_repr/schemas/config-schema.json
sgrava: Done.
Good point, I now require all fields.

pulser-core/pulser/json/abstract_repr/schemas/config-schema.json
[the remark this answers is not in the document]
HGSilveri: I think this is a good idea

pulser-core/pulser/json/abstract_repr/schemas/config-schema.json
```diff
@@ -125,7 +125,17 @@
...
             "amplitudes": {
-                "type": "object"
+                "type": "object",
+                "additionalProperties": {
+                    "oneOf": [
+                        {
+                            "type": "number"
+                        },
+                        {
+                            "$ref": "noise-schema.json#/definitions/ComplexNumber"
```
HGSilveri, line 135: Nice!

HGSilveri (commented):
Nice job!

pulser-core/pulser/backend/operator.py
sgrava: Done, please review the docstrings of StateRepr and OperatorRepr.

pulser-core/pulser/backend/state.py
sgrava: Done by adding  a _n_qudits attribute to StateRepr

tests/test_backend_abstract_repr.py
```diff
@@ -0,0 +1,200 @@
...
+    obs = observable(**expected_kwargs)
+    obs_repr = obs._to_abstract_repr()
+    assert isinstance(obs_repr, dict)
+
+    # test default values
+    assert obs_repr["observable"] == obs._base_tag
+    if "evaluation_times" not in expected_kwargs.keys():
+        assert obs_repr["evaluation_times"] is None
+    if "tag_suffix" not in expected_kwargs.keys():
+        assert obs_repr["tag_suffix"] is None
```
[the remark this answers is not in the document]
sgrava: Done as you suggested

tests/test_backend_abstract_repr.py
```diff
@@ -0,0 +1,200 @@
...
+    # test kwargs
+    for key, expected_value in expected_kwargs.items():
+        assert obs_repr[key] is expected_value
+
+
+def test_fidelity_repr():
+    evaluation_times = [0.1, 0.3, 0.9]
+    kwargs = {"evaluation_times": evaluation_times}
+
+    basis = {"r", "g"}
```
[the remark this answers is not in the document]
sgrava: Done, please, take a look.
I added a test in qutip_state to cover for this case.

tests/test_backend_abstract_repr.py
```diff
@@ -0,0 +1,200 @@
...
+    observables = (bitstrings, correlation)
+    with_modulation = True
+    default_evaluation_times = "Full"
+    expected_kwargs = {
+        "observables": observables,
+        "with_modulation": with_modulation,
+        "default_evaluation_times": default_evaluation_times,
+    }
+
+    config = EmulationConfig(**expected_kwargs)
```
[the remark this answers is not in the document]
sgrava: Since, `backend_options` is reserved, how should I call this additional dict? `emulation_options`?

tests/test_backend_abstract_repr.py
```diff
@@ -0,0 +1,200 @@
...
+        ),
+        (
+            EnergySecondMoment,
+            {"evaluation_times": [i * 0.1 for i in range(5)]},
+        ),
+    ],
+)
+def test_observable_repr(observable, expected_kwargs):
+    obs = observable(**expected_kwargs)
+    obs_repr = obs._to_abstract_repr()
```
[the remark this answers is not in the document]
sgrava: I test both ways now.
The only thing that is weird for me is that when testing for complex objects, you basically end up rewriting the serializer

tests/test_backend_abstract_repr.py
```diff
@@ -65,36 +71,90 @@
...
+            )
+        )
 
-    # test kwargs
-    for key, expected_value in expected_kwargs.items():
-        assert obs_repr[key] is expected_value
+    if "one_state" in obs_repr:
+        assert obs_repr["one_state"] == expected_kwargs.get("one_state", None)
+    if "num_shots" in obs_repr:
+        assert obs_repr["num_shots"] == expected_kwargs.get("num_shots", 1000)
```
HGSilveri, line 94: I tend to avoid placing `assert` within `if` blocks because you may accidentaly not ever be reaching them and not know it. In this case, you can again use `get()`
```suggestion
    assert obs_repr.get("one_state", None) == expected_kwargs.get("one_state", None)
    assert obs_repr.get("num_shots", 1000) == expected_kwargs.get("num_shots", 1000)
```

tests/test_backend_abstract_repr.py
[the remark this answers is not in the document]
sgrava: Hi @a-quelle @HGSilveri so 
- test_state_abstract_repr.py
- test_operator_abstract_repr.py
- test_config_abstract_repr.py
?
please, let me know

tests/test_backend_abstract_repr.py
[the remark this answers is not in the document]
HGSilveri: I saw, looks nice! Don't worry about that, it's normal in serialization tests

tests/test_backend_abstract_repr.py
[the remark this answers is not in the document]
HGSilveri: I think this is a bit too granular, just leave them all in this file, it's not problem

pulser-core/pulser/json/abstract_repr/schemas/config-schema.json
```diff
@@ -0,0 +1,472 @@
...
+                "description": "the noise model to use",
+                "$ref": "noise-schema.json#/definitions/NoiseModel"
+            }
+        },
+        "required": [
+            "observables",
+            "default_evaluation_times",
+            "with_modulation",
+            "prefer_device_noise_model",
+            "noise_model"
```
HGSilveri, line 105:
```suggestion
            "observables",
            "default_evaluation_times",
            "with_modulation",
            "prefer_device_noise_model",
            "noise_model",
            "interaction_matrix",
            "initial_state"
```

tests/test_backend_abstract_repr.py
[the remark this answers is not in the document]
sgrava: Sure, I'll leave it as it is

pulser-core/pulser/json/abstract_repr/schemas/config-schema.json
sgrava: done

tests/test_backend_abstract_repr.py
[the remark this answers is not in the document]
sgrava: Added, a skeleton for this test, please, take a look.
If the type in the repr is easy, just check that it is what you expect.
Otherwise check that 
```
expected = json.dumps(json.loads(expected["comlex_type_stuff"], cls=AbstractReprEncoder ))
repr["comlex_type_stuff"] == expected
```
Is this what you had in mind @HGSilveri ?

Now I only need to add some different config with mark.paametrize

pulser-core/pulser/json/abstract_repr/schemas/config-schema.json
```diff
@@ -0,0 +1,474 @@
...
+                }
+            }
+        },
+        "required": [
+            "eigenstates",
+            "n_qudits",
+            "operations"
+        ]
+    },
+    "expectation": {
```
HGSilveri, line 209: Can we make all observables not accept additional properties? I don't see any case where that would make sense 
For that, just set `additionalProperties: false` in each one
 @a-quelle do you have an objection?

tests/test_backend_abstract_repr.py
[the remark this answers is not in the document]
HGSilveri: Yes, this structure is fine, feel free to proceed

pulser-core/pulser/backend/default_observables.py
```diff
@@ -43,6 +43,14 @@ class StateResult(Observable):
     def _base_tag(self) -> str:
         return "state"
 
+    def _to_abstract_repr(self) -> dict[str, Any]:
+        raise TypeError(
```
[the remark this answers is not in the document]
sgrava: done

PR #828, opened 2025-03-13
[DevEx] Replacing ValueError with more specific errors wherever possible

### LINKED ISSUE #819: Rework errors when a register doesn't work on a device
In https://github.com/pasqal-io/quantum-evolution-kernel, we're using `Sequence` to test whether a geometry can run on a given device. Essentially, that's

```py
        try:
            pl.Sequence(register=reg, device=self.device)
        except ValueError as e:
            raise CompilationError(f"The graph is not compatible with {self.device}: {e}")
```

This works as intended, but there are two issues:

1. `ValueError` is too generic. We could accidentally be catching unrelated errors. I'd prefer a custom `PulserError` or `RegisterError`, etc.
2. Pulser logs `ERROR:...` messages on the console/notebook, despite the fact that this is actually not an error. I'd prefer if these messages were part of the exception raised.

As far as I can tell, there is no other mechanism that will let me validate a register. Could we either rework the above two points or introduce another mechanism (e.g. `Sequence.check_register(register=reg, device=device)`) that I could use for this purpose, without the two aforementioned issues?

Yoric:
This should make it easier for client libraries to catch error we raise and do something useful with them.

Closes #819

### DIFF
--- a/pulser-core/pulser/devices/_device_datacls.py
+++ b/pulser-core/pulser/devices/_device_datacls.py
@@ -515,7 +515,6 @@ class BaseDevice(ABC):
                 raise DistanceError(
                     device=self,
                     kind=kind,
-                    min=self.min_atom_distance,
                     precision_exp=COORD_PRECISION,
                     invalid=bad_qbt_pairs,
                 )
--- a/pulser-core/pulser/exceptions.py
+++ b/pulser-core/pulser/exceptions.py
@@ -181,20 +181,18 @@ class DistanceError(InvalidSequenceError):
 
     Attributes:
         kind: The kind of objects we're checking out (e.g. atoms, traps).
-        min: The minimal distance, in µm.
         precision_exp: A value P such that precision is 1e-P µm.
         pairs: A list of offending pairs.
     """
 
     kind: str
-    min: float
     precision_exp: int
     invalid: list[tuple[QubitId, QubitId]]
 
     def __str__(self) -> str:
         return (
             f"The minimal distance between {self.kind} in this device "
-            f"({self.min} µm) is not respected "
+            f"({self.device.min_atom_distance} µm) is not respected "
             f"(up to a precision of 1e{-self.precision_exp} µm) "
             f"for the pairs: {self.invalid}"
         )
@@ -206,7 +204,6 @@ class RadiusError(InvalidSequenceError):
 
     Attributes:
         kind: The kind of objects we're checking out (e.g. atoms, traps).
-        max: The maximal distance, in µm.
         invalid: A list of offending atoms (or traps, etc.)
     """
 


pulser-core/pulser/errors/__init__.py
```diff
@@ -0,0 +1,301 @@
+"""Errors raised by Pulser.
```
HGSilveri, line 1: Why use a module with just an `__init__.py` inside of a simple `errors.py` file?

pulser-core/pulser/errors/__init__.py
```diff
@@ -0,0 +1,301 @@
...
+    # separate for backwards compatibility.
+
+    # A number of dimensions.
+    DIMENSIONS = Literal[2, 3]
+    invalid: int
+
+
+@dataclass
+class DimensionChoiceError(DimensionError):
+    """An errr with the number of dimensions."""
```
HGSilveri, line 71:
```suggestion
    """An error with the number of dimensions."""
```

pulser-core/pulser/errors/__init__.py
HGSilveri: Also, I imagine we'll want to include this module in the API reference, right?

pulser-core/pulser/devices/_device_datacls.py
```diff
@@ -132,6 +147,19 @@ def __post_init__(self) -> None:
...
             param: str, type_: type, value_override: Any = None
         ) -> None:
+            """Check that one instance attribute has the right type.
+
+            Arguments:
+                param: The name of the instance attribute. The value
+                    checked is `getattr(self, param)`. Ignored
+                    if `value_override` is specified.
+                type_: The expected type.
+                value_override: If specified, instead `value_override`
```
HGSilveri, line 157:
```suggestion
                value_override: If specified, uses `value_override`
```

pulser-core/pulser/errors/__init__.py
```diff
@@ -0,0 +1,301 @@
...
+from typing import Literal, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    # Avoid circular dependencies due to type hints, part 2.
+    from pulser.devices._device_datacls import BaseDevice
+    from pulser.register.base_register import QubitId
+    from pulser.register.base_register import RegisterLayout
+
+
+class PulserError(ValueError):
```
HGSilveri, line 22: I'm not sure all `PulserError`s should be `ValueError`s of some kind. Perhaps it would be better to call it `PulserValueError`?

HGSilveri (commented):
Thanks for this @Yoric !
Changes are looking good and it's very nice to see that all existing unit tests pass with no modification. Now that that's verified though, I imagine we'll want to be more specific in these units tests and check that these custom exceptions are the ones being raised.

pulser-core/pulser/errors/__init__.py
Yoric: > Why use a module with just an `__init__.py` inside of a simple `errors.py` file?

In early prototypes, the module was spread across several submodules, so having an `__init__.py` made sense, but there's no good reason anymore. Happy to move this to an `errors.py` if you feel it's better.


> Also, I imagine we'll want to include this module in the API reference, right?

Absolutely. I'm not familiar with how the Pulser API is generated, how do I do that?

pulser-core/pulser/errors/__init__.py
Yoric: Not entirely sure I follow.

As far as I understand, we agree that all the errors raised by Pulser (at this stage, at least) should remain instances of `ValueError`, for the backwards compatibility. Do you just want me to rename this class to leave the name `PulserError` open for future errors? Happy to do that.

pulser-core/pulser/errors/__init__.py
HGSilveri: > In early prototypes, the module was spread across several submodules, so having an `__init__.py` made sense, but there's no good reason anymore. Happy to move this to an `errors.py` if you feel it's better.

Would it be okay to call it `exceptions.py` instead? I find the name less ambiguous

> Absolutely. I'm not familiar with how the Pulser API is generated, how do I do that?

In principle, adding the module [here](https://github.com/pasqal-io/Pulser/blob/f1bd30b7d8a5aadd1e06e3c4d12513584115c316/docs/source/apidoc/core.rst?plain=1#L61) should do it.
Also, I guess we'll want to export the module in the `pulser-core` too

pulser-core/pulser/errors/__init__.py
HGSilveri: Although all the errors you are replacing here are `ValueErrors`, not all the errors we raise in Pulser are `ValueError`s, so I think it would be misleading to have `PulseError` error be a `ValueError` sublclass. Renaming sounds good to me

pulser-core/pulser/errors/__init__.py
```diff
@@ -0,0 +1,301 @@
...
+    Attributes:
+        invalid: The invalid number of dimensions.
+    """
+
+    # Note: Ideally, all the DimensionXXXError error should be a single
+    # class, but since they have different error messages, we keep them
+    # separate for backwards compatibility.
+
+    # A number of dimensions.
+    DIMENSIONS = Literal[2, 3]
```
a-corni, line 65: I think `DIMENSIONS` should only be defined in one place, and not be defined in this class.
- Currently it is defined in https://github.com/pasqal-io/Pulser/blob/develop/pulser-core/pulser/devices/_device_datacls.py#L42. You might have circular imports if you try import from there
- I guess you should move the definition of DIMENSIONS in this file, and have _device_datacals import it.

pulser-core/pulser/errors/__init__.py
```diff
@@ -0,0 +1,301 @@
...
+            f"not {self.invalid}."
+        )
+
+
+@dataclass
+class DimensionTooHighError(DimensionError):
+    """An error with the number of dimensions: dimensions are too high.
+
+    Attribute:
+        max: The aximal number of dimensions
```
a-corni, line 87:
```suggestion
        max: The maximal number of dimensions
```

pulser-core/pulser/errors/__init__.py
```diff
@@ -0,0 +1,301 @@
...
+
+@dataclass
+class DimensionChoiceError(DimensionError):
+    """An errr with the number of dimensions."""
+
+    expected: list[DimensionError.DIMENSIONS]
+
+    def __str__(self) -> str:
+        return (
+            f"'dimensions' must be one of {self.expected}, "
```
a-corni, line 77: I believe self.expected should be obtained from self.device.dimensions

pulser-core/pulser/errors/__init__.py
```diff
@@ -0,0 +1,301 @@
...
+    Attribute:
+        max: The aximal number of dimensions
+    """
+
+    max: int
+
+    def __str__(self) -> str:
+        return (
+            "The device supports register layouts of at most "
+            f"{self.max} dimensions."
```
a-corni, line 95: self.max should be defined as max(self.device.dimensions)

pulser-core/pulser/errors/__init__.py
```diff
@@ -0,0 +1,301 @@
...
+@dataclass
+class TrapsNumberError(InvalidSequenceError):
+    """An error in the number of traps.
+
+    Attributes:
+        invalid: The invalid number of traps.
+        layout: The invalid layout.
+    """
+
+    invalid: int
```
a-corni, line 108: I think self.invalid is not needed since it is self.layout.number_of_traps

pulser-core/pulser/errors/__init__.py
```diff
@@ -0,0 +1,301 @@
...
+    Attributes:
+        min: The minimal number of traps for this device.
+    """
+
+    min: int
+
+    def __str__(self) -> str:
+        return (
+            "The device requires register layouts to have "
+            f"at least {self.min} traps; "
```
a-corni, line 125: self.min is self.device.min_layout_traps

pulser-core/pulser/errors/__init__.py
```diff
@@ -0,0 +1,301 @@
...
+            f"{self.layout!s} has only {self.invalid}."
```
a-corni, line 126: self.invalid is self.layout.number_of_traps

pulser-core/pulser/errors/__init__.py
```diff
@@ -0,0 +1,301 @@
...
+    Attributes:
+        max: The maximal number of traps for this device.
+    """
+
+    max: int
+
+    def __str__(self) -> str:
+        return (
+            "The device requires register layouts to have "
+            f"at most {self.max} traps; "
```
a-corni, line 143: self.max is self.device.max_number_of_traps

pulser-core/pulser/errors/__init__.py
```diff
@@ -0,0 +1,301 @@
...
+            f"{self.layout!s} has {self.invalid}."
```
a-corni, line 144: idem for invalid

pulser-core/pulser/errors/__init__.py
```diff
@@ -0,0 +1,301 @@
...
+
+    invalid: int
+    max: int
+
+    def __str__(self) -> str:
+        return (
+            f"The number of atoms ({self.invalid})"
+            " must be less than or equal to the maximum"
+            f" number of atoms supported by this device"
+            f" ({self.max})."
```
a-corni, line 187: That's self.device.max_atom_num :)

pulser-core/pulser/errors/__init__.py
```diff
@@ -0,0 +1,301 @@
...
+    Attributes:
+        kind: The kind of objects we're checking out (e.g. atoms, traps).
+        min: The minimal distance, in µm.
+        precision_exp: A value P such that precision is 1e-P µm.
+        pairs: A list of offending pairs.
+    """
+
+    kind: str
+    min: float
+    precision_exp: int
```
a-corni, line 204: That should always be COORD_PRECISION (can be imported without circular imports `from pulser.register.traps import COORD_PRECISION`, we placed it there to be consist when comparing distances through all pulser)

pulser-core/pulser/errors/__init__.py
```diff
@@ -0,0 +1,301 @@
...
+        invalid: A list of offending atoms (or traps, etc.)
+    """
+
+    kind: str
+    max: int
+    invalid: list[QubitId]
+
+    def __str__(self) -> str:
+        return (
+            f"All {self.kind} must be at most {self.max} μm "
```
a-corni, line 232:
```suggestion
            f"All {self.kind} must be at most {self.device.max_radial_distance} μm "
```

pulser-core/pulser/errors/__init__.py
```diff
@@ -0,0 +1,301 @@
...
+    min: int
+    max: int
+
+    def __str__(self) -> str:
+        return (
+            "The maximum number of layout traps "
+            f"({self.max}) must be greater than "
+            "or equal to the minimum number of layout traps "
+            f"({self.min})."
+        )
```
a-corni, line 301:
```suggestion
class MaxNumberOfTrapsError(InvalidSequenceError):
    """Invalid min/max number of traps.

    Raised whenever we attempt to
    specify a min number of traps that is larger than the max number
    of traps in a device.

    Attributes:
         device: The device that raises the error. 
    """

    def __str__(self) -> str:
        return (
            "The maximum number of layout traps "
            f"({self.device.max_layout_traps}) must be greater than "
            "or equal to the minimum number of layout traps "
            f"({self.device.min_layout_traps})."
        )
```

a-corni (commented):
Thanks for your work @Yoric ! I think you can improve your classes by using the device attribute. Perhaps you should add it to the attributes in the description ? 

Other custom errors I could think of we have in Pulser:
- related to remote connections in backends: https://github.com/pasqal-io/Pulser/blob/develop/pulser-core/pulser/backend/remote.py#L60
- related to serialization/deserialization https://github.com/pasqal-io/Pulser/blob/develop/pulser-core/pulser/json/exceptions.py

pulser-core/pulser/errors/__init__.py
```diff
@@ -0,0 +1,301 @@
...
+    kind: str
+    min: float
+    precision_exp: int
+    invalid: list[tuple[QubitId, QubitId]]
+
+    def __str__(self) -> str:
+        return (
+            f"The minimal distance between {self.kind} in this device "
+            f"({self.min} µm) is not respected "
+            f"(up to a precision of 1e{-self.precision_exp} µm) "
```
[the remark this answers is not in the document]
Yoric: I'd like to avoid hardcoding the precision within the error message.

pulser-core/pulser/errors/__init__.py
Yoric: > We already have custom exceptions to handle the serialization/deserialization errors (see https://github.com/pasqal-io/Pulser/blob/develop/pulser-core/pulser/json/exceptions.py), would it make sense to move these in this file ? These custom exceptions are ValueErrors, I agree with Henrique that it would be better to rename this class PulserValueError or something like of the kind :)

Can we maybe handle (de)serialization in a followup? I'd like to keep the size of this PR reasonable.

pulser-core/pulser/errors/__init__.py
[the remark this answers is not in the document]
a-corni: Makes sense, that's fine by me :)

Yoric:
> Thanks for this @Yoric ! Changes are looking good and it's very nice to see that all existing unit tests pass with no modification. Now that that's verified though, I imagine we'll want to be more specific in these units tests and check that these custom exceptions are the ones being raised.

Done. I might have gone a little overboard with this (see test_devices.py), how would you prefer we proceed?

HGSilveri:
> > Thanks for this @Yoric ! Changes are looking good and it's very nice to see that all existing unit tests pass with no modification. Now that that's verified though, I imagine we'll want to be more specific in these units tests and check that these custom exceptions are the ones being raised.
> 
> Done. I might have gone a little overboard with this (see test_devices.py), how would you prefer we proceed?

Hahaha it is a very fancy solution indeed but I have no problem with it :slightly_smiling_face:

pulser-core/pulser/devices/_device_datacls.py
```diff
@@ -186,13 +216,13 @@ def type_check(
                 valid = is_none or value > 0
             msg = prelude + f"'{param}' must be {comp}, not {value}."
             if not valid:
-                raise ValueError(msg)
```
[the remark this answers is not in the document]
Yoric: Fair enough.

pulser-core/pulser/devices/_device_datacls.py
```diff
@@ -386,19 +411,23 @@ def validate_register(self, register: BaseRegister) -> None:
...
-                    "The 'register' is associated with an incompatible "
-                    "register layout."
+            except (ValueError, TypeError) as e:
+                # For compatibility, we wrap all these errors as a WrapedError.
+                # If you need details, check field `wrapped`.
+                raise WrappedValueError(
+                    msg="The 'register' is associated with an incompatible "
+                    + "register layout.",
+                    wrapped=e,
                 )
```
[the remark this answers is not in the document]
Yoric: Good idea, thanks!

pulser-core/pulser/devices/_device_datacls.py
```diff
@@ -511,17 +533,20 @@ def _validate_radial_distance(
             > self.max_radial_distance
         )
         if np.any(too_far):
-            raise ValueError(
-                f"All {kind} must be at most {self.max_radial_distance} μm "
-                f"away from the center of the array, which is not the case "
-                f"for: {[ids[int(i)] for i in np.where(too_far)[0]]}"
+            assert self.max_radial_distance is not None
```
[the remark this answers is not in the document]
a-corni: Ah yes, let's go for the assert then, no need for the comment no worries :)

pulser-core/pulser/exceptions.py
```diff
@@ -0,0 +1,298 @@
...
+            f"not {self.invalid}."
+        )
+
+
+@dataclass
+class DimensionTooHighError(DimensionError):
+    """An error with the number of dimensions: dimensions are too high.
+
+    Attribute:
+        max: The maximal number of dimensions
```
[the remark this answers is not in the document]
Yoric: Removing from the documentation.

I understand what you mean by "I think it would be better to have it :)", though.

pulser-core/pulser/exceptions.py
```diff
@@ -0,0 +1,298 @@
...
+
+@dataclass
+class TrapsNumberTooLowError(TrapsNumberError):
+    """Not enough traps.
+
+    Attributes:
+        min: The minimal number of traps for this device.
+    """
+
+    min: int
```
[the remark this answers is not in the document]
Yoric: I may have gone a little bit overboard with customization :)

Yoric:
> I have one final question then: `InvalidSequenceError` is only used so far to check that a Register matches a Device. Do we agree that the idea would be to also use it to check when declaring a Channel/a DMM in a Sequence or when adding Pulses to these channels (basically, any error raised by checking an input with the value in the device of the sequence)

That's the idea, yes.

If we need to differentiate, we could introduce a intermediate classes `InvalidRegisterError` and `InvalidPulseError`, if this is necessary, but I don't see any urgent need and it feels easy to introduce as a followup without any breakage.

a-corni (commented):
Still have 3 little comments, otherwise I really like what you did with the tests !

pulser-core/pulser/exceptions.py
```diff
@@ -0,0 +1,277 @@
...
+
+@dataclass
+class DimensionPositionsTooHighError(DimensionError):
+    """An error with the number of dimensions: dimensions are too high."""
+
+    def __str__(self) -> str:
+        return (
+            f"All qubit positions must be at most {self.device.dimensions}D "
+            "vectors"
+        )
```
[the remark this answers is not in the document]
Yoric: I believe that `self.invalid` can be useful for more than `__str__`. If we have users, it's useful for them to be able to extract the offending value, even if we don't necessarily use this to display the error message.

I _think_ having `DimensionError` is marginally useful, but I can remove it if you prefer.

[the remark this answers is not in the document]
a-corni: No problem, if you use this self.invalid you can keep it. I was just fearing we would ask something no one cared about. I think like you on DimensionError, let's keep it.

a-corni (approved):
LGTM ! Thanks for the good job @Yoric

PR #829, opened 2025-03-14
Add global phase shift

### LINKED ISSUE #827: Add global phase shift
## Current situation
The current implementation of `Sequence.phase_shift` and `Sequence.phase_shift_index` require to specify the targets of the phase shift explicitly. If one wants to do a global phase shift, they need to explicitly name all qubits that are present on the register.

```python
qubits = {"q0": (0, 0), "q1": (0, 1), "q2": (1, 0), "q3": (1, 1)}
register = Register(qubits)
seq = Sequence(register, device)
seq.phase_shift(np.pi, *list(qubits.keys()))
# Or something like:
seq.phase_shift_index(np.pi, 0, 1, 2, 3)
```

## The problem
1. Since quite often users want to do a global phase shift, it is a bit  cumbersome to have to specify the targets.
2. By requiring explicit targets, the `Sequence` becomes harder to transfer to another `Register` using `switch_register`. If the `Register` doesn't use the same qubit ids or same number of qubits, the sequence cannot be transferred.

## Proposal
My proposal is to let the phase shift act as global when no targets are specified:
```python
qubits = {"q0": (0, 0), "q1": (0, 1), "q2": (1, 0), "q3": (1, 1)}
register = Register(qubits)
seq = Sequence(register, device)
seq.phase_shift(np.pi)
# Or:
seq.phase_shift_index(np.pi)
```

pimvenderbosch:
This proposes to change `Sequence.phase_shift` to accept not setting a specific target, so that the phase shift is applied to all qubits in the `Register`.
I noticed that it wouldn't raise an error if no targets were supplied in the current version. Instead, it just wouldn't have any effect as no targets have been selected. This behavior changes if this PR is merged. I don't foresee backwards compatibility issues because of it, because the function shouldn't have been used if no targets were supplied.

This PR Closes #827

### DIFF
--- a/pulser-core/pulser/json/abstract_repr/serializer.py
+++ b/pulser-core/pulser/json/abstract_repr/serializer.py
@@ -169,14 +169,12 @@ def serialize_abstract_sequence(
 
     def unfold_targets(
         target_ids: QubitId | Collection[QubitId],
-    ) -> QubitId | list[QubitId] | None:
+    ) -> QubitId | list[QubitId]:
         if isinstance(target_ids, (int, str)):
             return target_ids
 
         targets = list(cast(Collection, target_ids))
-        if len(targets) == 0:
-            return []
-        return targets if len(targets) > 1 else targets[0]
+        return targets if len(targets) != 1 else targets[0]
 
     def convert_targets(
         target_ids: Union[QubitId, Collection[QubitId]],
--- a/pulser-core/pulser/json/abstract_repr/serializer.py
+++ b/pulser-core/pulser/json/abstract_repr/serializer.py
@@ -169,11 +169,13 @@ def serialize_abstract_sequence(
 
     def unfold_targets(
         target_ids: QubitId | Collection[QubitId],
-    ) -> QubitId | list[QubitId]:
+    ) -> QubitId | list[QubitId] | None:
         if isinstance(target_ids, (int, str)):
             return target_ids
 
         targets = list(cast(Collection, target_ids))
+        if len(targets) == 0:
+            return []
         return targets if len(targets) > 1 else targets[0]
 
     def convert_targets(
--- a/pulser-core/pulser/sequence/sequence.py
+++ b/pulser-core/pulser/sequence/sequence.py
@@ -2172,6 +2172,14 @@ class Sequence(Generic[DeviceType]):
             )
 
         if not specific_targets:
+            warnings.warn(
+                "In version v1.4.0 the behavior of `Sequence.phase_shift` and "
+                "`Sequence.phase_shift_index` changed when called without "
+                "specifying targets. In previous versions calling without "
+                "targets wouldn't add a phase shift to any qubit, whereas in "
+                "versions v1.4.0 and up a phase shift will be added to all "
+                "qubits in the register if no specific targets are given."
+            )
             specific_targets = self._register.qubit_ids
             _index = False
 
--- a/pulser-core/pulser/sequence/sequence.py
+++ b/pulser-core/pulser/sequence/sequence.py
@@ -2178,7 +2178,8 @@ class Sequence(Generic[DeviceType]):
                 "specifying targets. In previous versions calling without "
                 "targets wouldn't add a phase shift to any qubit, whereas in "
                 "versions v1.4.0 and up a phase shift will be added to all "
-                "qubits in the register if no specific targets are given."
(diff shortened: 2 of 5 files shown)


HGSilveri:
> I noticed that it wouldn't raise an error if no targets were supplied in the current version. Instead, it just wouldn't have any effect as no targets have been selected. This behavior changes if this PR is merged. I don't foresee backwards compatibility issues because of it, because the function shouldn't have been used if no targets were supplied.

This is slightly concerning. I agree that it's unlikely someone is currently applying `phase_shift()` without any targets, but it's not impossible...

HGSilveri (commented):
These changes look good. Backwards compatibility concerns aside, I'd also like to see some tests on the serialization and deserialization to the abstract representation (in `test_abstract_repr.py`)

pimvenderbosch:
> > I noticed that it wouldn't raise an error if no targets were supplied in the current version. Instead, it just wouldn't have any effect as no targets have been selected. This behavior changes if this PR is merged. I don't foresee backwards compatibility issues because of it, because the function shouldn't have been used if no targets were supplied.
> 
> This is slightly concerning. I agree that it's unlikely someone is currently applying `phase_shift()` without any targets, but it's not impossible...

We could add a warning for a few versions, so that affected users get alerted?

pulser-core/pulser/sequence/sequence.py
```diff
@@ -2172,6 +2172,14 @@ def _phase_shift(
...
 
         if not specific_targets:
+            warnings.warn(
+                "In version v1.4.0 the behavior of `Sequence.phase_shift` and "
+                "`Sequence.phase_shift_index` changed when called without "
+                "specifying targets. In previous versions calling without "
+                "targets wouldn't add a phase shift to any qubit, whereas in "
+                "versions v1.4.0 and up a phase shift will be added to all "
+                "qubits in the register if no specific targets are given."
+            )
```
HGSilveri, line 2182: Let's specify the `stacklevel` (which I think is 3 in this case)

tests/test_abstract_repr.py
```diff
@@ -642,7 +643,12 @@ def sequence(self, request):
 
         seq.align("digital", "rydberg")
         seq.add(pi_pulse, "rydberg")
-        seq.phase_shift(1.0, basis="ground-rydberg")
+        with warnings.catch_warnings():
```
HGSilveri, line 646: You should use `pytest.warns()` instead

pulser-core/pulser/json/abstract_repr/serializer.py
```diff
@@ -169,11 +169,13 @@ def serialize_abstract_sequence(
 
     def unfold_targets(
         target_ids: QubitId | Collection[QubitId],
-    ) -> QubitId | list[QubitId]:
+    ) -> QubitId | list[QubitId] | None:
```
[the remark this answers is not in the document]
pimvenderbosch: Oh oops, forgot to delete after I fixed it in another way.

tests/test_abstract_repr.py
pimvenderbosch: I changed it at most locations. Couldn't do it for two of them. Because they should conditionally warn.

pulser-core/pulser/sequence/sequence.py
pimvenderbosch: Shouldn't it be 2, so that the warning comes at the level of `Sequence.phase_shift`?

pimvenderbosch: Or you want it on a line in which `phase_shift` is called. Guess that is level 3 then indeed.

HGSilveri (approved):
LGTM!

tests/test_abstract_repr.py
```diff
@@ -1782,8 +1796,14 @@ def test_deserialize_non_parametrized_op(self, op):
...
         )
-        _check_roundtrip(s)
-        seq = Sequence.from_abstract_repr(json.dumps(s))
+        with warnings.catch_warnings():
+            if op["op"] == "phase_shift":
+                warnings.filterwarnings(
+                    "ignore",
+                    "In version v1.4.0 the behavior of `Sequence.phase_shift`",
+                )
+            _check_roundtrip(s)
```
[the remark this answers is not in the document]
pimvenderbosch: Good suggestion, added this in my latest commit.

PR #832, opened 2025-03-18
Support State and Operator serialization

HGSilveri:
- [x] Keep track of the state amplitudes and use it to serialize `State` to the abstract representation
- [x] Do the same for `Operator`


pulser-simulation/pulser_simulation/qutip_state.py
```diff
@@ -45,11 +45,15 @@ class QutipState(State[SupportsComplex, complex]):
...
+        *,
+        eigenstates: Sequence[Eigenstate],
+        **kwargs: Any,
     ):
         """Initializes a QutipState."""
         self._validate_eigenstates(eigenstates)
-        self._eigenstates = tuple(eigenstates)
+        super().__init__(eigenstates=eigenstates, **kwargs)
         valid_types = ("ket", "bra", "oper")
         if not isinstance(state, qutip.Qobj) or state.type not in valid_types:
```
sgrava, line 58: Nice! 
This is the part that was worrying me.
It makes the _init_ of subclasses more complicated, right?
The same could be achieved with a couple of lines more in the `from_state_amplitudes` to append the eigenstates and the amplitudes in the class. And maybe also a flag for convenience `am_i_from_abstract_repr`.

This way you can still keep the `_to_abstract_repr` in the base class, which I really like.

HGSilveri: Yes, I'm basically just hiding everything under `**kwargs`. 
As you said, injecting these before returning in `from_state_amplitudes` would work too, that is what I originally had in mind actually. I would skip the flag though.
Let's see what others think, I'm fine with both options.

sgrava: Yeah I am also fine with both!

pulser-core/pulser/backend/state.py
```diff
@@ -172,3 +184,37 @@ def _validate_eigenstates(eigenstates: Sequence[Eigenstate]) -> None:
...
+        n_qudits = len(basis_states[0])
+        if not all(
+            len(bs) == n_qudits and set(bs) <= set(eigenstates)
+            for bs in basis_states
+        ):
+            raise ValueError(
+                "All basis states must be combinations of eigenstates with the"
+                f" same length. Expected combinations of {eigenstates}, each "
+                f"with {n_qudits} elements."
+            )
```
sgrava, line 207: Just curiosity here, how can you get away with the 100% coverage without catching this `ValueError` in a test?

HGSilveri: This is just code I moved from `QutipState` so it was already being tested

pulser-simulation/pulser_simulation/qutip_state.py
a-quelle: I would personally leave out the `__init__` and just hard initialize the fields to `None`. Then the `from_state_amplitudes` and equivalent from `Operator` can just set these fields.

a-quelle (commented):
I don't feel entirely capable to judge the changes in pulser simulation, but the base classes look good.

pulser-core/pulser/json/abstract_repr/serializer.py
```diff
@@ -49,6 +49,9 @@ def default(self, o: Any) -> dict[str, Any] | list | int:
         elif isinstance(o, set):
             return list(o)
         elif isinstance(o, complex):
+            if not o.imag:
+                # Try to return a real number when possible
+                return o.real
```
awennersteen, line 54: isn't this the wrong way around?

HGSilveri: How so @awennersteen ? If there is no imaginary part, we return just the real part

pulser-core/pulser/backend/operator.py
```diff
@@ -140,3 +151,16 @@ def from_operator_repr(
             The constructed operator.
         """
         pass
```
pablolh, line 153: Does it make sense to store the `self._eigenstates = eigenstates` etc... here and in the subclasses call `super().from_operator_repr(...)`?

Otherwise, only looking at this superclass, the error message below can be a bit confusing since it actually refers to subclass code storing `self._eigenstates`.

pulser-core/pulser/json/abstract_repr/serializer.py
awennersteen: I think I added a edit at the same time as you commented

HGSilveri: If you insist I can be explicit but `o.imag == 0` is the only case that will enter this `if`

pulser-core/pulser/backend/operator.py
```diff
@@ -140,3 +151,16 @@ def from_operator_repr(
...
+
+    def _to_abstract_repr(self) -> dict[str, Any]:
+        if not self._eigenstates or not self._n_qudits or not self._operations:
```
pablolh, line 156: I would choose the safer `self._eigenstates is None` etc given that any python value implicitely converts to bool

pulser-core/pulser/json/abstract_repr/serializer.py
awennersteen: I don't insist , but thanks

It really confused me as I at first assumed you wanted to always throw away the imaginary part

pablolh (commented):
I find this pretty cool way to serialize useful states regardless of implementation (mps, state vector, ...)!
However it's also relatively unsafe and relies on people implementing subclasses being well aware that they should not provide any way to modify a state or operator after its construction (e.g. in-place mul of an operator by a scalar).
The world (Python) is unsafe anyway...

pulser-core/pulser/backend/operator.py
HGSilveri: I'm avoiding this because 
1. This is a classmethod, which makes inheritance trickier (I believe just using `super()` won't work)
2. Because it is a classmethod, you don't really have access to the instance, just the class, so you can't do `self._eigenstates = ... ` here
3. It is an abstractmethod, meaning that would either force others to handle said inheritance or just overwrite the method altogether

> Otherwise, only looking at this superclass, the error message below can be a bit confusing since it actually refers to subclass code storing `self._eigenstates`.

The error message itself does not refer to `self._eigenstates` (or any of the other missing attributes) on purpose. It just points the user to using `from_operator_repr()` if they need to serialize the object.

pulser-core/pulser/backend/state.py
```diff
@@ -148,6 +147,28 @@ def from_state_amplitudes(
...
+        eigenstates: Sequence[Eigenstate],
+        amplitudes: Mapping[str, ArgScalarType],
+    ) -> tuple[
+        StateType, Sequence[Eigenstate], Mapping[str, ReturnScalarType]
+    ]:
+        """Implements the logic used in `from_state_amplitudes()`.
+
+        Expected to return the State instance alongside the eigenstates
+        and amplitudes used in serialization.
+        """
```
HGSilveri, line 171: Here's a way to make sure subclasses inject the necessary serialization information. 

Wdyt @pablolh , @a-quelle , @sgrava ? Should I do the same for `Operator` or just revert this?

HGSilveri:
> However it's also relatively unsafe and relies on people implementing subclasses being well aware that they should not provide any way to modify a state or operator after its construction (e.g. in-place mul of an operator by a scalar).

Ok, I actually see two ways around this. I'm not sure we want to go there, but I'll lay them out just in case:

1. Make `State` and `Operator` frozen dataclasses
2. Require `__eq__` to be an `abstractmethod` so we can check that the current `State/Operator` matches what we reconstruct from the stashed serialization information before we return in `_to_abstract_repr()`.

a-quelle:
> > However it's also relatively unsafe and relies on people implementing subclasses being well aware that they should not provide any way to modify a state or operator after its construction (e.g. in-place mul of an operator by a scalar).
> 
> I agree. But these are the same people that will implement `from_operator_repr()` and `from_state_amplitudes()` and inject the serialization information there. Let's hope they connect the dots

In emu_mps we had an issue where we evolved the initial state in-place, and people wanting to reuse it over many runs got very weird results. The fix we implemented for that would solve this issue, but it reminds me that emu-sv currently still suffers from this problem.

pulser-core/pulser/backend/state.py
a-quelle: I kinda like this

HGSilveri:
> > > However it's also relatively unsafe and relies on people implementing subclasses being well aware that they should not provide any way to modify a state or operator after its construction (e.g. in-place mul of an operator by a scalar).
> > 
> > 
> > I agree. But these are the same people that will implement `from_operator_repr()` and `from_state_amplitudes()` and inject the serialization information there. Let's hope they connect the dots
> 
> In emu_mps we had an issue where we evolved the initial state in-place, and people wanting to reuse it over many runs got very weird results. The fix we implemented for that would solve this issue, but it reminds me that emu-sv currently still suffers from this problem.

@a-quelle I don't think `State` and `Operator` should be changed in place. Do you have a compelling reason to do so? Why not just make new instances?
The fact that you are doing this suggests we need to forbid this at the base class level.

a-quelle:
> > > > However it's also relatively unsafe and relies on people implementing subclasses being well aware that they should not provide any way to modify a state or operator after its construction (e.g. in-place mul of an operator by a scalar).
> > > 
> > > 
> > > I agree. But these are the same people that will implement `from_operator_repr()` and `from_state_amplitudes()` and inject the serialization information there. Let's hope they connect the dots
> > 
> > 
> > In emu_mps we had an issue where we evolved the initial state in-place, and people wanting to reuse it over many runs got very weird results. The fix we implemented for that would solve this issue, but it reminds me that emu-sv currently still suffers from this problem.
> 
> @a-quelle I don't think `State` and `Operator` should be changed in place. Do you have a compelling reason to do so? Why not just make new instances? The fact that you are doing this suggests we need to forbid this at the base class level.

This is precisely the solution we used, to deepcopy the input when it's still negligible in memory, and modify in-place from there. But having the program internally recognise that it's still a `State` internally is useful, so I'd rather not have this be completely forbidden on the class.

HGSilveri:
> Ok, I actually see two ways around this. I'm not sure we want to go there, but I'll lay them out just in case:
> 
>     1. Make `State` and `Operator` frozen dataclasses
> 
>     2. Require `__eq__` to be an `abstractmethod` so we can check that the current `State/Operator` matches what we reconstruct from the stashed serialization information before we return in `_to_abstract_repr()`.

@a-quelle I sounds to me like from these two solutions, `2.` would be preferable.

pulser-core/pulser/backend/state.py
sgrava: Nice!

HGSilveri: Ok, I'll do it for `Operator` too

pulser-core/pulser/backend/operator.py
pablolh: Indeed I spoke too fast regarding the inheritance and classmethod.

regarding the error message I meant the code itself, you do check for `if not self._eigenstates` which is set in the subclasses.

A way around the problem could be to make different methods for the base class `from_operator_repr` and the ones from the subclasses, for instance something like:
```
from abc import abstractmethod

class State:
    @staticmethod
    @abstractmethod
    def _from_state_string(eigenstates, n_qubits, operations):
        pass


    @classmethod
    def from_state_string(cls, eigenstates, n_qubits, operations):
        result = cls._from_state_string(eigenstates=eigenstates, n_qubits=n_qubits, operations=operations)
        result._eigenstates = eigenstates
        result._n_qubits = n_qubits
        result._operations = operations
        return result

class QutipState(State):
    def __init__(self, arbitrary):
        print("making qutip state")
        self.type = "QutipState"

    @staticmethod
    def _from_state_string(eigenstates, n_qubits, operations):
        return QutipState(eigenstates + n_qubits + operations)

state = QutipState.from_state_string(1, 2, 3)
print(state.type)
print(state._eigenstates)
```

pablolh: Oops I think you did exactly that, my bad should read properly

a-quelle (approved):
lgtm

PR #833, opened 2025-03-21
Add (de)serializer to Results

a-quelle:
- Update the AbstractReprEncoder to support `torch.Tensor` if `torch` exists
- add `to_abstract_repr` and `from_abstract_repr` to `Results`
- tests

### DIFF
--- a/pulser-core/pulser/backend/results.py
+++ b/pulser-core/pulser/backend/results.py
@@ -13,12 +13,15 @@
 # limitations under the License.
 """Defines the base class for storing backend results."""
 from __future__ import annotations
+import json
+from pulser.json.abstract_repr.serializer import AbstractReprEncoder
 
 import collections.abc
 import typing
 import uuid
 from dataclasses import dataclass, field
 from typing import Any, TypeVar, overload
+from copy import deepcopy
 
 from pulser.backend.observable import Observable
 
@@ -134,7 +137,41 @@ class Results:
                 f"{observable!r} is not an Observable instance "
                 "nor a known observable tag in the results."
             )
+        
+    def _to_abstract_repr(self) -> dict:
+        d = deepcopy(self.__dict__)
+        for key, value in d["_tagmap"].items():
+            d["_tagmap"][key] = str(value)
+        tmp = {}
+        for key, value in d["_results"].items():
+            tmp[str(key)] = value
+        d["_results"] = tmp
+        tmp = {}
+        for key, value in d["_times"].items():
+            tmp[str(key)] = value
+        d["_times"] = tmp
+        return d
 
+    @classmethod
+    def _from_abstract_repr(cls, dict: dict) -> Results:
+        results = cls(
+            atom_order=tuple(dict["atom_order"]), total_duration=dict["total_duration"]
+        )
+        for key, value in dict["_tagmap"].items():
+            results._tagmap[key] = uuid.UUID(value)
+        for key, value in dict["_results"].items():
+            results._results[uuid.UUID(key)] = value
+        for key, value in dict["_times"].items():
+            results._times[uuid.UUID(key)] = value
+        return results
+
+    def to_abstract_repr(self) -> str:
+        return json.dumps(self._to_abstract_repr(), cls=AbstractReprEncoder)
+
+    @classmethod
+    def from_abstract_repr(cls, repr: str) -> Results:
+        d = json.loads(repr)
+        return cls._from_abstract_repr(d)
 
 ResultsType = TypeVar("ResultsType", bound=Results)
 
--- a/pulser-core/pulser/json/abstract_repr/serializer.py
+++ b/pulser-core/pulser/json/abstract_repr/serializer.py
@@ -34,6 +34,13 @@ if TYPE_CHECKING:
     from pulser.sequence import Sequence
     from pulser.sequence._call import _Call
 
+has_torch: bool
+try:
+    import torch
+    has_torch = True
+except ImportError: # pragma: no cover
+    has_torch = False
+
 
 class AbstractReprEncoder(json.JSONEncoder):
     """The custom encoder for abstract representation of Pulser objects."""
@@ -50,6 +57,11 @@ class AbstractReprEncoder(json.JSONEncoder):
             return list(o)
         elif isinstance(o, complex):
             return dict(real=o.real, imag=o.imag)
+        elif has_torch and isinstance(o, torch.Tensor):
+            if len(o.shape) == 0:
+                return o.item()
+            else:
+                return cast(list, o.tolist())
         else:  # pragma: no cover
             return cast(dict, json.JSONEncoder.default(self, o))
 
--- a/tests/test_abstract_repr.py
+++ b/tests/test_abstract_repr.py
@@ -88,6 +88,45 @@ phys_Chadoq2 = replace(
 )
 
 
+@pytest.mark.parametrize(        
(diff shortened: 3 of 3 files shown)


pulser-core/pulser/json/abstract_repr/serializer.py
```diff
@@ -53,6 +61,11 @@ def default(self, o: Any) -> dict[str, Any] | list | int | float:
                 # Try to return a real number when possible
                 return o.real
             return dict(real=o.real, imag=o.imag)
+        elif has_torch and isinstance(o, torch.Tensor):
```
HGSilveri, line 64:
```suggestion
        elif pm.AbstractArray.has_torch() and isinstance(o, pm.torch.Tensor):
```

pulser-core/pulser/json/abstract_repr/serializer.py
```diff
@@ -53,6 +61,11 @@ def default(self, o: Any) -> dict[str, Any] | list | int | float:
...
+            if len(o.shape) == 0:
+                return o.item()
+            else:
+                return o.tolist()
```
HGSilveri, line 68: This is not needed, `tolist()` on a 0D tensor returns a scalar
```suggestion
            return o.tolist()
```

tests/test_abstract_repr.py
```diff
@@ -89,6 +89,43 @@
...
 
 
+@pytest.mark.parametrize(
+    "test_torch",
+    [True, False],
+)
+def test_abstract_repr_encoder(test_torch: bool):
+    encoder = AbstractReprEncoder()
+
+    if not test_torch:
```
HGSilveri, line 99: You can avoid having everything in an if-else block by starting with 
```
if test_torch:
    torch = pytest.importorskip("torch")
```

pulser-core/pulser/json/abstract_repr/serializer.py
a-quelle: I see that indeed it does. But that is hideously misleading.

tests/test_abstract_repr.py
a-quelle: I don't see what you mean here. The way I set it up, it will test the non-torch part in the one path, and the torch path, if available, in the other. what you suggests seems to be an all-or-nothing approach. Am I missing something?

HGSilveri: What I suggest is that you do the tests that do not require `torch` regardless

tests/test_abstract_repr.py
```diff
@@ -89,6 +89,43 @@
...
+        result = encoder.default(1.0 + 2.0j)
+        assert result == dict(real=1.0, imag=2.0)
+    else:
+        torch = pytest.importorskip("torch")
+
+        result = encoder.default(torch.tensor(5.0))
+        assert result == 5.0
+
+        result = encoder.default(torch.tensor([1, 2, 3, 4]))
+        assert result == [1, 2, 3, 4]
```
HGSilveri, line 126:
```suggestion
    class Dummy:
        def _to_abstract_repr(self):
            return "_to_abstract_repr"

    result = encoder.default(Dummy())
    assert result == "_to_abstract_repr"

    result = encoder.default(np.array([1, 2, 3, 4]))
    assert result == [1, 2, 3, 4]

    result = encoder.default(np.intp(5))
    assert result == 5

    result = encoder.default({1, 2, 3, 4})
    assert result == [1, 2, 3, 4]

    result = encoder.default(1.0 + 2.0j)
    assert result == dict(real=1.0, imag=2.0)
    if test_torch:
        torch = pytest.importorskip("torch")

        result = encoder.default(torch.tensor(5.0))
        assert result == 5.0

        result = encoder.default(torch.tensor([1, 2, 3, 4]))
        assert result == [1, 2, 3, 4]
```

pulser-core/pulser/backend/results.py
```diff
@@ -142,6 +145,37 @@ def _find_uuid(self, observable: Observable | str) -> uuid.UUID:
                 "nor a known observable tag in the results."
             )
 
+    def _to_abstract_repr(self) -> dict:
+        d = deepcopy(self.__dict__)
```
HGSilveri, line 149: This is a bit hacky and we might get all sorts of unwanted stuff. How about just this?
```suggestion
        d = {"atom_order": self.atom_order, "total_duration": self.total_duration}
```

pulser-core/pulser/backend/results.py
```diff
@@ -142,6 +145,37 @@ def _find_uuid(self, observable: Observable | str) -> uuid.UUID:
...
+        d["_tagmap"] = {key: str(value) for key, value in d["_tagmap"].items()}
+        d["_results"] = {
+            str(key): value for key, value in d["_results"].items()
+        }
+        d["_times"] = {str(key): value for key, value in d["_times"].items()}
```
HGSilveri, line 154: nit: It seems like your schema will have fields `"_tagmap"`, `"_results"`, ... I think the `_` is unnecessary in a schema

pulser-core/pulser/backend/results.py
```diff
@@ -135,6 +138,42 @@ def _find_uuid(self, observable: Observable | str) -> uuid.UUID:
...
+            results._results[uuid.UUID(key)] = value
+        for key, value in dict["_times"].items():
+            results._times[uuid.UUID(key)] = value
+        return results
+
+    def to_abstract_repr(self) -> str:
+        return json.dumps(self._to_abstract_repr(), cls=AbstractReprEncoder)
+
+    @classmethod
+    def from_abstract_repr(cls, repr: str) -> Results:
```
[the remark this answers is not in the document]
a-quelle: I've added this

pulser-core/pulser/backend/results.py
```diff
@@ -171,10 +172,28 @@ def _from_abstract_repr(cls, dict: dict) -> Results:
...
+        Numpy arrays and torch Tensors are converted into lists,
+        and their original class is lost forever.
+
+        Returns:
+            The json string
+        """
+        abstr_str = json.dumps(
+            self._to_abstract_repr(), cls=AbstractReprEncoder
+        )
+        validate_abstract_repr(abstr_str, "results")
```
HGSilveri, line 186: For this validation, can you make it optional? Like Stefano did in EmulationConfig with `skip_validation`

pulser-core/pulser/backend/results.py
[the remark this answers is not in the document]
HGSilveri: You did in at serialization, which we want (optionally), but here I was suggesting to do it at deserialization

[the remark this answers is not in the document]
a-quelle: Yes, you're right of course. I'm adding it there, and making the other one optional.

tests/test_abstract_repr.py
```diff
@@ -89,6 +96,34 @@
 )
 
 
+def test_abstract_repr_encoder_non_torch():
```
HGSilveri, line 99: You can move all these tests to `test_backend_abstract_repr.py` now

a-quelle: even the one for the base encoder? The torch part I can see, but why the non-torch part? It's not backend specific.

HGSilveri: You're right, I meant just the tests for `Results`

HGSilveri (approved):
LGTM, thanks @a-quelle !

PR #834, opened 2025-03-21
Remove pulser-pasqal from repo

### LINKED ISSUE #820: Remove pulser-pasqal
Now that pulser-pasqal has been moved to `pasqal-cloud`, we can remove it from this repo.

HGSilveri:
Closes #820 .


pimvenderbosch:
Thanks, and looks good. I do notice that there are a few places where `pulser_pasqal` or `pulser-pasqal` are still mentioned. I think those mentions make sense for documentation purposes, but might be good to verify.

- in the API reference the file `pasqal.rst`: Pulls the API ref from the external `pulser_pasqal` package
- `backends.py`: Lists the available backends and mentions their source.
- backends for sequence execution notebook: Documentation purposes.
- in `setup.py` as a requirement: is this intended to be a requirement? Shouldn't it be optional?

If these are all intended, feel free to go ahead and merge.

pimvenderbosch (approved):

HGSilveri:
> Thanks, and looks good. I do notice that there are a few places where `pulser_pasqal` or `pulser-pasqal` are still mentioned. I think those mentions make sense for documentation purposes, but might be good to verify.
> 
>     * in the API reference the file `pasqal.rst`: Pulls the API ref from the external `pulser_pasqal` package
> 
>     * `backends.py`: Lists the available backends and mentions their source.

This is intended, the point of that module is to allow you to import all backends from the same place. 

> 
>     * backends for sequence execution notebook: Documentation purposes.
> 
>     * in `setup.py` as a requirement: is this intended to be a requirement? Shouldn't it be optional?

`pulser-pasqal` is included as part of the `pulser` metapackage, so that when you `pip install pulser` you get not only `pulser-core` but also `pulser-simulation` and `pulser-pasqal`. At some point, turning these extensions into optional dependencies is worth considering.

PR #835, opened 2025-03-24
[DevEx] Rework (de)serialization exceptions

### LINKED ISSUE #831: Consider reworking the errors raised by (de)serialization
As a followup to #819 / #828 , it might be a good idea to consider also making the errors raised by (de)serialization a bit higher-level.

Yoric:
Resolves #831. In #828, we reworked exceptions during the constructions of sequences. Let's now apply this design to (de)serialization.

### DIFF
--- a/pulser-core/pulser/json/exceptions.py
+++ b/pulser-core/pulser/json/exceptions.py
@@ -11,26 +11,26 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Custom exceptions for serialization errors."""
+"""Custom exceptions for serialization errors.
 
+Note: This module has migrated to pulser.exceptions.serialization. Please
+use the new path for any future work.
+"""
 
-class SerializationError(Exception):
-    """Exception raised when sequence serialization fails."""
+import warnings
 
-    pass
+from pulser.exceptions import serialization
 
+warnings.warn(
+    "module pulser.json.exceptions is deprecated, "
+    "please migrate your code to "
+    "use pulser.exceptions.serialization",
+    category=DeprecationWarning,
+    stacklevel=2,
+)
 
-class AbstractReprError(Exception):
-    """Exception raised for abstract representation errors.
-
-    Raised when an error occurs during the serialization to or deserialization
-    from the abstract representation.
-    """
-
-    pass
-
-
-class DeserializeDeviceError(Exception):
-    """Exception raised when device deserialization fails."""
-
-    pass
+# For backwards compatibility, re-export the exceptions
+# from their new module `pulser.exceptions`.
+SerializationError = serialization.SerializationError
+AbstractReprError = serialization.AbstractReprError
+DeserializeDeviceError = serialization.DeserializeDeviceError
--- a/tests/test_exceptions.py
+++ b/tests/test_exceptions.py
@@ -1,0 +1,17 @@
+"""Trivial tests on our exceptions."""
+
+import pytest
+
+
+def test_import_old_serialize_exceptions():
+    """Test importing from pulser.json.exceptions.
+
+    This should trigger a warning.
+    """
+    with pytest.warns(
+        expected_warning=DeprecationWarning,
+        match="module pulser.json.exceptions is deprecated",
+    ):
+        from pulser.json.exceptions import SerializationError
+
+        _ = SerializationError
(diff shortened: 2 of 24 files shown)


a-corni (commented):
That's indeed something like this I had in mind :) 
Eventually, we could raise a DeprecationWarning when calling pulser-core/pulser/json/exceptions.py -> even though that would be painful for us @HGSilveri

HGSilveri:
> That's indeed something like this I had in mind :)
> Eventually, we could raise a DeprecationWarning when calling pulser-core/pulser/json/exceptions.py -> even though that would be painful for us @HGSilveri

It shouldn't be painful, we'd just have to change the imports of these errors to point to their new location. I think we should do this

pulser-core/pulser/json/exceptions.py
```diff
@@ -11,26 +11,19 @@
...
-
-
-class DeserializeDeviceError(Exception):
-    """Exception raised when device deserialization fails."""
-
-    pass
+"""Custom exceptions for serialization errors.
+
+Note: This module has migrated to pulser.exceptions.serialization. Please
+use the new path for any future work.
```
a-corni, line 17: I would suggest to make this a DeprecationWarning, it should be raised each time someone is trying to import this module.

That will trigger some errors in the test, and will help you replace the tests that are importing this module.

Yoric:
> > That's indeed something like this I had in mind :)
> > Eventually, we could raise a DeprecationWarning when calling pulser-core/pulser/json/exceptions.py -> even though that would be painful for us @HGSilveri
> 
> It shouldn't be painful, we'd just have to change the imports of these errors to point to their new location. I think we should do this

If there are no libraries that raise these exceptions, it shouldn't be particularly hard.

Yoric:
Also, @HGSilveri, do I need to modify `apicore/core.rst` to add the new submodules of `pulser.exceptions`?

HGSilveri:
> Also, @HGSilveri, do I need to modify `apicore/core.rst` to add the new submodules of `pulser.exceptions`?

No, that should be automatic but you can check the CI docs build to see how they are rendering

pulser-core/pulser/json/exceptions.py
```diff
@@ -11,26 +11,24 @@
...
+import warnings
 
-    pass
+from pulser.exceptions import serialization
 
+warnings.warn(
+    "module pulser.json.exceptions is deprecated, "
+    "please migrate your code to "
+    "use pulser.exceptions.serialization"
+)
```
HGSilveri, line 28:
```suggestion
warnings.warn(
    "module pulser.json.exceptions is deprecated, "
    "please migrate your code to "
    "use pulser.exceptions.serialization", 
    category=DeprecationWarning
)
```
You can also try to adjust the `stacklevel` I'm not sure what's the best value in this case

Yoric: Good catch. At least testing on the REPL, `stacklevel=2` seems to be the only one to actually display a warning, so I'm putting that.

tests/test_exceptions.py
```diff
@@ -0,0 +1,14 @@
...
+
+import pytest
+
+
+def test_import_old_serialize_exceptions():
+    """Test importing from pulser.json.exceptions.
+
+    This should trigger a warning.
+    """
+    with pytest.warns(match="module pulser.json.exceptions is deprecated"):
```
HGSilveri, line 11: Can you check it's a `DeprecationWarning`?
```suggestion
    with pytest.warns(DeprecationWarning, match="module pulser.json.exceptions is deprecated"):
```

HGSilveri (commented):
Just one comment, otherwise LGTM

tests/test_json.py
```diff
@@ -157,16 +167,16 @@ def test_mappable_register():
...
     seq = Sequence(reg, DigitalAnalogDevice)
     var = seq.declare_variable("var")
 
     wf = BlackmanWaveform(var * 100 // 10, var)
     with pytest.warns(
         UserWarning, match="Calls to methods of parametrized objects"
-    ), pytest.raises(
-        ValueError, match="Serialization of calls to parametrized objects"
+    ), helpers.raises_all(
+        [ValueError], match="Serialization of calls to parametrized objects"
```
HGSilveri, line 179: Why are you using the `helpers` in this case?

HGSilveri (approved):
LGTM, thanks @Yoric !

HGSilveri:
Ah yes, it seems that recent changes are importing the errors you moved from the deprecated module... Can you change those please?

tests/test_json.py
Yoric: Oops, removing.

Yoric:
@HGSilveri This should now be fixed.

HGSilveri (approved):

PR #836, opened 2025-03-24
Store result from tag and uuid

a-quelle:
I need a method of storing a result just from what is available in other results, i.e. the tag and uuid of the observable, not the object they came from.


a-corni (approved):
LGTM !

pulser-core/pulser/backend/results.py
```diff
@@ -53,19 +70,7 @@ def _store(
...
-                f" at time {time}."
-            )
-        self._tagmap[observable.tag] = observable.uuid
-        assert (
-            _times == [] or _times[-1] < time
-        ), "Evaluation times are not sorted."
-        _times.append(time)
-        self._results.setdefault(observable.uuid, []).append(value)
-        assert len(_times) == len(self._results[observable.uuid])
+        self._store_raw(observable.uuid, observable.tag, time, value)
```
HGSilveri, line 73: You either use keywords here or take out the `*` in the  `_store_raw()`

PR #840, opened 2025-03-26
Support Python 3.13

### LINKED ISSUE #757: Pulser doesn't install on Python 3.13
Trying to install Pulser on Python 3.13 fails due to issues with building QuTiP and also the supported SciPy version of QuTiP. 

Here is a way to replicate via Docker containers, the same process works with Python 3.12 container (you don't need to install anything via apt with python 3.12 as everything comes pre-built)

The issue appears to come from restricting SciPy < 1.13 in pulser-simulation:
https://github.com/pasqal-io/Pulser/pull/668
I checked and the error in that issue error still persists in SciPy 1.14.1,  and is marked as unsupported also in QuTiP now from 4.7.6.

The QuTiP  issue of course is that we haven't upgraded to version 5, as detailed in #730 .

```bash
docker pull python:3.13
docker run -it python bash
root@bc1448bacafb:/# apt update && apt install -y build-essential gfortran libopenblas-dev
root@bc1448bacafb:/# pip install pulser
```

Also trying to build qutip, bypassing scipy restriction, seems to fail independently:

```bash
      qutip/control/cy_grape.cpp: In function ‘int __Pyx_PyInt_As_int(PyObject*)’:
      qutip/control/cy_grape.cpp:7770:46: error: too few arguments to function ‘int _PyLong_AsByteArray(PyLongObject*, unsigned char*, size_t, int, int, int)’
       7770 |                 int ret = _PyLong_AsByteArray((PyLongObject *)v,
            |                           ~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~
       7771 |                                               bytes, sizeof(val),
            |                                               ~~~~~~~~~~~~~~~~~~~
       7772 |                                               is_little, !is_unsigned);
            |                                               ~~~~~~~~~~~~~~~~~~~~~~~~
      In file included from /usr/local/include/python3.13/longobject.h:107,
                       from /usr/local/include/python3.13/Python.h:81,
                       from qutip/control/cy_grape.cpp:34:
      /usr/local/include/python3.13/cpython/longobject.h:111:17: note: declared here
        111 | PyAPI_FUNC(int) _PyLong_AsByteArray(PyLongObject* v,
            |                 ^~~~~~~~~~~~~~~~~~~
````

This appears to be a Python C-API issue. Seeing that they don't support version 4.x of QuTiP anymore I feel we will have to either upgrade to QuTiP 5, drop qutip as a backend, or restrict to at most Python 3.12 - which isn't a long-term solution (although not super urgent)

### PR DESCRIPTION
**Pull request by HGSilveri**
- Adds support for Python 3.13

Closes #757 .

### DIFF
--- a/pulser-core/requirements.txt
+++ b/pulser-core/requirements.txt
@@ -4,4 +4,5 @@ matplotlib < 4
 packaging   # This is already required by matplotlib but we use it too
 # Numpy 1.20 introduces type hints, 1.24.0 breaks matplotlib < 3.6.1
 numpy >= 1.20, != 1.24.0, < 2
+numpy >= 2; python_version >= '3.13' and platform_system=='Windows'
 scipy < 2
--- a/pulser-core/requirements.txt
+++ b/pulser-core/requirements.txt
@@ -4,5 +4,5 @@ matplotlib < 4
 packaging   # This is already required by matplotlib but we use it too
 # Numpy 1.20 introduces type hints, 1.24.0 breaks matplotlib < 3.6.1
 numpy >= 1.20, != 1.24.0
-numpy >= 2; python_version >= '3.13' and platform_system=='Windows'
+numpy >= 2; python_version >= '3.13'
 scipy < 2
--- a/pulser-core/setup.py
+++ b/pulser-core/setup.py
@@ -45,7 +45,7 @@ setup(
     name=distribution_name,
     version=__version__,
     install_requires=requirements,
-    extras_require={"torch": ["torch ~= 2.0"]},
+    extras_require={"torch": ["torch ~= 2.6"]},
     packages=find_packages(),
     package_data={package_name: ["py.typed"]},
     include_package_data=True,
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,7 +58,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.9", "3.12"]
+        python-version: ["3.9", "3.13"]
         with-torch: ["with-torch", "no-torch"]
     steps:
       - name: Check out Pulser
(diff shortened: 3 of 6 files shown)

### CODE REVIEW

File: pulser-core/setup.py [hunk 1]
```diff
@@ -45,7 +45,7 @@
     name=distribution_name,
     version=__version__,
     install_requires=requirements,
-    extras_require={"torch": ["torch ~= 2.0"]},
+    extras_require={"torch": ["torch ~= 2.6"]},
```
pimvenderbosch, line 48: To make sure torch is numpy2 compatible?

pulser-core/requirements.txt
```diff
@@ -4,4 +4,5 @@ matplotlib < 4
 packaging   # This is already required by matplotlib but we use it too
 # Numpy 1.20 introduces type hints, 1.24.0 breaks matplotlib < 3.6.1
 numpy >= 1.20, != 1.24.0
+numpy >= 2; python_version >= '3.13' and platform_system=='Windows'
```
pimvenderbosch, line 7: Is numpy1 only not supported on Windows? You exclude other platforms as well in the CI.

pimvenderbosch (commented):
LGTM!

pulser-core/requirements.txt
HGSilveri: Yeah, maybe I can remove the `Windows` marker... 
Python 3.13 only started being supported in numpy 2.1, but the CI still passed on Ubuntu and MacOS

pulser-core/setup.py
HGSilveri: Rather Python 3.13 compatible in all OSs

pulser-core/requirements.txt
HGSilveri: Yeah, I was automatically assuming we were always installing v2 on that tag actually

HGSilveri: Yes, I will

pimvenderbosch (approved):

PR #841, opened 2025-03-26
[FIX] Fix definition of InterpolatedWaveform with values as Variable 

a-corni:
Currently, defining an InterpolatedWaveform with both values and times as Variable fail, because the variable _values is not instantiated prior to the evaluation len(times) == len(values) in _check_values_times.

This fixes the issue ~and improves the tests: both numpy arrays and Variables have the attribute `size`, so we can actually check that the size of the given times and values are the same, whatever their type.~

It does not improve the tests, it only fixes the issue: taking into account the fact that the given objects can be ParamObj and not only Variable, and since ParamObj doesn't have the attribute `size`, a user can still provide value and times as Variable of different size, whose build will always fail (but at least the initialization is made possible if their size match).

### DIFF
--- a/pulser-core/pulser/waveforms.py
+++ b/pulser-core/pulser/waveforms.py
@@ -880,7 +880,7 @@ class InterpolatedWaveform(Waveform):
                 values_ = np.array(values, dtype=float)
             except TypeError as e:
                 raise TypeError(_err_message("values")) from e
-        else: 
+        else:
             values_ = values
         if times is None:
             return
--- a/pulser-core/pulser/waveforms.py
+++ b/pulser-core/pulser/waveforms.py
@@ -880,32 +880,30 @@ class InterpolatedWaveform(Waveform):
                 values_ = np.array(values, dtype=float)
             except TypeError as e:
                 raise TypeError(_err_message("values")) from e
-        else:
-            values_ = values
-        if times is None:
+        if times is None or isinstance(times, Parametrized):
             return
-        if not isinstance(times, Parametrized):
-            try:
-                times = cast(ArrayLike, times)
-                times_ = np.array(times, dtype=float)
-            except TypeError as e:
-                raise TypeError(_err_message("times")) from e
-            if np.any(times_ < 0):
-                raise ValueError(
-                    "All values in `times` must be greater than or equal to 0."
-                )
-            if np.any(times_ > 1):
-                raise ValueError(
-                    "All values in `times` must be less than or equal to 1."
-                )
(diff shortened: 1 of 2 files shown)


HGSilveri:
> This fixes the issue and improves the tests: both numpy arrays and Variables have the attribute `size`, so we can actually check that the size of the given times and values are the same, whatever their type.

But `ParamObj` doesn't have `size`. Did you check what happens if you don't give the variable directly but apply some operation to it before?

tests/test_waveforms.py
```diff
@@ -301,6 +301,15 @@ def test_interpolated():
         ),
     ):
         InterpolatedWaveform(duration, [0, 0.1, 0.2, 0.3], other_values)
+    times = seq.declare_variable("times", size=6)  # a Variable
```
HGSilveri, line 304: Try this and see if it still works (my guess is that it won't)
```suggestion
    times = seq.declare_variable("times", size=6) + 1
```

pulser-core/pulser/waveforms.py
```diff
@@ -905,6 +903,14 @@ def _err_message(argument_name: str) -> str:
                 raise ValueError(
                     "`times` must be an array of non-repeating values."
                 )
+        else:
+            times_ = times
+        if times_.size != values_.size:
```
HGSilveri, line 908: Maybe you can do this:
```suggestion
        try:
            size_mismatch = (times_.size != values_.size)
        except AttributeError:
            size_mismatch = False
        if size_mismatch:
```

a-corni:
> > This fixes the issue and improves the tests: both numpy arrays and Variables have the attribute `size`, so we can actually check that the size of the given times and values are the same, whatever their type.
> 
> But `ParamObj` doesn't have `size`. Did you check what happens if you don't give the variable directly but apply some operation to it before?

Ah damn didn't see your comment, it also came to this conclusion

tests/test_waveforms.py
a-corni: I know it won't ;)

pulser-core/pulser/waveforms.py
a-corni: I think it's going to be a nightmare for mypy, the fact that no one had an issue over the last 4 months makes me think that it's overkilled to make a too complex testing

HGSilveri: Ok, I'm fine with what you have right now

HGSilveri (approved):
LGTM, thanks for the quick fix!

PR #847, opened 2025-04-02
Add QutipBackendV2: using Qutip with new API

pablolh:
Create QutipConfig and QutipBackendV2 (Henrique)
Add handling of non-lindbladian noise

### DIFF
--- a/pulser-simulation/pulser_simulation/qutip_backend.py
+++ b/pulser-simulation/pulser_simulation/qutip_backend.py
@@ -16,17 +16,28 @@ from __future__ import annotations
 
 from typing import Any
 
-from pulser import Sequence
-from pulser.backend.abc import Backend
-from pulser.backend.config import EmulatorConfig
+import numpy as np
+import qutip
+
+import pulser
+from pulser.backend.abc import Backend, EmulatorBackend
+from pulser.backend.config import EmulationConfig, EmulatorConfig
+from pulser.backend.default_observables import StateResult
+from pulser.backend.results import Results
 from pulser.noise_model import NoiseModel
+from pulser_simulation.qutip_config import QutipConfig
+from pulser_simulation.qutip_op import QutipOperator
+from pulser_simulation.qutip_state import QutipState
 from pulser_simulation.simconfig import SimConfig
-from pulser_simulation.simresults import SimulationResults
+from pulser_simulation.simresults import CoherentResults, SimulationResults
 from pulser_simulation.simulation import QutipEmulator
 
 
 class QutipBackend(Backend):
     """A backend for emulating the sequences using qutip.
+
+    Warning:
+        Soon to be deprecated, please use ``QutipBackendV2``.
 
     Args:
         sequence: The sequence to emulate.
@@ -37,7 +48,7 @@ class QutipBackend(Backend):
 
     def __init__(
         self,
-        sequence: Sequence,
+        sequence: pulser.Sequence,
         config: EmulatorConfig = EmulatorConfig(),
         mimic_qpu: bool = False,
     ):
@@ -88,3 +99,140 @@ class QutipBackend(Backend):
             Otherwise, returns CoherentResults.
         """
         return self._sim_obj.run(progress_bar=progress_bar, **qutip_options)
+
+
+class QutipBackendV2(EmulatorBackend):
+    """A backend for emulating the sequences using qutip.
+
+    Conforms to the generic API from pulser.backend.
+
+    Args:
+        sequence: The sequence to emulate.
+        config: The configuration for the Qutip emulator.
+        mimic_qpu: Whether to mimic the validations necessary for
+            execution on a QPU.
+    """
+
+    default_config = QutipConfig(observables=[StateResult()])
+    _config: QutipConfig
+
+    def __init__(
+        self,
+        sequence: pulser.Sequence,
+        *,
+        config: EmulationConfig | None = None,
+        mimic_qpu: bool = False,
+    ) -> None:
+        """Initializes the backend."""
+        super().__init__(sequence, config=config, mimic_qpu=mimic_qpu)
+        noise_model: None | NoiseModel = None
+        if self._config.prefer_device_noise_model:
+            noise_model = sequence.device.default_noise_model
+        noise_model = noise_model or self._config.noise_model
+        simconfig = SimConfig.from_noise_model(noise_model)
+        self._sim_obj = QutipEmulator.from_sequence(
+            sequence,
+            sampling_rate=self._config.sampling_rate,
+            config=simconfig,
+            with_modulation=self._config.with_modulation,
+        )
+        self._sim_obj.set_evaluation_times(
+            self._config._get_legacy_evaluation_times(
+                self._sim_obj.total_duration_ns
+            ),
+        )
+        if self._config.initial_state:
+            self._sim_obj.set_initial_state(
+                self._config.initial_state.to_qobj()
+            )
+
+    def run(self) -> Results:
+        """Executes the sequence on the backend."""
+        res = Results(
+            atom_order=tuple(self._sequence.qubit_info),
+            total_duration=self._sim_obj.total_duration_ns,
+        )
+        eigenstates = self._sim_obj.samples_obj.eigenbasis
+
+        if (
+            ("doppler" not in self._sim_obj.config.noise)
+            and (
+                "amplitude" not in self._sim_obj.config.noise
+                or self._sim_obj.config.amp_sigma == 0.0
+            )
+            and (
+                "SPAM" not in self._sim_obj.config.noise
+                or self._sim_obj.config.eta == 0
+            )
+        ):
+            # A single run is needed, regardless of self.config.runs
+            single_res = self._sim_obj._run_solver(progress_bar=False)
+            assert isinstance(single_res, CoherentResults)
+
+            for qutip_res in single_res:
+                t = qutip_res.evaluation_time
+                state = QutipState(qutip_res.state, eigenstates=eigenstates)
+                ham: QutipOperator = QutipOperator(
+                    self._sim_obj.get_hamiltonian(
+                        t * res.total_duration, noiseless=True
+                    ),
+                    eigenstates=eigenstates,
+                )
+                for obs in self._config.observables:
+                    obs(
+                        config=self._config,
+                        t=t,
+                        state=state,
+                        hamiltonian=ham,
+                        result=res,
+                    )
+
+        else:
+            density_matrices: dict[float, qutip.Qobj] = {}
+            total_reps = 0
+            for cleanres_noisyseq, reps in self._sim_obj._noisy_runs(
+                progress_bar=False
+            ):
+                total_reps += reps
+                for index, qutip_res in enumerate(cleanres_noisyseq):
+                    t = qutip_res.evaluation_time
+
+                    if t not in density_matrices:
+                        density_matrices[t] = qutip.tensor(
+                            [
+                                qutip.Qobj(np.zeros((2, 2)))
+                                for _ in range(
+                                    self._sim_obj._hamiltonian._size
+                                )
+                            ]
+                        )
+
+                    if qutip_res.state.isoper:
+                        density_matrices[t] += reps * qutip_res.state
+                    else:
+                        density_matrices[t] += (
+                            reps * qutip_res.state * qutip_res.state.dag()
+                        )
+
+            density_matrices = {
+                t: m / total_reps for t, m in density_matrices.items()
+            }
+
+            for t, dm in density_matrices.items():
+                state = QutipState(dm, eigenstates=eigenstates)
+                ham = QutipOperator(
+                    self._sim_obj.get_hamiltonian(
+                        t * res.total_duration, noiseless=True
+                    ),
+                    eigenstates=eigenstates,
+                )
+                for obs in self._config.observables:
+                    obs(
+                        config=self._config,
+                        t=t,
+                        state=state,
+                        hamiltonian=ham,
+                        result=res,
+                    )
+
+        return res
--- a/pulser-simulation/pulser_simulation/qutip_config.py
+++ b/pulser-simulation/pulser_simulation/qutip_config.py
@@ -1,0 +1,139 @@
+# Copyright 2024 Pulser Development Team
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Defines the QutipConfig class."""
+from __future__ import annotations
+
+import warnings
+from typing import Any, Literal
+
+import numpy as np
+
+from pulser.backend.config import EmulationConfig
+from pulser_simulation.qutip_op import QutipOperator
+from pulser_simulation.qutip_state import QutipState
+
+
+class QutipConfig(EmulationConfig[QutipState]):
+    """The configuration of a QutipBackend.
+
+    Args:
+        observables: A sequence of observables to compute at specific
+            evaluation times. The observables without specified evaluation
+            times will use this configuration's 'default_evaluation_times'.
+        sampling_rate: The fraction of samples to extract from the pulse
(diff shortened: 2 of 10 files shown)


pulser-simulation/pulser_simulation/qutip_config.py
```diff
@@ -0,0 +1,127 @@
...
+            )
+        initial_state = backend_options.setdefault("initial_state")
+        if initial_state and not isinstance(initial_state, QutipState):
+            raise TypeError(
+                "If provided, `initial_state` must be an instance of "
+                f"`QutipState`, not {type(initial_state)}."
+            )
+        super().__init__(
+            sampling_rate=sampling_rate,
+            default_evaluation_times=default_evaluation_times,
```
a-quelle, line 83: I'm a bit confused about what's happening with `default_evaluation_times`. It exists in `EmulationConfig`, but there it's `np.ndarray` rather than `Sequence[float]`, so that could be the reason to redefine it here, but then I would expect some logic to cast it to `ndarray` before feeding it to `super`

pulser-simulation/pulser_simulation/qutip_backend.py
```diff
@@ -88,3 +96,138 @@ def run(
...
+
+                    density_matrices.setdefault(
+                        t_,
+                        qutip.tensor(
+                            [
+                                qutip.Qobj(np.zeros((2, 2)))
+                                for _ in range(
+                                    self._sim_obj._hamiltonian._size
+                                )
+                            ]
```
HGSilveri, line 201: nit: Is this going to be executed regardless of whether `t_` already exists in `density_matrices`? If so, the old-school `if t_ in density_matrices` or a `defaultdict` would be preferable

pulser-simulation/pulser_simulation/qutip_backend.py
```diff
@@ -88,3 +96,138 @@ def run(
...
+                    else:
+                        density_matrices[t_] += (
+                            reps * qutip_res.state * qutip_res.state.dag()
+                        )
+
+            density_matrices = {
+                t: m / total_reps for t, m in density_matrices.items()
+            }
+
+            for t, dm in density_matrices.items():
```
a-quelle, line 216: I might be misunderstanding the code, but should you aggregrate the `density_matrices` dict as
```
sampled_state = qutip.tensor(
                            [
                                qutip.Qobj(np.zeros((2, 2)))
                                for _ in range(
                                    self._sim_obj._hamiltonian._size
                                )
                            ]
                        )
for weight, dm in density_matrices.items():
    sampled_state += weight * dm
```
and then call observables on that?

pablolh: every shot gives you a density matrix for all times, so there are as many accumulating density matrices as there are sampling times and every new noisy run means adding new DMs for each time.
It would aggregate as you describe (for a given time) if I were to store every DM for every time and every run, and postprocess, but that would use too much memory, I think, without providing additional value afaics

pulser-simulation/pulser_simulation/qutip_backend.py
```diff
@@ -88,3 +96,138 @@ def run(
...
+        )
+        eigenstates = self._sim_obj.samples_obj.eigenbasis
+
+        if (
+            "amplitude" not in self._sim_obj.config.noise
+            or self._sim_obj.config.amp_sigma == 0.0
+        ) and (
+            "SPAM" not in self._sim_obj.config.noise
+            or self._sim_obj.config.eta == 0
+        ):
```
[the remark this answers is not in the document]
pablolh: agreed

pulser-simulation/pulser_simulation/qutip_backend.py
a-quelle: ah, I see, you do this aggregation in the previous loop, on the results yielded from `noisy_runs`. Never mind then.

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -479,6 +484,120 @@ def get_hamiltonian(self, time: float) -> qutip.Qobj:
             time / 1000
         )  # Creates new Qutip.Qobj
 
+    @staticmethod
+    def get_min_variation(ch_sample: ChannelSamples) -> int:
```
HGSilveri, line 488: I think this should stay internal, so I'd rename it to 
```suggestion
    def _get_min_variation(ch_sample: ChannelSamples) -> int:
```
Also, out of context it's hard to figure out what it is doing (and why) at a glance so a small docstring would help

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -479,6 +484,120 @@ def get_hamiltonian(self, time: float) -> qutip.Qobj:
...
+        return CoherentResults(
+            results,
+            self._hamiltonian._size,
+            self.basis_name,
+            self._eval_times_array,
+            self._meas_basis,
+            meas_errors,
+        )
+
+    def validate_options(self, options: Any) -> None:
```
HGSilveri, line 573: Same here, this shouldn't be a public method

pulser-simulation/pulser_simulation/qutip_backend.py
```diff
@@ -88,3 +96,138 @@ def run(
...
+            density_matrices = {
+                t: m / total_reps for t, m in density_matrices.items()
+            }
+
+            for t, dm in density_matrices.items():
+                state = QutipState(dm, eigenstates=eigenstates)
+                ham = QutipOperator(
+                    self._sim_obj.get_hamiltonian(
+                        t * res.total_duration
+                    ),  # FIXME: remove noise from this hamiltonian!
```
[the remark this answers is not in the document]
pablolh: There is an interesting question also regarding state preparation errors.
In emu-mps, the "noiseless" hamiltonian that the callbacks see has identity factors on the dark qubits.
So it "contains" the dark qubits still: the hamiltonian **doesn't act on them.**
Here we are supposed to give a hamiltonian for multiple combined runs that have potentially different randomly chosen dark qubits. So I think this hamiltonian shouldn't have dark qubits, because those are different for each of those runs. That would be a difference in behavior. I'm not shocked by either behaviors, I think at the time for emu-mps we thought it had to be like that.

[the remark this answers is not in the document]
HGSilveri: Good point! Unless we use the leakage state to emulate a state preparation error (which is currently not the case), I guess we will have to live with this approximation

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -157,6 +157,13 @@ def __init__(
             sampling_rate,
             noise_model,
         )
+        self._noiseless_hamiltonian = Hamiltonian(
+            self.samples_obj,
+            self._register.qubits,
+            device,
+            sampling_rate,
+            NoiseModel(),
+        )
```
HGSilveri, line 166: This can take a bit of time, might be better to calculate it and cache it only when it is first requested in `get_hamiltonian()`. A convenient way to do so is to define  `_noiseless_hamiltonian` as a `cached_property`

tests/pulser_simulation/test_qutip_backend_v2.py
```diff
@@ -0,0 +1,136 @@
...
+from pulser_simulation.simconfig import SimConfig
+from pulser_simulation.simulation import QutipEmulator
+
+Omega_max = 4 * 2 * math.pi
+U = Omega_max / 2
+delta_0 = -6 * U
+delta_f = 2 * U
+t_rise = 500
+t_fall = 1000
+t_sweep = (delta_f - delta_0) / (2 * np.pi * 10) * 1000
```
HGSilveri, line 35: nit: Not sure you need to make these global. I would move them inside the `sequence` fixture

tests/pulser_simulation/test_qutip_backend_v2.py
```diff
@@ -0,0 +1,136 @@
...
+        results.get_result("energy", 0.0)
+        == results.energy[0]
+        == pytest.approx(0.0)
+    )
+    assert (
+        results.get_result("energy", 1.0)
+        == results.energy[-1]
+        == pytest.approx(
+            qutip.expect(
+                backend._sim_obj.get_hamiltonian(t_rise + t_sweep + t_fall),
```
HGSilveri, line 89: Here you could do this instead
```suggestion
                backend._sim_obj.get_hamiltonian(sequence.get_duration()),
```

tests/pulser_simulation/test_qutip_backend_v2.py
```diff
@@ -0,0 +1,136 @@
...
+            Energy(evaluation_times=[0.001 * n for n in range(1001)]),
+        ],
+    )
+    backend = QutipBackendV2(sequence, config=config)
+    results = backend.run()
+    assert (
+        results.get_result("energy", 0.0)
+        == results.energy[0]
+        == pytest.approx(0.0)
+    )
```
HGSilveri, line 83: I think it'd be nice to check the energy at some point in the middle too

tests/pulser_simulation/test_qutip_backend_v2.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+def test_qutip_backend_v2_stochastic_noise(sequence):
+    np.random.seed(123)
+
+    noise_model = pulser.noise_model.NoiseModel(
+        temperature=50.0,
+        p_false_neg=0.01,
+        amp_sigma=1e-3,
+        runs=500,
```
HGSilveri, line 103: How long is this test taking to run? 500 does sounds like a lot of runs, even with just 4 qubits

tests/pulser_simulation/test_qutip_backend_v2.py
```diff
@@ -0,0 +1,136 @@
...
+    backend = QutipBackendV2(sequence, config=config)
+    results = backend.run()
+
+    # Same run with old API
+    sim_config = SimConfig.from_noise_model(noise_model)
+    qutip_emulator = QutipEmulator.from_sequence(sequence, config=sim_config)
+    results_old_api = qutip_emulator.run()
+
+    times = results.get_result_times("occupation")
+    occupation = np.array([x[0] for x in results.occupation])
```
HGSilveri, line 125: Since you are only using the first qubit, maybe you can reduce the number of qubits in the sequence. Maybe not to 1 so that you have interactions, but 2 should be fine

HGSilveri (commented):
Looking good!

tests/pulser_simulation/test_qutip_backend_v2.py
pablolh: it's taking a long time, but those occupation values from the sampled pseudo DM don't seem to make a lot of sense with much less unfortunately

pablolh: `runs=200` with only 2 qubits, it's taking 63 seconds on my machine now.

pablolh: and I had to change the tolerance, still reasonable

pablolh:
okay so I added all the tests to try and reach 100% coverage :sweat:

tests/pulser_simulation/test_qutip_backend_v2.py
pablolh: I think I just need to increase samples_per_run! let me work on that

pablolh: okay much better with samples_per_run=100, error is smaller and time is only 22 seconds on my machine

tests/pulser_simulation/test_qutip_backend_v2.py
```diff
@@ -0,0 +1,174 @@
...
+        Omega_max,
+        pulser.waveforms.RampWaveform(t_sweep, delta_0, delta_f),
+        0.0,
+    )
+    fall = pulser.Pulse.ConstantDetuning(
+        pulser.waveforms.RampWaveform(t_fall, Omega_max, 0.0), delta_f, 0.0
+    )
+
+    seq = pulser.Sequence(
+        reg, device if device is not None else pulser.devices.MockDevice
```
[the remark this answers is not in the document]
pablolh: image
I've been already bitten by this and I tend to agree that because many values can subtly evaluate to False, this should be discouraged.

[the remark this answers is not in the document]
pablolh: https://stackoverflow.com/questions/13710631/is-there-shorthand-for-returning-a-default-value-if-none-in-python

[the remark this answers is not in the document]
HGSilveri: To be clear, my suggestion was not to always use this, it should be considered on a case-by-case basis imo. I agree that because many things evaluate to `False` that can be problematic is some cases. I think that you will agree that in this case there is no danger but you are free not to use it, it was just a suggestion for brevity.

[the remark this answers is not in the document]
pablolh: It is a conscious choice of not using this seemingly beautiful `or` because it is in fact **not a shortcut**: it has a **different meaning** and is often called an "antipattern". See the comments [here](https://stackoverflow.com/questions/13710631/is-there-shorthand-for-returning-a-default-value-if-none-in-python)
The first thing is that when given a device, it could rely on `Device._bool_` and/or `Device.__len__` and then if someone somehow gives another data type it might behave even more differently.

pulser-simulation/pulser_simulation/qutip_backend.py
```diff
@@ -88,3 +96,142 @@ def run(
             Otherwise, returns CoherentResults.
         """
         return self._sim_obj.run(progress_bar=progress_bar, **qutip_options)
+
+
+class QutipBackendV2(EmulatorBackend):
```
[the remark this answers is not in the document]
pablolh: great, didn't think about this

pulser-simulation/pulser_simulation/qutip_backend.py
```diff
@@ -88,3 +96,142 @@ def run(
...
+    """A backend for emulating the sequences using qutip.
+
+    Warning:
+        Still experimental and a work-in-progress. Not all features of
+        QutipBackend are supported.
```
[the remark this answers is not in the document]
pablolh: agreed

pulser-simulation/pulser_simulation/qutip_config.py
```diff
@@ -0,0 +1,125 @@
...
+            model given with this configuration.
+        noise_model: An optional noise model to emulate the sequence with.
+            Ignored if the sequence's device has default noise model and
+            `prefer_device_noise_model=True`.
+
+    See Also:
+        EmulationConfig: The base configuration class for an EmulatorBackend.
+    """
+
+    sampling_rate: float
```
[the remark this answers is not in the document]
pablolh: ack

tests/pulser_simulation/test_qutip_backend_v2.py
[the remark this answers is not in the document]
HGSilveri: I know that, I only suggested this because is a test where you control everything, there is no "someone" to give another data type.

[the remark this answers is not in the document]
pablolh: damn you were fast :D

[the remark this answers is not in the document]
pablolh: thanks for the nice review btw

pulser-simulation/pulser_simulation/qutip_backend.py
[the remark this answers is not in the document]
pablolh: https://github.com/pasqal-io/Pulser/issues/850

pablolh (commented):
apologize for confused communication, I should def only use the "Files changed" tab of github always to let me edit my comments before sending review

pulser-simulation/pulser_simulation/qutip_backend.py
```diff
@@ -16,18 +16,28 @@
...
 from pulser_simulation.simconfig import SimConfig
-from pulser_simulation.simresults import SimulationResults
+from pulser_simulation.simresults import CoherentResults, SimulationResults
 from pulser_simulation.simulation import QutipEmulator
 
 
 class QutipBackend(Backend):
     """A backend for emulating the sequences using qutip.
 
+    Warning: soon to be deprecated, please use QutipBackendV2.
```
HGSilveri, line 39: nit: Formatting so it renders well in the docs
```suggestion
    Warning: 
        Soon to be deprecated, please use ``QutipBackendV2``.
```

HGSilveri (approved):
I just have a nit, otherwise LGTM! Very nice job @pablolh :rocket:

pulser-simulation/pulser_simulation/qutip_backend.py
pablolh: done! thanks!

PR #853, opened 2025-04-11
Replace wheel with build for packaging

### LINKED ISSUE #851: Update packaging to avoid bdist_wheel
`bdist_wheel` has been removed from the latest release of `wheel`, so we should move away from using it when we package and publish.

HGSilveri:
Closes #851 .


a-corni (approved):
LGTM !

Issue #852, opened 2025-04-11
Clarify `No DMM dmm_0 in the device` error message


Yoric:
When reading this error message, it's not clear to me whether:

a. There's no `dmm_0` because the device does not support DMM channels.
b. There's no `dmm_0` because of a typo.
c. There's no `emm_0` because I forgot to declare it somewhere.

Yoric:
Could we possibly tweak the message to the following? This would remove any ambiguity.

`No DMM dmm_0 in the device: this device supports DMMs [ ... ]` / `No DMM dmm_0 in the device: this device does not support DMM`

a-corni:
Yes, I would be in favour of using your first option. Combining your proposition and @awennersteen's:

f"No DMM called {dmm} is available in the device. The device supports the following DMM channels: {device.dmm_channels}"

That makes a good first issue !

PR #854, opened 2025-04-14
FIX: Clarify “No DMM” error message

Igoreduardobraga:
## Overview
This pull request addresses issue #852 by clarifying the error message raised when attempting to configure a non‑existent DMM channel on a device. Previously, the message simply stated:

<pre>
	f"No DMM {dmm_id} in the device."
</pre>

And now, the message has been changed to:

<pre>
	f"No DMM called {dmm_id} is available in the device. Your selected device {self._device.name} has the following DMM channels available: {list(self._device.dmm_channels.keys())}."
</pre>

Additionally, the related tests have been updated to match the new, detailed error message in the test_sequence file.

## Checklist

- [x] Closes https://github.com/pasqal-io/Pulser/issues/852
- [x] Executed tests
- [x] Adjusted tests in tests/test_sequence.py to match the new error message
- [x]  Executed tests and confirmed all tests pass

### DIFF
--- a/.github/scripts/package.sh
+++ b/.github/scripts/package.sh
@@ -15,12 +15,12 @@ packages=$(cat packages.txt)
 for pkg in $packages
 do
   echo "Packaging $pkg"
-  python $pkg/setup.py -q bdist_wheel -d "../dist"
+  python -m build $pkg --wheel -o "dist"
   rm -r $pkg/build
 done
 
 # Build the pulser metapackage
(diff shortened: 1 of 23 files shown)


a-corni (approved):
Thanks for tackling this issue @Igoreduardobraga !
It looks good to me !
Nit: It's preferable to use the property `self.device` instead of the internal attribute `self._device`, but I am fine with your changes :)

a-corni:
Closed by #864

Issue #858, opened 2025-04-15
Allow `Callback` instances in `EmulationConfig.observables`


HGSilveri:
Currently `EmulationConfig` enforces `observables: Sequence[Observable]`. Per user request, we should assess whether we can relax them to be `Sequence[Callback]` instead.

https://github.com/pasqal-io/Pulser/blob/63262de478721743897612ee43e50c9701532718/pulser-core/pulser/backend/config.py#L179-L185

At a glance, it looks like we are relying on every observable having a `tag`, which currently is specific to `Observable`. 

Cc @a-quelle

HGSilveri:
Closed by #862

PR #862, opened 2025-04-22
Allow general callback in EmulationConfig

a-quelle:
Currently, all the type hints say `Observable` and there is an `isinstance` check too. However, all the code will work with `Callback` too, it's just that the logic about checking evaluation times is not valid anymore for a general `Callback`.

I relaxed the typing a bit. There is currently no validation if the passed object is not an `Observable` since for a general `Callback` you cannot say very much about what it'll do.

### DIFF
--- a/pulser-core/pulser/backend/config.py
+++ b/pulser-core/pulser/backend/config.py
@@ -109,6 +109,7 @@ class EmulationConfig(BackendConfig, Generic[StateType]):
         callbacks: A general callback that is not an observable. Observables
             must be fed into the observables arg, since they all interact
             with the Results, and are subject to additional validation.
+            Unlike observables, these are called at every emulation step.
         default_evaluation_times: The default times at which observables
             are computed. Can be a sequence of unique relative times between 0
             (the start of the sequence) and 1 (the end of the sequence), in
--- a/pulser-core/pulser/backend/config.py
+++ b/pulser-core/pulser/backend/config.py
@@ -36,7 +36,7 @@ import numpy as np
 from numpy.typing import ArrayLike
 
 import pulser.math as pm
-from pulser.backend.observable import Observable
+from pulser.backend.observable import Callback, Observable
 from pulser.backend.operator import Operator, OperatorRepr
 from pulser.backend.state import State, StateRepr
 from pulser.json.abstract_repr.backend import _deserialize_emulation_config
@@ -138,6 +138,7 @@ class EmulationConfig(BackendConfig, Generic[StateType]):
 
     """
 
+    callbacks: Sequence[Callback]
     observables: Sequence[Observable]
     default_evaluation_times: np.ndarray | Literal["Full"]
     initial_state: StateType | None
@@ -154,6 +155,7 @@ class EmulationConfig(BackendConfig, Generic[StateType]):
     def __init__(
         self,
         *,
+        callbacks: Sequence[Callback] = (),
         observables: Sequence[Observable] = (),
         # Default evaluation times for observables that don't specify one
         default_evaluation_times: Sequence[SupportsFloat] | Literal["Full"] = (
@@ -175,7 +177,17 @@ class EmulationConfig(BackendConfig, Generic[StateType]):
                 " empty.",
                 stacklevel=2,
             )
-
+        for cb in callbacks:
+            if isinstance(cb, Observable):
+                raise TypeError(
+                    "All entries in 'callbacks' must not be instances of "
+                    "Observable, since those go in 'observables'."
+                )
+            if not isinstance(cb, Callback):
+                raise TypeError(
+                    "All entries in 'callbacks' must be instances of "
+                    f"Callback. Instead, got instance of type {type(cb)}."
+                )
         for obs in observables:
             if not isinstance(obs, Observable):
                 raise TypeError(
@@ -241,6 +253,7 @@ class EmulationConfig(BackendConfig, Generic[StateType]):
             )
 
         super().__init__(
+            callbacks=tuple(callbacks),
             observables=tuple(observables),
             default_evaluation_times=default_evaluation_times,
             initial_state=initial_state,
@@ -253,6 +266,7 @@ class EmulationConfig(BackendConfig, Generic[StateType]):
 
     def _expected_kwargs(self) -> set[str]:
         return super()._expected_kwargs() | {
+            "callbacks",
             "observables",
             "default_evaluation_times",
             "initial_state",
--- a/pulser-core/pulser/backend/config.py
+++ b/pulser-core/pulser/backend/config.py
@@ -106,6 +106,9 @@ class EmulationConfig(BackendConfig, Generic[StateType]):
         observables: A sequence of observables to compute at specific
             evaluation times. The observables without specified evaluation
             times will use this configuration's 'default_evaluation_times'.
+        callbacks: A general callback that is not an observable. Observables
+            must be fed into the observables arg, since they all interact
+            with the Results, and are subject to additional validation.
         default_evaluation_times: The default times at which observables
             are computed. Can be a sequence of unique relative times between 0
             (the start of the sequence) and 1 (the end of the sequence), in
@@ -170,7 +173,7 @@ class EmulationConfig(BackendConfig, Generic[StateType]):
     ) -> None:
         """Initializes the EmulationConfig."""
         obs_tags = []
-        if not observables:
+        if not observables and not callbacks:
             warnings.warn(
                 f"{self.__class__.__name__!r} was initialized without any "
                 "observables. The corresponding emulation results will be"
--- a/pulser-simulation/pulser_simulation/qutip_backend.py
+++ b/pulser-simulation/pulser_simulation/qutip_backend.py
@@ -178,6 +178,14 @@ class QutipBackendV2(EmulatorBackend):
                     ),
                     eigenstates=eigenstates,
                 )
+                for callback in self._config.callbacks:
+                    callback(
+                        config=self._config,
+                        t=t,
+                        state=state,
+                        hamiltonian=ham,
+                        result=res,
+                    )
                 for obs in self._config.observables:
                     obs(
                         config=self._config,
@@ -226,6 +234,14 @@ class QutipBackendV2(EmulatorBackend):
                     ),
                     eigenstates=eigenstates,
                 )
+                for callback in self._config.callbacks:
+                    callback(
+                        config=self._config,
+                        t=t,
+                        state=state,
+                        hamiltonian=ham,
+                        result=res,
+                    )
                 for obs in self._config.observables:
                     obs(
                         config=self._config,
--- a/pulser-simulation/pulser_simulation/qutip_config.py
+++ b/pulser-simulation/pulser_simulation/qutip_config.py
@@ -120,6 +120,8 @@ class QutipConfig(EmulationConfig[QutipState]):
         self, total_duration_ns: int
     ) -> Literal["Full"] | np.ndarray:
         extra_eval_times: set[float] = set()
+        if self.callbacks:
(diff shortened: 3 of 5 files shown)


pulser-core/pulser/backend/config.py
```diff
@@ -177,12 +179,13 @@ def __init__(
...
-            if not isinstance(obs, Observable):
+            if not isinstance(obs, Callback):
                 raise TypeError(
                     "All entries in 'observables' must be instances of "
-                    f"Observable. Instead, got instance of type {type(obs)}."
+                    f"Callback. Instead, got instance of type {type(obs)}."
                 )
-            obs_tags.append(obs.tag)
+            if isinstance(obs, Observable):
+                obs_tags.append(obs.tag)
```
a-corni, line 188: I think you should use uuid if the observable is a general callback -> we need to check that a given Callback is not used twice

pulser-core/pulser/backend/config.py
```diff
@@ -106,6 +106,8 @@ class EmulationConfig(BackendConfig, Generic[StateType]):
         observables: A sequence of observables to compute at specific
             evaluation times. The observables without specified evaluation
             times will use this configuration's 'default_evaluation_times'.
+            most also accept more general callbacks,
```
pablolh, line 109: change looks good but I think this docstring needs to be reworked
good to note that when given a generic callback, it is applied at every timestep, but first line mentions `to compute at evaluation times.`
also I am not sure of the meaning of this "at the cost of some validation", what do you mean?

Here is a suggestion:

```
observables: A sequence of callbacks to compute. Given `Observable` instances are computed at the specified evaluation times, and directly stored in the `Results` object. Those without explicit evaluation times will use this configuration's 'default_evaluation_times' instead. Generic `Callback` instances are computed at the end of each and every simulation time step.
```

tests/test_backend.py
```diff
@@ -407,7 +407,7 @@ def test_emulation_config():
 
     with pytest.raises(
         TypeError,
-        match="All entries in 'observables' must be instances of Observable",
+        match="All entries in 'observables' must be instances of Callback",
     ):
         EmulationConfig(observables=["fidelity"])
     with pytest.raises(
```
a-corni, line 413: I would like to see a test passing with a general callback, and a test with the same general callback used twice, raising an error "Some of the provided 'callbacks' share identical tags"

a-corni (commented):
I agree using general callbacks work, especially since we were already handling observables without evaluation times, which is very close to a general callback. However, the main differences is that we were using observable.tag, that should be replaced by callback.uuid to check that a callback is not called twice iiuc (users will the uuid of the callback to get back the results associated to it, right ?)

pulser-core/pulser/backend/config.py
a-corni: Don't general callbacks behave like a observable with evaluation times ?

pulser-core/pulser/backend/config.py
pablolh: I thought about it and I think that is debatable.

First of all, what is the use case for having Callbacks in your opinions (Antoine and Anton)? maybe we need to align on this because you don't say why this change in first place.

I think freedom should be left to advanced users who want to call a callback twice (per timestep), since callbacks are not meant to store in `Results`, but potentially perform some side-effect, like "print current time" or garbage collect or check memory usage or conditionally start pdb or... (god knows what else)
It is true that we didn't specify that Callbacks in the observables list are applied in the order specified, but I think that is a reasonable assumption.

Also, the behavior of `uuid` is different from `tag` because two different `Observable` instances can have same `tag` (with same `tag_suffix` even) but two different instances of `Callback` cannot have different `uuid`. So checking that uuids are not duplicate won't catch things like `observables=[MyCallback(), Energy(...), MyCallback()]`

What do you think?

pulser-core/pulser/backend/config.py
pablolh: > Don't general callbacks behave like an observable without evaluation times ?

I don't think so. For me with `Callback`'s you override the `__call__` abstract method to whatever you want, and this `operator()` is called at every timestep (in all emulators normally) because checking evaluation times is actually done in overriden `__call__` from `Observable` which then won't happen with `Callback`

pulser-core/pulser/backend/config.py
a-quelle: I'm inclined to agree with Pablo on this. Since `Observables` are put in `Results` by tag, it's important to prune duplicates. For callbacks, the mechanism is really that it just executes the given callback at a given time, and I think adding this kind of validation is potentially dangerous.

pulser-core/pulser/backend/config.py
a-quelle: no, the logic for putting a result in `Results` is specifically in `Observable`. The only thing the `Callback`specifies is the existence of a UUID, and a `call` method. Overriding `call` directly allows you to inject more general behaviour into the program without having to refer to the existence of tags, Results, and evaluation times

a-corni: Okay, thanks for the input, I now understand the difference and the goal of using general Callback :)

pulser-core/pulser/backend/config.py
a-corni: I agree with both of you, thanks for your inputs :)

pulser-simulation/pulser_simulation/qutip_config.py
```diff
@@ -121,7 +121,8 @@ def _get_legacy_evaluation_times(
     ) -> Literal["Full"] | np.ndarray:
         extra_eval_times: set[float] = set()
         for obs in self.observables:
-            extra_eval_times.update(obs.evaluation_times or [])
+            if isinstance(obs, Observable):
+                extra_eval_times.update(obs.evaluation_times or [])
```
a-corni, line 125: Following Pablo's comment, if one of the observable of the config is a Callback, then `extra_eval_times` is "Full": you want to evaluate the callback at all the evaluation times.

pablolh: I'm wondering if this piece of code shouldn't be in `pulser.backend`

What about setting `default_evaluation_times` to `Full` in `EmulationConfig` whenever there is at least one Callback?

pulser-core/pulser/backend/config.py
```diff
@@ -154,6 +155,7 @@ class EmulationConfig(BackendConfig, Generic[StateType]):
     def __init__(
         self,
         *,
+        callbacks: Sequence[Callback] = (),
```
a-corni, line 158: To be backwards compatible, callbacks should be the last args...

pulser-core/pulser/backend/config.py
```diff
@@ -138,6 +138,7 @@ class EmulationConfig(BackendConfig, Generic[StateType]):
 
     """
 
+    callbacks: Sequence[Callback]
```
a-corni, line 141: You should add a description for these callbacks 

callbacks: A sequence of callbacks to compute at each step. No Observable can be provided in this sequence (Observable must be provided in the observables argument).

a-corni:
Closes #858

a-corni (commented):
I am good with the changes, only asking to add a description of the new `callbacks` argument

pulser-simulation/pulser_simulation/qutip_config.py
a-quelle: this comment is outdated, since I force pushed over the branch after the slack discussion on tuesday. @pablolh I did what you suggested, except in `QutipConfig._get_legacy_evaluation_times" since I plan to do an implementation in emu-mps and emu-sv that avoids bringing evaluation times into the callbacks. In pulser-simulation this is hard to avoid, since it only calls the callbacks after the simulation is done.

pablolh:
For me looks good too, ready to approve

pulser-core/pulser/backend/config.py
a-quelle: I added the description, also test coverage was not 100%. This was not really visible since for some reason the pipeline didn't fully run after my force push.

pablolh (approved):

tests/pulser_simulation/test_qutip_backend_v2.py
```diff
@@ -22,13 +22,22 @@
...
 from pulser.backend.default_observables import Energy, Occupation, StateResult
+from pulser.backend.observable import Callback
 from pulser_simulation.qutip_backend import QutipBackendV2
 from pulser_simulation.qutip_config import QutipConfig
 from pulser_simulation.qutip_state import QutipState
 from pulser_simulation.simconfig import SimConfig
 from pulser_simulation.simulation import QutipEmulator
 
 
+class CountCalls(Callback):
```
a-corni, line 33: Looks like you need to add a description for this object :)

pulser-core/pulser/backend/config.py
```diff
@@ -106,6 +106,9 @@ class EmulationConfig(BackendConfig, Generic[StateType]):
         observables: A sequence of observables to compute at specific
             evaluation times. The observables without specified evaluation
             times will use this configuration's 'default_evaluation_times'.
+        callbacks: A general callback that is not an observable. Observables
+            must be fed into the observables arg, since they all interact
+            with the Results, and are subject to additional validation.
```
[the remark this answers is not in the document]
HGSilveri: I see where are putting the onus on the user to incorporate in its definition what evaluation times it should care about. This is fine, but it should be accompanied by the statement that each callback will be called at every emulation step (and not just at the `default_evaluation_times`, right?)

tests/pulser_simulation/test_qutip_backend_v2.py
a-quelle: done

tests/pulser_simulation/test_qutip_backend_v2.py
```diff
@@ -64,12 +73,29 @@ def sequence(device: pulser.devices.Device | None = None):
...
+    backend.run()
+    assert backend._config.callbacks[0].counter == seq.get_duration() + 1
+
+
 def test_qutip_backend_v2_energy():
     seq = sequence()
-    with pytest.raises(
-        TypeError, match="'config' must be an instance of 'EmulationConfig'"
-    ):
-        QutipBackendV2(seq, config="tralala")
```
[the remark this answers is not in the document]
a-quelle: I don't remember. Probably fallout from a false turn I took in my first attempt at an implementation. I put it back in and the test still passes.

pulser-simulation/pulser_simulation/qutip_backend.py
```diff
@@ -178,6 +178,14 @@ def run(self) -> Results:
...
                     eigenstates=eigenstates,
                 )
+                for callback in self._config.callbacks:
+                    callback(
+                        config=self._config,
+                        t=t,
+                        state=state,
+                        hamiltonian=ham,
+                        result=res,
+                    )
```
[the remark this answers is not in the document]
a-quelle: indeed. I added this to the description per your request above.

tests/pulser_simulation/test_qutip_backend_v2.py
```diff
@@ -22,13 +22,26 @@
...
 from pulser_simulation.simconfig import SimConfig
 from pulser_simulation.simulation import QutipEmulator
 
 
+class CountCalls(Callback):
+    def __init__(self):
+        """
+        Simple callback that counts how often it's been called.
+        The count can be queried after the simulation from self.counter.
+        """
```
HGSilveri, line 38: I'm afraid this will still be problematic, try this instead
```suggestion
class CountCalls(Callback):
    """Simple callback that counts how often it's been called.
    
    The count can be queried after the simulation from self.counter.
    """
    def __init__(self):
        """Initializes a CountCalls instance."""
```

PR #866, opened 2025-05-19
Make BackendConfig error when receiving unexpected kwargs

### LINKED ISSUE #857: Elevate warning on unexpected arguments in BackendConfig to an error
Currently, when a `BackendConfig` receives unexpected arguments (ie ones not in `BackendConfig._expected_kwargs()`), it raises an warning **if** `BackendConfig._enforce_expected_kwargs is True`. 
Since this is now controlled by the `BackendConfig` subclass, we can elevate it to an error.

HGSilveri:
Closes #857 .

### DIFF
--- a/pulser-core/pulser/register/register.py
+++ b/pulser-core/pulser/register/register.py
@@ -435,6 +435,7 @@ class Register(BaseRegister, RegDrawer):
         kwargs_savefig: dict = {},
         custom_ax: Optional[Axes] = None,
         show: bool = True,
+        draw_empty_sites: bool = False,
     ) -> None:
         """Draws the entire register.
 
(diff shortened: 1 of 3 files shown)


a-corni (approved):
LGTM !

tests/test_backend.py
```diff
@@ -508,8 +508,8 @@ def test_emulation_config():
     try:
         EmulationConfig._enforce_expected_kwargs = True
```
pimvenderbosch, line 509: If you use `mocker.patch` to set this value to `True`, you don't need the `try ... finally` block.

pimvenderbosch (approved):
Looks good! Approved 🎉

tests/test_backend.py
HGSilveri: Good point!

PR #867, opened 2025-05-20
Deny empty sequence for mimic_qpu=True

### LINKED ISSUE #863: QPUs do not accept empty sequences
Currently, in `Backend.validate_sequence(sequecne, mimic_qpu=True)`, we don't raise an error if the sequence is an empty sequence (sequence._empty_sequence is True if the sequence doesn't contain any pulse). However, the QPUs don't accept such Sequences. We should raise an error in `Backend.validate_sequence(sequence, mimic_qpu=True)` if `_empty_sequence=True`.

pimvenderbosch:
As described in #863, our QPUs don't accept empty sequences. Here I've added a check for this in `Backend.validate_sequence`. I've also added the `is_empty` property to conveniently check if the sequence is empty.

Closes #863

### DIFF
--- a/pulser-core/pulser/backend/abc.py
+++ b/pulser-core/pulser/backend/abc.py
@@ -53,10 +53,10 @@ class Backend(ABC):
         if not mimic_qpu:
             return
 
-        if sequence.is_empty:
+        if sequence.is_empty():
             raise ValueError(
-                "'sequence' should not be empty, please add content to it's "
-                "schedule."
+                "'sequence' should not be empty, please add an instruction "
+                "to a declared channel."
             )
 
         if not isinstance(device := sequence.device, Device):
--- a/pulser-core/pulser/backend/abc.py
+++ b/pulser-core/pulser/backend/abc.py
@@ -53,6 +53,12 @@ class Backend(ABC):
         if not mimic_qpu:
             return
 
+        if sequence.is_empty:
+            raise ValueError(
+                "'sequence' should not be empty, please add content to it's "
+                "schedule."
+            )
+
         if not isinstance(device := sequence.device, Device):
             raise TypeError(
                 "To be sent to a QPU, the device of the sequence "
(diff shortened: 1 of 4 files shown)


pulser-core/pulser/sequence/sequence.py
```diff
@@ -326,6 +326,16 @@ def available_channels(self) -> dict[str, Channel]:
...
 
+    @property
+    def is_empty(self) -> bool:
+        """States whether the sequence is empty.
+
+        Note:
+            This is equivalent to testing if the sequence duration is equal to
+            0.
+        """
+        return self.get_duration() == 0
```
HGSilveri, line 337: We have a bunch of `is_...` methods that are not properties, so maybe it would be best if this wasn't a property either

pulser-core/pulser/backend/abc.py
```diff
@@ -53,6 +53,12 @@ def validate_sequence(
         if not mimic_qpu:
             return
 
+        if sequence.is_empty:
+            raise ValueError(
+                "'sequence' should not be empty, please add content to it's "
+                "schedule."
```
HGSilveri, line 59: The notion of "schedule" is internal, how about
```suggestion
                "'sequence' should not be empty, please add an instruction "
                "to a declared channel."
```

tests/test_sequence.py
```diff
@@ -2923,3 +2923,17 @@ def test_sequence_diff(device, parametrized, with_modulation, with_eom):
...
+@pytest.mark.parametrize("channel", ["rydberg_global", "raman_local"])
+def test_sequence_is_empty(reg, device, channel):
+    sequence = Sequence(reg, device)
+    target = (
+        None
+        if channel == "rydberg_global"
+        else sequence._register.qubit_ids[0]
+    )
+    sequence.declare_channel(channel, channel, initial_target=target)
+    assert sequence.is_empty
```
HGSilveri, line 2937: You can check before the channel declaration too

HGSilveri (commented):
Thanks @pimvenderbosch !

HGSilveri (approved):
LGTM!

PR #869, opened 2025-05-21
Bump to v1.6dev0

pimvenderbosch:


pimvenderbosch:
Cannot directly merge from master because a commit needs to be added to bump to `dev` version. Retry is on #870

PR #870, opened 2025-05-21
Bump version to v1.6dev0

pimvenderbosch:


HGSilveri (approved):
LGTM! When you merge, consider changing the message to something like "Bring in changes from v1.5 Release" to avoid having two commits with then same message

PR #871, opened 2025-05-21
Fix implementation of amplitude fluctuation noise

HGSilveri:
- Makes the fluctuations change from run to run instead of pulse to pulse
- Applies the amplitude fluctuations to local channels as well


pulser-simulation/pulser_simulation/hamiltonian.py
```diff
@@ -299,14 +299,15 @@ def add_noise(
...
+                )
                 for slot in ch_samples.slots:
                     add_noise(
                         slot,
                         samples_dict,
                         _ch_obj.addressing == "Global",
-                        amp_fluctuation=max(
-                            0, np.random.normal(1.0, self.config.amp_sigma)
-                        ),
+                        amp_fluctuation=ch_amp_fluctuation,
```
a-corni, line 310: we actually no longer need add_noise to be for each slot, it could be 
add_noise(samples_dict, _ch_obj.addressing == "Global", amp_fluctuation=ch_amp_fluctuation, propagation_dir=_ch_obj.propagation_dir), and have in it samples_dict[qid]["amp"] *= amp_fraction and samples_dict[qid]["det"] += self._doppler_detune[qid]

a-corni (approved):
I have a small comment but otherwise it looks good to me :)

pulser-simulation/pulser_simulation/hamiltonian.py
HGSilveri: Good point!

HGSilveri: Actually, I think it's a bit more complicated than that... For local channels, we are applying the noise only on the pulses applied by the target qubits, which we take from the slots. I'll leave this like this for now

a-corni:
Ah yes sure, then all good, thanks for tackling this !

tests/pulser_simulation/test_simulation.py
```diff
@@ -1701,3 +1701,47 @@ def test_initial_state_sim():
             1e-2,
         )
     )
+
+
+def test_amp_sigma_noise(mod_device):
+    reg = Register({"q0": (0, 0), "q1": (10, 10)})
+    seq = Sequence(reg, mod_device)
+    seq.declare_channel("ch0", "rydberg_global")
+    seq.declare_channel("ch1", "raman_local", initial_target="q0")
```
a-quelle, line 1710: The implementation that Antoine suggested would suffer from the bug that two local channels targeting different qubits at the same timestep would not have their noise factors applied correctly. To test for this, it sounds like a good idea to add a second local channel in the test. I will do the same thing on my end.

tests/pulser_simulation/test_simulation.py
```diff
@@ -1731,17 +1735,46 @@ def test_amp_sigma_noise(mod_device):
...
+    )
+
+    # q1 is acted by both channels
+    expected_q1_dig_samples = noiseless_samples["Local"]["digital"]["q1"][
+        "amp"
+    ]
+    # First pulse on q1 is applied by ch2
+    expected_q1_dig_samples[: pulse1.duration] *= amp_factors["ch2"]
+    # Second pulse on q1 is applied by ch1
+    expected_q1_dig_samples[-pulse1.duration - 1 :] *= amp_factors["ch1"]
```
a-corni, line 1777: Very nice !

PR #872, opened 2025-05-22
Fix Sequence.is_empty() to work on parametrized sequences

HGSilveri:
In #867 , we forgot that `Sequence.get_duration()` can't be called on parametrized sequences. This alternative implementation works for both parametrized and non-parametrized sequences.


pimvenderbosch (approved):
Good catch! LGTM.

a-corni (approved):
Nice catch ! Looks good to me !

PR #877, opened 2025-06-02
[FIX] Enforce consistent dtype in coords

a-corni:
When running
```
import pulser
import numpy as np
reg = pulser.Register.from_coordinates(coords=[np.array([-2.501571, -0.003283], dtype=np.float32), np.array([2.50157 , 0.003283], dtype=np.float32)])
reg.with_automatic_layout(pulser.AnalogDevice)
```
We get an error `ValueError: The coordinate '[-2.5037005  -0.00256802]' is not a part of the RegisterLayout.`.
This is due to the fact that we are comparing exact coordinates in `get_trap_ids_from_coordinates`, and most likely due to an issue when going from an array to a list in generate_trap_coordinates.

This problem is solved when working with float64 arrays. The suggested fix here is then to make the type of the arrays/AbstractArray obtained through the properties of CoordsCollection consistent, make it be always float64.

### DIFF
--- a/tests/test_register.py
+++ b/tests/test_register.py
@@ -50,6 +50,7 @@ def test_creation():
     reg1 = Register(qubits)
     reg2 = Register.from_coordinates(coords, center=False, prefix="q")
     assert np.all(np.array(reg1._coords) == np.array(reg2._coords))
+    assert reg2.sorted_coords.dtype == np.float64
     assert reg1._ids == reg2._ids
 
     reg2b = Register.from_coordinates(coords, center=False, labels=["a", "b"])
@@ -58,9 +59,16 @@ def test_creation():
     with pytest.raises(ValueError, match="Label length"):
         Register.from_coordinates(coords, center=False, labels=["a", "b", "c"])
 
-    reg3 = Register.from_coordinates(np.array(coords), prefix="foo")
+    reg3 = Register.from_coordinates(
+        np.array(coords, dtype=np.float32), prefix="foo"
+    )
     coords_ = np.array([(-0.5, 0), (0.5, 0)])
     assert reg3._ids == ("foo0", "foo1")
+    assert (reg3._coords[0].dtype, reg3._coords[1].dtype) == (
+        np.float32,
+        np.float32,
+    )
+    assert reg3.sorted_coords.dtype == np.float64
     assert np.all(reg3._coords == coords_)
     assert not np.all(coords_ == coords)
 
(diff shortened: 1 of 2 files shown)


HGSilveri (commented):
The error is popping up in `Traps.get_traps_from_coordinates(*coordinates)`, so we should make sure that the input to this method is cast to the same dtype.
```
    def get_traps_from_coordinates(self, *coordinates: ArrayLike) -> list[int]:
        """Finds the trap ID for a given set of trap coordinates.

        Args:
            coordinates: The coordinates to return the trap IDs.

        Returns:
            The list of trap IDs corresponding to the coordinates.
        """
        traps = []
        rounded_coords = np.round(
            np.array(coordinates, dtype=float), decimals=COORD_PRECISION
        )
        ...
```

HGSilveri:
> Also found that `pulser.Register.from_coordinates([torch.tensor([1.0, -4.0], dtype=torch.long), torch.tensor([0.0, 0.0], dtype=torch.long)])` was failing as well

What was the issue here?

pulser-core/pulser/register/traps.py
```diff
@@ -95,7 +95,7 @@ def get_traps_from_coordinates(self, *coordinates: ArrayLike) -> list[int]:
         """
         traps = []
         rounded_coords = np.round(
-            np.array(coordinates), decimals=COORD_PRECISION
+            np.array(coordinates).astype(float), decimals=COORD_PRECISION
```
HGSilveri: nit: here it would be more natural to do
```suggestion
            np.array(coordinates, dtype=float), decimals=COORD_PRECISION
```

a-corni:
RuntimeError                              Traceback (most recent call last)
----> 1 reg = pulser.Register.from_coordinates([torch.tensor([1.0, -4.0], dtype=torch.long), torch.tensor([0.0, 0.0], dtype=torch.long)])

File .\pulser-core\pulser\register\base_register.py:182, in BaseRegister.from_coordinates(cls, coords, center, prefix, labels, **kwargs)
    180 coords_ = pm.vstack(cast(abcSequence, coords))
    181 if center:
--> 182     coords_ = coords_ - pm.mean(coords_, axis=0)  # Centers the array
    183 qubits: dict[str, pm.AbstractArray]
    184 if prefix is not None:

File .\pulser-core\pulser\math\__init__.py:185, in mean(a, axis)
    183 a = AbstractArray(a)
    184 if a.is_tensor:
--> 185     return AbstractArray(torch.mean(a.as_tensor(), dim=axis))
    186 return AbstractArray(np.mean(a.as_array(), axis=axis))

RuntimeError: mean(): could not infer output dtype. Input dtype must be either a floating point or complex dtype. Got: Long

tests/test_register.py
```diff
@@ -599,16 +614,43 @@ def test_register_recipes_torch(
...
+            8,
+        ),
+        (
+            Register.from_coordinates(
+                [
+                    np.array([-2.501571, -0.003283], dtype=np.float32),
+                    np.array([2.50157, 0.003283], dtype=np.float32),
+                    np.array([-2.501571, 5.0], dtype=np.float32),
+                    np.array([2.50157, 5.1], dtype=np.float32),
+                ]
```
HGSilveri, line 639: Could be nice to check what happens if not all coordinates have the same dtype

PR #880, opened 2025-06-06
Add detuning fluctuations to the noise model

a-quelle:
I propose we add the following field to the `NoiseModel`, based on the investigations by the simulation team. This is independent of the high frequency noise we discussed previously, because it only depends on a single standard deviation. It would function similarly to the `amp_sigma` parameter, except that the noise is additive, rather than multiplicative.

The pull request is a draft, because I first wanted to gather some opinions before taking the time to fix all the tests, and add it to pulser-simulation.


pulser-core/pulser/noise_model.py
```diff
@@ -93,6 +96,7 @@
     "dephasing_rate": 0.05,
     "hyperfine_dephasing_rate": 1e-3,
     "depolarizing_rate": 0.05,
+    "detuning_sigma": 0.0,
```
HGSilveri: No need for a legacy default as this is not a legacy parameter

pulser-core/pulser/noise_model.py
```diff
@@ -128,6 +132,8 @@ class NoiseModel:
     - **amplitude**: Gaussian damping due to finite laser waist and
       laser amplitude fluctuations. Parametrized by ``laser_waist``
       and ``amp_sigma``.
+    - **detuning**: Dephasing fluctuations, parametrized by
+      ``dephasing_sigma``.
```
HGSilveri: Sometimes you call it `dephasing_sigma`, other `detuning_sigma`. We should pick one, I think `detuning_sigma` might be better

pulser-core/pulser/noise_model.py
```diff
@@ -150,6 +156,11 @@ class NoiseModel:
             run to run as a standard deviation of a normal distribution
             centered in 1. Assumed to be the same for all channels (though
             each channel has its own randomly sampled value in each run).
+        detuning_sigma: Dictates the fluctuation in detuning of a channel from
+            run to run as a standard deviation of a normal distribution
+            centered in 0. Assumed to be the same for all channels (though
+            each channel has its own randomly sampled value in each run).
+        relaxation_rate: The rate of relaxation from the Rydberg to the
```
HGSilveri: This line got duplicated

pulser-core/pulser/noise_model.py
```diff
@@ -150,6 +156,11 @@ class NoiseModel:
             run to run as a standard deviation of a normal distribution
             centered in 1. Assumed to be the same for all channels (though
             each channel has its own randomly sampled value in each run).
+        detuning_sigma: Dictates the fluctuation in detuning of a channel from
+            run to run as a standard deviation of a normal distribution
+            centered in 0. Assumed to be the same for all channels (though
+            each channel has its own randomly sampled value in each run).
```
HGSilveri: We have to distinguish somehow that this is additive and not multiplicative

pulser-core/pulser/noise_model.py
```diff
@@ -76,6 +78,7 @@
     "p_false_pos",
     "p_false_neg",
     "amp_sigma",
+    "detuning_sigma",
```
HGSilveri: Since the detuning deviation is additive, does it actually make sense that it's probability like, ie in  [0,1]? Rather, it should just be a positive number, no?

pulser-core/pulser/noise_model.py
```diff
@@ -287,6 +303,7 @@ def _find_relevant_params(
                 nt_ == "doppler"
                 or (nt_ == "amplitude" and amp_sigma != 0.0)
                 or (nt_ == "SPAM" and state_prep_error != 0.0)
+                or (nt_ == "detuning" and detuning_sigma != 0.0)
```
HGSilveri: `detuning` noise should not be active if `detuning_sigma` is 0, so you can forgo asking for the `detuning_sigma` value and just check if `detuning` is in `noise_types` (like is done for `doppler`)

pulser-simulation/pulser_simulation/simconfig.py
```diff
@@ -112,6 +115,7 @@ class SimConfig:
     temperature: float = _LEGACY_DEFAULTS["temperature"]
     laser_waist: float = _LEGACY_DEFAULTS["laser_waist"]
     amp_sigma: float = _LEGACY_DEFAULTS["amp_sigma"]
+    detuning_sigma: float = _LEGACY_DEFAULTS["detuning_sigma"]
```
HGSilveri: Just put this here:
```suggestion
    detuning_sigma: float = 0.
```

pulser-simulation/pulser_simulation/simconfig.py
```diff
@@ -103,6 +104,8 @@ class SimConfig:
             pulses.
         amp_sigma: Dictates the fluctuations in amplitude as a standard
             deviation of a normal distribution centered in 1.
+        detuning_sigma: Dictates the fluctuations in detuning as a standard
+            deviation of a normal distribution centered in 0.
```
HGSilveri, line 108: Side note: it's about time we stop carrying around `SimConfig`...

pulser-core/pulser/noise_model.py
a-corni: If it is additive, then I guess you have to put units (in rad/µs)

pulser-core/pulser/noise_model.py
a-quelle: totally right. I initially called it `dephasing_sigma` due to a brainfart, and then realized I was wrong, and changed it, but not everywhere it seems.

pulser-core/pulser/noise_model.py
a-quelle: but why is `amp_sigma` probability like then? That is also just a positive number characterizing a distribution, not an actual probability. I put it there because `amps_sigma` was there.

pulser-simulation/pulser_simulation/simconfig.py
a-quelle: I'll take your suggestions on what to do with this, then. I agree that we need to get rid of it eventually, but what would be a good strategy for getting there? Don't people still depend on this? Is it deprecated yet?

HGSilveri: Not for this PR, but the idea would be to deprecate direct access to `QutipEmulator` and `SimConfig` and enforce usage of `QutipBackendV2` and `NoiseModel`

pulser-core/pulser/noise_model.py
HGSilveri: Because `amp_sigma` is multiplicative, so we just sample an adimensional normal distribution centered in 1 to get the multiplicative factor. In reality, being close to 1 is already problematic because you potentially run into having negative multiplicative factors which are nonsensical, but we didn't have a good justification for imposing an arbitrary upper-bound between 0 and 1.

HGSilveri: That being said, we do clip the sampled multiplicative factor to always be non-negative: https://github.com/pasqal-io/Pulser/blob/944b954ad6aee66d2385b1f0c8c0fbedfbd6343a/pulser-simulation/pulser_simulation/hamiltonian.py#L303

pulser-simulation/pulser_simulation/hamiltonian.py
```diff
@@ -302,12 +307,18 @@ def add_noise(
                 ch_amp_fluctuation = max(
                     0, np.random.normal(1.0, self.config.amp_sigma)
                 )
+                ch_det_fluctuation = (
+                    np.random.normal(0.0, self.config.detuning_sigma)
+                    if self.config.detuning_sigma
+                    else 0.0
+                )
```
HGSilveri, line 314: nit: Since `detuning_sigma` is a float, I think this would suffice
```suggestion
                ch_det_fluctuation = np.random.normal(0.0, self.config.detuning_sigma)
```

HGSilveri (approved):
LGTM! 
I have one nit and, if you don't mind, I'll change the PR title to keep the history nice and clean

pulser-core/pulser/json/abstract_repr/schemas/noise-schema.json
```diff
@@ -119,6 +122,7 @@
         "temperature",
         "laser_waist",
         "amp_sigma",
+        "detuning_sigma",
```
HGSilveri: The fact that this is marked as `required` tells me you updated the schema manually instead of taking the one generated from `pulser-schema`

pulser-core/pulser/noise_model.py
```diff
@@ -240,8 +255,8 @@ def to_tuple(obj: tuple) -> tuple:
         )
 
         # Get rid of unnecessary None's
-        for p_ in _POSITIVE | _PROBABILITY_LIKE:
-            param_vals[p_] = param_vals[p_] or 0.0
+        for p_ in param_vals.keys():
+            param_vals[p_] = param_vals[p_] or self._get_default_value(p_)
```
HGSilveri: You should rather do this to avoid overwriting non-`None` values

```suggestion
        for p_ in param_vals.keys():
            param_vals[p_] = param_vals[p_] or self._get_default_value(p_)
```
```suggestion
        for p_, val in param_vals.items():
            param_vals[p_] = self._get_default_value(p_) if val is None else val 
```

PR #888, opened 2025-06-18
Move the mimic-QPU validation to RemoteBackend.run()

MohamedAminMallek:
Move mimic-qpu validation from PasqalCloud to RemoteBackend

Potential breaking change:

If there are any connections that use mimic-qpu=true but do not implement the `fetch_available_devices` method, this change may cause the run call to fail

### DIFF
--- a/pulser-core/pulser/backend/qpu.py
+++ b/pulser-core/pulser/backend/qpu.py
@@ -64,23 +64,3 @@ class QPUBackend(RemoteBackend):
             job_params or [], self._sequence.device.max_runs
         )
         return super().run(job_params, wait)
-
-    @staticmethod
-    def validate_job_params(
-        job_params: list[JobParams], max_runs: int | None
-    ) -> None:
-        """Validates a list of job parameters prior to submission."""
-        suffix = " when executing a sequence on a real QPU."
-        if not job_params:
-            raise ValueError("'job_params' must be specified" + suffix)
-        RemoteBackend._type_check_job_params(job_params)
-        for j in job_params:
-            if "runs" not in j:
-                raise ValueError(
-                    "All elements of 'job_params' must specify 'runs'" + suffix
-                )
-            if max_runs is not None and j["runs"] > max_runs:
-                raise ValueError(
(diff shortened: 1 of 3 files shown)


pulser-core/pulser/backend/qpu.py
```diff
@@ -64,23 +64,3 @@ def run(
             job_params or [], self._sequence.device.max_runs
         )
```
HGSilveri: I guess you can get rid of this

pulser-core/pulser/backend/remote.py
```diff
@@ -339,6 +339,10 @@ def run(
             The results, which can be accessed once all sequences have been
             successfully executed.
         """
+        if self._mimic_qpu:
+            sequence = self._connection.update_sequence_device(self._sequence)
```
HGSilveri, line 352: Indeed, `fetch_available_device()` is not an abstract method so we can't guarantee this will be implemented. How about putting it in a `try...except` block?
```suggestion
            try:
                sequence = self._connection.update_sequence_device(self._sequence)
            except NotImplementedError:
                pass
```

HGSilveri (commented):
Very nice!

pulser-core/pulser/backend/remote.py
MatthieuMoreau0: Maybe with a warning if we think that's relevant for the user?

HGSilveri: What would be a relevant warning message for the user though?

MatthieuMoreau0: So I am not too sure but I guess it could be akin to say that the mimic_qpu device update didn't complete because available devices could not be fetched? but not much action the user could take eitherway so maybe it's more confusing than helpful tbh

HGSilveri: Perhaps something like
```
The chosen connection does not give access to the latest device specs; beware that execution might still fail if the specs in the chosen sequence are incompatible with the device's.
```
?
(The wording is likely poor, do not use as is)

PR #889, opened 2025-06-18
Avoid inconsistent behavior on operations between numpy arrays and parametrized objects

HGSilveri:
We have had a long standing issue with binary operations on abstract objects where the order of the arguments influenced the outcomes. 

For example,
```
>>> var = Variable("var", dtype=float, size=2)
>>> arr = np.arange(2)

>>> var + arr   # As expected
<pulser.parametrized.paramobj.ParamObj object at 0x7521dce93df0>

>>> arr + var   # An array of parametrized objects, not what we want
array([<pulser.parametrized.paramobj.ParamObj object at 0x7521dce92e30>,
       <pulser.parametrized.paramobj.ParamObj object at 0x7521dce93cd0>],
      dtype=object)
```

After some digging, it seems this can be avoided by writing a custom `__array_ufunc__` for the right-hand side object, so that's what we do here.

### DIFF
--- a/pulser-core/pulser/parametrized/paramobj.py
+++ b/pulser-core/pulser/parametrized/paramobj.py
@@ -39,9 +39,62 @@ from pulser.parametrized import Parametrized
 if TYPE_CHECKING:
     from pulser.parametrized import Variable
 
+# The full list of numpy ufuncs can be found at
+# https://numpy.org/devdocs//reference/ufuncs.html#math-operations
+
+# Mapping between ufunc and OpSupport method names, for those
+# that don't share the same name or require a dunder method
+# Note that for binary method we use the reverse method and invert
+# the input order accordingly
+_UFUNC_MAP = {
+    # Binary methods
+    "add": "add",
+    "subtract": "sub",
+    "multiply": "mul",
+    "divide": "truediv",
+    "true_divide": "truediv",
+    "floor_divide": "floordiv",
+    "power": "pow",
+    "float_power": "pow",
+    "remainder": "mod",
+    "mod": "mod",
+    "fmod": "mod",
+    # Special unary methods
+    "negative": "neg",
+    "absolute": "abs",
+    "fabs": "abs",
+    "floor": "floor",
+    "ceil": "ceil",
+}
+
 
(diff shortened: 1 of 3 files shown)


pulser-core/pulser/parametrized/paramobj.py
```diff
@@ -39,10 +39,63 @@
...
+        if method != "__call__" or len(inputs) > 2:
+            return NotImplemented
+
+        ufunc_name = ufunc.__name__
+        if ufunc_name in _UFUNC_MAP:
+            root_name = _UFUNC_MAP[ufunc_name]
+            if len(inputs) == 2 and inputs[1] is self:
+                # Use reverse method and inputs
+                root_name = "r" + root_name
+                inputs = inputs[::-1]
```
a-corni, line 86: I have trouble understanding this reverse of the inputs... Your inputs are supposed to be a tuple of lenth 1, right ?

pulser-core/pulser/parametrized/paramobj.py
```diff
@@ -39,10 +39,63 @@
...
+            method_name = f"__{root_name}__"
+        else:
+            method_name = ufunc_name
+
+        if inputs[0] is self:  # Make sure self is involved
```
a-corni, line 91: Can inputs[0] be different from self ? I have the feeling that it's an assertion

HGSilveri: It can probably be an assertion, I'm just playing it safe and letting it fallback to the `return NotImplemented` in case I'm missing something

pulser-core/pulser/parametrized/paramobj.py
HGSilveri: When it's a binary method you have two inputs (as checked in the `if` just above)

pulser-core/pulser/parametrized/paramobj.py
a-corni: Then I think you can't test it, you should add a # pragma: no cover

a-corni (approved):
LGTM !

PR #890, opened 2025-06-20
Adopt the parameters of the default backend config that are not provided

HGSilveri:
This will allow e.g. an Emu-MPS backend to define a default config with MPS specific parameters that are not in `EmulationConfig`'s signature and are adopted unless otherwise specifically included in the `config` given to `EmulatorBackend`.


tests/test_backend.py
```diff
@@ -389,7 +391,12 @@ def run(self):
         ),
     )
     assert emu._config.default_evaluation_times == "Full"
+    # with_modulation is not True because EmulationConfig has it in the
+    # signature as `with_modulation=False``
```
a-corni, line 405: Ah that's a shame... Should this signature be `with_modulation: bool | None = None` -> the behaviour will be kept, since we do `not with_modulation` or `with_modulation`, and we can check whether the user defined it or not 

I guess we could have all the attributes of EmulatorConfig as optional (and associate to a None value a default value in the code)

HGSilveri: That would make the default values obscure and I'm afraid it might be a bit late for that now...

a-corni: Definitely shady, but we could store the initial input in private variable, and expose only non-None value in public variable (what really matters for the user is the description and the public variables, not what was the value initially assigned to it)

a-corni (approved):
From our discussion, I approve the MR

PR #891, opened 2025-06-23
Handle when Results.atom_order entries are not strings

HGSilveri:
Even though using non-str qubit IDs is deprecated, it is still allowed and we were not handling it properly when serializing `Results`. This fixes that and also makes the `DeprecationWarning` to not use `int`s as qubit IDs louder.


a-corni (approved):
LGTM !

PR #892, opened 2025-06-25
Properly deserialize complex values in Results

HGSilveri:
Since the outcomes of some observables might be complex values (e.g. the operator in `Expecation` is non-hermitian or there is a residual imaginary part), we should properly deserialize them.


a-corni (approved):
LGTM !

PR #898, opened 2025-07-02
Add method to truncate a Sequence at a specific time

### LINKED ISSUE #873: Add method to truncate a Sequence's or channel's contents
When studying dynamics on a QPU, one has to truncate the Hamiltonian at different times. This is currently cumbersome to do outside of very simple cases (e.g. with constant waveforms) using a single parametrized sequence.

I propose we add a new `Sequence.truncate()` method that allows truncating a sequence at a given duration, which can be a given as a variable. Similar to `Sequence.get_duration()`, we can also give an optional `channel` argument that would truncate a single channel's contents rather than the entire sequence's. So the method's signature could be

```python
Sequence.truncate(duration: int | Parametrized, channel: str | None = None) -> None
```  

The way I see this working is:

1. For each channel, find the slot that includes the truncation threshold
2. Modify this slot to have the desired duration
3. Replace this slot and get rid of all the following slots

**Some things left to clarify:**

- What if the `duration` is not a multiple of the channel's clock period? Should we round up like we do for delays and pulses, or round down so as to not go over the `duration`?
- What if the truncated slot is a "target" slot? **Raising a `NotImplementedError` sounds reasonable.**
- What if the truncated slot ends up with a duration smaller than the channel's minimum instruction duration? Should we just discard it (easy) or try to merge it with the previous slot (hard, potentially messy)?
- What if the slot is an EOM start or end buffer? Within the EOM block, I think we can truncate normally but the start and end buffers shouldn't be messed with. Should we leave them be silently? Raise a warning? Or an error?

HGSilveri:
Note: Implemented for all declared channels only

Closes #873 .

### DIFF
--- a/pulser-core/pulser/sequence/_basis_ref.py
+++ b/pulser-core/pulser/sequence/_basis_ref.py
@@ -22,19 +22,23 @@ import numpy as np
 class _QubitRef:
     def __init__(self) -> None:
         self.phase = _PhaseTracker(0)
-        self.last_used = 0
+        self._last_used_history = [0]
+
+    @property
+    def last_used(self) -> int:
+        return max(self._last_used_history)
 
     def increment_phase(self, phi: float) -> None:
         self.phase[self.last_used] = self.phase.last_phase + phi
 
     def update_last_used(self, new_t: int) -> None:
-        self.last_used = max(self.last_used, new_t)
+        self._last_used_history.append(new_t)
 
     def truncate(self, t: int) -> None:
         self.phase.truncate(t)
-        self.last_used = (
-            self.phase.last_time if t < self.last_used else self.last_used
-        )
+        self._last_used_history = [
+            t_ for t_ in self._last_used_history if t_ <= t
+        ]
 
 
 class _PhaseTracker:
--- a/pulser-core/pulser/sequence/_basis_ref.py
+++ b/pulser-core/pulser/sequence/_basis_ref.py
@@ -30,6 +30,12 @@ class _QubitRef:
     def update_last_used(self, new_t: int) -> None:
         self.last_used = max(self.last_used, new_t)
 
+    def truncate(self, t: int) -> None:
+        self.phase.truncate(t)
+        self.last_used = (
+            self.phase.last_time if t < self.last_used else self.last_used
+        )
+
 
 class _PhaseTracker:
     """Tracks a phase reference over time."""
@@ -59,6 +65,17 @@ class _PhaseTracker:
         for i in range(start, end):
             change = self._phases[i] - self._phases[i - 1]
             yield (self._times[i] / time_scale, change)
+
+    def truncate(self, threshold: int) -> None:
+        new_times = []
+        new_phases = []
+        for t, ph in zip(self._times, self._phases):
+            if t > threshold:
+                break
+            new_times.append(t)
+            new_phases.append(ph)
+        self._times = new_times
+        self._phases = new_phases
 
     def _format(self, phi: float) -> float:
         return phi % (2 * np.pi)
--- a/pulser-core/pulser/sequence/_schedule.py
+++ b/pulser-core/pulser/sequence/_schedule.py
@@ -535,6 +535,27 @@ class _Schedule(Dict[str, _ChannelSchedule]):
         )
 
     def truncate(self, duration: int) -> None:
+        def adjust_eom_blocks(
+            threshold: int, ch_schedule: _ChannelSchedule
+        ) -> None:
+            for eom_ind, eom_block in enumerate(ch_schedule.eom_blocks):
+                # Threshold falls within an EOM block
+                if (
+                    eom_block.ti < threshold <= (eom_block.tf or threshold)
+                ):  # tf can be None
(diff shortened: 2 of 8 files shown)


pulser-core/pulser/sequence/sequence.py
```diff
@@ -1494,7 +1494,49 @@ def estimate_added_delay(
...
+            return
+
+        # Adjust the phase reference of all qubits
+        for basis_ref in self._basis_ref.values():
+            for qubit_ref in basis_ref.values():
+                qubit_ref.truncate(duration_)
+        self._schedule.truncate(duration_)
+
+    @seq_decorators.store
+    @seq_decorators.conditionally_block(if_parametrized_truncated=False)
```
a-corni, line 1539: 👍

pulser-core/pulser/sequence/_basis_ref.py
```diff
@@ -60,6 +70,17 @@ def changes(
             change = self._phases[i] - self._phases[i - 1]
             yield (self._times[i] / time_scale, change)
 
+    def truncate(self, threshold: int) -> None:
+        new_times = []
+        new_phases = []
+        for t, ph in zip(self._times, self._phases):
+            if t > threshold:
+                break
```
a-corni: could the times be unordered ? Won't it be better to have `continue` instead of a `break` ?

pulser-core/pulser/sequence/_schedule.py
```diff
@@ -534,6 +534,132 @@ def add_target(self, qubits_set: set[QubitId], channel: str) -> None:
...
+            assert (
+                ch_schedule.channel_obj.min_duration <= threshold <= duration
+            )
+            # Find the slot where to truncate
+            for slot_ind, slot in enumerate(all_slots):
+                if slot.ti < threshold <= slot.tf:
+                    break
+
+            # If the slot terminates at the threshold, just return
+            # the schedule up to this slot (inclusivly)
```
a-corni: nit (sorry to notice it): inclusively

pulser-core/pulser/sequence/_schedule.py
```diff
@@ -534,6 +534,132 @@ def add_target(self, qubits_set: set[QubitId], channel: str) -> None:
...
+                    f"at t = {threshold} ns was removed by a "
+                    "'truncate()' call.",
+                    stacklevel=3,
+                )
+                adjust_eom_blocks(threshold, ch_schedule)
+                # EOM end buffer, so we remove it and reopen EOM mode
+                ch_schedule.eom_blocks[-1] = replace(
+                    ch_schedule.eom_blocks[-1], tf=None
+                )
+                continue
```
a-corni, line 623: It seems to me that these two assume that the modulation bandwidth of the EOM config or at least of the channel were defined. If not, then there is no delay slot added before/after the EOM block by enable/disable_eom_mode (IIUC).

pulser-core/pulser/sequence/_schedule.py
```diff
@@ -534,6 +534,132 @@ def add_target(self, qubits_set: set[QubitId], channel: str) -> None:
...
+            adjust_eom_blocks(threshold, ch_schedule)
+
+            if slot.type == "target":
+                warnings.warn(
+                    f"'target()' instruction on channel {ch_name!r} at "
+                    f"t = {threshold} ns was removed by a "
+                    "'truncate()' call.",
+                    stacklevel=3,
+                )
+                continue
```
a-corni, line 635: If someone calls truncate with duration=target.tf, then the target slot is kept (even if target.tf=target.ti), perhaps we should have an exclusion line 584 `if slot.tf == threshold and slot.type !="target":`

pulser-core/pulser/sequence/_basis_ref.py
HGSilveri: That is safer indeed :)

pulser-core/pulser/sequence/_schedule.py
HGSilveri: I believe this is not a possibility, the modulation bandwidth must always be defined if an `eom_config` is present

pulser-core/pulser/sequence/_schedule.py
HGSilveri: Why is this an issue though? If you truncate right at the moment the target ends, then it should be kept no?

HGSilveri (commented):
Thanks for the comments!

pulser-core/pulser/sequence/_schedule.py
a-corni: But what if I want to truncate without keeping a target operation of duration 0 ? If I get it correctly, we will have slot.ti = slot.tf = slots[index-1].tf, so I would have to truncate at slots[index-1] -1 which will be then rounded down due to clock issues...

HGSilveri: I don't think you should use `truncate` as a way to delete a `target` operation, that's not the intention of this method. We only delete the target operation when we have to, so a target of duration 0 never has to be deleted

a-corni: Just thought about it, my case is already solved.

PR #904, opened 2025-07-08
Deprecate usage of  samples_per_run

a-corni:
- samples_per_run in noise_model can't have a value different than 1 (or it's default None).
- samples_per_run in SimConfig doesn't need to be deprecated, because SimConfig is deprecated.
- Checking that samples_per_run is not different than 1 in QutipEmulator is not needed, because it will be deprecated.
- QutipConfig already warns that samples_per_run different than 1 and None will not be used in the simulation.


pablolh (approved):

pulser-core/pulser/noise_model.py
```diff
@@ -422,6 +423,17 @@ def _validate_parameters(param_vals: dict[str, Any]) -> None:
                 comp = "a boolean"
             if not is_valid:
                 raise ValueError(f"'{param}' must be {comp}, not {value}.")
+            if param == "samples_per_run" and value != 1:
+                with warnings.catch_warnings():
+                    warnings.filterwarnings("always")
```
HGSilveri: Maybe "once" is enough
```suggestion
                    warnings.filterwarnings("once")
```

pulser-core/pulser/noise_model.py
```diff
@@ -145,7 +145,8 @@ class NoiseModel:
             bitstring distribution is sampled when calculating bitstring
             counts.
         samples_per_run: Number of samples per noisy Hamiltonian. Useful
-            for cutting down on computing time, but unrealistic.
+            for cutting down on computing time, but unrealistic. *Deprecated
+            since v1.6, use only `runs`.*
```
HGSilveri, line 150: Now that this is deprecated, it should default to 1 when it is needed (ie when there is stochastic noise). Otherwise it will still be requested which is very annoying

pulser-core/pulser/noise_model.py
HGSilveri: Thinking about it some more, I'm not sure it even needs to be this loud. A standard `DeprecationWarning` without the filter might be enough and it would be easier to handle in the tests (ie you could use module / function level `pytest.mark.filerwarnings` instead of a bunch of `with deprecated_call()`). Wdyt?

pulser-core/pulser/noise_model.py
a-corni: I came to this conclusion this morning ;)

pulser-core/pulser/noise_model.py
```diff
@@ -257,9 +260,14 @@ def to_tuple(obj: tuple) -> tuple:
...
-                self._get_default_value(p_) if val is None else val
-            )
+            if p_ in _PROBABILITY_LIKE | _POSITIVE and not isinstance(
+                val, (int, float)
+            ):
+                raise TypeError(
+                    f"Value for {p_} should be positive, "
+                    f"{'between 0 and 1, ' if p_ in _PROBABILITY_LIKE else ''}"
+                    f"not {val}."
+                )
```
HGSilveri: The value is being checked in `_validate_parameters()`, I think the message here should stick to the type check

pulser-core/pulser/noise_model.py
```diff
@@ -201,18 +204,18 @@ class NoiseModel:
...
-        relaxation_rate: float | None = None,
-        dephasing_rate: float | None = None,
-        hyperfine_dephasing_rate: float | None = None,
-        depolarizing_rate: float | None = None,
+        amp_sigma: float = 0.0,
+        detuning_sigma: float = 0.0,
+        relaxation_rate: float = 0.0,
+        dephasing_rate: float = 0.0,
+        hyperfine_dephasing_rate: float = 0.0,
+        depolarizing_rate: float = 0.0,
```
HGSilveri: Not sure it's worth it, but did you consider/try to go back to the usual `dataclass` syntax (ie going back to `init=True`)?

a-corni: That makes the code clearer ;)

pulser-core/pulser/noise_model.py
```diff
@@ -183,68 +183,49 @@ class NoiseModel:
...
+        param_vals["eff_noise_opers"] = to_tuple(self.eff_noise_opers)
+
+        # Checking the type of provided positive and probability parameters
+        for p_, val in param_vals.items():
+            if p_ in _PROBABILITY_LIKE | _POSITIVE and not isinstance(
+                val, (int, float)
+            ):
+                raise TypeError(
+                    f"Type for {p_} should be a float or int, not {type(val)}"
+                )
```
HGSilveri: Thinking about it some more, I wonder whether we want to do this type check. It will unnecessarily fail for numpy types and such, so I'd say it's too strict.
If it is not done, I think we'll just fail unexpectedly in `_validate_parameters()`, which is not perfect but not so bad either (in fact, it is what was happening before).
If we really want to keep it, I'd try casting to `float` in a `try...except` block instead

pulser-core/pulser/noise_model.py
```diff
@@ -430,7 +414,9 @@ def _to_abstract_repr(self) -> dict[str, Any]:
         eff_noise_opers = all_fields.pop("eff_noise_opers")
         all_fields["eff_noise"] = list(zip(eff_noise_rates, eff_noise_opers))
         for p in OPTIONAL_IN_ABSTR_REPR:
-            if all_fields[p] == self._get_default_value(p):
+            if (
+                p in _POSITIVE | _PROBABILITY_LIKE and all_fields[p] == 0
+            ) or all_fields[p] is None:
```
HGSilveri: You can now also take advantage of the dataclass defaults here :)

pulser-core/pulser/noise_model.py
a-corni: Yes I was also thinking about the numpy issues... I also consider it too restrictive. My issue is that _validate_parameters is only used to check that the defined parameters are of good type, so ultimately one could define a NoiseModel with a parameter set to None, and nothing would happen, it would be serialized/deserialized to None...
I am also missing testing the strictly positive parameters... Either I skip this, or I design a try...except block

HGSilveri: Just go for the try...except :)

pulser-core/pulser/noise_model.py
```diff
@@ -219,12 +219,14 @@ def to_tuple(obj: tuple) -> tuple:
...
-            if p_ in _PROBABILITY_LIKE | _POSITIVE and not isinstance(
-                val, (int, float)
-            ):
-                raise TypeError(
-                    f"Type for {p_} should be a float or int, not {type(val)}"
-                )
+            if p_ in _PROBABILITY_LIKE | _POSITIVE:
+                try:
+                    param_vals[p_] = float(val)
+                except Exception:
```
HGSilveri: I think it will always be a `TypeError` if it fails, no need to be so broad

pulser-core/pulser/noise_model.py
```diff
@@ -219,12 +219,14 @@ def to_tuple(obj: tuple) -> tuple:
...
+                    raise TypeError(
+                        f"Type for {p_} should be castable to float, not"
+                        f" {type(val)}."
+                    )
```
```suggestion
                    raise TypeError(
                        f"{p_} should be castable to float, not of type"
                        f" {type(val)}."
                    )
```

pulser-core/pulser/noise_model.py
```diff
@@ -408,16 +410,16 @@ def _validate_parameters(param_vals: dict[str, Any]) -> None:
                 )
 
     def _to_abstract_repr(self) -> dict[str, Any]:
-        all_fields = asdict(self)
+        all_fields = {}
+        for f in fields(self):
+            value = getattr(self, f.name)
+            if f.name in OPTIONAL_IN_ABSTR_REPR and f.default == value:
```
HGSilveri: Not relevant for now but this will fail if the field uses a `default_factory` instead of a `default`. You can use `pulser.json.utils.get_dataclass_defaults()` instead

pulser-core/pulser/noise_model.py
a-corni: There seems to ValueErrors as well... https://stackoverflow.com/questions/76120890/how-to-find-all-the-possible-exceptions-raised-by-float

a-corni: But it seems to be the only two exceptions

HGSilveri: My bad

a-corni: Actually, there is also OverflowException and others... Okay to leave it with Exception ?

pulser-core/pulser/noise_model.py
a-corni: Yeah I checked that there was no default_factory (such as list or arrays, right ?), everything is tuples. I am bringing this change.

HGSilveri:
#912 should fix the tests

pulser-core/pulser/noise_model.py
HGSilveri: Sure

tests/pulser_simulation/test_simulation.py
```diff
@@ -789,12 +789,6 @@ def test_config(matrices):
     )
 
 
-@pytest.mark.skipif(
-    np.lib.NumpyVersion(np.__version__) >= "2.0.0",
-    reason="Random number generation doesn't produce the same results in numpy"
-    " >v2",
-)
-@pytest.mark.filterwarnings("ignore:Setting samples_per_run different to 1 is")
```
a-corni: @HGSilveri I think the tests were failing only because I accidently deleted this filterwarning and the one line 1360.

PR #906, opened 2025-07-09
Improve exceptions in Sequence.switch_device()

HGSilveri:
- Fixes issue where the error message was not appearing
- Creates and introduces a custom `SwitchDeviceError`
- Explicitly checks for compatibility of the register with the new device


a-corni (approved):
LGTM :)

Issue #875, opened 2025-05-30
Update documentation once all the backends follow the new interface


HGSilveri:
## Requires

- https://github.com/pasqal-io/pasqal-cloud/issues/178
- https://github.com/pasqal-io/pasqal-cloud/issues/187
- #850 

## Tasks

- [x] Update `pulser.backends` to give access to all the new backends
- [x] Update the [Backend Execution page](https://pulser.readthedocs.io/en/stable/tutorials/backends.html) to refer only the latest backends
- [x] Add a dedicated tutorial for QPU execution
- [x] Replace usage of deprecated backend classes throughout the docs and tutorials with the new backends

a-quelle-pasqal:
Isn't the final bullet point already done? And the third is #970

PR #907, opened 2025-07-09
Deprecate QutipBackend

### LINKED ISSUE #850: Deprecate QutipBackend
QutipBackendV2 uses new generic API from `pulser.backend` and should be preferred over QutipBackend

HGSilveri:
- Deprecates `QutipBackend`
- Replaces it with `QutipBackendV2` on all tutorials **except for the Backend Execution page**, which will be tackled in #875 

Closes #850 .


pulser-core/pulser/backends.py
```diff
@@ -37,7 +38,7 @@
 if TYPE_CHECKING:
     from pulser.backend import QPUBackend as QPUBackend
     from pulser.backend.abc import Backend
-    from pulser_simulation import QutipBackend as QutipBackend
+    from pulser_simulation import QutipBackendV2 as QutipBackendV2
```
a-corni, line 41: Is this import really needed ? Same for QPUBackend...

HGSilveri: Not really needed, but it makes the class appear in the IDE when you do `pulser.backends.Q...` so it's nice.

PR #910, opened 2025-07-16
Fix emulation with state preparation errors on Numpy 2

HGSilveri:
When support for Numpy 2 was added in #837 , there was some mismatch between v1 and v2 results that was attributed to differences in the random number generation. Turns out it was actually a bug in how the "bad atoms" during state preparation were identified, which is fixed here.

This also explains the incorrect plots reported in #886, which we didn't manage to replicate because we all had Numpy 1 locally.


a-corni (approved):
I don't understand why specifying the dtype is making it work, but it looks good to me :)

a-corni:
Ah just ran it with numpy v2, it was converting "0" -> True and "1" -> True (I guess only checking if the string is empty or not, which makes sense

PR #912, opened 2025-07-18
Deal with Pillow 11.3 deprecation warning

HGSilveri:
- This should fix the failing tests from a `DeprecationWarning` raised by Pillow 1.3 due to usage inside `matplotlib`.
- `matplotlib 3.10.4` will handle this but it is not yet released
- It won't be dealt with for Python 3.9 because `matplotlib` stopped supporting it after v3.9


a-corni (approved):
LGTM !

PR #913, opened 2025-07-29
results aggregation

a-quelle-pasqal:
Ported the aggregation functionality from `emu-base` to the pulser backends.

There is now a `Results.aggregate` function. It works as follows:
Each `Observable` now has a `default_aggregation_type` property, which it stores in `Results` together with its results. This aggregation method is then used by `Results.aggregate` to aggregate the results. The user has the option to override the default behaviour.

The default behavior is:
- the mean of each result is taken for all results except `BitStrings` and `StateResult`
- for `BitStrings` the counters are joined, so 10 runs with 1000 bitstrings would yould an aggregated 10000 bitstrings
- for `StateResult` the default aggregator of `None` is kept, meaning that aggregation is skipped here.

Regarding the `StateResult`, the most sound aggregator in my opinion would be to sum them into a mixed state. However, the implementation would be sufficiently tricky that I decided to skip it for now.

I had to bump the version of python used in readthedocs, since the docstrings for `int` have a bug in them that causes the readthedocs to fail for python < 3.13. It is trying to access these docstrings because I use `IntEnum`

### DIFF
--- a/pulser-core/pulser/backend/results.py
+++ b/pulser-core/pulser/backend/results.py
@@ -220,13 +220,10 @@ class Results:
             results._results[uuid.UUID(key)] = deserialize_complex(value)
         for key, value in obj["times"].items():
             results._times[uuid.UUID(key)] = value
-        if (
-            "aggregation_methods" in obj
-        ):  # optional for backwards compatibility
-            for key, value in obj["aggregation_methods"].items():
-                results._aggregation_methods[uuid.UUID(key)] = (
-                    AggregationMethod(value)
-                )
+        for key, value in obj.get("aggregation_methods", {}).items():
+            results._aggregation_methods[uuid.UUID(key)] = AggregationMethod(
+                value
+            )
         return results
 
     def to_abstract_repr(self, skip_validation: bool = False) -> str:


pulser-core/pulser/backend/aggregators.py
```diff
@@ -0,0 +1,98 @@
...
+        acc = pm.torch.zeros_like(values[0])
+        for ten in values:
+            acc += ten
+        return acc / len(values)
+
+    if element_type == np.ndarray:
+        acc = np.zeros_like(values[0])
+        for ar in values:
+            acc += ar
+        return acc / len(values)
```
kbidzhiev, line 56: some crazy python thing

```python
if element_type in (pm.torch.Tensor, np.ndarray):
    lib = pm.torch if element_type == pm.torch.Tensor else np
    return lib.stack(values).mean(axis=0)
```

pulser-core/pulser/backend/aggregators.py
```diff
@@ -0,0 +1,98 @@
...
+) -> Union[Number, List[Number], List[List[Number]], ArrayLike]:
+    """Take the mean of the given Results.
+
+    Supported are numeric values, lists of numeric values
+    lists of lists of numeric values, torch Tensors and numpy arrays.
+    """
+    if values == []:
+        raise ValueError("Cannot average 0 samples")
+
+    element_type = type(values[0])
```
kbidzhiev: you use 2 styles in the code
`element_type == TARGET_TYPE`
and 
`isinstance(values[0], TARGET_TYPE)`

this can be uniformised with

```
element = values[0]
if isinstance(element, Number):
  ...
if pm.AbstractArray.has_torch() and isinstance(element, pm.torch.Tensor):
  ...
```
etc

pulser-core/pulser/backend/aggregators.py
```diff
@@ -0,0 +1,98 @@
...
+        return np.mean(values)  # type: ignore[no-any-return]
+
+    if pm.AbstractArray.has_torch() and element_type == pm.torch.Tensor:
+        acc = pm.torch.zeros_like(values[0])
+        for ten in values:
+            acc += ten
+        return acc / len(values)
+
+    if element_type == np.ndarray:
+        acc = np.zeros_like(values[0])
```
kbidzhiev: this should do the job
`return np.mean(values, axis=0)`

pulser-core/pulser/backend/aggregators.py
```diff
@@ -0,0 +1,98 @@
...
+    """
+    if values == []:
+        raise ValueError("Cannot average 0 samples")
+
+    element_type = type(values[0])
+
+    if isinstance(values[0], Number):
+        return np.mean(values)  # type: ignore[no-any-return]
+
+    if pm.AbstractArray.has_torch() and element_type == pm.torch.Tensor:
```
kbidzhiev, line 46: `return pm.torch.mean(pm.torch.stack(values), dim=0)`

pulser-core/pulser/backend/aggregators.py
kbidzhiev: `axis=0` can work for 2d arrays and 1d arrays simultaneously

pulser-core/pulser/backend/aggregators.py
```diff
@@ -0,0 +1,98 @@
...
+from numbers import Number
+from typing import Any, Callable, List, Union
+
+import numpy as np
+from numpy.typing import ArrayLike
+
+import pulser.math as pm
+from pulser.backend.observable import AggregationType
+
+_NUMERIC_TYPES = {int, float, complex}
```
murogrande: what is the purpose of this because the script is not using this?

pulser-core/pulser/backend/aggregators.py
murogrande: I would say something like: 
`numpy.stack(values, axis=0).mean(axis=0)` 
and for torch:
`torch.stack(values, dim=0).mean(dim=0)`

both are very similar, will it be possible to join them?

pulser-core/pulser/backend/aggregators.py
a-quelle-pasqal: good point

pulser-core/pulser/backend/aggregators.py
a-quelle-pasqal: it was needed before. It seems I forgot to remove it during a refactor

pulser-core/pulser/backend/aggregators.py
a-quelle-pasqal: the call to `stack` that Mauro adds is needed, because the core datatype is a `list`. I changed it to that.

pulser-core/pulser/backend/aggregators.py
```diff
@@ -0,0 +1,91 @@
...
+
+    if not isinstance(elt, list):
+        raise ValueError("Cannot average this type of data")
+
+    if values[0] == []:
+        raise ValueError("Cannot average list of empty lists")
+
+    sub_element_type = type(elt[0])
+
+    if isinstance(elt[0], Number):
```
kbidzhiev: as you did with 
```
    if isinstance(elt, Number):
        return np.mean(values)  # type: ignore[no-any-return]
```
the following should work
```
if isinstance(elt[0], Number):
    return np.mean(values, axis=0).tolist()
```

pulser-core/pulser/backend/aggregators.py
```diff
@@ -0,0 +1,91 @@
...
+        dim = len(elt)
+        return [np.mean([value[i] for value in values]) for i in range(dim)]
+
+    if sub_element_type != list:  # FIXME: ABC.Iterable? Collection? subclass?
```
kbidzhiev: `if not isinstance(elt[0], Sequence)` ?

pulser-core/pulser/backend/aggregators.py
```diff
@@ -0,0 +1,91 @@
...
+        raise ValueError(
+            f"Cannot average list of matrices of {type(elt[0][0])}"
+        )
+
+    dim1 = len(elt)
+    dim2 = len(elt[0])
+    return [
+        [np.mean([value[i][j] for value in values]) for j in range(dim2)]
+        for i in range(dim1)
+    ]
```
kbidzhiev: `mean` with `.tolist()`should do the job and return`list[list[Number]]`
```python
return np.mean(values, axis=0).tolist()
```

kbidzhiev: I've just tested,

`np.mean(values, axis=0).tolist()`
works for List[Number] and it's nested forms List[...List[Number]...]

pulser-core/pulser/backend/__init__.py
```diff
@@ -35,6 +36,8 @@
 )
 
 __all__ = [
+    "aggregation_type_definitions",
+    "AggregationType",
```
a-corni: Are these expected to be used by users ? We usually expose in __init__ only the objects that the user has to call directly (AggregationType and aggregation_type_definitions don't seem to be the case)

pulser-core/pulser/backend/aggregators.py
```diff
@@ -0,0 +1,91 @@
...
+from numpy.typing import ArrayLike
+
+import pulser.math as pm
+from pulser.backend.observable import AggregationType
+
+
+def mean_aggregator(
+    values: list[Any],
+) -> Union[Number, List[Number], List[List[Number]], ArrayLike]:
+    """Take the mean of the given Results.
```
a-corni: Nit: Since you are not taking as input Results, I suggest to have 
```suggestion
    """Take the mean of the given results.
```

pulser-core/pulser/backend/aggregators.py
```diff
@@ -0,0 +1,91 @@
...
+
+    Supported are numeric values, lists of numeric values
+    lists of lists of numeric values, torch Tensors and numpy arrays.
```
a-corni: Nit: This should be described in the argument

```suggestion
Argument:
        values: The results to average. Supported are numeric values, lists of numeric values
    lists of lists of numeric values, torch Tensors and numpy arrays.
Returns:
     The average over the first dimension of the provided results.
```

pulser-core/pulser/backend/aggregators.py
```diff
@@ -0,0 +1,91 @@
...
+import pulser.math as pm
+from pulser.backend.observable import AggregationType
+
+
+def mean_aggregator(
+    values: list[Any],
+) -> Union[Number, List[Number], List[List[Number]], ArrayLike]:
+    """Take the mean of the given Results.
+
+    Supported are numeric values, lists of numeric values
```
a-corni: I don't think you are supporting numeric values. You could just say "Supported are lists of numeric values, ..."

pulser-core/pulser/backend/aggregators.py
```diff
@@ -0,0 +1,91 @@
...
+
+import numpy as np
+from numpy.typing import ArrayLike
+
+import pulser.math as pm
+from pulser.backend.observable import AggregationType
+
+
+def mean_aggregator(
+    values: list[Any],
```
a-corni: Nit: Could the type be list[AbstractArrayLike] | list[list[AbstractArrayLike] ? https://github.com/pasqal-io/Pulser/blob/develop/pulser-core/pulser/math/abstract_array.py#L325C1-L325C19

pulser-core/pulser/backend/aggregators.py
```diff
@@ -0,0 +1,91 @@
...
+from typing import Any, Callable, List, Union
+
+import numpy as np
+from numpy.typing import ArrayLike
+
+import pulser.math as pm
+from pulser.backend.observable import AggregationType
+
+
+def mean_aggregator(
```
a-corni: Is this meant to be used by the user ? Can I suggest to make it private ?

pulser-core/pulser/backend/aggregators.py
```diff
@@ -0,0 +1,91 @@
...
+    values: list[Any],
+) -> Union[Number, List[Number], List[List[Number]], ArrayLike]:
+    """Take the mean of the given Results.
+
+    Supported are numeric values, lists of numeric values
+    lists of lists of numeric values, torch Tensors and numpy arrays.
+    """
+    if values == []:
```
a-corni, line 53: If it's meant to be exposed to the user, I would suggest checking that `values` is of type list before this comparison.

pulser-core/pulser/backend/aggregators.py
```diff
@@ -0,0 +1,91 @@
...
+        raise ValueError("Cannot average 0 samples")
+
+    elt = values[0]
```
a-corni, line 56: what about doing elt = pm.AbstractArray(values[0]) ? That way, I think you could use pm.mean https://github.com/pasqal-io/Pulser/blob/develop/pulser-core/pulser/math/__init__.py#L182 instead of the "if"s below

pulser-core/pulser/backend/aggregators.py
a-corni: Or pm.AbstractArrayLike

a-corni: sequence would allow tuples

pulser-core/pulser/backend/aggregators.py
```diff
@@ -0,0 +1,91 @@
...
+    if sub_element_type != list:  # FIXME: ABC.Iterable? Collection? subclass?
+        raise ValueError(f"Cannot average list of lists of {sub_element_type}")
+
+    if elt[0] == []:
+        raise ValueError("Cannot average list of matrices with empty columns")
+
+    if not isinstance(elt[0][0], Number):
+        raise ValueError(
+            f"Cannot average list of matrices of {type(elt[0][0])}"
+        )
```
a-corni: By having elt = pm.AbstractArray(values[0])
```
if len(elt.shape) > 1:
     raise ValueError
pm.mean(elt, axis=0)
```

pulser-core/pulser/backend/aggregators.py
```diff
@@ -0,0 +1,91 @@
...
+
+    dim1 = len(elt)
+    dim2 = len(elt[0])
+    return [
+        [np.mean([value[i][j] for value in values]) for j in range(dim2)]
+        for i in range(dim1)
+    ]
+
+
+def bag_union_aggregator(
```
a-corni: idem, I think it's best to make it a private function :)

pulser-core/pulser/backend/observable.py
```diff
@@ -69,6 +70,13 @@ def __call__(
         pass
 
 
+class AggregationType(IntEnum):
```
a-corni: - I find "Type" a bit misleading, since it's not a python type, but more a "Kind" ? I would suggest Aggregator, AggregationKind, AggregationMethod, wdyt ? 
- Do you really need it to be IntEnum ? Won't just an Enum work ?

pulser-core/pulser/backend/observable.py
```diff
@@ -204,3 +212,12 @@ def _validate_eval_times(
                 f"Instead, got {evaluation_times!r}."
             )
         return eval_times_arr
+
+    @property
+    def default_aggregation_type(self) -> Optional[AggregationType]:
```
a-corni: idem here I would suggest `default_aggregator` or `default_aggregation_method`

pulser-core/pulser/backend/observable.py
```diff
@@ -204,3 +212,12 @@ def _validate_eval_times(
...
+        """Defines how to combine values from multiple results.
+
+        This is used by `Results.aggregate`.
+        None means no default, therefore an aggregator must be user-provided.
+        """
+        return None
```
a-corni: I would like to suggest to raise a NotImplementedEroor instead of returning None ?

pulser-core/pulser/backend/default_observables.py
```diff
@@ -174,6 +176,8 @@ def apply(self, *, state: State, **kwargs: Any) -> Any:
         """Calculates the observable to store in the Results."""
         return self.state.overlap(state)
 
+    default_aggregation_type = AggregationType.MEAN
```
a-corni: Euh I find this pretty unorthodox haha
I am not exactly sure that this way, default_aggregation_type is still a property (I think it just became an attribute).

To me, you should re-write the property `default_aggregation_type` (or `default_aggregator`, see below) and have `return AggregationType.MEAN` or `return Aggregator.MEAN` (again, see below). 

Other option is to have `self._default_aggregator = None` in Observable.__init__, have `if _default_aggregator is None: raise NotImplementedError` in Observable.default_aggregator, and specify in the init here `self._default_aggregator=Aggregator.MEAN`

pulser-core/pulser/backend/aggregators.py
```diff
@@ -0,0 +1,91 @@
...
+    values: list[collections.Counter],
+) -> collections.Counter:
+    """Join a list of Counter objects."""
+    return sum(values, start=collections.Counter())
+
+
+aggregation_type_definitions: dict[AggregationType, Callable] = {
```
a-corni: Following my rebranding ideas (see below in observable.py), I would suggest to have AGGREGATOR_MAPPING:dict[Aggregator, Callable] (it's a constant, so it should be in capital letters)

pulser-core/pulser/backend/__init__.py
a-corni: IIUC, I think aggregation_type_definitions is not expected to be user-fancing, so it shouldn't be in this init. However, AggregationType (> suggested change Aggregator or others, see observable) is user facing and should stay.

pulser-core/pulser/backend/results.py
```diff
@@ -186,6 +203,9 @@ def _to_abstract_repr(self) -> dict:
             str(key): value for key, value in self._results.items()
         }
         d["times"] = {str(key): value for key, value in self._times.items()}
+        d["aggregation_types"] = {
+            str(key): value for key, value in self._aggregation_types.items()
+        }
```
a-corni: You have to modify the schema for it to accept an optionnal aggregation_types (> would rename it aggregators, following the renaming above ?)

pulser-core/pulser/backend/results.py
```diff
@@ -233,6 +255,98 @@ def from_abstract_repr(cls, repr: str) -> Results:
...
+        cls,
+        results_to_aggregate: typing.Sequence[Results],
+        **aggregation_functions: Callable[[Any], Any],
+    ) -> Results:
+        """Aggregate a Sequence of Results objects into a single Results.
+
+        This is meant to accumulate the results of several runs with
+        different noise trajectories into a single averaged Results.
+        By default, results are averaged, with the exception of BitStrings,
+        where the counters are joined. StateResult is not supported by default.
```
a-corni: You also have Variance not supported by default

pulser-core/pulser/backend/results.py
```diff
@@ -233,6 +255,98 @@ def from_abstract_repr(cls, repr: str) -> Results:
...
+
+        Args:
+            results_to_aggregate: The list of Results to aggregate
+            **aggregation_functions: Overrides for the default aggregator.
+                The argument name should be the tag of the Observable,
+                the value is a Callable that takes a list
```
a-corni: That's not an argument but a Keyword Argument. You can create another section to describe them (see an example [here](https://github.com/pasqal-io/Pulser/blob/develop/pulser-core/pulser/json/abstract_repr/serializer.py#L118)).

```suggestion
    Keyword Args:
            observable_tag: Overrides the default aggregator of the Observable with a custom function, that takes a list of values/ a list of list of values as input.
```

pulser-core/pulser/json/abstract_repr/serializer.py
```diff
@@ -54,6 +56,8 @@ def default(self, o: Any) -> dict[str, Any] | list | int | float:
                 # Try to return a real number when possible
                 return o.real
             return dict(real=o.real, imag=o.imag)
+        elif isinstance(o, IntEnum):
+            return o.value
```
a-corni: That would work with Enum only.

tests/test_abstract_repr.py
```diff
@@ -110,6 +111,13 @@ def _to_abstract_repr(self):
     result = encoder.default(1.0 + 2.0j)
     assert result == dict(real=1.0, imag=2.0)
 
+    class DummyEnum(IntEnum):
+        A = 5
+        B = 6
+
+    result = encoder.default(DummyEnum.A)
+    assert result == 5
+
```
a-corni: That's not really testing the results, right ? 😆

pulser-core/pulser/backend/results.py
```diff
@@ -233,6 +255,98 @@ def from_abstract_repr(cls, repr: str) -> Results:
...
+                "they do not all contain the same observables"
+            )
+        if not all(
+            results._aggregation_types == result_0._aggregation_types
+            for results in results_to_aggregate
+        ):
+            raise ValueError(
+                "You're trying to aggregate incompatible results: "
+                "they do not all contain the same aggregation functions"
+            )
```
a-corni: This does not take into account the custom function that the use could be defining, right ?

pulser-core/pulser/backend/results.py
```diff
@@ -233,6 +255,98 @@ def from_abstract_repr(cls, repr: str) -> Results:
...
+        aggregated = Results(
+            atom_order=result_0.atom_order,
+            total_duration=result_0.total_duration,
```
a-corni, line 351: I would prefer using self.atom_order and self.total_duration

pulser-core/pulser/backend/results.py
```diff
@@ -233,6 +255,98 @@ def from_abstract_repr(cls, repr: str) -> Results:
...
+        )
+        for tag in stored_callbacks:
+            default_aggregation_type = result_0._aggregation_types[
+                result_0._tagmap[tag]
+            ]
+            aggregation_type = aggregation_functions.get(
+                tag, default_aggregation_type
+            )
+            if aggregation_type is None:
+                logging.warning(f"Skipping aggregation of `{tag}`")
```
a-corni: It's something the user should be aware of, because he can act on it, then this should be a warnings.warn() (see [this discussion](https://stackoverflow.com/questions/9595009/warnings-warn-vs-logging-warning)). You should surely have a context manager, to raise always raise a warning.

a-corni (commented):
I have a batch of comments:
- check what we have in pulser.math.__init__ and pulser.math.AbstractArray for the aggregators
- If find "type" in the enum confusing, would rather have Aggregator
- See the comment on default_aggregation_type property
- need to change the schema for the serizalization to work.

pulser-core/pulser/backend/__init__.py
a-quelle-pasqal: I agree with this. Let's keep `AggregationType` and not the other.

pulser-core/pulser/backend/results.py
a-quelle-pasqal: that's true, but if the different `Results` in the list have different default aggregation types, that means they were not generated by the same `Observable`, but by different ones using the same `tag`. I'd still say this is dangerous, regardless of whether the user chooses to override the aggregation behaviour.

pulser-core/pulser/backend/results.py
a-quelle-pasqal: it's a class method there is no `self`

pulser-core/pulser/backend/results.py
a-quelle-pasqal: I find this issue a bit tricky. One the one hand, in emu-sv and emu-mps we store performance measures in `Results` and these do not aggregate, but neither can the user actually solve that. On the other hand, if the user stores a `StateResult` or `EnergyVariance` these also do not aggregate, and the user will probably want to be informed.

tests/test_abstract_repr.py
a-quelle-pasqal: no, it tests lines 59 and 60 from `pulser-core/pulser/json/abstract_repr/serializer.py`, and for this reason I think it's good to have it. It provides a direct test for this method. However, it does raise the question of why this line of code wasn't already covered by the serialization test for `Results`, which is a problem I will address.

pulser-core/pulser/backend/observable.py
a-quelle-pasqal: But the `None` indicates that aggregation of a result is skipped in `Results.aggregate` which is a totally valid thing, and we shouldn't have to start handling exceptions in `Results.aggregate` for that. Maybe add `AggregationType.SKIP` or something like that, and have the default return value be that?

pulser-core/pulser/backend/observable.py
a-quelle-pasqal: The reason I used `IntEnum` is so that the call to `value` has type `int` rather than `Any`. It allows the return type of the serializer to not be `Any`. See also

```
        elif isinstance(o, IntEnum):
            return o.value
```
where you had a comment about `Enum` that I'm not sure I understand, but is probably related to this issue.

pulser-core/pulser/backend/aggregators.py
a-quelle: I've looked at `pulser-core/math.init` and the functions defined in it. I agree that they can be used at the least to merge these two statements into one

```
    if pm.AbstractArray.has_torch() and isinstance(elt, pm.torch.Tensor):
        return pm.torch.stack(values).mean(dim=0)

    if isinstance(elt, np.ndarray):
        return np.stack(values).mean(axis=0)  # type: ignore[no-any-return]
```

I'm not so sure about all the lists stuff, which is needed because serialization of results turns everything into a `list`. However, the cost of doing this, is to introduce `AbstractArray` as a possible return type. This is very good within library code, because it allows us to make things generic, but I'm afraid it might make things more confusing for the user. because they would expect things to be aggregated into the same type as the input.

pulser-core/pulser/backend/aggregators.py
a-quelle: I do not think this function should change the type of the object being aggregated. If the user is aggregating `torch.Tensor` objects from their list of `Results`, it would lead to confusion if the aggregated result is an `AbstractArray` wrapping a `torch.Tensor` rather than a straight `torch.Tensor`

pulser-core/pulser/json/abstract_repr/serializer.py
a-quelle: turns out this is not needed, since the serializer supports `IntEnum` by default

tests/test_abstract_repr.py
a-quelle: no longer needed

pulser-core/pulser/backend/results.py
a-quelle: I've removed usage of `None` as an aggregation method, and rather added `AggregationMethod.SKIP` and `AggregationMethod.SKIP_WARN`. Using this, `StateResult` and `EnergyVariance` now skip aggregation and `warn` while the default unless overridden is `SKIP`. This will solve the issue with the statistics.

pulser-core/pulser/backend/aggregators.py
```diff
@@ -0,0 +1,90 @@
...
+        values: The results to average. Supported are lists of:
+        numeric values, lists of numeric values,
+        lists of lists of numeric values, torch Tensors and numpy arrays.
+
+    Returns:
+        The average over the first dimension of the provided results.
+    """
+    assert isinstance(
+        values, list
+    ), "Need to supply a list of values to average."
```
a-corni: Nit: Would be better to raise a TypeError instead of a AssertionError
 ```suggestion
 if not isinstance(values, list): # eventually, Sequence, if you don't want it to fail on a tuple
    with pytest.raises(TypeError,"Need to supply a list of values to average."):
```

File: pulser-core/pulser/backend/aggregators.py (hunk 15 above)
Comment (a-corni) [reply pulser-core/pulser/backend/aggregators.py:53]: if len(values) == 0 if you want to authorize tuples as well.

File: pulser-core/pulser/backend/aggregators.py (hunk 16 above)
Comment (a-corni) [reply pulser-core/pulser/backend/aggregators.py:56]: I you want to highlight this in the typing of the function, I suggest you use the decorator [overload](https://typing.python.org/en/latest/spec/overload.html) (see an example in pulser [here](https://github.com/pasqal-io/Pulser/blob/develop/pulser-core/pulser/channels/eom.py#L211-L238))

File: pulser-core/pulser/backend/results.py [hunk 33]
```diff
@@ -200,6 +220,10 @@ def _from_abstract_repr(cls, dict: dict) -> Results:
             results._results[uuid.UUID(key)] = deserialize_complex(value)
         for key, value in dict["times"].items():
             results._times[uuid.UUID(key)] = value
+        for key, value in dict["aggregation_types"].items():
```
a-corni: Euh I never saw it before but it's perhaps not a great idea to have a variable named "dict"... `obj` would be better ?

pulser-core/pulser/backend/results.py
```diff
@@ -200,6 +220,10 @@ def _from_abstract_repr(cls, dict: dict) -> Results:
...
+            results._aggregation_methods[uuid.UUID(key)] = AggregationMethod(
```
a-corni, line 224: Okay you need an IntEnum for this.

pulser-core/pulser/backend/results.py
```diff
@@ -233,6 +257,105 @@ def from_abstract_repr(cls, repr: str) -> Results:
...
+        where the counters are joined.
+        StateResult and EnergyVariance are not supported by default.
+
+        Args:
+            results_to_aggregate: The list of Results to aggregate
+
+        Keyword Args:
+            observable_tag: Overrides the default aggregator.
+            The argument name should be the tag of the Observable.
+            The value is a Callable taking a list of the type to aggregate.
```
a-corni: Nit: For the API doc, I think you should indent this.

pulser-core/pulser/backend/results.py
```diff
@@ -233,6 +257,105 @@ def from_abstract_repr(cls, repr: str) -> Results:
...
+        Returns:
+            The averaged Results object
+        """
+        if len(results_to_aggregate) == 0:
+            raise ValueError("no results to aggregate")
```
a-corni: nit:
```suggestion
            raise ValueError("No results to aggregate.")
```

pulser-core/pulser/backend/results.py
a-corni: Indeed 😅

pulser-core/pulser/backend/results.py
```diff
@@ -233,6 +257,105 @@ def from_abstract_repr(cls, repr: str) -> Results:
...
+            ]
+            aggregation_type = aggregation_functions.get(
+                tag, default_aggregation_method
+            )
+            if (
+                aggregation_type is AggregationMethod.SKIP
+                or aggregation_type is AggregationMethod.SKIP_WARN
+            ):
+                if aggregation_type is AggregationMethod.SKIP_WARN:
+                    warn(f"Skipping aggregation of `{tag}`")
```
a-corni: If you want to be sure to get one warning message per aggregation_type to SKIP_WARN, I think you should use a context manager:
```suggestion
                if aggregation_type is AggregationMethod.SKIP_WARN:
                    with warnings.catch_warnings():
                          warnings.simplefilter("always")
                          warnings.warn(f"Skipping aggregation of `{tag}`")
```

a-corni: Nit: For readibility, I would rather have aggregation_method instead of aggregation_type

pulser-core/pulser/backend/results.py
```diff
@@ -233,6 +257,105 @@ def from_abstract_repr(cls, repr: str) -> Results:
...
+                continue
+            aggregation_function: Any = (
+                aggregation_type
+                if callable(aggregation_type)
+                else AGGREGATOR_MAPPING[aggregation_type]
+            )
```
a-corni: I would be more in favor of checking if aggregation_type is among AGGREGATOR_MAPPING rather than checking if it's callable (notably, since aggregation_type is a TypeEnum I think it's a callable).
```suggestion
            aggregation_function: Callable = AGGREGATOR_MAPPING.get(aggregation_type, aggregation_type)
  
```

pulser-core/pulser/backend/results.py
```diff
@@ -233,6 +257,105 @@ def from_abstract_repr(cls, repr: str) -> Results:
...
+            evaluation_times = results_to_aggregate[0].get_result_times(tag)
+            if not all(
+                results.get_result_times(tag) == evaluation_times
+                for results in results_to_aggregate
+            ):
+                raise ValueError(
+                    "Monte-Carlo results seem to provide from "
+                    "incompatible simulations: "
+                    "the callbacks are not stored at the same times"
```
a-corni: Eventually, precise for which tag in this ValueError

pulser-core/pulser/backend/results.py
```diff
@@ -233,6 +257,105 @@ def from_abstract_repr(cls, repr: str) -> Results:
...
+                        for result in results_to_aggregate
+                    ]
+                )
+
+                aggregated._store_raw(
+                    uuid=uid,
+                    tag=tag,
+                    time=t,
+                    value=v,
+                    aggregation_type=default_aggregation_method,
```
a-corni: I think it would be good to precise in the docstring of the function that, whatever the aggregation method specified in the kwargs, the stored aggregation method is the default one for the observable.

pulser-core/pulser/backend/results.py
```diff
@@ -233,6 +257,105 @@ def from_abstract_repr(cls, repr: str) -> Results:
...
+            for results in results_to_aggregate
+        ):
+            raise ValueError(
+                "You're trying to aggregate incompatible results: "
+                "they do not all contain the same aggregation functions"
+            )
+        aggregated = Results(
+            atom_order=result_0.atom_order,
+            total_duration=result_0.total_duration,
+        )
```
a-corni: I think you should
- check that all the results have the same atom_order and same total_duration (since they should be from same experiment from different runs)
- constitute stored_callbacks as "the observable for which the aggregation method is not skipped". That would enable people that made different runs, all having at least one observable in common, but some observables being different, to aggregate a callback of interest that would be shared by all the runs. Wdyt ?

pulser-core/pulser/backend/aggregators.py
a-corni: I agree with you that we don't want to output an AbstractArray. However your current output type does not cover torch.Tensor. AbstractArrayLike enables you to handle Number | list | ndarray | torch.Tensor. As mentioned below, you can specify for some specific type that the input and output should be of the same type with the overload decorator.

pulser-core/pulser/backend/aggregators.py
```diff
@@ -0,0 +1,90 @@
...
+        return np.stack(values).mean(axis=0)  # type: ignore[no-any-return]
+
+    if isinstance(elt, Number):
+        return np.mean(values)  # type: ignore[no-any-return]
+
+    if not isinstance(elt, Sequence):
+        raise ValueError("Cannot average this type of data")
+
+    if values[0] == []:
+        raise ValueError("Cannot average list of empty lists")
```
a-corni: - You would fail if someone was providing a tuple 
- Add a . at the end of the error messages.
- I generally find this good practice to raise errors first, and then make returns (easier to maintain)
Here is my suggestion :) 

```{suggestion}
    if not isinstance(elt, AbstractArrayLike):  # Rejects tuple
        raise ValueError("Cannot average this type of data")
    if len(elt) == 0:
        raise ValueError("Cannot average list of empty lists")
   if pm.AbstractArray.has_torch() and isinstance(elt, pm.torch.Tensor) :
        return list(pm.torch.stack(values).mean(dim=0)) # list of torch tensors
    if isinstance(elt, np.ndarray):
        return list(np.stack(values).mean(axis=0))  # -> list of ndarray
    if isinstance(elt, Number):
        return cast(Number, np.mean(values))  # a number
```

To be checked as one-liner:
```{suggestion}
    if not isinstance(elt, AbstractArrayLike):  # Rejects tuple
        raise ValueError("Cannot average this type of data")
    if len(elt) == 0:
        raise ValueError("Cannot average list of empty lists")
   if isinstance(elt, Number):
       return np.mean(values)
   if pm.AbstractArray.has_torch() and isinstance(elt, pm.torch.Tensor) or isinstance(elt, np.ndarray):  # This command should also cover the case of elt being a list -> only add a check on the dimension of elt[0] before if elt is a list, and change this if to `if isinstance(elt, pm.AbstractArrayLike)`: 
   return list(pm.mean(pm.hstack(values), axis=0)._array)
```

pulser-core/pulser/json/abstract_repr/serializer.py
```diff
@@ -33,6 +33,7 @@
     from pulser.register.base_register import QubitId
     from pulser.sequence import Sequence
     from pulser.sequence._call import _Call
+
```
a-corni: Could you delete this line ?

pulser-core/pulser/backend/observable.py
a-corni: Okay I see it's from a typing issue of the value. I was just wondering why you were using an IntEnum and not just an Enum (the typing issue could have been solved by casting I guess).

pulser-core/pulser/backend/observable.py
a-corni: I agree with you, AggregationType.SKIP/SKIP_WARN was a great idea !

pulser-core/pulser/backend/results.py
a-corni: Ah okay to me, the user could ask to skip the handling of some observables where he did some nasty work.

pulser-core/pulser/backend/results.py
a-corni: Sounds good to me !

pulser-core/pulser/backend/aggregators.py
a-quelle-pasqal: but `Sequence` also allows strings, and then it will fail on a hard-to-read error in `np.mean`

a-quelle-pasqal: Essentially, I don't care about any type other than `list` because these functions are called from `Results.aggregate`, where they are fed a `list` specifically.

pulser-core/pulser/backend/aggregators.py
a-quelle-pasqal: never mind. I'm struggling with the type system, since there seems to be no real way to have it know about `torch.Tensor` in the case where `torch` is not available to the user. I'll get back to you on that.

pulser-core/pulser/backend/aggregators.py
a-quelle-pasqal: why do you care about tuples so much. We have full control over how this is called since it's a private function, and we know it's not going to get a tuple. Furthermore, type hints such as `Sequence` imply that the function can handle other indexable things such as `str`, which is just not true, so the suggestions you made in this direction all make the typing of the function worse, at the benefit of a feature that we do not actually use.

Furthermore, `AbstractArrayLike` also does not cover `torch.Tensor` since it's literally either `ArrayLike` or `AbstractArray`, and `AbstractArray` is not itself in any way related to `torch.Tensor` in it's type. I agree that the typing should indicate that a `torch.Tensor` could be returned, but the way to do it is not to indicate that the function could return an `AbstractArray` while it never will.

a-corni: Okay, I was just trying to help out to try to make the typing a bit more clear and easier to understand. Since it is now a private function, I care less about it's clarity with respect to the user.

a-corni: My bad regarding AbstractArrayLike, I think [TensorLike](https://github.com/pasqal-io/Pulser/blob/develop/pulser-core/pulser/math/__init__.py#L38C7-L38C17) covers torch.Tensor however

a-quelle-pasqal: I managed to get it to work by putting `"pm.torch.Tensor"` in the typing. This is apparently understood as the actual type provided it exists, and otherwise, it's ignored. `TensorLike` didn't behave properly: specifically, mypy didn't link it to `torch.Tensor` correctly in case `isinstance(elt, pm.torch.Tensor)`

a-corni: Thanks for having tried, I was wondering if "pm.torch.Tensor" was going to work. I am good with your implementation :)

pulser-core/pulser/backend/aggregators.py
a-corni: Looks like you found a way :)

pulser-core/pulser/backend/aggregators.py
```diff
@@ -0,0 +1,103 @@
...
+        raise ValueError(f"Cannot average list of lists of {type(elt[0])}")
+
+    if len(elt[0]) == 0:
+        raise ValueError("Cannot average list of matrices with empty columns")
+
+    if not isinstance(elt[0][0], (float, complex)):
+        raise ValueError(
+            f"Cannot average list of matrices of {type(elt[0][0])}"
+        )
+    return np.mean(values, axis=0).tolist()  # type: ignore[no-any-return]
```
a-corni: Maybe the typing will be solved with
```suggestion
    return list(np.mean(values, axis=0))  # type: ignore[no-any-return]
```

a-quelle-pasqal: a combination of both works. The `tolist` is needed because otherwise you get lists of numpy arrays, because it only flattens the outermost index, which is not the intended behaviour.

pulser-core/pulser/backend/results.py
a-quelle-pasqal: yes, that's a very good point. I'll go and do that

pulser-core/pulser/backend/results.py
```diff
@@ -233,6 +257,126 @@ def from_abstract_repr(cls, repr: str) -> Results:
...
+                if isinstance(aggregation_method, AggregationMethod)
+                else aggregation_method
+            )
+            evaluation_times = results_to_aggregate[0].get_result_times(tag)
+            if not all(
+                results.get_result_times(tag) == evaluation_times
+                for results in results_to_aggregate
+            ):
+                raise ValueError(
+                    "Monte-Carlo results seem to provide from "
```
a-corni: Nit: Are the results expected to be monte-carlo ? That could be results from the QPU, right ?

pulser-core/pulser/backend/aggregators.py
```diff
@@ -0,0 +1,103 @@
...
+        raise ValueError(f"Cannot average list of lists of {type(elt[0])}")
+
+    if len(elt[0]) == 0:
+        raise ValueError("Cannot average list of matrices with empty columns")
+
+    if not isinstance(elt[0][0], (float, complex)):
+        raise ValueError(
+            f"Cannot average list of matrices of {type(elt[0][0])}"
+        )
+    return list(np.mean(values, axis=0).tolist())
```
a-corni: Nit: Could you add a "." at the end of the errors ? (sorry for the pickiness)

pulser-core/pulser/backend/results.py
```diff
@@ -233,6 +257,126 @@ def from_abstract_repr(cls, repr: str) -> Results:
...
+                The argument name should be the tag of the Observable.
+                The value is a Callable taking a list of the type to aggregate.
+                Note that this does not override the default aggregation
+                behaviour of the aggregated results.
+
+        Returns:
+            The averaged Results object
+        """
+        if len(results_to_aggregate) == 0:
+            raise ValueError("no results to aggregate.")
```
a-corni:
```suggestion
            raise ValueError("No results to aggregate.")
```

pulser-core/pulser/backend/results.py
a-quelle-pasqal: done

pulser-core/pulser/backend/results.py
```diff
@@ -289,17 +289,33 @@ def aggregate(
...
+            *[set(x.get_result_tags()) for x in results_to_aggregate]
+        )
+        common_tags = all_tags.intersection(
+            *[set(x.get_result_tags()) for x in results_to_aggregate]
+        )
+
+        for results in results_to_aggregate:
+            for tag, uid in results._tagmap.items():
+                if tag not in common_tags and not (
+                    results._aggregation_methods[uid].value < 2
```
a-corni: This depends on the order of the values in AggregationMethod... Notably, if we add a new entry you will surely have to change this. For scalability reason, I would suggest:
```
results._aggregation_methods[uid] not in [AggregationMethod.SKIP, AggregationMethod.SKIP_MEAN]
```

pulser-core/pulser/backend/results.py
```diff
@@ -289,17 +289,33 @@ def aggregate(
...
+                ):
+                    raise ValueError(
+                        "You're trying to aggregate incompatible results: "
+                        f"result `{tag}` is not present in all results, "
+                        "but it's not marked to be skipped."
+                    )
```
a-corni: L 311-318 don't take into account the aggregation functions. I think you can have L309, with the same indentation as L 302:
```{suggestion}
        for results in results_to_aggregate:
            for tag, uid in results._tagmap.items():
                unspecified_tag = tag not in aggregation_functions
                if tag not in common_tags or (results._aggregation_methods[uid] if unspecified_tag else aggregation_functions[tag]) in [AggregationMethod.SKIP, Aggregation.SKIP_WARN]:
                    raise ValueError(
                        "You're trying to aggregate incompatible results: "
                        f"result `{tag}` is not present in all results, "
                        "but it's not marked to be skipped."
                    )
               if results_0._aggregation_methods[uid] != results._aggregation_methods[uid] and unspecified_tag:
                    raise ValueError(
                        "You're trying to aggregate incompatible results: "
                        "they do not all contain the same aggregation functions."
                    )
```

pulser-core/pulser/backend/results.py
```diff
@@ -186,20 +203,27 @@ def _to_abstract_repr(self) -> dict:
...
-        for key, value in dict["results"].items():
+        for key, value in obj["results"].items():
             results._results[uuid.UUID(key)] = deserialize_complex(value)
-        for key, value in dict["times"].items():
+        for key, value in obj["times"].items():
             results._times[uuid.UUID(key)] = value
+        for key, value in obj["aggregation_methods"].items():
+            results._aggregation_methods[uuid.UUID(key)] = AggregationMethod(
+                value
+            )
```
a-corni: I think you have to handle the case where "aggregation_methods" won't be present:
```suggestion
        for key, value in obj.get("aggregation_methods", {}).items():
            results._aggregation_methods[uuid.UUID(key)] = AggregationMethod(
                value
            )
```
Usecase: someone stored a result somewhere, wants to analyze it with the next release of pulser. I think it's fair to say that, if in this case, we don't support aggregation (aggregation methods is an empty dict)

tests/test_aggregators.py
```diff
@@ -0,0 +1,82 @@
...
+        ]
+        assert torch.allclose(
+            _mean_aggregator(input6), torch.tensor([2.0, 3.0, 4.0])
+        )
+
+
+def test__mean_aggregator_errors():
+    with pytest.raises(ValueError) as ex:
+        _mean_aggregator([])
+    assert str(ex.value) == "Cannot average 0 samples."
```
a-corni: You can check the value of the error message directly in pytest
```suggestion
    with pytest.raises(ValueError, match="Cannot average 0 samples."):
        _mean_aggregator([])
```

pulser-core/pulser/backend/results.py
a-quelle-pasqal: I agree that that line does not take into account the aggregation function, and this relates to a discussion we had already. Even though the user overrides the aggregation behaviour, if the default aggregation method is not the same for these results, it indicates that they came from different sources, and therefore the operation is suspect.

The way I wrote it, a "tag" has to be in common tags, or it has to be either `SKIP' or 'SKIP_WARN' and furthermore, all common tags have to agree, for the reason above.

pulser-core/pulser/backend/results.py
a-quelle-pasqal: yes, good point. In that case, I will raise a `NotImplementedError`

pulser-core/pulser/backend/results.py
a-corni: I am okay with the current state. Let's go with this and keep in mind that if someone asks for this behaviour to be modified later, we can easily implement it.

pulser-core/pulser/backend/results.py
a-corni: I think that's the only thing left :)

pulser-core/pulser/backend/results.py
```diff
@@ -220,10 +220,13 @@ def _from_abstract_repr(cls, obj: dict) -> Results:
...
-            results._aggregation_methods[uuid.UUID(key)] = AggregationMethod(
-                value
-            )
+        if (
+            "aggregation_methods" in obj
+        ):  # optional for backwards compatibility
+            for key, value in obj["aggregation_methods"].items():
+                results._aggregation_methods[uuid.UUID(key)] = (
+                    AggregationMethod(value)
+                )
```
a-corni: Very nit: We tend to try to decrease the number of indent, that's why I suggested:
```
           for key, value in get(obj["aggregation_methods"], {}).items():
                results._aggregation_methods[uuid.UUID(key)] = (
                    AggregationMethod(value)
                )
```

pulser-core/pulser/backend/results.py
```diff
@@ -298,7 +301,7 @@ def aggregate(
         )
 
         for results in results_to_aggregate:
-            if not hasattr(results, "_aggregation_methods"):
+            if results._results and (not results._aggregation_methods):
```
a-corni, line 301: Should we prevent people from providing empty results ? I think so, I am trying to understand where it will fail and it's not easy to find out...

a-quelle-pasqal: I think it will not fail, it will aggregate into an empty results.

`common_tags` will be empty, so the `ValueError` on `_aggregation_methods` and the `NotImplementedError` should not trigger. Neither should the ones on `total_duration` and `atom_order`. The aggregation loop over `common_tags` will simply be skipped, so `aggregated` will just be returned as is, which is empty

a-corni: But is it the behaviour you want ? If you provide a list of results, and one is empty, I don't think you want an empty aggregated result, but either to get an error or to get a Result aggregating all the others...

a-corni (approved):
LGTM ! Thanks for this MR, and going through all the comments !

pulser-core/pulser/backend/observable.py
```diff
@@ -204,3 +214,5 @@ def _validate_eval_times(
                 f"Instead, got {evaluation_times!r}."
             )
         return eval_times_arr
+
+    default_aggregation_method = AggregationMethod.SKIP
```
HGSilveri, line 218: I would rather see this properly typed as a `ClassVar`, preferably at the beginning of the class

pulser-core/pulser/backend/observable.py
```diff
@@ -69,6 +70,15 @@ def __call__(
         pass
 
 
+class AggregationMethod(IntEnum):
+    """Defines how to combine values from multiple results."""
+
+    SKIP = 0
+    SKIP_WARN = 1
+    MEAN = 2
+    BAG_UNION = 3
```
HGSilveri, line 79: nit: I feel like it would make more sense for this to be in `aggregators.py`

pulser-core/pulser/json/abstract_repr/schemas/results-schema.json
```diff
@@ -24,6 +24,12 @@
                        "$ref": "#/definitions/uuid"
                    }
                },
+               "aggregation_methods": {
+                   "type": "object",
+                   "additionalProperties": {
+                       "type": "integer"
+                   }
```
HGSilveri, line 31: Two questions here:
- Why not make this an Enum so that the validation is more robust?
- Why not use strings (e.g. `"SKIP", "SKIP_WARN",...) to make the serialization more human readable?

Issue #881, opened 2025-06-06
Implement atomic motion noise


HGSilveri:
Implement atomic motion noise by changing the slightly position of the atoms (potentially in 3D?) from shot-to-shot. In principle, the `temperature` is sufficient to parametrize this effect.

murogrande:
The relation between spatial position fluctuations, the temperature and trap parameters is given by:
$\sigma_r^{xy}=\sqrt{k_B T w^2/(4U_\textrm{trap})},\ \ \ 
    \sigma_r^{z}=\frac{\pi}{\lambda}\sqrt{2}w\sigma_r^{xy}.$

PR #915, opened 2025-08-07
noise register 

murogrande:
related to #881 
- adding support for register noise
- update the noise model with the following trap parameters: `trap_depth` and `trap_waist`

### DIFF
--- a/pulser-core/pulser/noise_model.py
+++ b/pulser-core/pulser/noise_model.py
@@ -15,7 +15,6 @@
 from __future__ import annotations
 
 import json
-import math
 import warnings
 from collections.abc import Collection, Sequence
 from dataclasses import dataclass, field, fields
@@ -28,7 +27,6 @@ import pulser.json.abstract_repr as pulser_abstract_repr
 from pulser.json.abstract_repr.serializer import AbstractReprEncoder
 from pulser.json.abstract_repr.validation import validate_abstract_repr
 from pulser.json.utils import get_dataclass_defaults
-
 
 __all__ = ["NoiseModel"]
 
@@ -48,7 +46,7 @@ NoiseTypes = Literal[
 _NOISE_TYPE_PARAMS: dict[NoiseTypes, tuple[str, ...]] = {
     "leakage": ("with_leakage",),
     "doppler": ("temperature",),
-    "register": ("trap_waist", "trap_depth"), # this need the temperature also
+    "register": ("trap_waist", "trap_depth"),  # this need the temperature also
     "amplitude": ("laser_waist", "amp_sigma"),
     "detuning": ("detuning_sigma",),
     "SPAM": ("p_false_pos", "p_false_neg", "state_prep_error"),
--- a/pulser-core/pulser/noise_model.py
+++ b/pulser-core/pulser/noise_model.py
@@ -15,6 +15,7 @@
 from __future__ import annotations
 
 import json
+import math
 import warnings
 from collections.abc import Collection, Sequence
 from dataclasses import dataclass, field, fields
@@ -28,6 +29,7 @@ from pulser.json.abstract_repr.serializer import AbstractReprEncoder
 from pulser.json.abstract_repr.validation import validate_abstract_repr
 from pulser.json.utils import get_dataclass_defaults
 
+
 __all__ = ["NoiseModel"]
 
 NoiseTypes = Literal[
@@ -35,6 +37,7 @@ NoiseTypes = Literal[
     "doppler",
     "amplitude",
     "detuning",
+    "register",
     "SPAM",
     "dephasing",
     "relaxation",
@@ -45,6 +48,7 @@ NoiseTypes = Literal[
 _NOISE_TYPE_PARAMS: dict[NoiseTypes, tuple[str, ...]] = {
     "leakage": ("with_leakage",),
     "doppler": ("temperature",),
+    "register": ("trap_waist", "trap_depth"), # this need the temperature also
     "amplitude": ("laser_waist", "amp_sigma"),
     "detuning": ("detuning_sigma",),
     "SPAM": ("p_false_pos", "p_false_neg", "state_prep_error"),
@@ -60,7 +64,6 @@ _PARAM_TO_NOISE_TYPE: dict[str, NoiseTypes] = {
     for param in params
 }
 
-# Parameter characterization
 
 _POSITIVE = {
     "dephasing_rate",
@@ -69,6 +72,8 @@ _POSITIVE = {
     "depolarizing_rate",
     "temperature",
     "detuning_sigma",
+    "trap_waist",
+    "trap_depth",
 }
 _STRICT_POSITIVE = {
     "runs",
@@ -84,6 +89,7 @@ _PROBABILITY_LIKE = {
 
 _BOOLEAN = {"with_leakage"}
 
+# legacy confing
 _LEGACY_DEFAULTS = {
     "runs": 15,
     "samples_per_run": 5,
@@ -99,7 +105,11 @@ _LEGACY_DEFAULTS = {
     "depolarizing_rate": 0.05,
 }
 
-OPTIONAL_IN_ABSTR_REPR = ("detuning_sigma",)
+OPTIONAL_IN_ABSTR_REPR = (
+    "detuning_sigma",
+    "trap_waist",
+    "trap_depth",
+)
 
 
 @dataclass(init=True, repr=False, frozen=True)
@@ -131,6 +141,9 @@ class NoiseModel:
     - **doppler**: Local atom detuning due to termal motion of the
       atoms and Doppler effect with respect to laser frequency.
       Parametrized by the ``temperature`` field.
+    - **register**: thermal fluctuations in the
+      register positions, parametrized by ``temperature``, ``trap_waist``
+      and, ``trap_depth``.
     - **amplitude**: Gaussian damping due to finite laser waist and
       laser amplitude fluctuations. Parametrized by ``laser_waist``
       and ``amp_sigma``.
@@ -166,6 +179,10 @@ class NoiseModel:
             distribution centered in 0. Assumed to be the same for all
             channels (though each channel has its own randomly sampled
             value in each run). This noise is additive. Defaults to 0.
+        trap_waist: Refers to the radius of the trapping laser beam at the
+            focus point (in µm). Defaults to 0.
+        trap_depth: The potential energy well depth that confines the atoms
+            (in µK). Defaults to 0.
         relaxation_rate: The rate of relaxation from the Rydberg to the
             ground state (in 1/µs). Corresponds to 1/T1. Defaults to 0.
         dephasing_rate: The rate of a dephasing occuring (in 1/µs) in a
@@ -196,6 +213,8 @@ class NoiseModel:
     detuning_sigma: float = 0.0
     relaxation_rate: float = 0.0
     dephasing_rate: float = 0.0
+    trap_waist: float = 0.0
+    trap_depth: float = 0.0
     hyperfine_dephasing_rate: float = 0.0
     depolarizing_rate: float = 0.0
     eff_noise_rates: tuple[float, ...] = ()
@@ -289,6 +308,7 @@ class NoiseModel:
                 or nt_ == "detuning"
                 or (nt_ == "amplitude" and amp_sigma != 0.0)
                 or (nt_ == "SPAM" and state_prep_error != 0.0)
+                or nt_ == "register"
             ):
                 relevant_params.update(("runs", "samples_per_run"))
         # Disregard laser_waist when not defined
@@ -410,6 +430,7 @@ class NoiseModel:
                     stacklevel=2,
                 )
 
+    # check noise_register, regsiter noise is optinal in the abstract repr
     def _to_abstract_repr(self) -> dict[str, Any]:
         all_fields = {}
         for f in fields(self):
--- a/pulser-simulation/pulser_simulation/hamiltonian.py
+++ b/pulser-simulation/pulser_simulation/hamiltonian.py
@@ -19,7 +19,6 @@ import functools
 import itertools
 from collections import defaultdict
 from collections.abc import Mapping
-
 from typing import Union, cast
 
 import numpy as np
@@ -34,8 +33,8 @@ from pulser.sampler.samples import SequenceSamples, _PulseTargetSlot
 from pulser_simulation.simconfig import (
     SUPPORTED_NOISES,
     doppler_sigma,
-    register_sigma_xy_z,
     noisy_register,
+    register_sigma_xy_z,
 )
 
 
--- a/pulser-simulation/pulser_simulation/hamiltonian.py
+++ b/pulser-simulation/pulser_simulation/hamiltonian.py
@@ -19,6 +19,7 @@ import functools
 import itertools
 from collections import defaultdict
 from collections.abc import Mapping
+
 from typing import Union, cast
 
 import numpy as np
@@ -30,7 +31,12 @@ from pulser.devices._device_datacls import BaseDevice
 from pulser.noise_model import NoiseModel
 from pulser.register.base_register import QubitId
 from pulser.sampler.samples import SequenceSamples, _PulseTargetSlot
-from pulser_simulation.simconfig import SUPPORTED_NOISES, doppler_sigma
+from pulser_simulation.simconfig import (
+    SUPPORTED_NOISES,
+    doppler_sigma,
+    register_sigma_xy_z,
+    noisy_register,
+)
 
 
 class Hamiltonian:
@@ -57,6 +63,7 @@ class Hamiltonian:
         """Instantiates a Hamiltonian object."""
         self.samples_obj = samples_obj
         self._qdict = {k: v.as_array(detach=True) for k, v in qdict.items()}
+
         self._device = device
         self._sampling_rate = sampling_rate
 
@@ -286,6 +293,7 @@ class Hamiltonian:
                     samples_dict[qid]["det"][slot.ti : slot.tf] += noise_det
                 # Gaussian beam loss in amplitude for global pulses only
                 # Noise is drawn at random for each pulse
+
                 if "amplitude" in self.config.noise_types:
                     amp_fraction = amp_fluctuation
                     if self.config.laser_waist is not None and is_global_pulse:
@@ -429,11 +437,21 @@ class Hamiltonian:
             )
             self._bad_atoms = dict(zip(self._qid_index, dist))
         if "doppler" in self.config.noise_types:
-            temp = self.config.temperature * 1e-6
+            temp = self.config.temperature * 1e-6  # Convert to K
             detune = np.random.normal(
                 0, doppler_sigma(temp), size=len(self._qid_index)
             )
             self._doppler_detune = dict(zip(self._qid_index, detune))
+
+        if "register" in self.config.noise_types:
+            # Register noise is applied to the coordinates of the atoms
+            xy_sigma, z_sigma = register_sigma_xy_z(
+                self.config.temperature,  # temperature in micro-Kelvin
+                self.config.trap_waist,
+                self.config.trap_depth,  # trap depth in micro-Kelvin
+            )
+            print("register noise:", xy_sigma, z_sigma)
+            self._qdict = noisy_register(self._qdict, xy_sigma, z_sigma)
 
     def _get_basis_name(self, with_leakage: bool) -> str:
         if len(self.samples_obj.used_bases) == 0:
@@ -478,7 +496,8 @@ class Hamiltonian:
 
         Warning:
             The refreshed noise parameters (when update=True) are only those
-            that change from shot to shot (ie doppler and state preparation).
+            that change from shot to shot (ie doppler, state preparation,
+            and register).
             Amplitude fluctuations change from pulse to pulse and are always
             applied in `_extract_samples()`.
 
@@ -486,9 +505,13 @@ class Hamiltonian:
             update: Whether to update the noise parameters.
         """
         if update:
+            print("update call")
             self._update_noise()
         self._extract_samples()
+        print("after extract samples:")
 
+        # add register noise use the _qdict to modify the register positions
+        # how to determine the register plane,
         def make_vdw_term(q1: QubitId, q2: QubitId) -> qutip.Qobj:
             """Construct the Van der Waals interaction Term.
 
@@ -531,6 +554,7 @@ class Hamiltonian:
             )
 
         def make_interaction_term(masked: bool = False) -> qutip.Qobj:
+            print("interactioncall:", self._qdict)
             if masked:
                 # Calculate the total number of good, unmasked qubits
                 effective_size = self._size - sum(self._bad_atoms.values())
--- a/pulser-simulation/pulser_simulation/simconfig.py
+++ b/pulser-simulation/pulser_simulation/simconfig.py
@@ -20,6 +20,7 @@ import warnings
 from dataclasses import dataclass, field, fields
 from typing import Any, Tuple, Type, TypeVar, Union, cast
 
+import numpy as np
 import qutip
 
 from pulser.noise_model import _LEGACY_DEFAULTS, NoiseModel, NoiseTypes
@@ -27,6 +28,7 @@ from pulser.noise_model import _LEGACY_DEFAULTS, NoiseModel, NoiseTypes
 MASS = 1.45e-25  # kg
 KB = 1.38e-23  # J/K
 KEFF = 8.7  # µm^-1
+WAVELENGH = 0.85  # µm
 
 T = TypeVar("T", bound="SimConfig")
 
@@ -41,8 +43,16 @@ SUPPORTED_NOISES: dict = {
         "eff_noise",
         "SPAM",
         "leakage",
+        "register",
     },
-    "XY": {"dephasing", "depolarizing", "eff_noise", "SPAM", "leakage"},
+    "XY": {
+        "dephasing",
+        "depolarizing",
+        "eff_noise",
+        "SPAM",
+        "leakage",
+        "register",
+    },
 }
 
 # Maps the noise model parameters with a different name in SimConfig
@@ -60,7 +70,36 @@ def doppler_sigma(temperature: float) -> float:
     Arg:
         temperature: The temperature in K.
     """
-    return KEFF * math.sqrt(KB * temperature / MASS)
+    return KEFF * math.sqrt(KB * temperature / MASS)    
+
+
+def register_sigma_xy_z(
+    temperature: float, trap_waist: float, trap_depth: float
+)-> tuple[float, float]:
+    """
+    Calculates the standar deviation for fluctions in atom position in the trap.
+    - 𝜎ˣʸ = √(T w²/(4 Uₜᵣₐₚ)), where T is temperaturae, w is the trap waist
+        and Uₜᵣₐₚ is the trap depth.
+    - 𝜎ᶻ = 𝜋 / 𝜆 √2 w 𝜎ˣʸ, 𝜆 is the wavelenght with a constant value of 0.85 µm
+
+    Parameters:
+    temperature (float): Temperature of the atoms in the trap (in Kelvin).
+    trap_depth (float): Depth of the trap (same units as temperature).
+    trap_waist (float): Waist of the trap (in µmeters).
+
+
+    Returns:
+    tuple: The standard deviations of the spatial position fluctuations
+    in the xy-plane (register_sigma_xy) and along the z-axis (register_sigma_z).
+    """
+
+    register_sigma_xy = math.sqrt(
+        temperature * trap_waist**2 / (4 * trap_depth)
+    )
+    register_sigma_z = (
+        math.pi / WAVELENGH * math.sqrt(2) * trap_waist * register_sigma_xy
+    )
+    return register_sigma_xy, register_sigma_z
 
 
 @dataclass(frozen=True)
@@ -91,6 +130,9 @@ class SimConfig:
               corresponding rates **eff_noise_rates** (in rad/µs).
             - "doppler": Local atom detuning due to finite speed of the
               atoms and Doppler effect with respect to laser frequency.
+            - register: Simulates the relation between fluctuations in the
+              position of the atoms in the trap, the temperature, and
+              trap parameters.
             - "amplitude": Gaussian damping due to finite laser waist
             - "SPAM": SPAM errors. Defined by **eta**, **epsilon** and
               **epsilon_prime**.
@@ -102,8 +144,12 @@ class SimConfig:
             noise.
         samples_per_run: Number of samples per noisy run.
             Useful for cutting down on computing time, but unrealistic.
-        temperature: Temperature, set in µK, of the Rydberg array.
+        temperature: Temperature, set in K, of the Rydberg array.
             Also sets the standard deviation of the speed of the atoms.
+        trap_depth: energy depth of the trap, set in the same units as
+            temperature (K). Used to calculate the register fluctuations.
+        trap_waist: Radius of the trapping laser beam at the focus point,
+            set in µm. Used to calculate the register fluctuations.
         laser_waist: Waist of the gaussian laser, set in µm, in global
             pulses.
         amp_sigma: Dictates the fluctuations in amplitude as a standard
@@ -120,6 +166,8 @@ class SimConfig:
     laser_waist: float = _LEGACY_DEFAULTS["laser_waist"]
     amp_sigma: float = _LEGACY_DEFAULTS["amp_sigma"]
     detuning_sigma: float = 0.0
+    trap_depth: float = 0.0
+    trap_waist: float = 0.0
     eta: float = _LEGACY_DEFAULTS["state_prep_error"]
     epsilon: float = _LEGACY_DEFAULTS["p_false_pos"]
     epsilon_prime: float = _LEGACY_DEFAULTS["p_false_neg"]
@@ -196,7 +244,15 @@ class SimConfig:
             raise TypeError(
                 f"'temperature' must be a float, not {type(self.temperature)}."
             )
+            
         self._change_attribute("temperature", self.temperature / 1e6)
+
+        if not isinstance(self.trap_depth, (int, float)):
+            raise TypeError(
+                f"'trap_depth' must be a float, not {type(self.trap_depth)}."
+            )
+        
+        #self._change_attribute("trap_depth", self.trap_depth / 1e6)
 
         NoiseModel._check_noise_types(cast(Tuple[NoiseTypes], self.noise))
         self._check_spam_dict()
@@ -223,6 +279,15 @@ class SimConfig:
     def doppler_sigma(self) -> float:
         """Standard deviation for Doppler shifting due to thermal motion."""
         return doppler_sigma(self.temperature)
+    
+    @property
+    def register_sigma_xy_z(self) -> tuple[float, float]:
+        """Standard deviation for fluctuations in the xy and z plane."""
+        register_sigma = register_sigma_xy_z(
+            self.temperature, self.trap_waist, self.trap_depth
+        )
+        return register_sigma[0],register_sigma[1]
+    
 
     def __str__(self, solver_options: bool = False) -> str:
         lines = [
@@ -244,6 +309,10 @@ class SimConfig:
             )
         if "doppler" in self.noise:
             lines.append(f"Temperature:           {self.temperature*1.e6}µK")
+        if "register" in self.noise:
+            lines.append(f"Temperature:          {self.temperature*1.e6}µK")
+            lines.append(f"Trap waist:           {self.trap_waist}µm")
+            lines.append(f"Trap depth:           {self.trap_depth}µK")
         if "amplitude" in self.noise:
             lines.append(f"Laser waist:           {self.laser_waist}μm")
             lines.append(f"Amplitude standard dev.:  {self.amp_sigma}")
@@ -294,3 +363,20 @@ class SimConfig:
     def supported_noises(self) -> dict:
         """Return the noises implemented on pulser."""
         return SUPPORTED_NOISES
+
+def noisy_register(q_dict: dict, register_sigma_xy, register_sigma_z):
+    "Add Gaussian noise to the positions of the register"
+    atoms = list(q_dict.keys())
+    num_atoms = len(list(atoms))
+    positions = list(q_dict.values())
+
+    if len(positions[0]) == 2:
+        positions = np.array(
+            [np.append(p, 0.0) for p in positions]
+        )  # Convert 2D positions to 3D
+
+    noise_xy = np.random.normal(0, register_sigma_xy, (num_atoms, 2))
+    noise_z = np.random.normal(0, register_sigma_z, num_atoms)
+    noise = np.column_stack((noise_xy, noise_z))
+    positions += noise
+    return {k: pos for (k, pos) in zip(atoms, positions)}
--- a/pulser-simulation/pulser_simulation/simconfig.py
+++ b/pulser-simulation/pulser_simulation/simconfig.py
@@ -70,14 +70,14 @@ def doppler_sigma(temperature: float) -> float:
     Arg:
         temperature: The temperature in K.
     """
-    return KEFF * math.sqrt(KB * temperature / MASS)    
+    return KEFF * math.sqrt(KB * temperature / MASS)
 
 
 def register_sigma_xy_z(
     temperature: float, trap_waist: float, trap_depth: float
-)-> tuple[float, float]:
+) -> tuple[float, float]:
     """
-    Calculates the standar deviation for fluctions in atom position in the trap.
+    Calculates the standar deviation for fluctions in atom position in the trap
     - 𝜎ˣʸ = √(T w²/(4 Uₜᵣₐₚ)), where T is temperaturae, w is the trap waist
         and Uₜᵣₐₚ is the trap depth.
     - 𝜎ᶻ = 𝜋 / 𝜆 √2 w 𝜎ˣʸ, 𝜆 is the wavelenght with a constant value of 0.85 µm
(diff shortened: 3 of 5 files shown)


pulser-core/pulser/noise_model.py
```diff
@@ -45,6 +46,7 @@
 _NOISE_TYPE_PARAMS: dict[NoiseTypes, tuple[str, ...]] = {
     "leakage": ("with_leakage",),
     "doppler": ("temperature",),
+    "register": ("trap_waist", "trap_depth"),  # temperature is also needed
```
a-quelle-pasqal: Looking at the `PARAM_TO_NOISE_TYPE` dict, it's assumed that each parameter only corresponds to a single noise type. How do we handle this?

pulser-core/pulser/noise_model.py
```diff
@@ -410,6 +427,7 @@ def _validate_parameters(param_vals: dict[str, Any]) -> None:
                     stacklevel=2,
                 )
 
+    # check noise_register, regsiter noise is optinal in the abstract repr
```
a-quelle-pasqal: typo, it should be "optional"

pulser-core/pulser/noise_model.py
```diff
@@ -196,6 +210,8 @@ class NoiseModel:
     detuning_sigma: float = 0.0
     relaxation_rate: float = 0.0
     dephasing_rate: float = 0.0
+    trap_waist: float = 0.0
```
a-quelle-pasqal, line 242: I think there needs to be some kind of validation in the init or somewhere that if any one of the 3 parameters is non-zero, they should all be non-zero.

pulser-simulation/pulser_simulation/hamiltonian.py
```diff
@@ -435,6 +441,15 @@ def _update_noise(self) -> None:
             )
             self._doppler_detune = dict(zip(self._qid_index, detune))
 
+        if "register" in self.config.noise_types:
+            # should be applied before the Hamiltonian is built
+            xy_sigma, z_sigma = register_sigma_xy_z(
```
a-quelle-pasqal: why not make a single function that spits out the noisy register positions directly? This two stage rocket makes things more convoluted than needed, since you don't use the sigma values after this.

pulser-simulation/pulser_simulation/hamiltonian.py
```diff
@@ -489,6 +505,8 @@ def _construct_hamiltonian(self, update: bool = True) -> None:
             self._update_noise()
         self._extract_samples()
 
+        # add register noise use the _qdict to modify the register positions
+        # how to determine the register plane,
```
a-quelle-pasqal: what about this comment? It seems to suggest some work still needs to be done, or can it be deleted?

pulser-simulation/pulser_simulation/simconfig.py
```diff
@@ -63,6 +73,40 @@ def doppler_sigma(temperature: float) -> float:
     return KEFF * math.sqrt(KB * temperature / MASS)
 
 
+def register_sigma_xy_z(
```
a-quelle-pasqal: why this function? The associated property in `SimConfig` does nothing but call this one, and the existence of both is a bit confusing. Maybe just move this into the implementation of the property.

tests/test_noise_model.py
```diff
@@ -307,10 +307,18 @@ def test_relevant_params(self):
             "samples_per_run",
         }
 
+        assert NoiseModel._find_relevant_params(
+            {"register"}, 0.0, 0.0, None
+        ) == {
```
a-quelle-pasqal, line 408: I think we should fix it somehow so that this also returns temperature. See also my comments on noise_model.py

tests/pulser_simulation/test_simulation.py
```diff
@@ -1947,6 +1948,8 @@ def test_amp_sigma_noise():
 
 
 def test_detuning_noise():
+    # NOTE: changing values in detuning_sigma will not change the results :(
+    # is this expected?
```
a-quelle-pasqal: there are asserts missing in the test cases. Also, two of the test cases are not correct. Please replace the test cases with

```
assert np.all(np.isclose(rydberg_0, np.array([-0.04902824] * (2 * duration ) + [0.])))
    assert np.all(np.isclose(rydberg_1, np.array([-0.04902824] * (2 * duration ) + [0.])))
    assert np.all(
        np.isclose(
            digital_0,
            np.array([-0.17550787] * (duration) + [0.0] * (duration + 1)),
        )
    )
    assert np.all(
        np.isclose(
            digital_1,
            np.array([-0.20112646] * (duration) + [0.0] * (duration + 1)),
        )
    )
```

a-quelle-pasqal: You can verify that with the above replacement, the test only passes if you put back the old values.

pulser-core/pulser/noise_model.py
murogrande: this was a reminder, I will delete this comment

pulser-simulation/pulser_simulation/hamiltonian.py
murogrande: I followed `doppler_sigma` implementation. It is built in the same way. In [`simconfig.py` ](https://github.com/pasqal-io/Pulser/blob/develop/pulser-simulation/pulser_simulation/simconfig.py#L57) you have the equation as a function, then in line 223 is [called](https://github.com/pasqal-io/Pulser/blob/develop/pulser-simulation/pulser_simulation/simconfig.py#L223) as property.

In `hamiltonian.py`, it is used only as [equation](https://github.com/pasqal-io/Pulser/blob/develop/pulser-simulation/pulser_simulation/hamiltonian.py#L434)

pulser-simulation/pulser_simulation/hamiltonian.py
murogrande: I will delete this

pulser-simulation/pulser_simulation/simconfig.py
murogrande: check the information above related to `doppler_sigma`

tests/pulser_simulation/test_simulation.py
murogrande: I update the test and it is working

pulser-core/pulser/noise_model.py
a-quelle-pasqal: Let's just leave this as is for now. As long as `_find_relevant_params` correctly returns `temperature` for `register` noise, we're probably good.

.pre-commit-config.yaml
```diff
@@ -31,3 +31,4 @@ repos:
         name: mypy
         entry: ./.mypy_script
         language: script
+        pass_filenames: false
```
HGSilveri: What is this for?

pulser-core/pulser/noise_model.py
```diff
@@ -196,6 +210,8 @@ class NoiseModel:
...
+    trap_depth: float = 0.0
```
HGSilveri: This is not a good default for this value because it would blow up the noise, it should rather be 
```suggestion
    trap_depth: float | None = None
```
and `STRICT_POSITIVE` instead of just `POSITIVE`

pulser-core/pulser/noise_model.py
HGSilveri: I think that if `temperature` is non-zero but `trap_waist` and `trap_depth` are, only `doppler` noise should be activated. So `temperature > 0` does not require `trap_*` to be defined but defined `trap_*` requires `temperature > 0`

pulser-core/pulser/noise_model.py
HGSilveri: I'd say it would be better to include `temperature` here too and substitute `_PARAM_NOISE_TYPE` with an explicit map between a param and a noise type. 
That being said, I do think that `temperature` should "activate" `doppler` noise while `trap_depth` and `trap_waist` "activate" `register` noise.
If you end up keeping this current form, make sure to make a special case in `_find_relevant_params` as @a-quelle-pasqal suggested.

pulser-core/pulser/noise_model.py
```diff
@@ -131,6 +138,9 @@ class NoiseModel:
     - **doppler**: Local atom detuning due to termal motion of the
       atoms and Doppler effect with respect to laser frequency.
       Parametrized by the ``temperature`` field.
+    - **register**: thermal fluctuations in the
+      register positions, parametrized by ``temperature``, ``trap_waist``
+      and, ``trap_depth``.
```
```suggestion
    - **register**: Thermal fluctuations in the
      register positions; parametrized by ``temperature``, ``trap_waist``
      and ``trap_depth``, which must all be defined.
```

pulser-simulation/pulser_simulation/simconfig.py
HGSilveri: We have actually deprecated `SimConfig` so I would refrain from adding any new parameters to it and keep it's usage exclusive to the `NoiseModel`

pulser-simulation/pulser_simulation/simconfig.py
```diff
@@ -294,3 +360,23 @@ def _check_eff_noise(self) -> None:
...
+    if len(positions[0]) == 2:
+        positions = np.array(
+            [np.append(p, 0.0) for p in positions]
+        )  # Convert 2D positions to 3D
+
+    noise_xy = np.random.normal(0, register_sigma_xy, (num_atoms, 2))
+    noise_z = np.random.normal(0, register_sigma_z, num_atoms)
+    noise = np.column_stack((noise_xy, noise_z))
+    positions += noise
+    return {k: pos for (k, pos) in zip(atoms, positions)}
```
HGSilveri: This is a bit of an odd placement for this function... I sounds like this (alongside `doppler_sigma`, `register_sigma_xy_z` and the associated constants) would be better placed inside a new file, e.g. `pulser_simulation/noise_functions.py`

pulser-simulation/pulser_simulation/hamiltonian.py
HGSilveri: Since we don't want to support new functionality in `SimConfig`, I second @a-quelle-pasqal 's suggestion

.pre-commit-config.yaml
murogrande: I had some issues with the `pre-commit` locally, I will delete this change.

pulser-simulation/pulser_simulation/simconfig.py
HGSilveri: Since we will have to move these again anyway in #916 , let's leave them here for now

pulser-core/pulser/noise_model.py
HGSilveri: Following discussion with @a-quelle-pasqal , I agree that if `trap_*` are defined but `temperature==0`, a warning should be raised instead of requiring `temperature>0`

pulser-core/pulser/noise_model.py
HGSilveri: Slight modification after discussion with @a-quelle-pasqal : we can keep `"temperature"` out of `NOISE_TYPE_PARAMS["register"]` since we won't enforce it to be non-zero. That will also allow us to keep the current `_PARAM_NOISE_TYPE` format

pulser-core/pulser/noise_model.py
```diff
@@ -30,6 +31,8 @@
 
 __all__ = ["NoiseModel"]
 
+WAVELENGTH = 0.85  # µm
```
HGSilveri: Let's be more specific and call this 
```suggestion
TRAP_WAVELENGTH = 0.85  # µm
```

.pre-commit-config.yaml
HGSilveri: This is still here, did you forget to push?

pulser-core/pulser/noise_model.py
```diff
@@ -196,6 +242,12 @@ class NoiseModel:
     detuning_sigma: float = 0.0
     relaxation_rate: float = 0.0
     dephasing_rate: float = 0.0
+    trap_waist: float = (
+        0.0  # if the trap depth > the trap waist should be not none
+    )
+    trap_depth: float | None = (
+        None  # can be 0.0 has to be non zero, it is none raise a error
+    )
```
HGSilveri: nit: let's clarify the comments and put them on their own line
```suggestion
    # if the trap depth is not None the trap waist should be 0.0
    trap_waist: float = 0.0
    # Must be defined when trap_waist > 0
    trap_depth: float | None = None
```

pulser-core/pulser/noise_model.py
```diff
@@ -274,6 +330,43 @@ def to_tuple(obj: tuple) -> tuple:
...
                 )
 
+    def register_sigma_xy_z(self) -> tuple[float, float]:
+        """Standard deviation for fluctuations in the xy and z plane."""
+        if self.trap_waist is None or self.trap_depth is None:
+            raise ValueError("Trap waist and depth must be defined.")
+        register_sigma = register_sigma_xy_z(
+            self.temperature, self.trap_waist, self.trap_depth
+        )
+        return register_sigma[0], register_sigma[1]
```
HGSilveri: We don't have methods for sampling noise for the parameters, I don't think we should start doing it for this one

pulser-core/pulser/noise_model.py
```diff
@@ -274,6 +330,43 @@ def to_tuple(obj: tuple) -> tuple:
...
+            "register" in true_noise_types
+            and (self.trap_waist == 0.0 or self.trap_depth is None)
+            and self.temperature != 0.0
+        ):
+            warnings.warn(
+                "Trap waist, depth, and temperature must be defined in "
+                + "order to "
+                + "simulate register noise. Only doppler noise will be used",
+                Warning,
+                stacklevel=0,
```
HGSilveri: Why are you setting these parameters in this way?

pulser-core/pulser/noise_model.py
```diff
@@ -274,6 +330,43 @@ def to_tuple(obj: tuple) -> tuple:
...
+            self.temperature, self.trap_waist, self.trap_depth
+        )
+        return register_sigma[0], register_sigma[1]
+
+    def _check_register_and_doppler_noise_params(
+        self, true_noise_types: set
+    ) -> None:
+        if (
+            "register" in true_noise_types
+            and (self.trap_waist == 0.0 or self.trap_depth is None)
```
HGSilveri: These conditions in isolation should result in an error, not a warning

pulser-core/pulser/noise_model.py
```diff
@@ -274,6 +330,43 @@ def to_tuple(obj: tuple) -> tuple:
...
+            warnings.warn(
+                "Trap waist, depth, and temperature must be defined in "
+                + "order to "
+                + "simulate register noise. Only doppler noise will be used",
+                Warning,
+                stacklevel=0,
+            )
+        elif (
+            "register" in true_noise_types
+            and (self.trap_waist == 0.0 or self.trap_waist != 0.0)
```
HGSilveri: This is a logical tautology (ie always true) :sweat_smile:

pulser-core/pulser/noise_model.py
```diff
@@ -274,6 +330,43 @@ def to_tuple(obj: tuple) -> tuple:
...
+    def register_sigma_xy_z(self) -> tuple[float, float]:
+        """Standard deviation for fluctuations in the xy and z plane."""
+        if self.trap_waist is None or self.trap_depth is None:
+            raise ValueError("Trap waist and depth must be defined.")
+        register_sigma = register_sigma_xy_z(
+            self.temperature, self.trap_waist, self.trap_depth
+        )
+        return register_sigma[0], register_sigma[1]
+
+    def _check_register_and_doppler_noise_params(
```
HGSilveri: This method should be checking two things **(when "register" is in `true_noise_types`)**:
if `self.trap_waist == 0.0 or self.trap_depth is None` -> Error
elif temperature == 0.0 -> Warning

HGSilveri: For consistency, it would also be better to make it a `staticmethod` like the others

.pre-commit-config.yaml
murogrande: sorry, I will remove it. But the MR was not ready. I will DM you when it is ready

pulser-core/pulser/noise_model.py
a-quelle-pasqal: what are the values that indicate absence? Since `trap_depth` is strictly positive, I'd expect `0.0` rather than `None` and similarly, since `trap_waist` is only non-negative, I'd expect `None` rather than `0.0`. See also the default values defined at the start of the `NoiseModel`

pulser-core/pulser/noise_model.py
```diff
@@ -131,6 +136,9 @@ class NoiseModel:
     - **doppler**: Local atom detuning due to termal motion of the
       atoms and Doppler effect with respect to laser frequency.
       Parametrized by the ``temperature`` field.
+    - **register**: thermal fluctuations in the
+      register positions, parametrized by ``temperature``, ``trap_waist``
+      and, ``trap_depth``, which must all be defined
```
HGSilveri: nit: Formatting so it appears nicely on the docs
```suggestion
    - **register**: Thermal fluctuations in the
      register positions, parametrized by ``temperature``, ``trap_waist``
      and, ``trap_depth``, which must all be defined.
```

pulser-core/pulser/noise_model.py
```diff
@@ -166,6 +174,10 @@ class NoiseModel:
             distribution centered in 0. Assumed to be the same for all
             channels (though each channel has its own randomly sampled
             value in each run). This noise is additive. Defaults to 0.
+        trap_waist: Refers to the radius of the trapping laser beam at the
+            focus point (in µm). Defaults to 0.
```
HGSilveri: Isn't this the waist of each individual trap?
```suggestion
        trap_waist: The waist of each optical trap at the focal point (in µm). 
            Defaults to 0.
```

pulser-core/pulser/noise_model.py
```diff
@@ -466,3 +552,23 @@ def from_abstract_repr(obj_str: str) -> NoiseModel:
                 obj_str
             )
         )
+
+
+def noisy_register(
```
HGSilveri: This function should be private. 
Actually, as @a-quelle-pasqal mentioned in the past, can't we just have on function that takes in the atom coordinates and the noise model and returns the new coordinates? Do we really need `_register_sigma_xy_z()` and `noisy_register()`?

pulser-simulation/pulser_simulation/hamiltonian.py
```diff
@@ -435,6 +436,14 @@ def _update_noise(self) -> None:
             )
             self._doppler_detune = dict(zip(self._qid_index, detune))
 
+        if (
+            "register" in self.config.noise_types
+            and self.config.trap_depth is not None
+        ):
+            # should be applied before the Hamiltonian is built
+            xy_sigma, z_sigma = self.config._register_sigma_xy_z()
+            self._qdict = noisy_register(self._qdict, xy_sigma, z_sigma)
```
HGSilveri: Exactly here, couldn't it simply be:
```suggestion
            self._qdict = noisy_register(self._qdict, self.config)
```

pulser-core/pulser/noise_model.py
murogrande: I see your point, but I don’t think merging them is a good idea because one function should ideally do one thing:

 - `_register_sigma_xy_z()` is responsible for computing the noise parameters from the trap properties.

- `_noisy_register()` is responsible for applying those parameters to the coordinates.

If we compress everything into `_noisy_register()`, the function will grow too large and mix different concerns (parameter computation + coordinate update). 

Keeping them separate makes the code clearer, easier to test, and easier to maintain.

pulser-simulation/pulser_simulation/hamiltonian.py
murogrande: I see your point here, let me change the code.

pulser-core/pulser/noise_model.py
HGSilveri: I agree that keeping the computation of a `sigma_xy_z` in a function is good. What I'm not so sure about is that we shouldn't call that function within `noisy_register()`, but let's leave it like this.

pulser-core/pulser/noise_model.py
murogrande: I was checking how the output looks like. I will change to staklevel=2, it will trigger the warning directly in the function the user calls

pulser-core/pulser/noise_model.py
murogrande: indeed, let me change the code

pulser-core/pulser/noise_model.py
murogrande: Indeed, this method only returns the sigmas. The sampling is done in `_noisy_register` to keep things consistent with how we handle other noise parameters.

pulser-core/pulser/noise_model.py
```diff
@@ -274,6 +299,65 @@ def to_tuple(obj: tuple) -> tuple:
                     stacklevel=2,
                 )
 
+    def _register_sigma_xy_z(self) -> tuple[float, float]:
```
HGSilveri: Can you make this a function instead of a method (like you did for `_noisy_register()`)?

pulser-core/pulser/noise_model.py
```diff
@@ -274,6 +299,65 @@ def to_tuple(obj: tuple) -> tuple:
...
+
+        Note: a k_B factor is absorbed in the trap depth (Uₜᵣₐₚ), so the units
+        of temperature and trap depth are the same.
+
+        Args:
+            temperature (float): Temperature (T) of the atoms in the trap
+            (in Kelvin).
+            trap_depth (float): Depth of the trap (Uₜᵣₐₚ)
+            (same units as temperature).
+            trap_waist (float): Waist of the trap (w) (in µmeters).
```
HGSilveri: Formatting
```suggestion
            temperature (float): Temperature (T) of the atoms in the trap
                (in Kelvin).
            trap_depth (float): Depth of the trap (Uₜᵣₐₚ)
                (same units as temperature).
            trap_waist (float): Waist of the trap (w) (in µmeters).
```

pulser-core/pulser/noise_model.py
```diff
@@ -274,6 +299,65 @@ def to_tuple(obj: tuple) -> tuple:
...
+                "order to simulate register noise."
+            )
+        elif temperature == 0.0:
+            warnings.warn(
+                "trap_waist, trap_depth, and temperature must be defined in "
+                "order to simulate register noise. Temperature is "
+                f"{temperature} µK. Register noise will have no effect",
+                UserWarning,
+                stacklevel=2,
+            )
```
HGSilveri: @a-quelle-pasqal @murogrande I know we agreed on this and it will work, but it can be quite inefficient as it will trigger the "noisy simulation" even when `temperature==0.0`(so there is actually no noise). Given that `"register"` noise is activated iff `"trap_waist"` or `"trap_depth"` are given, can't we just enforce that all three parameters must be given in that case?

```suggestion
        if trap_waist == 0.0 or trap_depth is None or temperature == 0.0:
            raise ValueError(
                "trap_waist, trap_depth and temperature must all"
                " be defined in order to simulate register noise."
            )
```

pulser-simulation/pulser_simulation/hamiltonian.py
```diff
@@ -435,6 +436,13 @@ def _update_noise(self) -> None:
             )
             self._doppler_detune = dict(zip(self._qid_index, detune))
 
+        if (
+            "register" in self.config.noise_types
+            and self.config.trap_depth is not None
```
HGSilveri: You don't need to check this, it should be guaranteed by the noise model

tests/pulser_simulation/test_simulation.py
```diff
@@ -1973,17 +1975,60 @@ def test_detuning_noise():
...
-    np.all(np.isclose(rydberg_0, np.array([-0.04902824] * (2 * duration + 1))))
-    np.all(np.isclose(rydberg_1, np.array([-0.04902824] * (2 * duration + 1))))
-    np.all(
+
+    assert np.all(
+        np.isclose(rydberg_0, np.array([-0.04902824] * (2 * duration) + [0.0]))
+    )
+    assert np.all(
+        np.isclose(rydberg_1, np.array([-0.04902824] * (2 * duration) + [0.0]))
+    )
```
HGSilveri: nit: For these cases you can use `np.testing.assert_allclose()`, e.g.
```suggestion
    np.testing.assert_allclose(
        rydberg_0, np.array([-0.04902824] * (2 * duration) + [0.0])
    )
    np.testing.assert_allclose(
        rydberg_1, np.array([-0.04902824] * (2 * duration) + [0.0])
    )
```

tests/test_noise_model.py
```diff
@@ -374,3 +431,169 @@ def test_repr(self):
...
+                    trap_depth=150.0,  # same units as temperature
+                    trap_waist=1.0,
+                    runs=1,
+                    samples_per_run=1,
+                )
+            )
+            == "NoiseModel(noise_types=('doppler', 'register'), runs=1, "
+            "samples_per_run=1, temperature=15.0, trap_waist=1.0, "
+            "trap_depth=150.0)"
+        )
```
HGSilveri, line 528: In this case, if you had not defined `temperature`, it would not have appeared in the repr and I think that's a mistake. 
We should still update `_find_relevant_params()` to add `temperature` to the relevant params when `"register"` noise is present, that will fix this
(also, add a test with this case)

tests/test_noise_model.py
```diff
@@ -374,3 +431,169 @@ def test_repr(self):
...
+            trap_waist=1.0,
+            trap_depth=150.0,
+            runs=1,
+            samples_per_run=1,
+        )
+        assert "register" in noise_model.noise_types
+
+
+def test_check_register_and_doppler_noise_params_error():
+    """Test ValueError is raised when register noise parameters are invalid."""
```
HGSilveri: nit: This description doesn't match the test

tests/test_noise_model.py
```diff
@@ -374,3 +431,169 @@ def test_repr(self):
...
+        ValueError, match="'trap_depth' must be greater than zero, not None"
+    ):
+        noise_model = NoiseModel(
+            temperature=1.0,
+            trap_depth=None,
+            trap_waist=1.0,
+            runs=1,
+            samples_per_run=1,
+        )
+        noise_model._register_sigma_xy_z()
```
HGSilveri: It's not `_register_sigma_xy_z()` that's raising the error, it's `NoiseModel`. You can probably delete this test

HGSilveri (commented):
Don't forget to update the JSON schema and test the (de)serialization in `test_abstract_repr.py`

pulser-core/pulser/noise_model.py
murogrande: is it possible to close this comment? or you can elaborate more about "sampling noise for the parameters"

pulser-core/pulser/noise_model.py
murogrande: let me change the code

pulser-core/pulser/noise_model.py
murogrande: let me change the code, thanks

tests/test_noise_model.py
murogrande: you are right, let me change the code

tests/test_noise_model.py
murogrande: If I add your above [comment](https://github.com/pasqal-io/Pulser/pull/915#discussion_r2309503812), we do not need to worry about the `_find_relevant_params()` because we are forcing the user to define the temperature when the trap parameters are different from the default value

murogrande:
Things to do:
- add another test to get 100% coverage. Specifically, line 446 of `hamiltonian.py`
- update the pulser schema

tests/test_noise_model.py
HGSilveri: I think you should still make  `_find_relevant_params()` return "temperature" for "register" noise

pulser-core/pulser/noise_model.py
```diff
@@ -136,6 +139,9 @@ class NoiseModel:
     - **doppler**: Local atom detuning due to termal motion of the
       atoms and Doppler effect with respect to laser frequency.
       Parametrized by the ``temperature`` field.
+    - **register**: Thermal fluctuations in the
+      register positions, parametrized by ``temperature``, ``trap_waist``
+      and, ``trap_depth``, which must all be defined
```
a-corni: Do you want to add the formula used to compute the shift in position from temperature, trap_waist and trap_depth ?

.pre-commit-config.yaml
```diff
@@ -30,4 +30,5 @@ repos:
       - id: mypy
         name: mypy
         entry: ./.mypy_script
-        language: script
\ No newline at end of file
+        language: script
+        pass_filenames : false
```
HGSilveri, line 34: I see this is back :sweat_smile:

pulser-core/pulser/noise_model.py
```diff
@@ -321,6 +399,7 @@ def _find_relevant_params(
                 or nt_ == "detuning"
                 or (nt_ == "amplitude" and amp_sigma != 0.0)
                 or (nt_ == "SPAM" and state_prep_error != 0.0)
+                or (nt_ == "register" and temperature != 0.0)
```
HGSilveri: Is this still needed now that `"register"` requires `temperature > 0.0`?
Sorry, I know we have been going back and forth on this

pulser-core/pulser/noise_model.py
```diff
@@ -306,12 +330,66 @@ def to_tuple(obj: tuple) -> tuple:
                     stacklevel=2,
                 )
 
+    @staticmethod
+    def _register_sigma_xy_z(
```
HGSilveri: nit: Do you have any objection to making this a regular function (ie not a staticmethod)? 
It's kind of an odd exception like this

pulser-core/pulser/noise_model.py
murogrande: sure let me change the code

pulser-core/pulser/noise_model.py
murogrande: I wrote you about this, probably, I was not clear enough. I will take out the temperature.

pulser-core/pulser/noise_model.py
murogrande: I will change the code

.pre-commit-config.yaml
murogrande: is it a bad idea to leave this? I do not understand well what is the reason to have a `.mypy.ini` script (instead of using directly mypy mirror), but when I run `pre-commit`, mypy is launching a lot of processes making my computer slow and the only way to stop this was using this `pass_filenames`.
Will be a problem for you if we add `pass_filenames`?

pulser-core/pulser/noise_model.py
HGSilveri: Sorry, I either missed it or misunderstood

pulser-core/pulser/noise_model.py
a-corni: (Also very minor, but since it's being displayed in the documentation)
```suggestion
      and ``trap_depth``, which must all be defined.
```

.pre-commit-config.yaml
HGSilveri: My issue was that I didn't really know what this was doing. I took a look at the pre-commit docs and this looks fine, we can keep it

pulser-core/pulser/noise_model.py
```diff
@@ -31,11 +32,14 @@
 
 __all__ = ["NoiseModel"]
 
+TRAP_WAVELENGTH = 0.85  # µm
```
a-corni, line 35: We could leave this as an environment variable, such that users can modify it from a notebook easily (see what was done [here](https://github.com/pasqal-io/Pulser-myQLM/blob/main/pulser_myqlm/constants.py#L10))

pulser-core/pulser/noise_model.py
```diff
@@ -306,12 +330,66 @@ def to_tuple(obj: tuple) -> tuple:
...
+        return register_sigma_xy, register_sigma_z
+
+    @staticmethod
+    def _check_register_noise_params(
+        true_noise_types: Collection[NoiseTypes],
+        trap_waist: float,
+        trap_depth: float | None,
+        temperature: float,
+    ) -> None:
+        if "register" not in true_noise_types:
```
a-corni, line 346: Since I had to note it down do understand what was going on, I would suggest to add a comment
```suggestion
        if "register" not in true_noise_types:
            # trap_waist and trap_depth have default values
```

pulser-core/pulser/noise_model.py
```diff
@@ -306,12 +330,66 @@ def to_tuple(obj: tuple) -> tuple:
...
+            return
+        if trap_waist == 0.0 or trap_depth is None or temperature == 0.0:
```
a-corni, line 349: Since `sigma_xy` is proportional to the temperature, it's not really an issue if temperature is equal to 0 (there will simply be no modification of the register). It's more a check to perform at the level of the generation of the noisy register: if the temperature is 0, skip the sampling to skip doing unnecessary operations. (note: that's what we do with doppler noise IIRC)

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -521,6 +522,7 @@ def get_hamiltonian(
         Args:
             time: The specific time at which we want to extract the
                 Hamiltonian (in ns).
+            noiseless: If True, returns the Hamiltonian without noise.
```
a-corni, line 524: Thanks !

pulser-simulation/pulser_simulation/simconfig.py
```diff
@@ -28,6 +28,7 @@
 KB = 1.38e-23  # J/K
 KEFF = 8.7  # µm^-1
```
a-corni, line 29: @HGSilveri Should we move these constants + doppler_sigma to noise_model.py ? (perhaps not in this MR, can open an issue with this) + we could make these environment variables (if, for instance, someone wants to work with another specie or refine the value of MASS, KB, KEFF)

a-corni (commented):
Hey @murogrande ! I don't understand the changes you did in 
- pulser-core/pulser/register/register_layout.py
- pulser-core/pulser/sampler/samples.py
- pulser-core/pulser/waveforms.py (this one especially)
Were they mandatory ? :)

pulser-core/pulser/noise_model.py
murogrande: sorry, `regular function` do mean a method of the class where I can use self?
or you mean a function outside of the class?

a-corni: To me, a function outside of the class

HGSilveri: Yes, a function outside the class

pulser-simulation/pulser_simulation/simconfig.py
HGSilveri: This will all be moved in #916 once this is merged

pulser-core/pulser/noise_model.py
murogrande: I like the idea, let me change the code

pulser-core/pulser/noise_model.py
HGSilveri: We have been going back and forth on this. Currently, the consensus was that if `trap_depth` and `trap_waist` are defined but temperature is zero it should throw an error because the user wants to have register noise but failed to specify one of the parameters

pulser-core/pulser/noise_model.py
HGSilveri: I think this should be done in a future PR tbh (namely in #916)

pulser-core/pulser/noise_model.py
```diff
@@ -32,7 +33,7 @@
 
 __all__ = ["NoiseModel"]
 
-TRAP_WAVELENGTH = 0.85  # µm
+TRAP_WAVELENGTH = float(os.getenv("TRAP_WAVELENGTH", 0.85))  # µm
```
HGSilveri: This is not exactly what @a-corni  meant, I think. How about we leave this for a future PR?

tests/test_abstract_repr.py
```diff
@@ -2819,15 +2830,18 @@ def test_noise_optional_params(detuning_sigma, det_hf_psd, det_hf_freqs):
...
         detuning_hf_freqs=det_hf_freqs,
-        temperature=10.0,
+        temperature=temperature,
         with_leakage=True,
         eff_noise_rates=(0.1,),
         eff_noise_opers=(np.random.rand(3, 3),),
         hyperfine_dephasing_rate=1.5,
     )
     repr = noise._to_abstract_repr()
+    print(repr)
```
HGSilveri: nit: You forgot this print

pulser-core/pulser/noise_model.py
a-corni: Sounds good to me :)

pulser-core/pulser/noise_model.py
a-corni: Sounds good to me to do it there.

pulser-core/pulser/noise_model.py
```diff
@@ -136,6 +139,13 @@ class NoiseModel:
     - **doppler**: Local atom detuning due to termal motion of the
       atoms and Doppler effect with respect to laser frequency.
       Parametrized by the ``temperature`` field.
+    - **register**: Thermal fluctuations in the
+      register positions, parametrized by ``temperature``, ``trap_waist``
+      and, ``trap_depth``, which must all be defined,
+      - Plane standard deviation: 𝜎ˣʸ = √(T w²/(4 Uₜᵣₐₚ)), where T is
+      temperature, w is the trap waist and Uₜᵣₐₚ is the trap depth.
+      - Off plane standard deviation: 𝜎ᶻ = 𝜋 / 𝜆 √2 w 𝜎ˣʸ, where 𝜆 is the
+      trap wavelength with a constant value of 0.85 µm
```
a-corni: I let you check it [in the documentation](https://pulser--915.org.readthedocs.build/en/915/noise_model.html), but it's not rendering great. May I suggest
```suggestion
    - **register**: Thermal fluctuations in the
      register positions, parametrized by ``temperature``, ``trap_waist``
      and, ``trap_depth``, which must all be defined,

      - Plane standard deviation: :math:`\sigma^{xy} = \sqrt{\frac{T w²}{4 U_{trap}}}`, where T is
      temperature, w is the trap waist and :math:`U_{trap}` is the trap depth.
      - Off plane standard deviation: :math:`\sigma^z = \frac{\pi}{\lambda \sqrt{2} w \sigma^{xy}}`, where :math:`\lambda` is the
      trap wavelength with a constant value of 0.85 µm
```

murogrande: I have issues with the latex code, flake does not like it
`pulser-core/pulser/noise_model.py:147:14: W605 invalid escape sequence '\s'
pulser-core/pulser/noise_model.py:147:28: W605 invalid escape sequence '\s'
pulser-core/pulser/noise_model.py:150:14: W605 invalid escape sequence '\s'
pulser-core/pulser/noise_model.py:150:31: W605 invalid escape sequence '\p'
pulser-core/pulser/noise_model.py:150:36: W605 invalid escape sequence '\l'
pulser-core/pulser/noise_model.py:150:44: W605 invalid escape sequence '\s'
pulser-core/pulser/noise_model.py:150:55: W605 invalid escape sequence '\s'`

murogrande: is this only happening to me?

murogrande: I fix this, let's see how the pipeline goes

a-corni: Indeed, you need to add r at the beginning ;)

murogrande: just in case the second equation: is 
$\sigma^z = \frac{\pi}{\lambda} \sqrt{2} w \sigma^{xy}$

a-corni: Equations look great like this ;) 
If you want to make a sub list, you should normally skip one line before the first "- ..."
image

a-corni: (see my suggestion above ;) )

pulser-core/pulser/noise_model.py
```diff
@@ -136,6 +139,16 @@ class NoiseModel:
...
+    - **register**: Thermal fluctuations in the
+      register positions, parametrized by ``temperature``, ``trap_waist``
+      and, ``trap_depth``, which must all be defined.
+      (1) Plane standard deviation fluctuation given by:
+      :math:`\sigma^{xy} = \sqrt{\frac{T w²}{4 U_{trap}}}`, where T is
+      temperature, w is the trap waist and :math:`U_{trap}` is
+      the trap depth.
+      (2) Off plane standard deviation fluctuation given by:
+      :math:`\sigma^z = \frac{\pi}{\lambda}\sqrt{2} w \sigma^{xy}`, where
+      :math:`\lambda` is the trap wavelength with a constant value of 0.85 µm.
```
a-corni, line 151: That works, otherwise I would have suggested to follow https://github.com/mkdocstrings/mkdocstrings/issues/135
```suggestion
    - **register**: Thermal fluctuations in the
      register positions, parametrized by ``temperature``, ``trap_waist``
      and, ``trap_depth``, which must all be defined.
      
          - Plane standard deviation fluctuation given by:
          :math:`\sigma^{xy} = \sqrt{\frac{T w²}{4 U_{trap}}}`, where T is
          temperature, w is the trap waist and :math:`U_{trap}` is
          the trap depth.
          - Off plane standard deviation fluctuation given by:
          :math:`\sigma^z = \frac{\pi}{\lambda}\sqrt{2} w \sigma^{xy}`, where
          :math:`\lambda` is the trap wavelength with a constant value of 0.85 µm.
```

a-corni: Note: this is what was tried in [this commit](https://github.com/pasqal-io/Pulser/pull/915/commits/5941002f2cb20290200dd239eb623f7c5904b5ce) and [this commit](https://github.com/pasqal-io/Pulser/pull/915/commits/7a08b811fb34f2c03f6ad22631ad3916203277c7), without being successful.

a-corni (approved):
I will wait for @HGSilveri's approval but it overall looks good to me ! Thanks for the impressive work @murogrande !

tests/test_noise_model.py
a-quelle-pasqal: Looking at the code, this is now done.

PR #917, opened 2025-08-18
Support time-dependent high-frequency detuning noise

kbidzhiev:
Time dependent high frequency **detuning** noise.

$\delta_{hf}(t) = \sqrt{2} \sum \sqrt{(f_{i+1} - f_i) * S(f_i)} * \cos(2\pi [f_i t + \varphi_i])$,
where $\varphi_i$ is a random phase sampled from uniform distribution $\varphi_i \sim U[0,1)$

To use high frequency detuning noise in the `NoiseModel`, the user is expected to provide 2 ArrayLike containers (list, tuple or array), namely Power Spectral Density `detuning_hf_psd` and relevant frequency support `detuning_hf_freqs`

```python
noise_mod = NoiseModel(
   detuning_hf_psd=(1, 4, 2),  # list, tuple, np.array
   detuning_hf_freqs=(3, 6, 7),  # list, tuple, np.array
   runs=1
)
```

`detuning_hf_psd` and `detuning_hf_freqs` must either be empty (default: `()` empty tuple) or 1D tuple(ArrayLike) with at least two positive elements of the same length. The frequency array `detuning_hf_freqs ` must be strictly increasing.


pulser-core/pulser/json/abstract_repr/schemas/noise-schema.json
```diff
@@ -36,6 +36,15 @@
...
         },
+        "detuning_high_freq": {
+          "type": "array",
+          "items": {
+            "type": "array",
+            "minItems": 2,
+            "maxItems": 2,
+            "items": { "type": "number" }
+          }
+        },
```
HGSilveri: You'll need to open an MR on Pulser-Schema with these changes, rather than adding them directly here

pulser-core/pulser/noise_model.py
```diff
@@ -166,6 +172,9 @@ class NoiseModel:
             distribution centered in 0. Assumed to be the same for all
             channels (though each channel has its own randomly sampled
             value in each run). This noise is additive. Defaults to 0.
+        detuning_high_freq: Defines high frequency fluctuations of detuning
```
HGSilveri: A few remarks here:

- I think a more appropriate name for this parameter would be `detuning_psd` 
- The format and order of the `freq` and `psd` should be clear in the docstring
- I find it unintuitive that the `psd` (ie the y-axis) comes before the `freq`. Wouldn't it be more natural to give `(freq, psd)`?
- The units of both `freq` and `psd` should be specified
- Both arrays needing to have the same len and dim=1 should be specified too

pulser-core/pulser/noise_model.py
```diff
@@ -194,6 +203,7 @@ class NoiseModel:
     laser_waist: float | None = None
     amp_sigma: float = 0.0
     detuning_sigma: float = 0.0
+    detuning_high_freq: Sequence[ArrayLike] = ()
```
HGSilveri: I think the type-hint can be clearer:
```suggestion
    detuning_high_freq: tuple[ArrayLike, ArrayLike] = ()
```
I think in this case it might even make sense to go a step further a use a typed `NamedTuple` for this quantity, but we can discuss that

pulser-core/pulser/noise_model.py
```diff
@@ -317,6 +330,35 @@ def _check_noise_types(noise_types: Sequence[NoiseTypes]) -> None:
...
+    @staticmethod
+    def _check_detuning_high_frequency_noise(
+        detuning_hf: Sequence[ArrayLike],
+    ) -> None:
+        if len(detuning_hf) == 2:
+            psd = detuning_hf[0]
+            freqs = detuning_hf[1]
+            if not (
+                isinstance(psd, np.ndarray)
+                and isinstance(freqs, np.ndarray)
```
HGSilveri: This will prevent you from using `torch.Tensor`s, is that not a concern?

pulser-core/pulser/noise_model.py
```diff
@@ -317,6 +330,35 @@ def _check_noise_types(noise_types: Sequence[NoiseTypes]) -> None:
...
+                and len(psd) == len(freqs)
+                and len(psd) > 1
+                and len(freqs) > 1
+                and psd.ndim == freqs.ndim == 1
+            ):
+                raise ValueError(
+                    "Power Spectral Density and the frequencies"
+                    " expected be 1D NumPy arrays of the same"
```
HGSilveri: The requirement that they are Numpy arrays seems unnecessary but if you have a good justification for it, it should be reflected in the type-hint

pulser-core/pulser/noise_model.py
```diff
@@ -466,3 +508,50 @@ def from_abstract_repr(obj_str: str) -> NoiseModel:
...
         )
+
+    @staticmethod
+    def generate_detuning(
+        detuning_sigma: float,
+        detuning_high_freq: tuple[ArrayLike, ArrayLike],
+        times: ArrayLike,
+        rng: Generator = np.random.default_rng(),
+    ) -> ArrayLike:
+        """Compute δ_hf(t) + δ_σ.
```
HGSilveri: This method looks out of place here... I believe it will be moved in #916 anyway, so maybe just make it a function instead.

pulser-core/pulser/noise_model.py
```diff
@@ -466,3 +508,50 @@ def from_abstract_repr(obj_str: str) -> NoiseModel:
                 obj_str
             )
         )
+
+    @staticmethod
+    def generate_detuning(
```
HGSilveri: The name could be more descriptive (and private), e.g.
```suggestion
    def _generate_detuning_fluctuations(
```

pulser-core/pulser/noise_model.py
```diff
@@ -466,3 +508,50 @@ def from_abstract_repr(obj_str: str) -> NoiseModel:
...
+            det_cst_term = rng.normal(0.0, detuning_sigma)
+
+        if detuning_high_freq:
+            psd = (detuning_high_freq[0])[:-1]
+            freqs = (detuning_high_freq[1])[:-1]
+            df = np.diff(detuning_high_freq[1])
+            scale = np.sqrt(2.0 * df * psd)
+            phases = rng.uniform(0.0, 1.0, size=len(freqs) - 1)
+            arg = 2.0 * np.pi * (np.outer(freqs, times) + phases[:, None])
+            det_hf = (scale[:, None] * np.cos(arg)).sum(axis=0)
```
HGSilveri: I'll trust you on this for now, just commenting so I come back to it later

pulser-core/pulser/noise_model.py
kbidzhiev: moved to a function

pulser-core/pulser/noise_model.py
kbidzhiev: I used Sequence to avoid ugly default value for 
```python
tuple[ArrayLike, ArrayLike] = ([], [])

Comment (kbidzhiev) [reply pulser-core/pulser/noise_model.py]: anyway, tuple is more natural

File: pulser-core/pulser/noise_model.py (hunk 4 above)
Comment (a-quelle-pasqal) [reply pulser-core/pulser/noise_model.py]: yes I would say let's keep the options open. I don't think we're ever going to implement differentiation w.r.t. the PSD, but we could at least let the user specify it using torch if they feel more comfortable that way.

File: pulser-simulation/pulser_simulation/hamiltonian.py [hunk 9]
```diff
@@ -309,10 +309,10 @@ def add_noise(
                 ch_amp_fluctuation = max(
                     0, np.random.normal(1.0, self.config.amp_sigma)
                 )
-                ch_det_fluctuation = (
-                    np.random.normal(0.0, self.config.detuning_sigma)
-                    if self.config.detuning_sigma
-                    else 0.0
+                ch_det_fluctuation = self.config.generate_detuning(
```
a-quelle-pasqal: this is no longer a number, but an array. Does the logic in here still function as is?

tests/pulser_simulation/test_simulation.py
```diff
@@ -1987,3 +1987,88 @@ def test_detuning_noise():
...
+
+    sim = QutipEmulator.from_sequence(
+        seq,
+        noise_model=noise_mod,
+    )
+    h = sim._hamiltonian._hamiltonian
+
+    sim_samples = sim._hamiltonian.samples
+
+    # rydberg_0 = sim_samples["Local"]["ground-rydberg"]["q0"]["det"]
```
a-quelle-pasqal: is this test finished? I don't see any asserts. Furthermore, all your pipelines are failing because code-coverage for the tests is not 100%. You'll need to add more.

pulser-core/pulser/json/abstract_repr/schemas/noise-schema.json
kbidzhiev: ok, done one schema side.

So does it make sense to add it here ?

pulser-core/pulser/noise_model.py
kbidzhiev: Done! thanks for your comments

pulser-core/pulser/noise_model.py
kbidzhiev: initial design with tuple of two ArrayLike is changed to be 2 separate ArrayLike.

We discussed `NamedTuple ` and dataclass. Both create additional burden for UX.

pulser-core/pulser/noise_model.py
kbidzhiev: removed explicit check on instance

pulser-core/pulser/noise_model.py
kbidzhiev: removed requirement :)

pulser-core/pulser/noise_model.py
kbidzhiev: I use non-private function `generate_detuning_fluctuations` to be able to call it later in `class Hamiltonian`

pulser-core/pulser/noise_model.py
kbidzhiev: I've added explicit test with very simplified formula. 

`test_noise_hf_detuning_generation()`

pulser-simulation/pulser_simulation/hamiltonian.py
kbidzhiev: In add_noise, the scalar input `ch_detuning` was broadcast against an array, so operations like `array += number` worked via NumPy broadcasting. In the current version, `ch_detuning` is explicitly an array, so the update is performed as `array += array instead`.

tests/pulser_simulation/test_simulation.py
kbidzhiev: it was in a draft stage. Now you can have a look on tests as well

tests/pulser_simulation/test_simulation.py
```diff
@@ -1987,3 +1987,125 @@ def test_detuning_noise():
             np.array([-0.20112646] * (duration) + [0.0] * (duration + 1)),
         )
     )
+
+
+def test_hf_detuning_noise_validation():
```
kbidzhiev: @HGSilveri I wrote tests here because I saw above `test_detuning_noise()`.
Does it make sense to move this test to `test_noise_model.py` ?

pulser-core/pulser/noise_model.py
```diff
@@ -166,6 +176,16 @@ class NoiseModel:
             distribution centered in 0. Assumed to be the same for all
             channels (though each channel has its own randomly sampled
             value in each run). This noise is additive. Defaults to 0.
+        detuning_hf_psd: Power Spectral Density(PSD) is 1D array (in Hz²/Hz)
+            provided together with `detuning_hf_freqs` define high frequency
+            noise contribution of time dependent detuning (in rad/µs).
+            Must either be empty or an array with at least two values,
```
a-quelle-pasqal: it's not an array, but a tuple, looking at the below type hints. Same goes for the corresponding tuple of frequencies. does it make sense to allow `Sequence[Number]`?

tests/pulser_simulation/test_simulation.py
a-quelle-pasqal: I think we'll have to move this to `test_noisy_samples` (although that should be renamed also) as part of #916 . I don't really care where this goes for now.

pulser-core/pulser/noise_model.py
kbidzhiev: Sequence doesn't take into account account numpy arrays

pulser-core/pulser/json/abstract_repr/schemas/noise-schema.json
HGSilveri: Once the changes in Pulser-schema are merged, you should take the new schema and copy-paste it here.

pulser-core/pulser/noise_model.py
HGSilveri: Actually, you turned it into a normal method instead, right? It's a bit out of place since we don't do that for any other noise types, making it a standalone (private) function would be more appropriate imo

pulser-core/pulser/noise_model.py
```diff
@@ -134,8 +139,13 @@ class NoiseModel:
...
       and ``amp_sigma``.
-    - **detuning**: Detuning fluctuations, parametrized by
-      ``detuning_sigma``.
+    - **detuning**: Detuning fluctuations consisting of two
+      components:
+      (1) constant offset (zero-frequency), parameterized by
+      ``detuning_sigma``;
+      (2) time-dependent high-frequency fluctuations, defined by the
+      power spectral density ``detuning_hf_psd`` over the relevant
+      ``detuning_hf_freqs`` frequencies support.
```
HGSilveri, line 148: Could you maybe add the formula here?

pulser-core/pulser/noise_model.py
HGSilveri: This will be moved to pulser-core in #916 so I wouldn't worry about it being private. If you make it public, it will appear in the docs and I don't think that's a good idea

tests/pulser_simulation/test_simulation.py
HGSilveri: I'd say the validation should be moved to `test_noise_model.py`, the other one you can leave here for now

pulser-core/pulser/noise_model.py
```diff
@@ -317,6 +346,40 @@ def _check_noise_types(noise_types: Sequence[NoiseTypes]) -> None:
...
+    @staticmethod
+    def _check_detuning_hf_noise(
+        psd: tuple[float, ...],
+        freqs: tuple[float, ...],
+    ) -> None:
+        if (psd == ()) ^ (freqs == ()):
+            raise ValueError(
+                "psd and freqs must either both be"
+                " empty tuples or both be provided."
+            )
```
```suggestion
            raise ValueError(
                "`detuning_hf_psd` and `detuning_hf_freqs` must either "
                "both be empty tuples or both be provided."
            )
```

pulser-core/pulser/noise_model.py
```diff
@@ -317,6 +346,40 @@ def _check_noise_types(noise_types: Sequence[NoiseTypes]) -> None:
...
+        if psd_a.size != freqs_a.size:
+            raise ValueError("psd and freqs are expected have same length.")
+
+        if not (np.all(psd_a > 0) and np.all(freqs_a > 0)):
+            raise ValueError(
+                "psd and freqs are expected have positive values."
+            )
+
+        if np.any(np.diff(freqs_a) < 0):
+            raise ValueError("freqs are expected be monotonously growing.")
```
HGSilveri: On all of these, I would replace
- `psd -> 'detuning_hf_psd'`
- `freqs -> 'detuning_hf_freqs'`

pulser-core/pulser/noise_model.py
```diff
@@ -317,6 +346,40 @@ def _check_noise_types(noise_types: Sequence[NoiseTypes]) -> None:
...
+            )
+
+        if psd == ():
+            return
+
+        psd_a = np.asarray(psd)
+        freqs_a = np.asarray(freqs)
+
+        if psd_a.ndim != 1 or freqs_a.ndim != 1:
+            raise ValueError("psd and freqs are expected be 1D arrays.")
```
HGSilveri: It's a bit misleading to call them arrays when they have to be tuples, no?

pulser-core/pulser/noise_model.py
```diff
@@ -317,6 +346,40 @@ def _check_noise_types(noise_types: Sequence[NoiseTypes]) -> None:
...
+
+        if psd_a.size <= 1 or freqs_a.size <= 1:
+            raise ValueError("psd and freqs are expected be length > 1.")
+
+        if psd_a.size != freqs_a.size:
+            raise ValueError("psd and freqs are expected have same length.")
```
HGSilveri: nit: If you make this check up you can simplify the one above

pulser-core/pulser/noise_model.py
kbidzhiev: ok, it means I'm using private function as public in the hamiltonian class

HGSilveri: That's true, but it's only temporary and will be handled in #916

pulser-core/pulser/noise_model.py
kbidzhiev: changed to be 1d tuples

pulser-core/pulser/noise_model.py
kbidzhiev: done

pulser-core/pulser/noise_model.py
kbidzhiev: Which one above ?

I need them to be :
- empty or not
- if not empty :
  - 1D
  - of the same size
  - not be size == 1

pulser-core/pulser/noise_model.py
kbidzhiev: done

tests/pulser_simulation/test_simulation.py
kbidzhiev: done

pulser-core/pulser/json/abstract_repr/schemas/noise-schema.json
kbidzhiev: Generated and copypasted

pulser-core/pulser/noise_model.py
kbidzhiev: done

pulser-core/pulser/noise_model.py
kbidzhiev: done. it's private

pulser-core/pulser/noise_model.py
kbidzhiev: done

pulser-core/pulser/noise_model.py
kbidzhiev: changed to explicitly require tuple. 

Inside there is a function that translates lists, arrays and tuples into a tuple

kbidzhiev:
> I think you could still use more tests, namely:
> 
> * One where you include this noise in an emulation
> * Tests for (de)serialization of a NoiseModel with these new parameters in `test_abstract_repr.py`

Could you be more specific ? In which file are emulation with Noise are tested ? 
I've added test in `test_abstract_repr.py`. Did you mean that one ?

pulser-core/pulser/noise_model.py
HGSilveri: If you check that they have the same size first, you can then check the length of just one of them. It's a nit, feel free to resolve this without addressing it

HGSilveri:
> > I think you could still use more tests, namely:
> > 
> > * One where you include this noise in an emulation
> > * Tests for (de)serialization of a NoiseModel with these new parameters in `test_abstract_repr.py`
> 
> Could you be more specific ? In which file are emulation with Noise are tested ? I've added test in `test_abstract_repr.py`. Did you mean that one ?

In `test_simulation.py` we test emulations with different types of noise. It could be nice to have a test with this new noise type as well

pulser-core/pulser/noise_model.py
kbidzhiev: done, thanks, it was a good comment.

Firstly I misunderstood it

pulser-core/pulser/noise_model.py
```diff
@@ -364,21 +367,33 @@ def _check_detuning_hf_noise(
...
+            raise ValueError(
+                "`detuning_hf_psd` and `detuning_hf_freqs`"
+                " are expected be 1D tuples."
+            )
 
         if psd_a.size != freqs_a.size:
-            raise ValueError("psd and freqs are expected have same length.")
+            raise ValueError(
+                "`detuning_hf_psd` and `detuning_hf_freqs`"
+                " are expected have same length."
```
```suggestion
                "`detuning_hf_psd` and `detuning_hf_freqs`"
                " are expected to have the same length."
```

pulser-core/pulser/noise_model.py
```diff
@@ -364,21 +367,33 @@ def _check_detuning_hf_noise(
...
+            )
+
+        if psd_a.size <= 1:
+            raise ValueError(
+                "`detuning_hf_psd` and `detuning_hf_freqs`"
+                " are expected be length > 1."
```
```suggestion
                "`detuning_hf_psd` and `detuning_hf_freqs`"
                " are expected to have length > 1."
```

pulser-core/pulser/noise_model.py
```diff
@@ -364,21 +367,33 @@ def _check_detuning_hf_noise(
...
+            )
 
         if not (np.all(psd_a > 0) and np.all(freqs_a > 0)):
             raise ValueError(
-                "psd and freqs are expected have positive values."
+                "`detuning_hf_psd` and `detuning_hf_freqs`"
+                " are expected have positive values."
```
```suggestion
                "`detuning_hf_psd` and `detuning_hf_freqs`"
                " are expected to have positive values."
```

pulser-core/pulser/noise_model.py
```diff
@@ -364,21 +367,33 @@ def _check_detuning_hf_noise(
...
             )
 
         if np.any(np.diff(freqs_a) < 0):
-            raise ValueError("freqs are expected be monotonously growing.")
+            raise ValueError(
+                "`detuning_hf_freqs` are expected be monotonously growing."
```
```suggestion
                "`detuning_hf_freqs` are expected to be monotonously growing."
```

pulser-core/pulser/noise_model.py
```diff
@@ -530,7 +545,7 @@ def from_abstract_repr(obj_str: str) -> NoiseModel:
             )
         )
 
-    def generate_detuning_fluctuations(
+    def _generate_detuning_fluctuations(
```
HGSilveri: This is indeed private now but it's still a `NoiseModel` method. My issue with that is just one of consistency, since we don't do that for any of the other noise types. I would prefer it if it were a regular function that takes in the `NoiseModel` as an argument instead, ie
```
def _generate_detuning_fluctuations(noise_model: NoiseModel, times: ArrayLike, ...) -> np.ndarray
```

tests/test_abstract_repr.py
```diff
@@ -2808,3 +2816,9 @@ def test_noise_optional_params(detuning_sigma):
     )
     repr = noise._to_abstract_repr()
     assert ("detuning_sigma" in repr) == (detuning_sigma != 0.0)
+    assert ("detuning_hf_psd" in repr) == (
+        detuning_hf_psd != () and detuning_hf_freqs != ()
+    )
+    assert ("detuning_hf_freqs" in repr) == (
+        detuning_hf_psd != () and detuning_hf_freqs != ()
+    )
```
HGSilveri: This is no longer valid under the new schema, you'll have to update the serialization and deserialization logic.
Also, a test where you deserialize would be useful

pulser-core/pulser/noise_model.py
kbidzhiev: done

pulser-core/pulser/noise_model.py
kbidzhiev: done

pulser-core/pulser/noise_model.py
kbidzhiev: done

pulser-core/pulser/noise_model.py
kbidzhiev: done

pulser-core/pulser/noise_model.py
kbidzhiev: done

tests/test_abstract_repr.py
kbidzhiev: DONE

kbidzhiev:
@HGSilveri done

pulser-core/pulser/noise_model.py
a-corni: [This](https://pulser.readthedocs.io/en/latest/noise_model.html) is the way it renders, did you want to make two lists ? Should we try to have the formula appearing as latex form ? image

HGSilveri: There will be fixes to this PR in #922, perhaps add this there so we don't forgot to address it too

PR #918, opened 2025-08-22
Clarify DMM tutorial

HGSilveri:
Following user feedback that the sign of the local detuning term was not clear.


a-corni (approved):
Thanks for tackling these ! Much clearer this way :)

PR #919, opened 2025-08-22
Test coverage with Numpy 2 as well

### LINKED ISSUE #845: Migrate test coverage requirement to numpy v2
With the migration to numpy v2 (see #837), we only get 100 % test coverage on tests when using numpy v1. Due to a difference in random number generation in v2 some tests fail because the random number generator doesn't evaluate to the same noise values even though the seem seed is set. The chosen workaround is that only 100 % test coverage is required for the the combination 'with-torch'/'numpy-v1'. 

I propose to change the expected data in the tests to match random number generation in numpy v2, so that we can migrate the test coverage requirement to the combination 'with-torch'/'numpy-v2'. Overall I think it's cleaner to have the most important test combination aligned with the newest version of dependencies.

HGSilveri:
Closes #845 .

### DIFF
--- a/tests/pulser_simulation/test_simulation.py
+++ b/tests/pulser_simulation/test_simulation.py
@@ -20,9 +20,14 @@ import numpy as np
 import pytest
 import qutip
 
+import pulser.noise_model  # for monkeypatch in test_detuning_noise
 from pulser import Pulse, Register, Sequence
 from pulser.devices import AnalogDevice, DigitalAnalogDevice, MockDevice
-from pulser.noise_model import _LEGACY_DEFAULTS, NoiseModel
+from pulser.noise_model import (
+    _LEGACY_DEFAULTS,
+    NoiseModel,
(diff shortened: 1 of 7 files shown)


tests/pulser_simulation/test_simulation.py
```diff
@@ -833,9 +833,9 @@ def test_noise_with_zero_epsilons(seq, matrices):
     sim2 = QutipEmulator.from_sequence(
         seq,
         sampling_rate=0.01,
-        config=SimConfig(
+        noise_model=SimConfig(
             noise=("SPAM"), eta=0.0, epsilon=0.0, epsilon_prime=0.0
-        ),
+        ).to_noise_model(),
```
a-corni, line 843: why aren't these shouting errors more often ?

HGSilveri: I think I either got rid of them or explicitly ignored the warning when I did the deprecation

PR #894, opened 2025-06-25
Fix evaluation times rounding error in QutipBackendV2

HGSilveri:
There was a sporadic rounding error issue when converting evaluation times to the "legacy evaluation times" expected by `QutipEmulator`.

PR #920, opened 2025-08-27
Antecipate fixes from 1.6 release

HGSilveri:
These were non-critical fixes planned for release with Pulser 1.6 but since that is taking a while, we'll release them ahead of time.

- Properly deserialize complex values in Results (#892)
- Handle when Results.atom_order entries are not strings (#891)
- Fix evaluation times rounding error in QutipBackendV2 (#894)


a-corni (approved):
LGTM 🚀

PR #926, opened 2025-09-10
optional sequence

a-quelle-pasqal:


pulser-core/pulser/backend/remote.py
```diff
@@ -314,8 +314,8 @@ class RemoteBackend(Backend):
 
     def __init__(
         self,
-        sequence: Sequence,
         connection: RemoteConnection,
+        sequence: Sequence | None = None,
         mimic_qpu: bool = False,
     ) -> None:
```
a-corni, line 320: For backwards compatibility, I would be more enclined to set both sequence and connection as optional. As a refinement, if we want people to be able to provide only a connection and a mimic_qpu, maybe we could use [overload](https://typing.python.org/en/latest/spec/overload.html) 
```
@overload
def __init__(
       self,
       connection:RemoteConnection,
       mimic_qpu:bool = False,
)
@overload
def __init__(
        self, 
        sequence: Sequence | None = None, 
        connection: RemoteConnection | None = None, 
        mimic_qpu: bool=False,
) -> None
```

a-corni: My bad regarding overload. It's more complex than this:
```
@overload
def __init__(
       self,
       connection: RemoteConnection 
       sequence: Sequence | None,
       mimic_qpu:bool,
)
     pass
@overload
def __init__(
       self,
       connection: Sequence | None, 
       sequence: RemoteConnection,
       mimic_qpu:bool,
)
     connection, sequence = sequence, connection
     pass
@overload
def __init__(
        self, 
        connection: RemoteConnection | Sequence | None, 
        sequence: RemoteConnection | Sequence | None = None, 
        mimic_qpu: bool=False,
) -> None

a-corni: @MatthieuMoreau0 @HGSilveri

MatthieuMoreau0: Yeah for backwards compatibility you need to keep the order consistent, so as far as I know we don't have a choice but to set connection to kwarg, and raise an error in the init function if it's unset by the user

a-quelle-pasqal: @MatthieuMoreau0 I think for kwargs you have to specify them as `connection=something` etc., so in that case you could just as well do 
`def __init__(self, *, sequence = None, connection)`
but this does have the problem that it breaks original code of the form
`QPUBackend(sequence,connection)` since you'd have to change it to
`QPUBackend(sequence=sequence, connection=connection)`

PR #930, opened 2025-09-15
Isolate EmulationConfig validation in EmulatorBackend.validate_config()

HGSilveri:
This permits the validation of an `EmulationConfig` through the target backend class alone (ie. without needing to instantiate it). It should make it easier to perform an early check on a config when only the target backend class is known.


MatthieuMoreau0:
> notifying @MatthieuMoreau0 there will be deprecation calls in the tests of pulser_pasqal

Thanks for the ping! I looked at the code changes, I am not sure I understand what you mean, sorry

PR #931, opened 2025-09-16
Fix bad qubit masking in interaction matrix

a-quelle-pasqal:


tests/test_hamiltonian_data.py
```diff
@@ -447,19 +450,29 @@ def test_noisy_interaction_matrix():
...
         ham.noisy_interaction_matrix._array,
-        np.zeros_like(ham._interaction_matrix),
+        np.array(
+            [
+                [0.0000, 0.0000, 0.0000, 0.0000],
+                [0.0000, 0.0000, 26.4198, 0.0000],
+                [0.0000, 26.4198, 0.0000, 0.0000],
+                [0.0000, 0.0000, 0.0000, 0.0000],
+            ]
+        ),
```
HGSilveri, line 473: nit: Asserting that the bad atoms are the first and fourth and maybe comparing with the (noiseless) interaction matrix would be good additions imo

HGSilveri (approved):

PR #932, opened 2025-09-17
Stop detaching when converting AbstractArray to float

### LINKED ISSUE #921: Consider not detaching on `AbstractArray.__float__()`
In PR #914 , we started explicitly detaching a `torch.Tensor` before casting to `float` so that we avoid an `UserWarning` that `torch` started raising. 
However, I would argue that this warnings has its place, so I suggest we revert the change in `AbstractArray.__float__()`.
In fact, the only place we cast Tensors to float in our codebase is in `_cast_check()` and that is being handled separately, so the tests will still pass without warnings.

WDYT, @a-corni ?

a-corni:
Closes #921


HGSilveri (approved):
LGTM, thanks @a-corni !

PR #933, opened 2025-09-17
Convert hf detuning noise units from Hz to rad/microsec

a-quelle-pasqal:
All other detuning related quantities are in `rad/microsec` so the choice of `Hz` was inconsistent. In addition, in the new units, all quantities in the power spectral density are of order unity, which makes things easier to sanity check.

### DIFF
--- a/pulser-core/pulser/noise_model.py
+++ b/pulser-core/pulser/noise_model.py
@@ -48,7 +48,7 @@ _NOISE_TYPE_PARAMS: dict[NoiseTypes, tuple[str, ...]] = {
     "doppler": ("temperature",),
     "register": ("trap_waist", "trap_depth"),
     "amplitude": ("laser_waist", "amp_sigma"),
-    "detuning": ("detuning_sigma", "detuning_hf_psd", "detuning_hf_freqs"),
+    "detuning": ("detuning_sigma", "detuning_hf_psd", "detuning_hf_omegas"),
     "SPAM": ("p_false_pos", "p_false_neg", "state_prep_error"),
     "dephasing": ("dephasing_rate", "hyperfine_dephasing_rate"),
     "relaxation": ("relaxation_rate",),
@@ -102,7 +102,7 @@ OPTIONAL_IN_ABSTR_REPR = (
     "trap_waist",
     "trap_depth",
     "detuning_hf_psd",
-    "detuning_hf_freqs",
+    "detuning_hf_omegas",
 )
 
 
@@ -154,7 +154,7 @@ class NoiseModel:
       ``detuning_sigma``;
       (2) time-dependent high-frequency fluctuations, defined by the
       power spectral density ``detuning_hf_psd`` over the relevant
-      ``detuning_hf_freqs`` frequencies support.
+      ``detuning_hf_omegas`` frequencies support.
       δ_hf(t) = Σ_k sqrt(2 * Δf_k * psd_k) * cos(2π(f_k * t + φ_k))
       where φ_k ~ U[0, 1) (uniform random phase),
       Δf_k = freqs[k+1] - freqs[k].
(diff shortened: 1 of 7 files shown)


pulser-core/pulser/noise_model.py
```diff
@@ -154,7 +154,7 @@ class NoiseModel:
       ``detuning_sigma``;
       (2) time-dependent high-frequency fluctuations, defined by the
       power spectral density ``detuning_hf_psd`` over the relevant
-      ``detuning_hf_freqs`` frequencies support.
+      ``detuning_hf_omegas`` frequencies support.
       δ_hf(t) = Σ_k sqrt(2 * Δf_k * psd_k) * cos(2π(f_k * t + φ_k))
       where φ_k ~ U[0, 1) (uniform random phase),
       Δf_k = freqs[k+1] - freqs[k].
```
kbidzhiev: Formulas can be translated to latex style. For instance above - `**register**` uses it

pulser-core/pulser/noise_model.py
```diff
@@ -154,10 +154,11 @@ class NoiseModel:
...
       power spectral density ``detuning_hf_psd`` over the relevant
-      ``detuning_hf_freqs`` frequencies support.
-      δ_hf(t) = Σ_k sqrt(2 * Δf_k * psd_k) * cos(2π(f_k * t + φ_k))
-      where φ_k ~ U[0, 1) (uniform random phase),
-      Δf_k = freqs[k+1] - freqs[k].
+      ``detuning_hf_omegas`` frequencies support.
+      :math:`\delta_{hf}(t) = \Sigma_k \sqrt(2 * \Delta f_k * psd_k)
+      * \cos(2\pi(f_k * t + \phi_k))`
+      where :math:`\phi_k \backsim U[0, 1)` (uniform random phase),
+      and :math:`\Delta f_k = freqs[k+1] - freqs[k]`.
```
kbidzhiev: previous name was `hf_freq` and thats why `f_k` and `freqs[k]`. maybe rename them into omegas ? and properly adjust the formulas ? 
and why \Sigma ? why not \sum ?

murogrande: the documentation is printed as 
Screenshot from 2025-09-18 11-21-18

here in power spectral density I will add PSD in parenthesis and in the equation I will write `psd_k` as `PSD_k`. I have the feeling that PSD is more common name

pulser-core/pulser/noise_model.py
```diff
@@ -192,16 +193,16 @@ class NoiseModel:
             Defaults to 0.
         trap_depth: The potential energy well depth that confines the atoms
             (in µK). Defaults to None.
-        detuning_hf_psd: Power Spectral Density(PSD) is 1D tuple (in Hz²/Hz)
-            provided together with `detuning_hf_freqs` define high frequency
+        detuning_hf_psd: Power Spectral Density(PSD) is 1D tuple (in rad/µs)
```
murogrande: here as documentation,  you have PSD, not psd. Probably, also in the variable I will add PSD in capitals in order to be consistent.

pulser-core/pulser/noise_model.py
```diff
@@ -154,10 +154,11 @@ class NoiseModel:
       ``detuning_sigma``;
       (2) time-dependent high-frequency fluctuations, defined by the
```
kbidzhiev, line 159: Would it be possible to put (1) and (2) in new line `\\`-wise

murogrande: I think flake8 will complain. I tried last time and I could not do it, right. Ping me if you find the way how to do it.

pulser-core/pulser/noise_model.py
```diff
@@ -153,11 +153,12 @@ class NoiseModel:
...
-      power spectral density ``detuning_hf_psd`` over the relevant
-      ``detuning_hf_freqs`` frequencies support.
-      δ_hf(t) = Σ_k sqrt(2 * Δf_k * psd_k) * cos(2π(f_k * t + φ_k))
-      where φ_k ~ U[0, 1) (uniform random phase),
-      Δf_k = freqs[k+1] - freqs[k].
+      power spectral density (PSD) ``detuning_hf_psd`` over the relevant
+      ``detuning_hf_omegas`` angular frequency support.
+      :math:`\delta_{hf}(t) = \sum_k \sqrt(2*\Delta \omega_k*\mathrm{PSD}_k)
+      * \cos(2\pi(\omega_k * t + \phi_k))`
+      where :math:`\phi_k \backsim U[0, 1)` (uniform random phase),
```
kbidzhiev: remove 2pi from cos
also you can update the uniform to be form 0 to 2pi `U[0, 2pi)`

pulser-core/pulser/noise_model.py
kbidzhiev: Nice ! if you can do the same for the "register" above that would be great. it also has (1) and (2)

pulser-core/pulser/noise_model.py
kbidzhiev: instead 
\sqrt(2 * \Delta f_k * psd_k)
try
\sqrt{2 * \Delta f_k * psd_k}


I think  \sqrt applies only for the "("

kbidzhiev (approved):

a-corni (approved):
LGTM

PR #934, opened 2025-09-18
Implement disable_doppler

### LINKED ISSUE #923: Add option to disable doppler noise in the NoiseModel
With the introduction of `"register"` noise in #915 , the `temperature` parameter will be shared by `"register"` and `"doppler"` noise. Given the implicit activation of the noise types in `NoiseModel`, we need to ensure that individual noise types can still be activated. There are three cases:
- Only "doppler" noise -> Ok, define only `temperature` as it won't activate `"register"` noise
- "doppler" + "register" -> Ok, define `trap_waist`, `trap_depth` and `temperature`
- **Only "register"** -> NOK, as "register" noise requires `temperature`, which in turn activates `"doppler"` noise

As such, I propose we add a new `NoiseModel.disable_doppler: bool` parameter, which would allow us to define `temperature` without activating `"doppler"` noise.

a-corni:
- disable_doppler is added to the NoiseModel attributes. It defaults to False. There is no need to modify the schema of the NoiseModel because it doesn't appear in its abstract representation and it is deduced from the noise types and the value of the temperature when deserializing a NoiseModel.
- It is possible to define disable_doppler, while no doppler noise was going to be implemented. However, if you deserialize an abstract representation obtained from the serialization of a NoiseModel with an unnecessary disable_doppler, disable_doppler will be set to False in the newly built NoiseModel.
- "doppler" is deleted from the noise_types if disable_doppler is set to True. The associated temperature and the number of runs are still entered in the abstract representation, even if they are no longer needed (usecase: QPU exposes temperature with by default disable_doppler = True -> user can still fetch temperature, and rebuild a NoiseModel where he replaces disable_doppler by True).

Closes #923

### DIFF
--- a/pulser-core/pulser/json/abstract_repr/deserializer.py
+++ b/pulser-core/pulser/json/abstract_repr/deserializer.py
@@ -449,6 +449,7 @@ def _deserialize_noise_model(noise_model_obj: dict[str, Any]) -> NoiseModel:
         noise_model_obj["temperature"] > 0 and "doppler" not in noise_types
     )
     relevant_params = pulser.NoiseModel._find_relevant_params(
+        # doppler parameters are relevant even if doppler is disabled
         noise_types + (["doppler"] if disable_doppler else []),
         noise_model_obj["state_prep_error"],
         noise_model_obj["amp_sigma"],
--- a/pulser-core/pulser/json/abstract_repr/deserializer.py
+++ b/pulser-core/pulser/json/abstract_repr/deserializer.py
@@ -445,8 +445,11 @@ def _deserialize_noise_model(noise_model_obj: dict[str, Any]) -> NoiseModel:
 
     noise_types = noise_model_obj.pop("noise_types")
     with_leakage = "leakage" in noise_types
+    disable_doppler = (
+        noise_model_obj["temperature"] > 0 and "doppler" not in noise_types
+    )
     relevant_params = pulser.NoiseModel._find_relevant_params(
-        noise_types,
+        noise_types + (["doppler"] if disable_doppler else []),
         noise_model_obj["state_prep_error"],
         noise_model_obj["amp_sigma"],
         noise_model_obj["laser_waist"],
@@ -475,6 +478,7 @@ def _deserialize_noise_model(noise_model_obj: dict[str, Any]) -> NoiseModel:
         eff_noise_rates=tuple(eff_noise_rates),
         eff_noise_opers=tuple(eff_noise_opers),
         with_leakage=with_leakage,
+        disable_doppler=disable_doppler,
         detuning_hf_psd=tuple(detuning_hf_psd),
         detuning_hf_freqs=tuple(detuning_hf_freqs),
         detuning_sigma=detuning_sigma,
--- a/pulser-core/pulser/json/abstract_repr/deserializer.py
+++ b/pulser-core/pulser/json/abstract_repr/deserializer.py
@@ -464,14 +464,14 @@ def _deserialize_noise_model(noise_model_obj: dict[str, Any]) -> NoiseModel:
     relevant_params -= {"detuning_sigma"}  # Handled separately, optional arg
 
     detuning_hf_psd = []
-    detuning_hf_freqs = []
+    detuning_hf_omegas = []
     if "detuning_hf" in noise_model_obj:
         for psd, freq in noise_model_obj.pop("detuning_hf"):
             detuning_hf_psd.append(psd)
-            detuning_hf_freqs.append(freq)
+            detuning_hf_omegas.append(freq)
     relevant_params -= {  # Handled separately
         "detuning_hf_psd",
-        "detuning_hf_freqs",
+        "detuning_hf_omegas",
     }
 
     noise_model = pulser.NoiseModel(
@@ -481,7 +481,7 @@ def _deserialize_noise_model(noise_model_obj: dict[str, Any]) -> NoiseModel:
         with_leakage=with_leakage,
         disable_doppler=disable_doppler,
         detuning_hf_psd=tuple(detuning_hf_psd),
-        detuning_hf_freqs=tuple(detuning_hf_freqs),
+        detuning_hf_omegas=tuple(detuning_hf_omegas),
         detuning_sigma=detuning_sigma,
     )
     assert set(noise_model.noise_types) == set(noise_types)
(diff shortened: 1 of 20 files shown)


pulser-core/pulser/json/abstract_repr/deserializer.py
```diff
@@ -445,8 +445,11 @@ def _deserialize_noise_model(noise_model_obj: dict[str, Any]) -> NoiseModel:
 
     noise_types = noise_model_obj.pop("noise_types")
     with_leakage = "leakage" in noise_types
+    disable_doppler = (
+        noise_model_obj["temperature"] > 0 and "doppler" not in noise_types
+    )
     relevant_params = pulser.NoiseModel._find_relevant_params(
-        noise_types,
+        noise_types + (["doppler"] if disable_doppler else []),
```
HGSilveri, line 453: The fact that you are adding doppler when it is set to be disabled may be confusing. I would suggest adding a clarifying comment and/or explicitly adding the `temperature` to `relevant_params` instead

pulser-core/pulser/noise_model.py
```diff
@@ -322,8 +328,10 @@ def to_tuple(obj: tuple) -> tuple:
         ]
         for param_, val_ in param_vals.items():
             object.__setattr__(self, param_, val_)
-            if param_ not in relevant_params and (
-                val_ if param_ != "samples_per_run" else val_ != 1
+            if (
+                param_ != "disable_doppler"  # disable_doppler is not used
```
HGSilveri: I find this comment confusing, how about
```suggestion
                # disable_doppler is not a noise parameter
                param_ != "disable_doppler"
```

tests/test_abstract_repr.py
```diff
@@ -264,10 +286,23 @@ def test_noise_model(noise_model: NoiseModel):
...
     assert (
-        NoiseModel.from_abstract_repr(json.dumps(ser_noise_model_obj))
-        == re_noise_model
+        dict_legacy.pop("runs") == noise_model.runs or _LEGACY_DEFAULTS["runs"]
+    )
+    dict_re_noise_model.pop("runs")
+    assert dict_legacy.pop("disable_doppler") == (
+        temp > 0 and "doppler" not in re_noise_model.noise_types
     )
+    dict_re_noise_model.pop("disable_doppler")
```
HGSilveri, line 318: Maybe add a comment here saying why you are checking these values manually

pulser-core/pulser/json/abstract_repr/deserializer.py
a-corni: I will add a comment :) Actually, it's more than adding the temperature, you also have to add the runs and I guess at the moment the samples_per_run... Anyway, that was making lots of code replication, I figured adding doppler to noise type would be better.

pulser-core/pulser/noise_model.py
a-corni: Sounds good !

tests/test_abstract_repr.py
a-corni: Yes I agree it's a mess... It's backward compatible in the sense that the noise types still agree, but now the values of temperature and runs are stored whereas they weren't previously

pulser-core/pulser/noise_model.py
```diff
@@ -543,7 +552,10 @@ def _to_abstract_repr(self) -> dict[str, Any]:
             ):
                 continue
             all_fields[f.name] = value
+        # Remove fields that can be deduced from the noise_types
```
```suggestion
        # Removed fields that can be deduced from the noise_types
```

pulser-core/pulser/noise_model.py
```diff
@@ -218,6 +219,8 @@ class NoiseModel:
             Defaults to 0.
         with_leakage: Whether or not to include an error state in the
             computations (default to False).
+        disable_doppler: Whether or not to disable the doppler noise,
+            even if the temperature is defined.
```
HGSilveri: Would be nice to mention why this can be useful
```suggestion
            even if the temperature is defined. In this way,
            'register' noise (which requires 'temperature') can be
            activated on its own (i.e. without 'doppler').
```

tests/test_abstract_repr.py
```diff
@@ -264,10 +287,32 @@ def test_noise_model(noise_model: NoiseModel):
...
+        temp = dict_legacy.pop("temperature")
+        assert temp == (
+            noise_model.temperature or _LEGACY_DEFAULTS["temperature"]
+        )
+        dict_re_noise_model.pop("temperature")
+        assert (
+            dict_legacy.pop("runs") == noise_model.runs
+            or _LEGACY_DEFAULTS["runs"]
+        )
+        dict_re_noise_model.pop("runs")
```
HGSilveri, line 330: The purpose of this part of the test was to make sure old noise models are still deserialized correctly, so allowing `disable_noise_model` to be `True` in the legacy does not make much sense imo... 
Instead of this complicated thing, I'd consider splitting this "legacy" section into a separate test where you purposefully specify only the parameters and noise types that were there before v0.20 (ie those in `_LEGACY_DEFAULTS`)

a-corni: I am okay with making a new test function and testing without disable_noise_model. However, I still need this complicated tests because the default behaviour has changed: the default temperature was >0, but there was no doppler in the noise types, so it's interpreted as a NoiseModel with a temperature but with disable_doppler=True (it works because the default number of runs is greater than 1)

tests/test_abstract_repr.py
```diff
@@ -242,17 +247,79 @@ def test_register(reg: Register | Register3D):
...
+            runs=1,
+        ),
+        NoiseModel(
+            temperature=50.0,
+            trap_depth=150.0,
+            trap_waist=1.0,
+            runs=1,
+            samples_per_run=1,
+        ),
+        NoiseModel(temperature=50.0, trap_depth=150.0, trap_waist=1.0, runs=1),
```
HGSilveri: nit: These parameters were all introduced post v0.20, I'd say they don't need to be tested

tests/test_abstract_repr.py
HGSilveri: Damn, ok... :disappointed:

tests/test_abstract_repr.py
a-corni: Solved 👍

PR #938, opened 2025-09-24
Fix HamiltonianData when qubit ids are ints

a-quelle-pasqal:


a-corni (commented):
It looks good to me, but you could eventually make a test with a Register with integer qubit IDs such that we keep track of this possible error ?

PR #940, opened 2025-09-25
Fix creation of Hamiltonian from sequence for 3d register

HaroldErbin:
Calling `QutipEmulator.from_sequence()` with a 3d register fails because `Hamiltonian.__init__` assumes that the register is 2d. This is a regression in v1.6 as it used to work in v1.5 and below.

Here is a MWE:
```python
import dataclasses
from pulser import Register3D, Sequence, Pulse
from pulser.devices import AnalogDevice
from pulser_simulation import QutipEmulator

device = device = dataclasses.replace(AnalogDevice, dimensions=3)
register = Register3D.cubic(side=2, spacing=6, prefix="q")

seq = Sequence(register, device)
seq.declare_channel("ising", "rydberg_global")
channel = seq.declared_channels["ising"]
cst_pulse = Pulse.ConstantPulse(duration=500, amplitude=2., detuning=1., phase=0.)
seq.add(cst_pulse, "ising")

computation = QutipEmulator.from_sequence(seq)
```
The error message in v1.6 is:
```
  File "pulser/pulser-simulation/pulser_simulation/simulation.py", line 917, in from_sequence
    return cls(
           ^^^^
  File "pulser/pulser-simulation/pulser_simulation/simulation.py", line 172, in __init__
    self._hamiltonian = Hamiltonian(
                        ^^^^^^^^^^^^
  File "pulser/pulser-simulation/pulser_simulation/hamiltonian.py", line 55, in __init__
    register = Register.from_coordinates(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "pulser/pulser-core/pulser/register/base_register.py", line 205, in from_coordinates
    return cls(qubits, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "pulser/pulser-core/pulser/register/register.py", line 63, in __init__
    raise ValueError(
ValueError: All coordinates must be specified as vectors of size 2.
```

### DIFF
--- a/pulser-simulation/pulser_simulation/hamiltonian.py
+++ b/pulser-simulation/pulser_simulation/hamiltonian.py
@@ -26,7 +26,7 @@ from pulser._hamiltonian_data import HamiltonianData
 from pulser.channels.base_channel import States
 from pulser.devices._device_datacls import BaseDevice
 from pulser.noise_model import NoiseModel
-from pulser.register import QubitId, Register
+from pulser.register import QubitId, Register, Register3D
 from pulser.sampler.samples import SequenceSamples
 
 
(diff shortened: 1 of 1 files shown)


HaroldErbin:
Sounds good to me. In fact, I was wondering why coordinates instead of the register were passed, since the latter contains more information. My first attempt used `BaseRegister`, but I was not sure how this would fit with other parts of the code, so I kept the changes as minimal as possible.

PR #941, opened 2025-09-26
Fix that Hamiltonian does not work with Register3D

a-quelle-pasqal:
You need to instantiate a `Register` or `Register3D` depending on the length of the position vector. `Hamiltonian` always tried `Register`. Fixed the issue by making `Hamiltonian` take a `BaseRegister` rather than a `qdict`.


HGSilveri (approved):

PR #942, opened 2025-09-26
Fix detection of shot to shot noise

a-quelle-pasqal:
`QutipEmulator` only does repeated runs if it detects the presence of shot to shot noise, since in its absence, the results of the emulator are deterministically all the same over many runs. The new noise types of `register`, `detuning`, but due to an earlier regression also `doppler` were missing from this list. A version of this check is used in several places, and I've done a small refactor to reuse the same code for all of them. In the unit tests, since mostly `amp_sigma` is present in the noise model, this bug didn't trigger.

The MR is a draft because I still need to add a test that fails without the above code changes.

### DIFF
--- a/tests/pulser_simulation/test_simulation.py
+++ b/tests/pulser_simulation/test_simulation.py
@@ -28,6 +28,7 @@ from pulser.register.register_layout import RegisterLayout
 from pulser.sampler import sampler
 from pulser.waveforms import BlackmanWaveform, ConstantWaveform, RampWaveform
 from pulser_simulation import QutipEmulator, SimConfig
+from pulser_simulation.simresults import NoisyResults
 
 
 @pytest.fixture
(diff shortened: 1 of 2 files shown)


HGSilveri:
@a-corni We could still use the test that's added here though...

a-corni:
> @a-corni We could still use the test that's added here though...

Yes, and adding shot to shot noise if register noise is defined ;) My bad

tests/pulser_simulation/test_simulation.py
```diff
@@ -2214,3 +2215,53 @@ def test_detuning_hf_noise(monkeypatch):
...
+    ),
+)
+def test_noisy_runs(noise):
+    count = 0
+    old_run = QutipEmulator._run_solver
+
+    def mock_run_solver(self, progress_bar: bool = False, **options):
+        nonlocal count
+        count += 1
+        return old_run(self, progress_bar, **options)
```
HGSilveri: nit: Is there any reason why you chose to make a mock manually instead of using `mocker`?
From what I can tell, it could be as simple as
```suggestion
def test_noisy_runs(noise, mocker):
    mock_run_solver = mocker.patch(
        "pulser_simulation.simulation.QutipEmulator._run_solver",
        wraps=QutipEmulator._run_solver
      )
```
Then later you could access the number of calls via `mock_run_solver.call_count`

a-corni (approved):
LGTM !

PR #943, opened 2025-09-28
Fix detection of doppler noise

a-quelle-pasqal:
see #942


HGSilveri:
Not sure what's going on with the CI here... We might need to push something for it to be triggered

PR #899, opened 2025-07-03
Deprecate SimConfig

HGSilveri:
Closes #884.

PR #944, opened 2025-09-29
Release v1.6.0

HGSilveri:
**Main changes:**
- Add detuning fluctuations to the noise model (#880)
- Deprecate SimConfig (#899) 
- Add shortcut in Results for the final bitstrings and state (#905)
- Add method to truncate a Sequence at a specific time (#898)
- Deprecate QutipBackend (#907) 
- Support Results aggregation (#913) 
- Support time-dependent high-frequency detuning noise (#917) 
- Support noise from atom motion in the Register (#915) 
- Deprecate and replace misleading method names (#929) 
- Isolate the data needed to generate an Hamiltonian in HamiltonianData (#916)


a-corni (approved):
LGTM !

PR #947, opened 2025-09-30
Fix CI tkinter

a-corni:
Our CI tests involving drawing with matplotlib fail again, even though we are using the AGG backend, which is still the one advised to solve the issue (see [issue on matplotlib](https://github.com/matplotlib/matplotlib/issues/28957) and [matplotlib doc](https://matplotlib.org/stable/users/explain/figure/backends.html)). 
Trying some stuff to fix it, enable temporarily the CI jobs to run on this PR (to be reverted)

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,33 +1,10 @@
 name: Release to PyPI
 
 on:
-  release:
-    types: [released]
   pull_request:
 
 jobs:
-  deploy:
-    runs-on: ubuntu-latest
-    environment: release
-    permissions:
-      # IMPORTANT: this permission is mandatory for trusted publishing
-      id-token: write
-    steps:
-      - name: Check out Pulser
-        uses: actions/checkout@v4
-        with:
-          ref: ${{ github.ref }}
-      - name: Set up Python
-        uses: actions/setup-python@v5
-        with:
-          python-version: 3.9
-      - name: Install Python dependencies
-        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -5,13 +5,14 @@ on:
...
       - develop
+  pull_request:
 
 jobs:
   full-tests:
     runs-on: ${{ matrix.os }}
     env:
       # Set Matplotlib backend to fix flaky execution on Windows
-      MPLBACKEND: agg
+      MPLBACKEND: Agg
```
HGSilveri, line 14: Tbh I think this was already working before. We just need to replicate it in the `publish.yml` file, as that is what is running the tests when we release

.github/workflows/test.yml
```diff
@@ -5,13 +5,14 @@ on:
     branches:
       - master
       - develop
+  pull_request:
```
HGSilveri: Don't forget to remove this

.github/workflows/publish.yml
```diff
@@ -63,6 +63,9 @@ jobs:
 
   check-release:
     needs: deploy
```
HGSilveri: I wonder if we can trigger this without triggering `depoly:` to check if it's working

a-corni:
@HGSilveri at least all the windows tests are okay now, but the issue with the macos runners doesn't seem related... Do you agree ?

PR #951, opened 2025-10-17
Add support for metadata in the serialized Sequence

MatthieuMoreau0:


pulser-core/pulser/json/abstract_repr/schemas/sequence-schema.json
```diff
@@ -1068,6 +1068,10 @@
               "description": "Variables and expressions that can be used in expressions or parametrized values.",
               "type": "object"
             },
+            "metadata": {
```
MatthieuMoreau0: The json schema needs a proper update to make metadata optional, this was a simple POC

pulser-core/pulser/sequence/sequence.py
```diff
@@ -76,6 +77,15 @@
 
 PROTOCOLS = Literal["min-delay", "no-delay", "wait-for-all"]
 
+_metadata = contextvars.ContextVar("metadata", default={"library": "pulser", "pulser_version": pulser.__version__})
```
HGSilveri: There is already a dedicated `pulser_version` field, I wonder if we want to replicate the info in the metadata

HGSilveri (commented):
Thanks for the POC @MatthieuMoreau0 !
I think we should also try to impose some structure to the `metadata` field via the schema, rather than allowing an arbitrary object

pulser-core/pulser/sequence/sequence.py
MatthieuMoreau0: No point in duplicating that information imo. Libraries could add their version to the context however

HGSilveri:
@MatthieuMoreau0 I took the liberty of continuing this. I think what I did is self-explanatory from the commit messages.

Not sure yet if we should release this as an hotfix or not. I'm keeping the branch pointing to `develop` for now and I'll (re)mark it as a Draft to avoid it being merged by accident

a-corni (approved):
LGTM ! Good job :)

HGSilveri:
Let's release this together with #961 . I'll skip the version bump on this one

PR #954, opened 2025-10-27
Drop Python 3.9, introduce Python 3.14 + mypy==1.18.2

HGSilveri:
- Python 3.9 reaches its EOL at the end of October 2025
- Python 3.14 has been officially released, so we should start properly testing against it
- Since CI `mypy` checks now happen with Numpy v2 and type hints needed to be refreshed, also takes the opportunity to upgrade `mypy` to 1.18.2


a-corni (approved):
Thanks for tackling this ! LGTM !

PR #961, opened 2025-11-13
Fix the inclusion of noise in QutipBackendV2

### LINKED ISSUE #960: Fix the inclusion of noise in `QutipBackendV2`
There seems to be at least more than one noise-related bug in the current `QutipBackendV2` implementation:

## `leakage` noise is not handled properly

The following script (courtesy of @murogrande ) showcases how adding leakage noise to `QutipBackendV2` fails. 

[script_leakage.py](https://github.com/user-attachments/files/23501595/script_leakage.py)

Looking at the code, I could pinpoint two points of failure (I did not dig deeper so there could be more):

  1. On [line 161](https://github.com/pasqal-io/Pulser/blob/hs/pprint-ch-specs/pulser-simulation/pulser_simulation/qutip_backend.py#L161), the `eigenstates` are extracted from the `SequenceSamples` instead of the `_HamiltonianData` so they don't include the garbage state

  2. On [line 188](https://github.com/pasqal-io/Pulser/blob/hs/pprint-ch-specs/pulser-simulation/pulser_simulation/qutip_backend.py#L188), `noiseless=True` disregards the garbage state, so the operator dimension won't match the state dimensions

## New stochastic noises can be silently ignored

On [lines 167-174 ](https://github.com/pasqal-io/Pulser/blob/c046c947a73c4961c823de174661bfe01d8d521b/pulser-simulation/pulser_simulation/qutip_backend.py#L167-L174), the decision on whether to do a single emulation run or to average over multiple runs is **based on the stochastic noises available at the time the backend was defined**. So, notably, a `NoiseModel` with only `detuning` or `register` noise will incorrectly return the results of a single run.

In this case, although we could just harcode the new noise types, I think it would be much better to add a method like `NoiseModel.requires_averaging()` (there is probably a better name for this) that we could just poll inside `QutipBackendV2`.

a-quelle-pasqal:
This solves the two bugs mentioned in #960 
The bugs are solved, and for each there is a test that fails on the old code, and passes now.


pulser-core/pulser/backend/state.py
```diff
@@ -186,6 +186,8 @@ def infer_one_state(self) -> Eigenstate:
             return "1"
         if eigenstates == {"r", "g"}:
             return "r"
+        if eigenstates == {"r", "g", "x"}:
+            return "r"
```
HGSilveri: I think you could even do 

```
eigenstates = set(self.eigenstates) - {"x"}
```
and cover all the leakage cases

pulser-simulation/pulser_simulation/qutip_backend.py
```diff
@@ -185,7 +181,7 @@ def run(self) -> Results:
                 state = QutipState(qutip_res.state, eigenstates=eigenstates)
                 ham: QutipOperator = QutipOperator(
                     self._sim_obj.get_hamiltonian(
-                        t * res.total_duration, noiseless=True
+                        t * res.total_duration, noiseless=False
```
HGSilveri: nit: Ah yes, since we don't have any shot to shot noise, having `noiseseless=False` here will only increase the Hamiltonian dimension in case leakage is present, which is what we want. 
I would add a small comment stating that, otherwise at first glance it looks like we are calculating the observables off of a noisy Hamiltonian (which would be wrong IIUC)

tests/pulser_simulation/test_qutip_backend_v2.py
```diff
@@ -241,3 +242,108 @@ def test_qutip_backend_v2_eval_times_rounding():
...
+    # pulser convention of basis
+    basisx = np.array([0.0, 0.0, 1.0]).reshape(3, 1)
+    basisg = np.array([0.0, 1.0, 0.0]).reshape(3, 1)
+    basisr = np.array([1.0, 0.0, 0.0]).reshape(3, 1)
+
+    rate = 0.5
+    eff_rate = [rate, rate]
+    eff_ops = [basisx @ basisr.T, basisx @ basisg.T]  # |x><r| and |x><g|
+
+    # eff_ops must be a torch.tensor if you want to work with emu_mps
```
HGSilveri: nit: This comment was left behind (also in the next test)

tests/pulser_simulation/test_qutip_backend_v2.py
```diff
@@ -241,3 +242,108 @@ def test_qutip_backend_v2_eval_times_rounding():
...
+    noise_model = pulser.NoiseModel(
+        trap_depth=1.0,
+        trap_waist=1.0,
+        temperature=50.0,
+        disable_doppler=True,
+        detuning_sigma=5.0,
+        runs=10,
+    )
```
HGSilveri, line 339: nit: Can you add
```suggestion
    )
    assert set(noise_model.noise_types) == {"register", "detuning"} 
```

HGSilveri (commented):
~~You'll also want to bump the version to 1.6.3 in the PR~~
Just saw that's done

pulser-core/pulser/backend/state.py
a-quelle-pasqal: yes, very nice.

pulser-simulation/pulser_simulation/qutip_backend.py
a-quelle-pasqal: Your comment here made me realize that the same issue is present in `else` path where there is shot-to-shot noise. Adding an `amp_sigma` to the noise model in `test_leakage` will cause the test to fail again. I'm going back to the drawing board.

tests/pulser_simulation/test_qutip_backend_v2.py
```diff
@@ -241,3 +242,108 @@ def test_qutip_backend_v2_eval_times_rounding():
...
+    both_leaked = QutipOperator(
+        qutip.tensor(
+            [qutip.Qobj(basisx @ basisx.T), qutip.Qobj(basisx @ basisx.T)]
+        ),
+        eigenstates,
+    )
+
+    p_no_leaked = np.zeros((3, 3))
+    p_no_leaked[0, 0] = 1.0
+    p_no_leaked[1, 1] = 1.0
```
murogrande, line 295: You don't need to add another 1.0 in [2,2] position?

tests/pulser_simulation/test_qutip_backend_v2.py
a-quelle-pasqal: done

tests/pulser_simulation/test_qutip_backend_v2.py
a-quelle-pasqal: no, because you want to measure the projector onto the non-leakage states.

tests/pulser_simulation/test_qutip_backend_v2.py
a-quelle-pasqal: done

tests/pulser_simulation/test_qutip_backend_v2.py
```diff
@@ -210,7 +211,7 @@ def get_noise_model(samples_per_run: int) -> pulser.NoiseModel:
         [qutip.tensor([qutip.basis(2, 0).proj(), qutip.qeye(2)])]
     )[0][indices]
 
-    assert np.max(np.abs(occupation - occupation_old_api)) < 0.03
+    assert np.max(np.abs(occupation - occupation_old_api)) < 0.0341
```
a-quelle-pasqal: The changed implementation of `QutipEmulator._noiseless_hamiltonian(self)` means it cannot be a cached property any longer. That means each time it is computed, noise trajectories are created, which change the RNG. There is no way to preserve this test.

pulser-simulation/pulser_simulation/qutip_backend.py
```diff
@@ -158,22 +161,16 @@ def run(self) -> Results:
             atom_order=tuple(self._sequence.qubit_info),
             total_duration=self._sim_obj.total_duration_ns,
         )
-        eigenstates = self._sim_obj.samples_obj.eigenbasis
+        eigenstates = self._sim_obj._hamiltonian.data.eigenbasis
         options: dict = {}
+        noiseless_hamiltonian = self._sim_obj._noiseless_hamiltonian
```
a-quelle-pasqal: It's no longer a cached property, so compute it once. Skipping this step makes the tests VERY slow.

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -527,9 +534,7 @@ def build_operator(self, operations: Union[list, tuple]) -> qutip.Qobj:
         """
         return self._hamiltonian.build_operator(operations)
 
-    def get_hamiltonian(
-        self, time: float, noiseless: bool = False
-    ) -> qutip.Qobj:
+    def get_hamiltonian(self, time: float) -> qutip.Qobj:
```
HGSilveri: Hum, do we need this breaking change?

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -201,14 +200,22 @@ def __init__(
                 self._meas_basis = self.basis_name.replace("_with_error", "")
         self.set_initial_state("all-ground")
 
-    @cached_property
+    @property  # the basis can change when set_config is called
```
HGSilveri: Can't we keep it cached and just clear the cache on every `set_config` call?

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -201,14 +201,25 @@ def __init__(
                 self._meas_basis = self.basis_name.replace("_with_error", "")
         self.set_initial_state("all-ground")
 
-    @cached_property
-    def _noiseless_hamiltonian(self) -> Hamiltonian:
+    @lru_cache(maxsize=2)
+    def _noiseless_hamiltonian(self, leakage: bool) -> Hamiltonian:
```
HGSilveri: To really minimize disruption (and potential confusion, since leakage being included in something named "noiseless" is counterintuitive), how about we do this?

```suggestion
    @property
    def _noiseless_hamiltonian(self, leakage) -> Hamiltonian:
        return self._get_noiseless_hamiltonian(leakage=False)
        
    @lru_cache(maxsize=2)
    def _get_noiseless_hamiltonian(self, leakage: bool) -> Hamiltonian:
```

You can then use this in `QutipBackendV2`

```
self._get_noiseless_hamiltonian(leakage=noise_model.with_leakage)
```

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -559,7 +570,9 @@ def get_hamiltonian(
             )
 
         if noiseless:
-            return self._noiseless_hamiltonian._hamiltonian(time / 1000)
+            return self._noiseless_hamiltonian(
+                self.noise_model.with_leakage
+            )._hamiltonian(time / 1000)
```
HGSilveri: I would not change this behavior in the hotfix

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -201,14 +201,29 @@ def __init__(
...
         self.set_initial_state("all-ground")
 
-    @cached_property
+    @property
     def _noiseless_hamiltonian(self) -> Hamiltonian:
+        return self._get_noiseless_hamiltonian(False)
+
+    @lru_cache(maxsize=2)
+    def _get_noiseless_hamiltonian(self, leakage: bool) -> Hamiltonian:
+        """The garbage state IS INCLUDED if leakage is in the noise model."""
```
HGSilveri: nit: Let's update this docstring

HGSilveri (commented):
As a hotfix, I think this is fine. Just one last change and I'll approve

HGSilveri (approved):
Nice, LGTM!

PR #962, opened 2025-11-24
Fix bag union aggregation with non-Counter dicts

a-quelle-pasqal:
The bag union aggregator only works when all objects are `Counter`s. Which should ideally be the case for all returned bitstring counts, but in `SampledResult` this is currently not always the case.

I am changing both the bag union aggregator to also allow dict, and the `SamplesResult` class to store bitstrings as a `Counter` even when fed a `dict`
I also added tests for both changes.

### DIFF
--- a/tests/test_math.py
+++ b/tests/test_math.py
@@ -211,10 +211,11 @@ class TestAbstractArray:
         # pow
         np.testing.assert_array_equal(arr**5.0, val**5.0)
 
+        # rounding errors here
         np.testing.assert_array_almost_equal(
             abs(arr) ** values, abs(val_np) ** values
-        )  # rounding errors here
-        np.testing.assert_array_equal(2.0**arr, 2.0**val)
+        )
+        np.testing.assert_array_almost_equal(2.0**arr, 2.0**val)
 
         # mod
         np.testing.assert_array_equal(arr % 5.0, val % 5.0)


pulser-core/pulser/backend/results.py
```diff
@@ -103,12 +104,10 @@ def __getattr__(self, name: str) -> list[Any]:
         raise AttributeError(f"{name!r} is not in the results.")
 
     @property
-    def final_bitstrings(self) -> dict[str, int]:
+    def final_bitstrings(self) -> Counter[str]:
         """Returns the bitstrings at the end of the sequence, if available."""
         try:
-            return cast(
-                typing.Dict[str, int], self.get_result("bitstrings", time=1.0)
-            )
```
a-corni, line 111: This would change the API a bit... What about discarding the changes here and only keep the one in aggregators ?

a-quelle-pasqal: It's  mostly returning `Counter` already. I'm fine with keeping this `dict` since `Counter` is a `dict`, but I do think we need to keep the change in `SampledResults`

pulser-core/pulser/backend/aggregators.py
```diff
@@ -91,10 +91,10 @@ def _mean_aggregator(
 
 
 def _bag_union_aggregator(
-    values: list[collections.Counter],
-) -> collections.Counter:
+    values: list[Counter],
+) -> Counter:
     """Join a list of Counter objects."""
-    return sum(values, start=collections.Counter())
+    return sum((Counter(x) for x in values), start=Counter())
```
HGSilveri: nit: The "pythonic" way would be (feel free to disregard this, of course)
```suggestion
    return sum(map(Counter, values), start=Counter())
```

pulser-core/pulser/backend/results.py
```diff
@@ -103,12 +104,10 @@ def __getattr__(self, name: str) -> list[Any]:
...
+            return cast(Counter[str], self.get_result("bitstrings", time=1.0))
```
HGSilveri: I think to do this, we should rather do
```suggestion
            return Counter(cast(typing.Dict[str, int], self.get_result("bitstrings", time=1.0)))
```
Seems a bit beyond the scope of this hotfix and I don't think we really need it. If the user wants a counter, they can always cast it themselves, right?

PR #963, opened 2025-11-25
Fix timing mismatches when switching a Sequence's device strictly

HGSilveri:
Upon realizing that `Sequence.with_new_device()` passed even when a difference in `phase_jump_time` resulted in a different sequence after the switch, I figured it would be better to check that the time slots are identical when we can (i.e. when the sequence is not parametrized) so that we don't miss anything. 
This new check makes some of the existing checks redundant (e.g. checking the samples when there is an active EOM) but I figured getting rid of those was beyond the scope of an hotfix. I'll open an issue to do that.

- Check timing parameters only on parametrized sequences (including `phase_jump_time`)
- Check time slots match on non-parametrized sequences 

Also, a strict switch with an SLM mask that used to work before will no longer work now because the value of detuning chosen for the `DMM` will generally be different. I think it makes sense to fail on this.


tests/test_sequence.py
```diff
@@ -1365,6 +1388,149 @@ def test_switch_device_eom(
     (seq.build(**build_kwargs) if build_kwargs else seq).draw()
 
 
+def test_switch_device_strict_time_slots_check(reg):
```
HGSilveri, line 1393: Full disclosure: this was written using Claude Code. 
I was quite happy with the result so I kept it with minimal modifications

a-corni: 😆

.gitignore
```diff
@@ -18,4 +18,5 @@ dist/
 env*
 *.egg-info/
 __venv__/
-venv
\ No newline at end of file
+venv
+CLAUDE.md
```
a-corni, line 22: And here is our good friend Claude indeed

pulser-core/pulser/sequence/helpers/_switch_device.py
```diff
@@ -344,6 +352,20 @@ def build_sequence_from_matching(
...
+                    new_ch_name = old_to_new_ch_name.setdefault(
+                        old_ch_name, old_ch_name
+                    )
+                    if (
+                        new_seq._schedule[new_ch_name].slots
+                        != seq._schedule[old_ch_name].slots
+                    ):
+                        raise SwitchDeviceError(
+                            "Changing the device produced a sequence with "
+                            f"different samples for channel {old_ch_name!r}."
```
a-corni, line 367: Note: The error message can be improved here by outputting the parameters that differ in some channels. We could consider raising a warning instead of an error when the sequence is non parametrized and some timing parameters differ, saying they might induce non-matching sequence)

tests/test_sequence.py
```diff
@@ -733,9 +733,13 @@ def test_switch_device_down(
...
-    with helpers.raises_all(
-        [ValueError, SwitchDeviceError],
-        match="No match for channel raman_1 with the same clock_period.",
+    with (
+        helpers.raises_all(
+            [ValueError, SwitchDeviceError],
+            match="No match for channel raman_1 with the same clock_period.",
+        )
+        if parametrized
+        else contextlib.nullcontext()
```
a-corni, line 744: We could evenutally add a comment `# can switch since sequence contains only channel declaration` ?

tests/test_sequence.py
```diff
@@ -1365,6 +1388,149 @@ def test_switch_device_eom(
     (seq.build(**build_kwargs) if build_kwargs else seq).draw()
 
 
+def test_switch_device_strict_time_slots_check(reg):
+    """Test that strict device switch verifies time slots are preserved.
+
+    This test specifically checks the logic at lines 355-368 in
```
a-corni: nit: Is it a good idea to mention the lines ?

tests/test_sequence.py
a-corni: Damn that test is good 😅 I think it can replace test_switch_device_down (that only uses declared channels without any pulses added to it).

tests/test_sequence.py
```diff
@@ -1092,7 +1115,7 @@ def test_switch_device_up(
                         nested_s_loc[:100]
                         == (-10.0 if trap_id in mod_trap_ids else 0)
                     )
-                else:
```
a-corni, line 1095: I guess this is related to:
> Also, a strict switch with an SLM mask that used to work before will no longer work now because the value of detuning chosen for the `DMM` will generally be different. I think it makes sense to fail on this.

HGSilveri: Yes, that's right

pulser-core/pulser/sequence/helpers/_switch_device.py
HGSilveri: I'll add this to #964

PR #974, opened 2025-12-05
Fix output amplitude modulation

HGSilveri:
# Issue

There was a widespread confusion between *amplitude* and *intensity* output modulation in the code. Namely, 
- What was being shown in Pulser plots and provided in samples was the expected output modulation of _intensity_, **not of amplitude** (as it should have been).
- The conversion between `mod_bandwidth` and `rise_time` was valid for the intensity rise time, not the amplitude's

# Changes
- A new `modulation.py` submodule is introduced to isolate the modulation conversion functions
- Direct import of the `MODBW_TO_TR` is deprecated
- The output modulation formula's are fixed in the least disruptive way possible, **as is outlined below**
- UT for modulation was completed (previously skipping the specific modulation in the EOM buffer)

### Intensity rise time

The intensity rise definition still holds: *the time it takes the **intensity** signal (e.g. as seen on a photodiode) take to go from 10% to 90% of its top value in a step response.*

**However**, the `rise_time` parameter in Pulser will now be the **amplitude** rise time, which corresponds to $\sqrt{2}\cdot T_{r,I}$.

### Modulation bandwidth formula

To avoid changing the `mod_bandwidth` value in all the datasheets/devices, we’ll **keep its formula** and instead **change its definition** (see next section)**.** This means that **Pulser will use an unorthodox definition of “modulation bandwidth”**, which will be properly documented.

So, the `mod_bandwidth` value in Pulser will still correspond to

$$
F_{mod} =\frac{\sqrt{8\ln 2}}{\pi}\frac{v_a}{d}
$$

### Bandwidth × Intensity Rise Time Product

Stays the same, at

$$
F_{mod} \times T_{r,I} = 0.48
$$

This means all existing calculations of `mod_bandwidth` from the intensity rise time (i.e. the usual experimental procedure) will remain valid.

## What changes

These changes are prompted by:

- The introduction of a proper distinction between amplitude vs. intensity rise time
- A correction of the amplitude modulation transfer function

|  | Before | **After** |
| --- | --- | --- |
| Amplitude modulation transfer function | $MTF(f) = \exp\left(-\frac{\pi^2 f^2 d^2}{8 v_a^2}\right) = \exp\left(-\frac{f^2}{F_{mod}^2}\cdot\ln2\right)$ | $H_A(f) = \exp\left(-\frac{\pi^2 f^2 d^2}{4 v_a^2}\right) =\exp\left(-\frac{f^2}{(F_{mod}/\sqrt{2}) ^2}\cdot \ln2\right)$ |
| Amplitude rise time (aka `rise_time` in Pulser), $T_{r,A}$ | Same as $T_{r,I}$ | $\sqrt{2}\cdot T_{r,I}$ |
| `mod_bandwidth` definition | “Frequency component experiencing -3dB (50%) reduction.” (it was not explicit that the reduction was in *power* but the use of -3dB implied it). | “the frequency component at which the amplitude is attenuated by 75%.” |


pulser-core/pulser/channels/__init__.py
```diff
@@ -17,6 +17,9 @@
 from pulser.channels.dmm import DMM
 from pulser.channels.eom import BaseEOM, RydbergEOM, RydbergBeam
 
+# Expose the modulation submodule
+from pulser.channels import modulation as modulation
+
```
a-corni, line 22: Is it necessary to have it exposed ?

pulser-core/pulser/channels/modulation.py
```diff
@@ -0,0 +1,140 @@
...
+        NotImplementedError: If mod_bandwidth exceeds the maximum allowed
+        value.
+    """
+    if mod_bandwidth <= 0.0:
+        raise ValueError(
+            "'mod_bandwidth' must be greater than zero, not"
+            f" {mod_bandwidth}."
+        )
+    if mod_bandwidth > (
+        max_bw := calculate_mod_bandwidth_from_amplitude_rise_time(1)
```
a-corni, line 136: I have the feeling `calculate_mod_bandwidth_from_amplitude_rise_time(1)` is a constant that should be stored outside this function, not called each time...

pulser-core/pulser/channels/modulation.py
```diff
@@ -0,0 +1,140 @@
...
+    Returns:
+        The amplitude rise time (in ns), defined as the time taken to go
+        from 10% to 90% output amplitude in response to a step change.
+
+    Note:
+        The calculation is based on the modulation bandwidth at which
+        there is a 75% attenuation in amplitude. The relationship between
+        amplitude and intensity rise times is: t_amp = sqrt(2) * t_int.
+    """
+    return int(round(_mod_bw_rise_time_conversion(mod_bandwidth) * np.sqrt(2)))
```
a-corni, line 116: It nitpicking over 1ns, but this `round` means that we will sometime take a smaller rise time than asked by the user. Don't we want to be more conservative and have:
```suggestion
    return int(_mod_bw_rise_time_conversion(mod_bandwidth) * np.sqrt(2))+1
```

a-corni: As far as I have understood we used to take the opposite (shorter rise time) 🤔 `return int(MODBW_TO_TR / self.mod_bandwidth * 1e3)`

tests/test_channels.py
```diff
@@ -275,7 +285,7 @@ def test_modulation_errors():
 @pytest.mark.parametrize(
     "channel, tr, eom, side_buffer_len",
     [
-        (_raman_local, _raman_local.rise_time, False, 45),
+        (_raman_local, _raman_local.rise_time, False, 85),
```
a-corni, line 288: Why is it 85 and not 90 ? (I thought the rise time would now be 2*former rise times)

tests/test_pulse.py
```diff
@@ -104,12 +105,17 @@ def eom_channel():
 
 
 def test_fall_time(eom_channel):
-    assert eom_channel.eom_config.rise_time == 20
-    assert eom_channel.rise_time == 120
+    eom_rise_time = calculate_amplitude_rise_time(
+        eom_channel.eom_config.mod_bandwidth
+    )
+    aom_rise_time = calculate_amplitude_rise_time(eom_channel.mod_bandwidth)
```
a-corni, line 111: These two have changed, right ? So all our switch_device operations with EOM will fail, no ?
- see below how these two have failed

tests/test_sequence.py
```diff
@@ -1809,8 +1809,8 @@ def test_str(reg, device, mod_device, det_map):
     msg_ch0 = (
         "Channel: ch0\nt: 0 | Initial targets: q0 | Phase Reference: 0.0 "
         + "\nt: 0->500 | Pulse(Amp=2 rad/µs, Detuning=-10 rad/µs, Phase=0) "
-        + "| Targets: q0\nt: 500->800 | Delay \nt: 800->800 | Target: q7 | "
-        + "Phase Reference: 0.0"
+        + "| Targets: q0\nt: 500->800 | Delay \nt: 800->840 | Delay "
+        + "\nt: 840->840 | Target: q7 | Phase Reference: 0.0"
```
a-corni, line 1813: Idem on switch_device (although not important since that's a retarget operation)

tests/test_sequence_sampler.py
```diff
@@ -286,26 +309,45 @@ def test_eom_modulation(mod_device, disable_eom):
...
+            want = eom_output + aom_output
 
         # Check that modulation through sample() = sample() + modulation
         got = getattr(mod_samples.channel_samples["ch0"], qty).as_array()
         alt_got = getattr(
             input_samples.modulate(chan, full_duration), qty
         ).as_array()
         np.testing.assert_array_equal(got, alt_got)
-        np.testing.assert_allclose(want.as_array(), got, atol=1e-10)
+        np.testing.assert_allclose(want.as_array(), got, atol=1e-15)
```
a-corni, line 350: You needed more precision ? 😅

HGSilveri: Now that the test was properly done, I could afford it :sunglasses:

pulser-core/pulser/channels/__init__.py
HGSilveri: Yes, I want HW enginners to use `calculate_mod_bandwidth_from_intensity_rise_time()` for example

pulser-core/pulser/channels/modulation.py
HGSilveri: Given the deprecation I had to do with `MODBW_TO_TR`, I'm wary of constants... The limitation here is really one of precision, since we can't handle rise times below 1ns. Also, this function just does some basic float arithmetic, it's basically free

pulser-core/pulser/channels/modulation.py
HGSilveri: Yes, we used to just cast to `int` and it was "fine". I figured the fairest thing to do would to round to the nearest int, I don't think it makes a difference honestly

tests/test_channels.py
HGSilveri: These side buffers are trimmed in `calc_modulation_buffer()` to not waste time, so it was never the full `2*rise_time`. I admit here I just took copied the new value it gave me
Side note: The new rise time is sqrt(2)*former rise times

tests/test_pulse.py
HGSilveri: I don't think they will fail... Each time you construct the sequence with the current version, so the timings will always be consistent within that version. Am I missing something?

pulser-core/pulser/channels/modulation.py
a-corni: By casting to the next integer, you could get rid of this test (it would be at least 1). I am not against it, it's only used in post_init of channels.

tests/test_channels.py
a-corni: Yes but that's ever more puzzling, sqrt(2)*45 doesn't give 85, more like 64...

HGSilveri: I don't think the effect is linear in this case. 

**Before**:
image

**After**:
image

pulser-core/pulser/channels/modulation.py
HGSilveri: I think it's still worth mentioning that there is a maximum `mod_bandwidth` we support, I would prefer keeping this

tests/test_channels.py
a-corni: Okay I got the point !

tests/test_pulse.py
a-corni: No you are not :)

a-corni (approved):
Thanks for your replies ! LGTM !

PR #980, opened 2026-01-06
Fix mypy changing default behaviour to no_implicit_optional=True

a-corni:
Pipeline https://github.com/pasqal-io/Pulser/actions/runs/20742117803/job/59550890998?pr=972 is failing because of issues with mypy. It seems it changed its default behaviour to no_implicit_optional=True (to follow PEP 484).

### DIFF
--- a/pulser-core/requirements.txt
+++ b/pulser-core/requirements.txt
@@ -3,6 +3,6 @@ referencing  # A dependency of jsonschema we use. We let jsonschema set the vers
 matplotlib >= 3.10.5, < 4
 packaging   # This is already required by matplotlib but we use it too
 # Numpy 1.20 introduces type hints, 1.24.0 breaks matplotlib < 3.6.1
-numpy >= 1.20, != 1.24.0, < 2.4.0
-numpy >= 2, < 2.4.0; python_version >= '3.13'
+numpy >= 1.20, != 1.24.0
+numpy >= 2; python_version >= '3.13'
 scipy < 2


pulser-core/pulser/devices/_device_datacls.py
```diff
@@ -153,7 +153,7 @@ class BaseDevice(ABC):
 
     def __post_init__(self) -> None:
         def type_check(
-            param: str, type_: type, value_override: Any = None
+            param: str, type_: type, value_override: Optional[Any] = None
```
HGSilveri: Why are you using `Optional` instead of `| None`?

a-corni: I followed the package advised by mypy, https://github.com/hauntsaninja/no_implicit_optional, that generated Optional instead of |None. I can go for | None if you prefer

HGSilveri: It's equivalent and it is the preferred syntax afaik, so I'd say to use `| None`

pulser-core/pulser/register/_reg_drawer.py
HGSilveri, line 1: The diff on this file is all messed up, can you help me identify what you changed exactly?

pulser-core/pulser/register/register.py
```diff
@@ -424,6 +424,31 @@ def rotated(self, degrees: float) -> Register:
...
+            if trap_id not in filled_traps_ids
+        ]
+        return cast(
+            Register,
+            self.layout.define_register(
+                *empty_traps_ids,
+                qubit_ids=[str(trap_id) for trap_id in empty_traps_ids],
+            ),
+        )
+
```
HGSilveri: I think you started out from the wrong branch so you some extra changes here

HGSilveri (commented):
Thanks for taking this on @a-corni

pulser-core/pulser/register/_reg_drawer.py
a-corni: Ah yes it was stupid of mine but I replaced a if...else ... by a if: ... return ... to save one indentation.
Then I only added the assertion the fig.add_subplot(projection="3d", ...) creates an Axes3d

pulser-core/pulser/register/register.py
a-corni: Indeed, now I recall I did the commit from #978 in my local develop ^^' changing the direction of the target branch to #978's branch, I will merge them one after another

HGSilveri: Somehow there's a bunch of different unrelated changes now :upside_down_face:

HGSilveri: I guess it's because you merged `develop` into this branch

PR #981, opened 2026-01-06
Replace NoiseModel.runs with EmulationConfig.n_trajectories

### LINKED ISSUE #956: Move `runs` out of `NoiseModel`
- [ ] Introduce `EmulationConfig.n_trajectories` to substitute `NoiseModel.runs`
  -  Make `EmulationConfig.n_trajectories: int | None = None` to force the user to pick a value
- [ ] Deprecate `NoiseModel.runs`
  - Allow `runs` to be left as `None` under any circumstance
  - If `runs` is defined, raise a `DeprecationWarning` and point to `EmulationConfig.n_trajectories`
  - If both `runs` and `n_trajectories` are defined, raise an error and ask for just `n_trajectories`

HGSilveri:
- [x] Introduces `EmulationConfig.n_trajectories`, which defaults to `NoiseModel.runs` when not defined and then to 40 if that is not defined either.
- [x] Deprecates `NoiseModel.runs` and replaces it with `EmulationConfig.n_trajectories`
- [x] Adds dedicated UTs

Closes #956 .

### DIFF
--- a/pulser-core/pulser/backend/config.py
+++ b/pulser-core/pulser/backend/config.py
@@ -134,10 +134,14 @@ class EmulationConfig(BackendConfig, Generic[StateType]):
             `prefer_device_noise_model=True`.
         n_trajectories: The number of trajectories to average over when the
             emulation includes stochastic noise or is using a Monte Carlo
-            solver. If left undefined, tries to use the (now deprecated) value
-            of `NoiseModel.runs`. If that is also not defined,
-            **defaults to 40** trajectories. Whenever averaging over multiple
-            trajectories is not needed, this value is ignored.
+            solver. When left undefined:
+
+            (1) If 'prefer_device_noise_model=False', tries to use the (now
+            deprecated) value of `NoiseModel.runs`. If that is also not
+            defined, defaults to 1 trajectory.
+
+            (2) If 'prefer_device_noise_model=True', **defaults to 40**
+            trajectories.
 
     Note:
         Additional parameters may be provided. It is up to the emulation
@@ -274,13 +278,14 @@ class EmulationConfig(BackendConfig, Generic[StateType]):
                 "can't be simultaneously defined. Please favour using only"
                 " `EmulationConfig.n_trajectories`."
             )
-        n_trajectories = (
-            n_trajectories
-            if n_trajectories is not None
-            # NoiseModel.runs is type and value checked so we don't have to
-            # explicitly check for None
-            else (noise_model.runs or DEFAULT_N_TRAJECTORIES)
-        )
+
+        if n_trajectories is None:
+            if prefer_device_noise_model:
+                n_trajectories = DEFAULT_N_TRAJECTORIES
+            else:
+                n_trajectories = (
+                    noise_model.runs if noise_model.runs is not None else 1
+                )
 
         if n_trajectories < 1 or n_trajectories != int(n_trajectories):
             raise ValueError(
--- a/pulser-simulation/pulser_simulation/simulation.py
+++ b/pulser-simulation/pulser_simulation/simulation.py
@@ -829,7 +829,7 @@ class QutipEmulator:
             return self._run_solver(
                 self._current_hamiltonian,
                 progress_bar,
-                mcsolve_ntraj=self.n_trajectories,
+                mcsolve_ntraj=self.n_trajectories or 1,
                 **options,
             )
 
--- a/pulser-core/pulser/backend/abc.py
+++ b/pulser-core/pulser/backend/abc.py
@@ -14,6 +14,7 @@
 """Base class for the backend interface."""
 from __future__ import annotations
 
+import warnings
 from abc import ABC, abstractmethod
 from collections.abc import Sequence
 from typing import ClassVar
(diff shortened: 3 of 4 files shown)


pulser-core/pulser/backend/config.py
```diff
@@ -255,6 +264,30 @@ def __init__(
...
+            n_trajectories is not None
+            and noise_model.runs is not None
+            and n_trajectories != noise_model.runs
+        ):
+            raise ValueError(
+                "`EmulationConfig.n_trajectories` and `NoiseModel.runs` "
+                "can't be simultaneously defined. Please favour using only"
+                " `EmulationConfig.n_trajectories`."
+            )
+        n_trajectories = (
```
a-quelle-pasqal: This implementation relies on the current functioning of pulser-simulation, which currently ignores `n_trajectories` if no noise is present for example. I'd prefer if the value of `n_trajectories` that ends up in the config is appropriate given the context. It means that applications built on this config can take the value of `n_trajectories` at face value, making them easier to maintain.

I this particular case, I'd suggest putting `n_trajectories = 1` if no noise is present. In this scenario. when we implement monte carlo in emu-sv, I will make custom logic in `SVConfig` to set `n_trajectories=1` if only lindblad noise is present and it uses the lindblad solver. I don't think that particular change is needed in `QutipConfig` since the situation is already handled in `QutipEmulator` (which is not the ideal location but why change it).

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -783,10 +823,7 @@ def run(
         """
         self._validate_options(options)
 
-        if (
-            "SPAM" not in self.noise_model.noise_types
-            or self.noise_model.state_prep_error == 0
-        ) and not has_shot_to_shot_except_spam(self.noise_model):
+        if not _has_stochastic_noise(self.noise_model):
```
a-quelle-pasqal, line 827: this is good. You can make the same change in `QutipBackendV2.run` see `qutip_backend.py:172-175`

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -783,10 +823,7 @@ def run(
...
             # A single run is needed, regardless of self.config.runs
             return self._run_solver(
```
a-quelle-pasqal, line 829: In `run_solver`, you can now set `ntraj` for `mcsolve` to `n_trajectories`, see line `731`. Your docstring implies that you intended to do this, but simply forgot.

tests/pulser_simulation/test_simulation.py
```diff
@@ -1860,12 +1875,12 @@ def test_effective_size_disjoint(channel_type):
         seq,
         sampling_rate=0.01,
         noise_model=NoiseModel(
-            runs=15,
             samples_per_run=5,
             state_prep_error=0.4,
             p_false_pos=0.01,
             p_false_neg=0.05,
         ),
+        n_trajectories=1,
```
a-quelle-pasqal: This used to have `runs=15`, what's happening here?

tests/test_abstract_repr.py
```diff
@@ -210,60 +210,66 @@ def test_register(reg: Register | Register3D):
...
+        lambda: NoiseModel(
             temperature=50.0,
             trap_depth=150.0,
             trap_waist=1.0,
             runs=1,
             samples_per_run=1,
             disable_doppler=True,
         ),
-        NoiseModel(temperature=50.0, trap_depth=150.0, trap_waist=1.0, runs=1),
+        lambda: NoiseModel(
```
a-quelle-pasqal, line 266: I'm assuming you tried without making all the noise models a factory function? But did you also try switching the order of the annotations in that case? maybe there's no way to make it better than this, but I do know order of annotations matters in how code executes.

pulser-core/pulser/backend/config.py
```diff
@@ -131,6 +132,12 @@ class EmulationConfig(BackendConfig, Generic[StateType]):
         noise_model: An optional noise model to emulate the sequence with.
             Ignored if the sequence's device has default noise model and
             `prefer_device_noise_model=True`.
+        n_trajectories: The number of trajectories to average over when the
+            emulation includes stochastic noise or is using a Monte Carlo
+            solver. If left undefined, tries to use the (now deprecated) value
+            of `NoiseModel.runs`. If that is also not defined,
+            **defaults to 40** trajectories. Whenever averaging over multiple
```
a-corni: Why 40 ? :) I would have suggested 1

pulser-simulation/pulser_simulation/simulation.py
HGSilveri: It's not that simple though, is it? I should only directly set `ntraj = n_trajectories` if there is no stochastic noise

I will do this

tests/pulser_simulation/test_simulation.py
HGSilveri: Just a mistake, thanks for spotting it

tests/test_abstract_repr.py
HGSilveri: Yes, I tried to use `pytestmark` to ignore the deprecation warnings for the whole file but for some reason the deprecation warning was never caught correctly, so I had to resort to this

pulser-core/pulser/backend/config.py
HGSilveri: In thinking about this, I realized I overlooked something: when `prefer_device_noise_model: int = True`, I don't even know what the noise model will be so I can't access that info to decide what to put in `n_trajectories`; in fact, I shouldn't even be using `noise_model.runs` in this case (like I currently am). 

In my mind, one advantage of having `n_trajectories` in `EmulationConfig` is that we could have it be independent of the contents of `NoiseModel` (and thus avoid logic like we used to have for `runs` in `NoiseModel`). What you are suggesting goes in the opposite direction though

a-quelle-pasqal: I agree that this is a good long-term goal. In that case though, I would always default to `1`, since it gives `n_trajectories` more of an opt-in flavour, have it do more runs by explicitly asking for it, and if you do, well it's up to you to ensure it actually makes sense to do it.

PR #982, opened 2026-01-08
Add validation of expected kwargs in QutipConfig

a-quelle-pasqal:
The `n_trajectories` kwarg in `EmulationConfig` was not added to the list of expected kwargs. This slipped past the tests because validation of expected kwargs was turned off in `QutipConfig`, where it should be turned on. I turned on the validation, and added `n_trajectories` to the list.

### DIFF
--- a/pulser-simulation/pulser_simulation/qutip_config.py
+++ b/pulser-simulation/pulser_simulation/qutip_config.py
@@ -72,7 +72,7 @@ class QutipConfig(EmulationConfig[QutipState]):
         EmulationConfig: The base configuration class for an EmulatorBackend.
     """
 
-    # Whether to warn if unexpected kwargs are received
+    # Whether to error if unexpected kwargs are received
     _enforce_expected_kwargs: ClassVar[bool] = True
     sampling_rate: float
     _state_type = QutipState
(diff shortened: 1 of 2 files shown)


pulser-core/pulser/backend/observable.py
```diff
@@ -152,8 +152,6 @@ def __call__(
             state: The current state.
             hamiltonian: The Hamiltonian at this time.
             result: The Results object to store the result in.
-            time_tol: Tolerance below which two time values are considered
```
a-quelle-pasqal, line 155: this arg is not actually in the function anymore. As part of a separate discussion:

What to do here?

HGSilveri: It should be deleted, I probably just forgot

pulser-simulation/pulser_simulation/qutip_config.py
```diff
@@ -72,6 +72,8 @@ class QutipConfig(EmulationConfig[QutipState]):
         EmulationConfig: The base configuration class for an EmulatorBackend.
     """
 
+    # Whether to warn if unexpected kwargs are received
```
HGSilveri: It actually doesn't warn anymore, it errors. Let's update the comment here and in `EmulationConfig`

PR #983, opened 2026-01-09
Use Results.aggregate in QutipBackendV2

### LINKED ISSUE #957: Make the emulator backends automatically aggregate the results over the `n_trajectories`
It would be nice if we could have all emulator backends aggregate the results for simulations over multiple trajectories. 

This would also greatly benefit from a uniform way to aggregate `StateResults`, which is currently not implemented.  

**Note**: Requires #956.

a-quelle-pasqal:
- implements `density_matrix_aggregator` for use with `StateResult` in `Results.aggregate`
- refactors `QutipBackendV2.run` to use `Results.aggregate` rather than the existing functionality.

Behavioral regressions:
- The number of bitstrings sampled is now `n_trajectories * n_samples`, before it was `n_samples`
- `EnergyVariance` is not aggregated, so it will be absent from the results if `n_trajectories > 1`. Note that this cannot be solved within the aggregation framework because the aggregator would need access the the mean of the energy, the presence of which cannot be guaranteed.

Closes #957

### DIFF
--- a/tests/pulser_simulation/test_qutip_backend_v2.py
+++ b/tests/pulser_simulation/test_qutip_backend_v2.py
@@ -15,7 +15,6 @@ from __future__ import annotations
 
 import dataclasses
 import math
-import warnings
 from unittest.mock import patch
 
 import numpy as np
@@ -398,7 +397,9 @@ def test_aggregation():
         n_trajectories=5,
         noise_model=pulser.NoiseModel(state_prep_error=1 / 3),
     )
-    with warnings.catch_warnings(record=True) as w:
+    with pytest.warns(
+        UserWarning, match="Skipping aggregation of `energy_variance`."
+    ):
         with patch(
             "pulser._hamiltonian_data.hamiltonian_data.np.random.uniform"
         ) as bad_atoms_mock:
@@ -437,5 +438,3 @@ def test_aggregation():
         "110": 1000,
     }
     assert "energy_variance" not in qutip_results.get_result_tags()
-    assert issubclass(w[0].category, UserWarning)
-    assert "Skipping aggregation of `energy_variance`." in str(w[0].message)


HGSilveri:
> * `EnergyVariance` is not aggregated, so it will be absent from the results if `n_trajectories > 1`. Note that this cannot be solved within the aggregation framework because the aggregator would need access the the mean of the energy, the presence of which cannot be guaranteed.

Can't we add `Energy` and `EnergySecondMoment` to the observables (or even replace `EnergyVariance` by them) when `EnergyVariance` is present and aggregation will occur, so that we can then calculate it off of them?

pulser-simulation/pulser_simulation/aggregators.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def density_matrix_aggregator(values: list[QutipState]) -> QutipState:
+    """Average states into a mixed state.
+
+    Argument:
+        values: The results to average.
+
+    Returns:
+        The mixed state containing each input state with probability 1/n.
```
HGSilveri, line 26: You're not really averaging here though, just summing the density matrices

pulser-simulation/pulser_simulation/qutip_backend.py
```diff
@@ -220,40 +233,37 @@ def run(self) -> Results:
...
+                        )
+                    for obs in self._config.observables:
+                        obs(
+                            config=self._config,
+                            t=float(t),
+                            state=state,
+                            hamiltonian=ham,
+                            result=res,
+                        )
+                results.append(res)
```
HGSilveri: It seems that, unlike before, you are not weighing the contribution of a given result based on the number of reps. The simplest way to do it would be to 
```suggestion
                results += [res] * reps
```

pulser-simulation/pulser_simulation/qutip_backend.py
```diff
@@ -220,40 +233,37 @@ def run(self) -> Results:
...
-                        config=self._config,
-                        t=float(t),
-                        state=state,
-                        hamiltonian=ham,
-                        result=res,
+                    ham = QutipOperator(
+                        self._sim_obj._get_noiseless_hamiltonian(
+                            self._config.noise_model.with_leakage
+                        )._hamiltonian(t * res.total_duration / 1000),
+                        eigenstates=eigenstates,
```
HGSilveri: Is this perhaps costly to create? Could be worth implementing a simple cache, just to avoid the redundant recreation

pulser-simulation/pulser_simulation/qutip_backend.py
```diff
@@ -220,40 +233,37 @@ def run(self) -> Results:
...
+                        obs(
+                            config=self._config,
+                            t=float(t),
+                            state=state,
+                            hamiltonian=ham,
+                            result=res,
+                        )
+                results.append(res)
+            if len(results) == 1:
+                # aggregation does not work on EnergyVariance for example.
```
HGSilveri: I'm missing why this comment is here... Even if `EnergyVariance` was the only observable in the config, it would still be computed for every noise trajectory, right?

pulser-simulation/pulser_simulation/qutip_backend.py
```diff
@@ -26,13 +26,21 @@
...
 from pulser_simulation.qutip_config import QutipConfig
 from pulser_simulation.qutip_op import QutipOperator
 from pulser_simulation.qutip_state import QutipState
 from pulser_simulation.simresults import CoherentResults, SimulationResults
 from pulser_simulation.simulation import QutipEmulator, _has_stochastic_noise
 
 
+def _get_state_tag(results: Results) -> str | None:
+    for tag in results.get_result_tags():
+        if tag.startswith("state"):
```
HGSilveri: Can we make this robust to an eventual change of `StateResult._base_tag`?
```suggestion
        if tag.startswith(StateResult()._base_tag):
```

tests/pulser_simulation/test_qutip_backend_v2.py
```diff
@@ -363,3 +364,40 @@ def test_register_detuning_detection():
...
+        "ryd",
+    )
+
+    occup = Occupation(evaluation_times=[1.0])
+
+    qutip_config = QutipConfig(
+        observables=(occup,),
+        n_trajectories=3,
+        noise_model=pulser.NoiseModel(state_prep_error=1 / 3),
+    )
```
HGSilveri, line 399: I think it's important to also explicitly test the density matrix aggregation and the behavior when `EnergyVariance` is among the observables

pulser-simulation/pulser_simulation/qutip_backend.py
a-quelle-pasqal: yes, but if you call `aggregate` even on a single result, it would not put `EnergyVariance` in there.

pulser-simulation/pulser_simulation/aggregators.py
a-quelle-pasqal: good catch, what disturbs me, is that there is no test that fails on this. The issue is only a normalization, so I'm guessing that qutip compensated for it in all observables, but I'll add a test that fails on this.

pulser-simulation/pulser_simulation/qutip_backend.py
a-quelle-pasqal: `self._sim_obj._get_noiseless_hamiltonian(self._config.noise_model.with_leakage)` is cached. I'm assuming calling `_hamiltonian(t * res.total_duration / 1000)` is fast, probably just some list indexing, because it'll be used by qutip inside the solvers. Wrapping in `QutipOperator` is quite cheap.

pulser-simulation/pulser_simulation/qutip_backend.py
a-quelle-pasqal: indeed, and I'll also add a test that catches this issue.

pulser-simulation/pulser_simulation/qutip_backend.py
HGSilveri: Shouldn't aggregate just not do anything when there is a single result? That's what I would expect at least

a-quelle-pasqal: that's a fair point, it's essentially moving those two lines to `aggregate`. I'll do that, and add a test

tests/pulser_simulation/test_qutip_backend_v2.py
a-quelle-pasqal: I added some asserts to the test

pulser-simulation/pulser_simulation/qutip_backend.py
a-quelle-pasqal: turns out it already existed, I just missed it because i was looking in the wrong place.

pulser-simulation/pulser_simulation/qutip_backend.py
a-quelle-pasqal: In the end, I decided to just create an entirely new `Results` for each rep. It was the easiest way to obtain separate bitstring sampling for each rep. The solution you outline works, but you're not fully leveraging the total number of bitstrings that you're doing.

tests/pulser_simulation/test_qutip_backend_v2.py
```diff
@@ -363,3 +371,67 @@ def test_register_detuning_detection():
...
+        qutip_results.occupation[-1], np.array([0.6, 0.6, 0.8]), atol=1e-4
+    )
+    assert qutip_results.final_bitstrings == {
+        "011": 2000,
+        "101": 2000,
+        "110": 1000,
+    }
+    assert "energy_variance" not in qutip_results.get_result_tags()
+    assert issubclass(w[0].category, UserWarning)
+    assert "Skipping aggregation of `energy_variance`." in str(w[0].message)
```
HGSilveri: nit: This is an unorthodox to catch a warning for testing, did you consider using `with pytest.warns()`?

HGSilveri (approved):
LGTM, I have only one nit

pulser-simulation/pulser_simulation/qutip_backend.py
HGSilveri: Ok, let's not optimize prematurely I guess

PR #984, opened 2026-01-09
Include the "pulser_version" in the serialized Sequence

HGSilveri:
This information was already included inside the `"device"` object, but not having it at the serialized sequence level was nonetheless an oversight.


a-corni (commented):
LGTM :) Nice catch !

a-corni (approved):
Approved !

PR #987, opened 2026-01-14
Allow noise_model=None in EmulationConfig

### LINKED ISSUE #968: Error with QutipConfig if noise_model is None
I have an issue when initializing `QutipConfig` because the test on the `noise_model` in `__init__` is not robust:
```python
if noise_params_conv is not None:
    noise_params, runs, with_modulation = noise_params_conv
    noise_model = NoiseModel(
        runs=runs, **noise_params
    )
else:
    noise_model = None
config = QutipConfig(noise_model=noise_model)
```
gives the error:
```
File "rydberg-spin-chains/rydberg_spin_chains/providers/pasqal/simulations.py", line 133, in create_computation
    config = QutipConfig(
             ^^^^^^^^^^^^
  File "rydberg-spin-chains/.venv/lib/python3.12/site-packages/pulser_simulation/qutip_config.py", line 90, in __init__
    ].samples_per_run not in [None, 1]:
      ^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'samples_per_run'
```

The reason comes from the test
```python
if "noise_model" in backend_options and backend_options[
    "noise_model"
].samples_per_run not in [None, 1]:
```
which is not robust. The first condition should be:
```python
if backend_options.get("noise_model", None) is not None:
```

There is a similar issue in
```
pulser/backend/config.py", line 253, in __init__
TypeError: 'noise_model' must be a NoiseModel instance, not <class 'NoneType'>.
```
One needs to use:
```python
if noise_model is not None and not isinstance(noise_model, NoiseModel):
```

Again another:
```python
rydberg-spin-chains/rydberg-spin-chains/.venv/lib/python3.12/site-packages/pulser_simulation/qutip_backend.py", line 187, in run
    self._config.noise_model.with_leakage
AttributeError: 'NoneType' object has no attribute 'with_leakage'
```

Obviously these are just suggestions, maybe there are shorter options. Another alternative is to check from scratch if `noise_model` is `None` and remove from the arguments.

HGSilveri:
Closes #968.

### DIFF
--- a/.github/workflows/pulser-setup/action.yml
+++ b/.github/workflows/pulser-setup/action.yml
@@ -41,7 +41,7 @@ runs:
       if: ${{ inputs.numpy == 'numpy-v1'}}
       shell: bash
       run: |
-        pip install numpy==1.26
+        pip install numpy==1.26.4
     - name: Install numpy v2
       if: ${{ inputs.numpy == 'numpy-v2'}}
       shell: bash


a-corni (approved):
LGTM!

PR #988, opened 2026-01-15
Track progress of run execution

### LINKED ISSUE #967: Improve verbosity of run() for noisy simulations
I would suggest improving the verbose output when running simulations for which the parameters in the Hamiltonian vary. In this case, there is a loop with `runs` iteration, and the verbose argument just prints that number of time the evolution:
```
...
Total run time:   3.29s
10.1%. Run time:   0.35s. Est. time left: 00:00:00:03
20.1%. Run time:   0.68s. Est. time left: 00:00:00:02
30.0%. Run time:   1.01s. Est. time left: 00:00:00:02
40.0%. Run time:   1.34s. Est. time left: 00:00:00:02
50.0%. Run time:   1.68s. Est. time left: 00:00:00:01
60.1%. Run time:   2.01s. Est. time left: 00:00:00:01
70.1%. Run time:   2.34s. Est. time left: 00:00:00:00
80.0%. Run time:   2.66s. Est. time left: 00:00:00:00
90.0%. Run time:   2.99s. Est. time left: 00:00:00:00
100.0%. Run time:   3.31s. Est. time left: 00:00:00:00
Total run time:   3.31s
10.1%. Run time:   0.34s. Est. time left: 00:00:00:03
20.1%. Run time:   0.68s. Est. time left: 00:00:00:02
30.0%. Run time:   1.01s. Est. time left: 00:00:00:02
40.0%. Run time:   1.34s. Est. time left: 00:00:00:02
50.0%. Run time:   1.67s. Est. time left: 00:00:00:01
60.1%. Run time:   2.00s. Est. time left: 00:00:00:01
70.1%. Run time:   2.33s. Est. time left: 00:00:00:00
80.0%. Run time:   2.66s. Est. time left: 00:00:00:00
90.0%. Run time:   2.98s. Est. time left: 00:00:00:00
100.0%. Run time:   3.29s. Est. time left: 00:00:00:00
...
```
It would be more informative to display
 ```
...
run 3/40
run 4/40
...
```
A full verbose mode could still display the information above, but with additional information on the iteration.

### PR DESCRIPTION
**Pull request by a-corni**
Fixes #967 
Prints the number of "run" ie qutip run we are performing as "Run .../number of runs". 
Can be discussed:
- I am printing "Run 1/1" if it's a coherent execution. It is speaking to me but is it for everyone ?
- "Run .../..." could be replaced by "Qutip run .../..." or "Qutip execution .../..."
- It is always being displayed for QutipBackend and QutipBackendV2. Is it expected behaviour for QutipBackendV2 ? Should we add a "show_progress" option to QutipConfig ? Btw, should we add a progress_bar option to QutipConfig to display the progress of the Qutip solver per trajectory ?

### CODE REVIEW

**Comment by HGSilveri** [comment]
I will let @a-quelle-pasqal have the last word, but I think you should at least use "Trajectory" instead of "Run"

File: pulser-simulation/pulser_simulation/simulation.py [hunk 1]
```diff
@@ -825,6 +825,7 @@ def run(
         self._validate_options(options)
 
         if not _has_stochastic_noise(self.noise_model):
+            print("Running Trajectory 1/1")
```
HGSilveri: How about 
```suggestion
            print("Emulating Trajectory 1/1")
```
?

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -872,7 +873,9 @@ def run(
     def _noisy_runs(
         self, progress_bar: bool, **options: Any
     ) -> Iterator[tuple[SimulationResults, int]]:
-        for ham, reps in self._hamiltonians:
+        n_trajectories = self.n_trajectories
+        for i, (ham, reps) in enumerate(self._hamiltonians):
+            print(f"Running Trajectory {i+1}/{n_trajectories}")
```
HGSilveri: There is a gotcha here: when `reps > 1` (i.e. when your only shot-to-shot noise is state preparation errors), you actually simulate the trajectory once but weigh it `reps` times in the result, and so you won't run the solver `n_trajectories` times. 
I think the most explicit way to show this would be to write it like so:
```suggestion
        n_trajectories = self.n_trajectories
        i = 0
        for ham, reps in self._hamiltonians:
            if reps > 1:
                print(f"Emulating Trajectories [{i+1} - {i + reps}]/{n_trajectories}"
            else:
                print(f"Running Trajectory {i+1}/{n_trajectories}")
            i += reps
```

And of course, this should be appropriately tested

pulser-simulation/pulser_simulation/simulation.py
a-quelle-pasqal: I like "Emulating Trajectory"

pulser-simulation/pulser_simulation/simulation.py
a-quelle-pasqal: yes, that's nice!

tutorials/advanced_features/Backends for Sequence Execution.ipynb
```diff
@@ -29,7 +29,7 @@
   },
```
a-quelle-pasqal, line 29: <p>this should be rerun to show "Emulating Trajectory 1/1` no?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pasqal-io/Pulser/pull/988/discussion/#comment-2707579337'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='4'/>

pulser-simulation/pulser_simulation/simulation.py
a-corni: Thanks for having caught this :) It will work as a comment (I had completely missed the purpose of `reps`) ^^'

tests/pulser_simulation/test_qutip_backend_v2.py
HGSilveri, line 1: nit: Do you think there is additional value on testing the print here wrt to `test_simulation.py` only? It looks to me that you are doing it better and more comprehensively over there

tutorials/retired/Simulating Sequences with Errors and Noise.ipynb
HGSilveri, line 1: nit: There is no need to go back and rerun the retired tutorials. I'm even afraid that we will modify them unwillingly

HGSilveri (commented):
This is in good shape, I only have nits now

tutorials/retired/Simulating Sequences with Errors and Noise.ipynb
a-corni: Reverted.

tests/pulser_simulation/test_qutip_backend_v2.py
a-corni: If we want to add an option to turn off the print, it will be in the QutipConfig and we will be happy to have the test here already. Other than this more tests never causes any harm (especially when it doesn't consume more time like this one).

PR #990, opened 2026-01-16
[FEAT] Add NoiseModel.summary()

### LINKED ISSUE #966: Add report generation functionality for NoiseModel
### Motivation
Currently, when comparing noise parameters to their actual usage in emulation, it's difficult to see the derived/computed quantities. Most noise parameters are used directly, but some (like sigmas for doppler and atom positions in register noise) are derived from the input parameters. Having a report would make it easier to:
- Understand what values are actually being used in the emulation
- Debug and validate noise model configurations
- Compare expected vs actual noise parameters

### Proposed Solution
Create a method (e.g., `NoiseModel.report()` or similar) that outputs:
- All noise parameters as directly used in emulation
- Derived quantities (particularly sigmas for doppler and atom position register noise)
- Clear labeling to distinguish input parameters from computed values

### Additional Context
This would primarily benefit users debugging simulations or validating that their noise model configurations produce the expected behavior in emulation.

a-corni:
Summarize the effect of the noises in the NoiseModel on the atoms, in order of appearance in the Sequence (Register, state initialization, amplitude, detuning, noise channels, measurement).
It's a verbose summary, as a string. I find this is the most useful to summarize the noise effects. I think we can have another issue where we provide the most noise quantities as a dictionnary (if needed).
Sums up the quantities that will be resampled when generating the trajectories at the end of the summary (signals them with **, not to confound T2* with such quantity). Didn't write NoiseModel.runs since it's now deprecated. 
Fixes #966

### DIFF
--- a/pulser-core/pulser/json/abstract_repr/schemas/config-schema.json
+++ b/pulser-core/pulser/json/abstract_repr/schemas/config-schema.json
@@ -447,14 +447,14 @@
             },
             "evaluation_times": {
                 "$ref": "#/evaluation_times"
+            },
+            "tag_suffix": {
+                "description": "added to base_tag to distinguish different instances",
+                "type": [
+                    "string",
+                    "null"
+                ]
             }
-        },
-        "tag_suffix": {
-            "description": "added to base_tag to distinguish different instances",
-            "type": [
-                "string",
-                "null"
-            ]
         },
         "required": [
             "observable",
--- a/tests/test_backend_abstract_repr.py
+++ b/tests/test_backend_abstract_repr.py
@@ -134,6 +134,11 @@ class TestObservableRepr:
                 "num_shots", None
             )
 
+        # Check observable against the schema via config serialization
+        EmulationConfig(observables=[obs]).to_abstract_repr(
+            skip_validation=False
+        )
+
     @mark.parametrize(
         "state_kwargs",
         [


pulser-core/pulser/noise_model.py
```diff
@@ -628,3 +677,129 @@ def from_abstract_repr(obj_str: str) -> NoiseModel:
...
+                    _summary += (
+                        "       - Doppler fluctuations**: "
+                        f"{doppler_sigma(self.temperature)} rad/µs\n"
+                    )
+            if len(self.detuning_hf_psd) > 0:
+                psd = list(zip(self.detuning_hf_omegas, self.detuning_hf_psd))
+                _summary += (
+                    "  - High-Frequency Detuning fluctuations**, PSD: "
+                    f"{psd} (rad/µs, rad/µs)\n"
+                )
```
HGSilveri: Isn't this going to screw up the summary given the amount of points? Maybe you can just mention how to access the points instead of including them in the string

pulser-core/pulser/noise_model.py
```diff
@@ -106,6 +108,53 @@
 )
 
 
+def doppler_sigma(temperature: float) -> float:
```
HGSilveri: As discussed, let's take the opportunity to make this private

HGSilveri:
I think it would be nice to showcase this in the `Noise Model` tutorial (and it would also give me a good example to look at :sweat_smile: )

pulser-core/pulser/noise_model.py
```diff
@@ -628,3 +677,191 @@ def from_abstract_repr(obj_str: str) -> NoiseModel:
                 obj_str
             )
         )
+
+    def get_noise_table(self) -> dict[str, tuple[Any, str]]:
+        """Maps noise quantities impacting simu. with their value and units."""
```
HGSilveri: I think we could shorten this to
```suggestion
        """Maps non-zero noise quantities with their value and units."""
```

HGSilveri: Also, a `Returns` section could be nice

pulser-core/pulser/noise_model.py
```diff
@@ -628,3 +677,191 @@ def from_abstract_repr(obj_str: str) -> NoiseModel:
...
+            _summary += "- Register Position Fluctuations**:\n"
+            _summary += (
+                "  - XY-Plane Position Fluctuations: "
+                f"{_repr_value_unit(*noise_table['register_sigma_xy'])}\n"
+            )
+            _summary += (
+                "  - Z-Axis Position Fluctuations: "
+                f"{_repr_value_unit(*noise_table['register_sigma_z'])}\n"
+            )
+            add_to_traj_summary.append("register, ")
```
HGSilveri: Here my general reccomendation would be to create list of lines that you join at the end with `"\n"`, so you would go like this

```suggestion
        _summary = ["Noise summary:"]
        noise_table = self.get_noise_table()
        add_to_traj_summary = []
        # Follow the impact as in step-by-step tutorial
        if "register_sigma_xy" in noise_table:
            assert "register_sigma_z" in noise_table
            # 1. Register
            _summary += [
                "- Register Position Fluctuations**:",
                "  - XY-Plane Position Fluctuations: "
                f"{_repr_value_unit(*noise_table['register_sigma_xy'])}",
                "  - Z-Axis Position Fluctuations: "
                f"{_repr_value_unit(*noise_table['register_sigma_z'])}",
            ]
            add_to_traj_summary.append("register, ")
```

I prefer this because I find it more readable and you don't have to care about the `\n` everywhere.

pulser-core/pulser/noise_model.py
```diff
@@ -628,3 +677,191 @@ def from_abstract_repr(obj_str: str) -> NoiseModel:
...
+            table["p_false_neg"] = (self.p_false_neg, "")
+        return table
+
+    def summary(self) -> str:
+        """A readable summary of the impact of the noise on the simulation."""
+
+        def _repr_value_unit(value: Any, unit: str) -> str:
+            if unit == "":
+                return f"{value}"
+            return f"{value} {unit}"
```
HGSilveri: I think we should round floats to a few significant digits (maybe the Python default of 6 is fine):

```suggestion
        def _repr_value_unit(value: Any, unit: str) -> str:
            value_str = f"{value:g}" if isinstance(value, float) else f"{value}"
            if unit == "":
                return value_str
            return f"{value_str} {unit}"
```

pulser-core/pulser/noise_model.py
```diff
@@ -628,3 +677,191 @@ def from_abstract_repr(obj_str: str) -> NoiseModel:
...
+                    "   - Custom Lindblad operators (in 1/µs)"
+                    + (
+                        " including a leakage state"
+                        if noise_table["with_leakage"][0]
+                        else ""
+                    )
+                    + ":\n"
+                )
+                for rate, oper in noise_table["eff_noise"][0]:
+                    _summary += f"       - {rate} * {oper}\n"
```
HGSilveri: How well does this render?

a-corni: See below `       - 0.1 * ((0.0, 1.0, 0.0), (1.0, 0, 0), (0.0, 0, -1))` or in the tests: https://github.com/pasqal-io/Pulser/pull/990/changes#diff-f73d1200ccf4e4cea483647b602907846d2cc65008f64439d54daaee9654ba8cR787-R818

docs/source/noise_model.ipynb
```diff
@@ -60,7 +60,8 @@
    "id": "725df1eb",
    "metadata": {},
    "source": [
-    "If you have a `NoiseModel`, you can know the types of noise that it implements checking its property `noise_types`."
+    "If you have a `NoiseModel`, you can know the types of noise that it implements by checking its property `noise_types`. This offers a qualitative description of the noises. If you want to know the qualitative impact of each noise on your `Sequence`, the `get_noise_table` method will return you a mapping between noise quantities to their value, and the `summary` method will provide you with a readable version of this table."
```
HGSilveri: I guess you meant

```suggestion
    "If you have a `NoiseModel`, you can know the types of noise that it implements by checking its property `noise_types`. This offers a qualitative description of the noises. If you want to know the quantitative impact of each noise on your `Sequence`, the `get_noise_table()` method will return you a mapping between noise quantities to their value, and the `summary()` method will provide you with a readable version of this table."
```

pulser-core/pulser/noise_model.py
```diff
@@ -636,3 +689,196 @@ def from_abstract_repr(obj_str: str) -> NoiseModel:
...
+            # 2. State Preparation
+            summary_list.append(
+                "- State Preparation Error Probability**: "
+                f"{_repr_value_unit(*noise_table['state_prep_error'])}"
+            )
+            add_to_traj_summary.append("initial state")
+
+        # 3. Pulse Shaping
+        if "laser_waist" in noise_table or "amp_sigma" in noise_table:
+            summary_list.append("- Amplitude fluctuations:")
```
HGSilveri: To encompass the finite-waist, maybe 
```suggestion
            summary_list.append("- Amplitude inhomogeneities:")
```
is more comprehensive. To me, it better encompasses both spatial and temporal inhomogeneities

pulser-core/pulser/noise_model.py
```diff
@@ -636,3 +689,196 @@ def from_abstract_repr(obj_str: str) -> NoiseModel:
...
+            if "laser_waist" in noise_table:
+                summary_list += [
+                    "  - Finite-waist Gaussian damping σ="
+                    f"{_repr_value_unit(*noise_table['laser_waist'])}"
+                ]
+            if "amp_sigma" in noise_table:
+                summary_list += [
+                    "  - Shot-to-shot Amplitude Fluctuations**:"
+                    f" {_repr_value_unit(*noise_table['amp_sigma'])}"
+                    " Amplitude"
```
HGSilveri: I think the % implies well enough that it is a fraction of the amplitude, we can probaly skip this line

pulser-core/pulser/noise_model.py
```diff
@@ -636,3 +689,196 @@ def from_abstract_repr(obj_str: str) -> NoiseModel:
                 obj_str
             )
         )
+
+    def get_noise_table(self) -> dict[str, tuple[Any, str]]:
```
HGSilveri, line 693: This ended up becoming quite a long method but it gets the job done, let's not waste time trying to make it more compact

EDIT: I meant to put this on the `summary()` method actually

docs/source/noise_model.ipynb
```diff
@@ -60,7 +60,26 @@
    "id": "725df1eb",
    "metadata": {},
    "source": [
-    "If you have a `NoiseModel`, you can know the types of noise that it implements checking its property `noise_types`."
+    "If you have a `NoiseModel`, you can know the types of noise that it implements by checking its property `noise_types`. This offers a qualitative description of the noises. If you want to know the quantitative impact of each noise on your `Sequence`, the `get_noise_table()` method will return you a mapping between noise quantities and their value, and the `summary()` method will provide you with a readable version of this table."
```
HGSilveri, line 71: Let's put this in a subsection, e.g. "Inspecting a NoiseModel"

docs/source/noise_model.ipynb
```diff
@@ -176,7 +184,7 @@
     "\n",
     "**Notes**:\n",
     "\n",
-    "- Stochastic noises can be identified by a `**` in `NoiseModel.summary()` (see example in the first part).\n",
+    "- Stochastic noises can be identified by a `**` in `NoiseModel.summary()` (see example in `Inspecting a NoiseModel`).\n",
```
HGSilveri: nit: Link to the section could be nice

PR #992, opened 2026-01-22
[FIX] Make BackendConfig read-only + Add `BackendConfig.with_changes()`

HGSilveri:
`BackendConfig` currently allows unsafe in-place changes. This PR:

- Blocks `__setattr__` to make the class read-only
- Adds a `BackendConfig.with_changes()` method (**the name is up for discussion**) to make a copy with modifications


a-quelle-pasqal (approved):
good. This was bugging me a bit.

pulser-core/pulser/backend/config.py
```diff
@@ -109,6 +112,10 @@ def __init__(
         # Store in _backend_options together with all the other paramters
         self._backend_options["default_num_shots"] = default_num_shots
 
+    def with_changes(self: Self, **changes: Any) -> Self:
```
a-corni, line 115: Nice typing
nit: I found `replace` more intuitive than `with_changes` but

tests/test_backend.py
```diff
@@ -665,8 +677,20 @@ def test_emulation_config():
...
+    with pytest.raises(
+        AttributeError,
+        match=re.escape(
+            "'EmulationConfig' is read-only. Please use "
+            "'EmulationConfig.with_changes(n_trajectories=...)'"
+        ),
+    ):
+        config.n_trajectories = 10
+
+    assert config.with_changes(n_trajectories=10).n_trajectories == 10
```
a-corni, line 693: nit: Eventually just check that config is unchanged `assert config.n_trajectories == 1`

a-corni (commented):
I have two nits but otherwise it looks good to me !

pulser-core/pulser/backend/config.py
HGSilveri: My issue is that `replace()` as a method implies in-place substitution, while if you put `with_...` it's generally understood that it will return a new instance (and it's also what we follow on e.g. `with_new_device`).

With that in mind, I was torn between `with_changes()` and `with_updates()`, do you have a preference?

a-corni (approved):

PR #993, opened 2026-01-22
[FIX] EOM: Allow for the definition of |detuning_off|>Channel.max_abs_detuning

### LINKED ISSUE #953: Detuning goes beyond the accepted range in EOM mode at extreme values of detuning_on
# Description

Currently, the `Device.max_abs_detuning` is enforced to all values of detuning in a channel. This causes a failure at a particular edge-case:

- When `Device.eom_config.controlled_beams` has a single entry, there will be a single option of `detuning_off = detuning_on + lightshift`, where: 
  - `lightshift > 0` if the controlled beam is the RED
  - `lightshift < 0` if the controlled beam is the BLUE

- Taking the more common case where `lightshift < 0` let's imagine `detuning_on` is asked to be `-max_abs_detuning`
- Then, `detuning_off = -max_abs_detuning + lightshift < -max_abs_detuning` i.e. the `detuning_off` unavoidable falls outside the `max_abs_detuning` range
- **Currently, this fails ungracefully**
- Note that `detuning_on` doesn't even have to be set at the extreme value, anywhere between `]-max_abs_detuning-lightshift, -max_abs_detuning]` would trigger the same error

## MRE
With `pulser-core == 1.6.2`
```python
import pulser

seq = pulser.Sequence(pulser.Register({"q0": (0, 0)}), pulser.AnalogDevice)
seq.declare_channel("rydberg_global", "rydberg_global")
max_abs_detuning = seq.declared_channels["rydberg_global"].max_abs_detuning
seq.enable_eom_mode("rydberg_global", amp_on=1., detuning_on=-max_abs_detuning)
``` 
<details>
  <summary>Traceback</summary>

```
ValueError                                Traceback (most recent call last)
Cell In[6], [line 6](vscode-notebook-cell:?execution_count=6&line=6)
      4 seq.declare_channel("rydberg_global", "rydberg_global")
      5 max_abs_detuning = seq.declared_channels["rydberg_global"].max_abs_detuning
----> [6](vscode-notebook-cell:?execution_count=6&line=6) seq.enable_eom_mode("rydberg_global", amp_on=1., detuning_on=-max_abs_detuning)

File ~/git/Pulser/pulser-core/pulser/sequence/_decorators.py:82, in verify_parametrization.<locals>.wrapper(self, *args, **kwargs)
     80 for x in chain(args, kwargs.values()):
     81     verify_variable(self, x)
---> [82](https://file+.vscode-resource.vscode-cdn.net/home/henrique/git/Pulser/~/git/Pulser/pulser-core/pulser/sequence/_decorators.py:82) func(self, *args, **kwargs)

File ~/git/Pulser/pulser-core/pulser/sequence/_decorators.py:154, in conditionally_block.<locals>.decorator.<locals>.wrapper(self, *args, **kwargs)
    134     if (
    135         # If truncate is a parametrized call and one of the blocker
    136         # calls is before it, no more instructions are accepted
   (...)
    146         )
    147     ):
    148         raise RuntimeError(
    149             "The sequence can only be measured. This is because it"
    150             f"is parametrized and one or more of {blocker_calls} "
    151             "was called before a `truncate()` call."
    152         )
--> [154](https://file+.vscode-resource.vscode-cdn.net/home/henrique/git/Pulser/~/git/Pulser/pulser-core/pulser/sequence/_decorators.py:154) return func(self, *args, **kwargs)

File ~/git/Pulser/pulser-core/pulser/sequence/sequence.py:1044, in Sequence.enable_eom_mode(self, channel, amp_on, detuning_on, optimal_detuning_off, correct_phase_drift)
   1041 if not channel_obj.supports_eom():
   1042     raise TypeError(f"Channel '{channel}' does not have an EOM.")
-> [1044](https://file+.vscode-resource.vscode-cdn.net/home/henrique/git/Pulser/~/git/Pulser/pulser-core/pulser/sequence/sequence.py:1044) detuning_off, switching_beams = self._process_eom_parameters(
   1045     channel_obj, amp_on, detuning_on, optimal_detuning_off
   1046 )
   1047 if not self.is_parametrized():
   1048     assert not isinstance(amp_on, Parametrized)

File ~/git/Pulser/pulser-core/pulser/sequence/sequence.py:2505, in Sequence._process_eom_parameters(self, channel_obj, amp_on, detuning_on, optimal_detuning_off)
   2493 (
   2494     detuning_off,
   2495     switching_beams,
   (...)
   2500     return_switching_beams=True,
   2501 )
   2502 off_pulse = Pulse.ConstantPulse(
   2503     channel_obj.min_duration, 0.0, detuning_off, 0.0
   2504 )
...
    486     )
    487 avg_amp = np.average(amp_samples_np)
    488 if 0 < avg_amp < self.min_avg_amp:

ValueError: The pulse's detuning values go out of the range allowed for the chosen channel.
```
</details>

## The patch

The easiest thing to do would be to restrict the allowed values beyond `max_abs_detuning`, informing the user that `detuning_on` needs to be within a more restrict range to ensure `detuning_off` still respects `max_abs_detuning`.
This is not an ideal solution because
1. It requires the user to run into the error (unless we add a way to access the restricted valid `detuning_on` range + document this)
2. It restricts `detuning_on` beyond what is necessary

## The proper fix

The proper thing to do would be to **allow `detuning_off` to go beyond `max_abs_detuning`**. This is physically possible, since the spill over is coming from the lightshift contribution.
However, a change like this can have widespread ramifications. In particular, I'm concerned about the reliance on the assumption, both within our codebase or even outside of it, that the detuning is always within `[-max_abs_detuning, max_abs_detuning]`.

a-corni:
In EOM mode, the detuning is constant = detuning_on, but the detuning seen by the atom varies between detuning_on (when a pulse is applied ie when the switch beam(s) is(are) switched ON) and detuning off during enable_eom/disable_eom/delays (when the switch beam(s) is(are) switched OFF). 

We allow the definition of detuning_on=max_abs_detuning and allow to have |detuning_off|>Channel.max_abs_detuning. This doesn't have an impact on the rest of the stack.

PR starts by introducing a failing test, and then the feature that fixes it.

Fixes #953


pulser-core/pulser/sequence/sequence.py
```diff
@@ -2501,10 +2501,16 @@ def _process_eom_parameters(
                     float(optimal_detuning_off),
                     return_switching_beams=True,
                 )
-                off_pulse = Pulse.ConstantPulse(
-                    channel_obj.min_duration, 0.0, detuning_off, 0.0
+                # Detuning sent by laser is constant equal to detuning_on
+                assert np.isclose(
```
HGSilveri: This is assert is basically just testing `calculate_detuning_off()`, I'm not sure there is much need for it.

tests/pulser_simulation/test_simulation.py
```diff
@@ -2470,3 +2471,65 @@ def test_qutip_invalid_solver_error(seq):
             n_trajectories=1,
         )
         sim.run()
+
+
+@pytest.mark.parametrize("min_detuning_on", [False, True])
+def test_eom(mod_device, reg, min_detuning_on):
```
HGSilveri: nit: 

```suggestion
def test_eom_limit_det(mod_device, reg, min_detuning_on):
```

tests/pulser_simulation/test_simulation.py
```diff
@@ -2470,3 +2471,65 @@ def test_qutip_invalid_solver_error(seq):
...
+            "101": 9,
+        }
+    else:
+        assert final_state == {"000": 879, "010": 49, "100": 40, "001": 32}
+    # Also with noisy detuning
+    sim = QutipEmulator.from_sequence(
+        seq, noise_model=NoiseModel(detuning_sigma=0.1), n_trajectories=3
+    )
+    res = sim.run()
+    assert res[-1].bitstring_counts == {"000": 3}
```
HGSilveri: Honest question: do we actually care about the results here? Don't we just want to make sure QutipEmulator can handle it?

If we don't care, I would avoid the maintenance overhead of the hardcoded results

tests/test_sequence.py
```diff
@@ -71,6 +71,14 @@ def device():
     )
 
 
+def _eom_detuning_on_dict(max_abs_detuning: float) -> dict[str, float]:
+    """A dict of detuning_on possibilities in EOM mode."""
+    return {"NULL": 0, "MAX": -max_abs_detuning}
+
+
+EOM_DETUNING_ON_CHOICES: list[str] = list(_eom_detuning_on_dict(0))
```
HGSilveri: nit: I think it'd be fine and likely preferable to place these next to the test where you use them

tests/test_sequence.py
```diff
@@ -1155,6 +1163,7 @@ def test_switch_device_up(
 )
 
 
+@pytest.mark.parametrize("detuning_on_str", EOM_DETUNING_ON_CHOICES)
```
HGSilveri: I found this quite convoluted, could we be more explicit? E.g.


```suggestion
@pytest.mark.parametrize("detuning_on_option", ["NULL", "MAX"])
```

tests/test_sequence.py
```diff
@@ -1615,17 +1634,46 @@ def test_delay_at_rest(in_eom, at_rest, delay_duration):
...
     assert pulse.duration == 100
     if in_eom:
-        seq.enable_eom_mode("ryd", 1, 0, 0)
+        seq.enable_eom_mode("ryd", amp, det, 0)
         seq.add_eom_pulse("ryd", pulse.duration, 0)
+        assert len(seq._schedule["ryd"].eom_blocks) == 1
+        assert seq._schedule["ryd"].eom_blocks[0].detuning_on == det
+        assert (
+            det_off := seq._schedule["ryd"].eom_blocks[0].detuning_off
+        ) < det
```
HGSilveri: nit: I think these checks are a bit out of place here, considering that they don't have much to do with "delay at rest". You can perhaps move them further down (inside the `if in_eom:` you have at the end) and add a comment explaining why you are checking this

tests/test_sequence.py
```diff
@@ -3009,7 +3070,16 @@ def test_modify_eom_setpoint(
...
+    detuning_on_match = _eom_detuning_on_dict(ryd_ch_obj.max_abs_detuning)
+    detuning_on = detuning_on_match[detuning_on_str]
+    param_vals = [1.0, detuning_on]
+    if detuning_on + det_diff < -ryd_ch_obj.max_abs_detuning:
+        with pytest.raises(
+            ValueError, match="The pulse's detuning values go out of the range"
+        ):
+            seq.build(params=param_vals)
+        # Nothing else to check
+        return
```
HGSilveri: The issue with this construction is that you might inadvertedly always be entering this `if` and never reaching the lines after it. It would be cleaner and safer to test this separately

pulser-core/pulser/sequence/sequence.py
a-corni: I agree, I first used it to check my understanding and decided to let it to remind that if we someday change the computation of detuning_off, this has an impact here. Wdyt ?

tests/pulser_simulation/test_simulation.py
a-corni: Okay, but we actually don't have any tests with a Sequence using EOM.

tests/pulser_simulation/test_simulation.py
a-corni: I agree that the outcome with only n_trajectories=3 isn't meaningful enough. Let's delete this result comparison.

pulser-core/pulser/sequence/sequence.py
HGSilveri: If the purpose is to have a guardrail, I'd rather have us check that `detuning_off >= -max_abs_detuning + lightshift` if `lightshift < 0` or `detuning_off <= max_abs_detuning + lightshift` if `lightshift > 0`

a-corni: That would be similar to having `channel_obj.validate_pulse(off_pulse)` with `off_pulse = Pulse.ConstantPulse(channel_obj.min_duration, 0.0, detuning_off-lightshift, 0.0)`, right ?

HGSilveri: Yes, although that's perhaps more convoluted. The point is that we want to make sure the spill over comes exclusively from lightshift. I guess you can do `|detuning_off| - max_abs_detuning <= |lightshift|`

a-corni: I went for your first proposition in the end.

tests/test_sequence.py
a-corni: It's actually targetting only one of the cases I made by looping over [det_diff possibilities]x["NULL", "MAX"]. I have made it explicit that I am catching this case and this case only.

HGSilveri: Good enough for me :)

PR #994, opened 2026-01-23
default_evaluation_times should not be used when evaluation_times is passed for an Observable

a-quelle-pasqal:
I updated the logic in `Observable.__call__`, and updated one of the tests with another assert that used to rely on the old behaviour.

### DIFF
--- a/pulser-simulation/pulser_simulation/__init__.py
+++ b/pulser-simulation/pulser_simulation/__init__.py
@@ -16,6 +16,7 @@
 from pulser import EmulatorConfig, NoiseModel
 
 from pulser_simulation._version import __version__ as __version__
+from pulser_simulation.aggregators import density_matrix_aggregator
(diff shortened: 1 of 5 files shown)


tests/test_backend.py
```diff
@@ -669,11 +669,12 @@ def test_emulation_config():
...
 
     # EmulationConfig would error on receiving a numpy array of len > 1
+    times = np.array([0.5, 1.0])
     conf = EmulationConfig(
-        default_evaluation_times=np.array([0.5, 1.0]),
+        default_evaluation_times=times,
         observables=(BitStrings(),),
     )
-    assert conf.default_evaluation_times == [0.5, 1.0]
+    assert np.allclose(conf.default_evaluation_times, times)
```
HGSilveri: nit: I guess "all equal" would work too?
```suggestion
    np.testing.assert_equal(conf.default_evaluation_times, times)
```

PR #972, opened 2025-12-04
Add tutorial on QPU execution

a-quelle-pasqal:
Added a notebook on qpu execution.

Solves #970

PR #995, opened 2026-01-23
Results class improvements + Docs page on Results and Observables

HGSilveri:
- [x] Creates a dedicated `Results.__str__()` 
- [x] Improves backend interoperability via classproperties `EmulationConfig.state_type/operator_type` and `EmulatorBackend.config_type` (see example in the new docs page)
- [x] Creates a docs page on how to use observables and access results
- [ ] Updates the `Backend Execution` page to take advantage of these changes (and potentially also #992)
- [x] Dedicated UTs

### DIFF
--- a/pulser-core/pulser/backend/config.py
+++ b/pulser-core/pulser/backend/config.py
@@ -50,6 +50,9 @@ EVAL_TIMES_LITERAL = Literal["Full", "Minimal", "Final"]
 
 StateType = TypeVar("StateType", bound=State)
 
+# TODO: Replace with built-in Self when python >= 3.11
+Self = TypeVar("Self", bound="BackendConfig")
+
 
 class BackendConfig:
     """The base backend configuration.
@@ -87,7 +90,7 @@ class BackendConfig:
             )
         # Store the abstract repr of the config in _backend_options
         # Prevents potential issues with mutable arguments
-        self._backend_options = copy.deepcopy(backend_options)
+        super().__setattr__("_backend_options", copy.deepcopy(backend_options))
         if "backend_options" in backend_options:
             with warnings.catch_warnings():
                 warnings.filterwarnings("always")
@@ -110,6 +113,10 @@ class BackendConfig:
         # Store in _backend_options together with all the other paramters
         self._backend_options["default_num_shots"] = default_num_shots
 
+    def with_changes(self: Self, **changes: Any) -> Self:
+        """Returns a copy of the config with the given changes."""
+        return type(self)(**(self._backend_options | changes))
+
     def _expected_kwargs(self) -> set[str]:
         return set()
 
@@ -121,6 +128,14 @@ class BackendConfig:
         ):
             return self._backend_options[name]
         raise AttributeError(f"{name!r} has not been passed to {self!r}.")
+
+    def __setattr__(self, name: str, value: Any) -> None:
+        cls_name = type(self).__name__
+        raise AttributeError(
+            f"{cls_name!r} is read-only. Please use "
+            f"'{cls_name}.with_changes({name}=...)' to make a copy with the "
+            "desired changes."
+        )
 
 
 class EmulationConfig(BackendConfig, Generic[StateType]):
@@ -420,7 +435,7 @@ class EmulationConfig(BackendConfig, Generic[StateType]):
 # Legacy class
 
 
-@dataclass
+@dataclass(frozen=True)
 class EmulatorConfig(BackendConfig):
     """The configuration for emulator backends.
 
--- a/pulser-core/pulser/sequence/helpers/_switch_device.py
+++ b/pulser-core/pulser/sequence/helpers/_switch_device.py
@@ -15,11 +15,10 @@
 from __future__ import annotations
 
 import dataclasses
+import functools
 import itertools
 import warnings
 from typing import TYPE_CHECKING, Any, cast
-
-import numpy as np
 
 from pulser.channels.base_channel import Channel
 from pulser.channels.dmm import _get_dmm_name
@@ -98,12 +97,13 @@ def switch_device(
             int, ch_obj.fixed_retarget_t
         ) < cast(int, ch_obj.min_retarget_interval)
 
+    @functools.cache
     def check_channels_match(
         old_ch_name: str,
         new_ch_obj: Channel,
-        active_eom_channels: list,
+        active_eom_channels: tuple[str, ...],
         strict: bool,
-    ) -> tuple[str, str]:
+    ) -> tuple[str, str, list[str]]:
         """Check whether two channels match.
 
         Returns a tuple that contains a non-strict error message and a
@@ -113,6 +113,7 @@ def switch_device(
         will eventually be filled. If strict=True, all the conditions are
         checked - the returned error can either be non-strict or strict.
         """
+        diff_params = []
         old_ch_obj = seq.declared_channels[old_ch_name]
         # We verify the channel class then
         # check whether the addressing is Global or Local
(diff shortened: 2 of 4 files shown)


docs/source/results.ipynb
```diff
@@ -0,0 +1,464 @@
+{
```
a-quelle-pasqal, line 1: <p>Recently, we added a <code>SparseOperator</code> to emu-sv, in addition to the <code>DenseOperator</code> which you will get via the procedure described here. This is required to work with <code>Operator</code> objects at all for 20+ qubits. I'll leave it up to you how much of the specifics to go into here, but it's probably a good idea to mention that multiple <code>Operator</code> objects can exist for a backend, and to look at the backend specific documentation for more information. There are several instances where in the emulator docs we refer to the documentation here, and I'm now reminded that this might be problematic in the cloud mirror of all the documentation. As a consequence I can imagine if you don't want to link to the emulator docs here, and that raises the question of how much detail to go into.   </p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pasqal-io/Pulser/pull/995/discussion/#comment-2721195306'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='7'/>

a-quelle-pasqal, line 1: <pre class="ql-syntax" spellcheck="false">Line #29.        evaluation_times=[0.0, 0.25, 0.5, 0.75, 1.0],
</pre><p>due to the current interest in rotating the measurement basis (see the questions raised in the emulators channel by joseph for example, of which there were several in this direction this month), maybe measuring sigma_x would cause these lines to stick with the reader better.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pasqal-io/Pulser/pull/995/discussion/#comment-2721195309'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='8'/>

a-quelle-pasqal, line 1: <p>should this be a raw cell? Some formatting is a bit weird as a consequence. See the occurances of :py:class:</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pasqal-io/Pulser/pull/995/discussion/#comment-2721195311'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='10'/>

a-quelle-pasqal, line 1: <p>I assume you're going to remove all the duplicates with the above cells from here later, since you mentioned you're not with the notebook yet. However, just to be safe, I'm mentioning it.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pasqal-io/Pulser/pull/995/discussion/#comment-2721195314'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

HGSilveri: Yes, this should render properly in the RTD build (which I hopefully just fixed with my latest push)

HGSilveri: <p>I'm torn on that, I know that some people like self-contained scripts that they can just copy paste, but I'm not a fan of the long cell nor the code duplication. I went with the the former option but it's up to debate</p><p> </p>

HGSilveri: <p>I have no problem with that</p>

HGSilveri: I'm trying to minimize the amount of backend-specific decisions we force upon the users, and how much they have to dig in the documentation.

One idea I had but didn't address here because of scope creep (among other things) was to automatically attempt casting between `State/Operator` of different types, which should be doable whenever they have been created via the abstract repr.
This would allow e.g.
- `QutipBackendV2` taking in a fidelity state as a `StateRepr`
- `EmuSVBackend` converting between `DenseOperator` and `SparseOperator` conditionally on system size.

This still needs to be fleshed out further but I believe it would be a nice improvement to interoperability.

I will make sure to open an issue for this.

HGSilveri: I'm hidding the code in this cell and instead putting it in a toggled section. Let me know what you think of this solution

docs/source/results.ipynb
```diff
@@ -0,0 +1,580 @@
...
+   "metadata": {
+    "editable": true,
+    "raw_mimetype": "text/x-rst",
+    "slideshow": {
+     "slide_type": ""
+    },
+    "tags": []
+   },
+   "source": [
+    "On every Pulser backend, a call to ``Backend.run()`` will return an instance of :py:class:`~pulser.backend.Results` (or in the case of a ``RemoteBackend``, a sequence of :py:class:`~pulser.backend.Results`). \n",
```
a-corni: Perhaps we could add a reference to the Backend execution tutorial here ?

docs/source/results.ipynb
```diff
@@ -0,0 +1,580 @@
...
+    "- what the `Observable` mechanism is and how to use it;\n",
+    "- what observables are available by default;\n",
+    "- how to retrieve the results in a `Results` instance."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## The `Observable` mechanism\n",
```
a-corni: I think we are lacking a little bit of introduction at the beggining of the tutorial on when do you care about Observables -> Just before the first part, I would put a small sentence along those lines:

`"As introduced in the [backend execution tutorial](./backends.nblink), emulators measure certain quantities along time, called "Observables", and return these measurements inside a :py:class:`~pulser.backend.Results` class (or in the case of a ``RemoteBackend``, a sequence of :py:class:`~pulser.backend.Results`). Let's dive into `Observables` and the `Results` class. \n",

docs/source/results.ipynb
```diff
@@ -0,0 +1,580 @@
...
+    "    amplitudes={\"rr\": 1.0},\n",
+    ")\n",
+    "# Use `tag_suffix` to better identify the observable in the Results\n",
+    "fidelity = Fidelity(r_state, tag_suffix=\"rr\")\n",
+    "\n",
+    "# Arbitrarily chosen operator XX (where X = |r><g|+|g><r|)\n",
+    "pauli_x = operator_class.from_operator_repr(\n",
+    "    eigenstates=(\"r\", \"g\"),\n",
+    "    n_qudits=2,\n",
+    "    operations=[(1.0, [({\"rg\": 1.0, \"gr\": 1.0}, {0, 1})])],\n",
```
a-corni: Should we have a little subsection to explain how to define the operations ?

docs/source/results.ipynb
```diff
@@ -0,0 +1,580 @@
...
+   "outputs": [],
+   "source": [
+    "results.get_tagged_results()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "To get the list of values for a particular observable, we can either access the dictionary directly or use a convenience shortcut:"
```
a-corni: nit
```suggestion
    "To get the list of values for a particular observable, we can either access the dictionary directly or use a convenient shortcut:"
```

docs/source/results.ipynb
```diff
@@ -0,0 +1,580 @@
...
+   "file_extension": ".py",
+   "mimetype": "text/x-python",
+   "name": "python",
+   "nbconvert_exporter": "python",
+   "pygments_lexer": "ipython3",
+   "version": "3.12.3"
+  }
+ },
+ "nbformat": 4,
+ "nbformat_minor": 4
```
a-corni: nit: You are not talking about RemoteResults and how to find Results out of them/how to create them from a batch id. I think this has its place in the execution on Emulator/execution on QPU backend (I have added it there), we do agree its place is in #972 ? Do we want to make a "To go further" section here ?

a-corni: Or a simple note

pulser-core/pulser/backend/results.py
```diff
@@ -41,13 +41,15 @@ class Results:
     """
 
     atom_order: tuple[str, ...]
+    """The order of the atoms/qudits in the results."""
     total_duration: int
-    _results: dict[uuid.UUID, list[Any]] = field(init=False)
-    _times: dict[uuid.UUID, list[float]] = field(init=False)
+    """The total duration of the sequence, in ns."""
```
a-corni, line 46: I don't get these two comments, aren't they redundant with the docstring ?

pulser-core/pulser/backend/results.py
```diff
@@ -401,6 +403,22 @@ def aggregate(
 
         return aggregated
 
+    def __str__(self) -> str:
+        evaluation_times = {
+            tag: self._times[_uuid] for tag, _uuid in self._tagmap.items()
+        }
```
a-corni, line 409: What was the issue with `self._times` ?

HGSilveri: `self._times` has UUIDs as keys, which is highly unreadable. This substitutes the UUIDs with the tags

pulser-core/pulser/backend/results.py
HGSilveri: It's to make them appear nicely here

image

docs/source/results.ipynb
HGSilveri: What if instead I improve the "What you will learn" to convey the same info? e.g.

*What you will learn:*

- how to configure an `Observable` to measure quantities of interest in an emulation;
- what observables are available by default;
- how to retrieve the measured observables from a `Results` instance.

docs/source/results.ipynb
HGSilveri: Since that is not relevant only in obversables (e.g. you can use `State` to define a custom initial state) I think this deserves either its own page and/or a more detailed docstring. 

I will open an issue to do this

docs/source/results.ipynb
HGSilveri: I just took out "convenience", it is a bit redundant

docs/source/results.ipynb
HGSilveri: I think I would link the QPU tutorial, but we'll need it first

docs/source/results.ipynb
HGSilveri: I sneaked the reference in the callout box a bit below, let me know what you think

pulser-core/pulser/backend/results.py
a-corni: Aaaah thanks !

docs/source/results.ipynb
a-corni: Even though I usually like a sentence at the beginning of the tutorial, I agree the What you will learn is now improved and is surely sufficient ;)

docs/source/results.ipynb
a-corni: I will try to wrap it up today.

a-corni: So you will merge first.

PR #998, opened 2026-01-27
Make `job_params` optional for built sequences on every remote backend

### LINKED ISSUE #959: Stop using 'job_params' as the way to specify the runs on a `RemoteBackend`
- For `EmulatorBackend`s on the new interface, the number of runs is not required and is even misleading. If it is defined, it should be via the `Bitstrings` observable, not via `job_params`. When given in `job_params`, we should raise a warning saying we are ignoring the value.
- For `QPUBackend`, we could introduce a `QPUConfig` class with a parameter to specify a default number of runs. We should still allow this parameter to be overwritten in `job_params` in case the user requires a different number of shots for different combinations of variables, but at least we wouldn't require this value to be specified.

**Open question:**

In the past, we have considered standardizing what I'm here calling "runs". In the emulators we usually refer to these as "shots" while in other places it is even referred to as "samples" (*not* a good option, let's not consider it). 
Since moving from "shots" to "runs" in the emulators would be more disruptive, would this be a good opportunity to start calling these "shots" instead?

HGSilveri:
- [x] On `QPUBackend`, defaults `runs` to `BackendConfig.default_num_shots`
- [x] On `RemoteEmulatorBackend`, raises warning whenever `runs` is defined (i.e. merge https://github.com/pasqal-io/pasqal-cloud/pull/229)
- [x] Upper-bounds `pulser-pasqal` once https://github.com/pasqal-io/pasqal-cloud/pull/229 is merged and release
- [x] ~Updates `Emulator execution` tutorial to stop using the `runs=1` hack~ TBD in #1003 instead

Closes #959 .

### DIFF
--- a/pulser-core/pulser/backend/qpu.py
+++ b/pulser-core/pulser/backend/qpu.py
@@ -57,7 +57,7 @@ class QPUBackend(RemoteBackend):
                 job with 'BackendConfig.default_num_shots'.
                 If the sequence is parametrized, the values for all the
                 variables necessary to build the sequence must be given in
-                it's own dictionary, for each job, under the 'variables' field.
+                its own dictionary, for each job, under the 'variables' field.
                 Each dictionary may contain a custom 'runs' field specifying
                 the number of shots to take of the same sequence; if not given,
                 'BackendConfig.default_num_shots' is used when available.
--- a/pulser-core/pulser/backend/remote.py
+++ b/pulser-core/pulser/backend/remote.py
@@ -346,12 +346,10 @@ class RemoteBackend(Backend):
         """Runs the sequence on the remote backend and returns the result.
 
         Args:
-            job_params: A list of parameters for each job to execute. Each
-                mapping must contain a defined 'runs' field specifying
-                the number of times to run the same sequence. If the sequence
-                is parametrized, the values for all the variables necessary
-                to build the sequence must be given in it's own mapping, for
-                each job, under the 'variables' field.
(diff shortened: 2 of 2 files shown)


pulser-core/pulser/backend/__init__.py
```diff
@@ -18,7 +18,11 @@
     Backend as Backend,
     EmulatorBackend as EmulatorBackend,
 )
-from pulser.backend.config import EmulatorConfig, EmulationConfig
+from pulser.backend.config import (
+    BackendConfig,
+    EmulatorConfig,
+    EmulationConfig,
```
a-corni, line 24: Note for #972 we should add RemoteResults in backend.__init__

pulser-core/pulser/backend/qpu.py
```diff
@@ -31,26 +32,35 @@ class QPUBackend(RemoteBackend):
...
-                the number of times to run the same sequence. If the sequence
-                is parametrized, the values for all the variables necessary
-                to build the sequence must be given in it's own mapping, for
-                each job, under the 'variables' field.
+            job_params: A list of dictionaries with the parameters to execute
+                each job. If not given, the backend will attempt to run one
+                job with 'BackendConfig.default_num_shots'.
+                If the sequence is parametrized, the values for all the
+                variables necessary to build the sequence must be given in
+                it's own dictionary, for each job, under the 'variables' field.
```
a-corni: Nit: its

PR #1005, opened 2026-02-02
Fix definition of eff_noise_opers with torch tensor

a-corni:


pulser-core/pulser/noise_model.py
```diff
@@ -25,11 +25,17 @@
...
+import pulser.math as pm
 from pulser.constants import KB, KEFF, MASS, TRAP_WAVELENGTH
 from pulser.json.abstract_repr.serializer import AbstractReprEncoder
 from pulser.json.abstract_repr.validation import validate_abstract_repr
 from pulser.json.utils import get_dataclass_defaults
 
+try:
+    import torch
+except ImportError:  # pragma: no cover
+    pass
```
HGSilveri: nit: You can rely on `pm.torch` instead of importing explicitly

pulser-core/pulser/noise_model.py
HGSilveri, line 326: I think this would be preferable:

```suggestion
            if isinstance(obj, qutip.Qobj):
                obj = np.array(
            obj.to("Dense").data_as("ndarray"),  # type: ignore
            dtype=complex,
                )
            if pm.AbstractArray.has_torch() and isinstance(obj, pm.torch.Tensor):
                # Convert torch objects into np.ndarray
                obj = pm.AbstractArray(obj).as_array(detach=False)
```

I would not detach to explicitly fail if the torch tensor is differentiable (let's add a test for that too btw)

HGSilveri (commented):
Thanks for tackling this @a-corni !

pulser-core/pulser/noise_model.py
a-corni: That would have been preferable since the beginning, but I can't have qutip.Qobj here...

a-corni: We are in pulser-core

a-corni: Okay for the detach

HGSilveri: Ah of course, I forgot about that

HGSilveri: To avoid clashing with other classes in the future, let's relax the exception from `AttributeError` to `Exception`

```suggestion
            except Exception:
```

HGSilveri (approved):
LGTM, thanks @a-corni !

PR #1010, opened 2026-02-06
Preserve observables' UUID in (de)serialization and Results.aggregate

### LINKED ISSUE #1009: Preserve observable's UUID in Results
Every `Observable` has an UUID attached at instantiation. This was originally added to have a unique correspondence between an observable and its entry in `Results`, since:

1. What an observable measures can depend on its arguments - critically so in state- or operator-dependent observables like `Fidelity` and `Expectation`
2. Robust hashing of any callback/observable was deemed unfeasible

To avoid having the user interact with an observable's UUID directly, methods like `Results.get_result(observable: Observable)` take the observable instance and access its UUID to find the corresponding result (if present) - the idea being that they can use the same object they created and gave to `EmulationConfig` to retrieve the corresponding results, thus avoiding all ambiguity.

## Problem

The problem is that the config's observable's UUID is **being lost** in certain cases:

1. Serialization/deserialization: We currently don't include the original UUID in the serialized object and have the deserialized instance just create a new UUID.

**Solution**: we could simply start preserving the UUID in serialization/deserialization by including it in the observable's abstract repr.

2. Results aggregation: when aggregating multiple results, the tags are preserved but a new UUID is created.

**Solution**: First, we must start ensuring that the observable UUIDs over the multiple results match, not just their tags. Then, we can take the matching UUID instead of creating a new one.

*Side note*: 
The fact that we aggregate by matching only tags can be in itself error prone (e.g. if aggregating by the `fidelity` tag `Results` that actually correspond to different `Fidelity` observables). However, I can also see value in it e.g. when storing results over multiple sessions to later aggregate.
 
To accommodate, I suggest we add a `match_uuids: bool` flag to `Results.aggregate()` that can even be `False` by default but, crucially, **should  be `True` when we automatically aggregate within a backend**.

HGSilveri:
- Updates the observables' JSON schema to support an "uuid" field (not required for backwards compatibility)
- Includes the UUID in the serialization of every observable
- Preserve the UUID in `Results.aggregate` when it is common to all results

Closes #1009

### DIFF
--- a/pulser-core/pulser/backend/results.py
+++ b/pulser-core/pulser/backend/results.py
@@ -273,6 +273,13 @@ class Results:
         where the counters are joined.
         StateResult and EnergyVariance are not supported by default.
 
+        Warning:
+            The ability to access a result from an observable instance
+            (e.g. via Results.get_result(obs)) is only preserved if all
+            aggregated results originated from the same observable instance.
+            When that is not the case, the aggregated result can still
+            be accessed via the observable's tag.
+
         Args:
             results_to_aggregate: The list of Results to aggregate
 
@@ -383,7 +390,14 @@ class Results:
                     "incompatible simulations: "
                     f"the times for `{tag}` are not all the same."
                 )
-            uid = uuid.uuid4()
+
+            _uuids = set(res._tagmap[tag] for res in results_to_aggregate)
+            if len(_uuids) == 1:
+                # Preserve the UUID when all results share the same
+                uid = list(_uuids)[0]
+            else:
+                # Otherwise, create a new one
+                uid = uuid.uuid4()
 
             for t in result_0.get_result_times(tag):
                 v = aggregation_function(
--- a/pulser-core/pulser/backend/observable.py
+++ b/pulser-core/pulser/backend/observable.py
@@ -114,6 +114,7 @@ class Observable(Callback):
             "observable": self._base_tag,
             "evaluation_times": self.evaluation_times,
             "tag_suffix": self._tag_suffix,
+            "uuid": str(self._uuid),
         }
 
     @property
--- a/pulser-core/pulser/json/abstract_repr/backend.py
+++ b/pulser-core/pulser/json/abstract_repr/backend.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import uuid
 from typing import TYPE_CHECKING, Type, TypeVar
 
 from pulser.backend.default_observables import (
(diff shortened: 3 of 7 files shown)


a-quelle-pasqal (approved):
I'm okay with this

a-corni:
Tackles first point of #1009

a-corni (approved):
LGTM

pulser-core/pulser/backend/results.py
```diff
@@ -273,6 +273,13 @@ def aggregate(
         where the counters are joined.
         StateResult and EnergyVariance are not supported by default.
 
+        Warning:
+            The ability to access a result from an observable instance
+            (e.g. via Results.get_result(obs)) is only preserved if all
+            aggregated results originated from the same observable instance.
+            When that is not the case, the aggregated result can still
+            be accessed via the observable's tag.
```
a-corni: I have the feeling this "same" can be a bit misleading:
```suggestion
            The ability to access a result from an observable instance
            (e.g. via Results.get_result(obs)) is only preserved if all
            aggregated results originated from this observable instance `obs`.
            When that is not the case, the aggregated result can still
            be accessed via the observable's tag.
```

pulser-core/pulser/backend/results.py
```diff
@@ -383,7 +390,14 @@ def aggregate(
...
                 )
-            uid = uuid.uuid4()
+
+            _uuids = set(res._tagmap[tag] for res in results_to_aggregate)
+            if len(_uuids) == 1:
+                # Preserve the UUID when all results share the same
+                uid = list(_uuids)[0]
+            else:
+                # Otherwise, create a new one
+                uid = uuid.uuid4()
```
a-corni, line 400: - Should we raise a UserWarning, as mentioned in the docstring ?
- Another option is to aggregate with the uuid of the first observable, which can help for the `get_result(obs)` ?

HGSilveri: - I think a UserWarning is too loud, when a user is manually aggregating results from different emulations they should not expect references via one particular observable instance to work anyway (and the docstring is already there to remind them of this). Worst case, they still try to access via the observable, fail and then just use the tag instead
- That we would introduce the notion that the order of the results to aggregate matters. I don't think it's sound enough to justify the change of behavior, I think in this case not choosing and just going for a new UUID makes more sense

a-corni: - I agree
- I am enclined to say it would be an arrangement. Let's forget this.

a-corni (approved):

a-quelle-pasqal (approved):
LGTM

PR #1011, opened 2026-02-10
Make BackendConfig pickleable again

a-quelle-pasqal:
It needs to have a `__setstate__` method, otherwise unpickling will go into an infinite recursion of `__setattr__` and `__repr__`. This behaviour was introduced with the addition of `__repr__`.

### DIFF
--- a/pulser-core/pulser/backend/config.py
+++ b/pulser-core/pulser/backend/config.py
@@ -138,6 +138,7 @@ class BackendConfig:
         )
 
     def __setstate__(self, d: dict) -> None:
+        # Allows the config to be unpickled
         super().__setattr__("__dict__", d)
 
     def __repr__(self) -> str:


tests/test_backend.py
```diff
@@ -539,6 +540,28 @@ def test_backend_config():
...
+    assert set(cf._backend_options.keys()) == set(
+        new_cf._backend_options.keys()
+    )
+    for k in cf._backend_options.keys():
+        if k != "observables":
+            assert cf._backend_options[k] == new_cf._backend_options[k]
+        else:
+            assert set(o.uuid for o in cf._backend_options[k]) == set(
+                o.uuid for o in new_cf._backend_options[k]
+            )
```
HGSilveri, line 562: nit: I feel like you could just check that the backend options match exactly _and_ that the observables UUIDs were kept.
Also, I would expect the order to be kept too so we can take the check further (though this is only relevant if there were more than one observable in the config)


```suggestion
    assert cf._backend_options == new_cf._backend_options
    assert all(obs.uuid == new_obs.uuid for obs, new_obs in zip(cf.observables, new_cf.observables))
```

pulser-core/pulser/backend/config.py
```diff
@@ -137,6 +137,9 @@ def __setattr__(self, name: str, value: Any) -> None:
             "desired changes."
         )
 
+    def __setstate__(self, d: dict) -> None:
+        super().__setattr__("__dict__", d)
```
HGSilveri, line 142: nit: Can we add a comment for this?

```suggestion
        # Allows the config to be unpickled
        super().__setattr__("__dict__", d)
```

HGSilveri (approved):
This LGTM, I only have a couple of nits

tests/test_backend.py
a-quelle-pasqal: I tried `assert cf._backend_options == new_cf._backend_options` but it doesn't pass since the objects in the `observables` fields are not identical, so I had to resort to either this or implementing equality in `Observable`.

Issue #1012, opened 2026-02-13
Complement the `MinQubitNumberError` message


HGSilveri:
The `MinQubitNumberError` [message](https://github.com/pasqal-io/Pulser/blob/develop/pulser-core/pulser/exceptions/sequence.py#L117) shouldinform the user that it is still possible to create arbitrarily small registers _if_ the layout has exactly the minimum of traps allowed (i.e. that `min_layout_filling_fraction` is not enforced in this case).

PR #1013, opened 2026-02-13
fix(exceptions): complement MinQubitNumberError message

facepunchh:
#1012

### DIFF
--- a/pulser-core/pulser/exceptions/sequence.py
+++ b/pulser-core/pulser/exceptions/sequence.py
@@ -131,7 +131,9 @@ class MinQubitNumberError(QubitsNumberError):
             "device's minimum layout filling fraction, the given"
             f" register has too few qubits ({self.invalid}). "
             "On this device, this layout must hold at least "
-            f"{self.min} qubits."
+            f"{self.min} qubits. Note that arbitrarily small "
+            "registers can still be created if the layout has "
+            "exactly the minimum number of traps allowed."
         )
 
 


pulser-core/pulser/exceptions/sequence.py
```diff
@@ -131,7 +131,9 @@ def __str__(self) -> str:
             "device's minimum layout filling fraction, the given"
             f" register has too few qubits ({self.invalid}). "
             "On this device, this layout must hold at least "
-            f"{self.min} qubits."
+            f"{self.min} qubits. Note that arbitrarily small "
+            "registers can still be created if the layout has "
+            "exactly the minimum number of traps allowed."
```
a-corni, line 136: Including the minimum number of traps allowed would be a good additionnal information, for users to fix their layout.

```suggestion
class MinQubitNumberError(QubitsNumberError):
"""Too few qubits for the layout.

   Attributes:
       invalid: The invalid number of qubits.
       min_qubits: The minimum number of qubits.
       min_traps: The minimum number of traps.
   """

invalid: int
min_qubits: int
min_traps: int

def __str__(self) -> str:
return (
"Given the number of traps in the layout and the "
"device's minimum layout filling fraction, the given"
f" register has too few qubits ({self.invalid}). "
"On this device, this layout must hold at least "
            f"{self.min_qubits} qubits. Note that arbitrarily small "
            "registers can still be created if the layout has "
            f"exactly the minimum number of traps allowed: {self.min_traps}."
```

You would need to also include this min_traps when [raising the error](https://github.com/pasqal-io/Pulser/blob/develop/pulser-core/pulser/devices/_device_datacls.py#L502) (min_traps=self.min_layout_traps)

HGSilveri:
Hey @facepunchh , this has been pending for a while. Are you planning on picking it back up soon? If not, we will take the liberty to finish it

HGSilveri:
Addressed in #1048

PR #1018, opened 2026-02-28
Rename Device.default_noise_model to Device.noise_model (#1014)

### LINKED ISSUE #1014: Rename `Device.default_noise_model` to `Device.noise_model`
The change should be fully backwards compatible, i.e.
- Initializing with `Device(..., default_noise_model=...,...)` should still work (with a deprecation warning)
- Accessing `Device.default_noise_model` should still work (by making it a property)
- (De)serialization should be handled appropriately - ideally avoiding changes to the JSON schema

jose-compu:
Closes #1014

Renames `Device.default_noise_model` to `Device.noise_model` while keeping full backward compatibility:

- **New API:** `noise_model` is the primary attribute and constructor parameter
- **Init:** `Device(..., default_noise_model=...)` still works, with a deprecation warning
- **Access:** `device.default_noise_model` still works via a deprecated property that returns `noise_model`
- **Serialization:** JSON still uses the `default_noise_model` key to avoid schema changes
- **Deserialization:** Supports both `default_noise_model` and `noise_model` in the JSON input

### DIFF
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -448,24 +448,24 @@ def test_emulator_backend(sequence):
         my_param="bar",
     )
 
-    with pytest.warns((UserWarning, DeprecationWarning)) as record:
+    with pytest.warns(
+        UserWarning,
+        match="'sequence.device.noise_model.runs=3' is being "
+        "ignored; 'config.n_trajectories=40' will be used instead",
+    ):
         _config = EmulationConfig(
             observables=(BitStrings(),), prefer_device_noise_model=True
         )
         assert _config.n_trajectories == 40
-        _device = dataclasses.replace(
-            sequence.device,
-            noise_model=pulser.NoiseModel(amp_sigma=0.1, runs=3),
-        )
+        with pytest.deprecated_call():
+            _device = dataclasses.replace(
+                sequence.device,
+                noise_model=pulser.NoiseModel(amp_sigma=0.1, runs=3),
+            )
         ConcreteEmulator(
             pulser.Sequence(sequence.register, _device),
             config=_config,
         )
-    assert any(
-        "sequence.device.noise_model.runs=3" in str(w.message)
-        for w in record
-        if w.category is UserWarning
-    )
 
     emu = ConcreteEmulator(sequence, config=concrete_config)
 


pulser-core/pulser/devices/_device_datacls.py
```diff
@@ -795,6 +809,30 @@ def _specs(self, for_docs: bool = False) -> str:
         )
 
 
+def _wrap_init_for_default_noise_model(original_init):
+    """Wrap __init__ to accept deprecated default_noise_model parameter."""
+
+    def wrapped_init(self, *args, default_noise_model=None, **kwargs):
+        if default_noise_model is not None:
+            if kwargs.get("noise_model") is not None:
+                raise TypeError(
```
a-corni: Isn't it more a ValueError than a TypeError ? :)

pulser-core/pulser/devices/_device_datacls.py
```diff
@@ -795,6 +809,30 @@ def _specs(self, for_docs: bool = False) -> str:
...
+                    "Cannot specify both 'noise_model' and 'default_noise_model'"
+                )
+            warnings.warn(
+                "'default_noise_model' is deprecated, use 'noise_model' instead.",
```
a-corni: nit: We usually like to specify the version from which something was deprecated

```suggestion
                "'default_noise_model' is deprecated since pulser 1.8.0, use 'noise_model' instead.",
```

pulser-core/pulser/devices/_device_datacls.py
```diff
@@ -795,6 +809,30 @@ def _specs(self, for_docs: bool = False) -> str:
         )
 
 
+def _wrap_init_for_default_noise_model(original_init):
+    """Wrap __init__ to accept deprecated default_noise_model parameter."""
+
+    def wrapped_init(self, *args, default_noise_model=None, **kwargs):
```
HGSilveri: Let's use @wraps here to preserve the original signature

```suggestion
    @functools.wraps(original_init)
    def wrapped_init(self, *args, default_noise_model=None, **kwargs):
```

pulser-core/pulser/json/abstract_repr/deserializer.py
```diff
@@ -519,8 +523,14 @@ def _deserialize_device_object(obj: dict[str, Any]) -> Device | VirtualDevice:
...
-        elif param.name == "default_noise_model":
-            params[param.name] = _deserialize_noise_model(obj[param.name])
+        elif param.name == "noise_model":
+            # JSON schema uses default_noise_model for compatibility
+            json_key = (
+                "default_noise_model"
+                if "default_noise_model" in obj
+                else "noise_model"
+            )
+            params["noise_model"] = _deserialize_noise_model(obj[json_key])
```
HGSilveri: We're not changing the JSON schema so the `json_key` will remain `default_noise_model`

tests/test_backend.py
HGSilveri, line 1: I'm failing to see the need for the changes on this file (apart from the renaming of `default_noise_model -> noise_model`). Could you please explain?

pulser-core/pulser/json/abstract_repr/deserializer.py
jose-compu: okey fixed here: https://github.com/pasqal-io/Pulser/pull/1018/changes/07b7d3d9caa0743a517f5192fc52329558551019

tests/test_backend.py
jose-compu: The only change beyond the rename is the `warnings.catch_warnings()` block. The test builds a device with `noise_model=pulser.NoiseModel(amp_sigma=0.1, runs=3)`. In current Pulser, using `NoiseModel(..., runs=...)` triggers a `DeprecationWarning` (“NoiseModel.runs is deprecated”). The test is written to assert only on the `UserWarning` about the device noise model’s runs being ignored. Without filtering out that deprecation, the first warning seen is the `DeprecationWarning`, so `pytest.warns(UserWarning, ...)` fails. The filter is there so we can still check the intended `UserWarning` while the `runs` deprecation exists.

If you’d rather not add this, we can drop the filter and relax the test (e.g. allow multiple warnings or assert on the presence of the UserWarning among them).

pulser-core/pulser/devices/_device_datacls.py
jose-compu: ok change here, thanks https://github.com/pasqal-io/Pulser/pull/1018/changes/ada0b86720eea56ca949fa907eda76f09654afc0

tests/test_backend.py
HGSilveri: I think it would be better to just do the argument renaming and leave everything else untouched.

pulser-core/pulser/devices/_device_datacls.py
```diff
@@ -795,6 +810,35 @@ def _specs(self, for_docs: bool = False) -> str:
...
+
+    @functools.wraps(original_init)
+    def wrapped_init(
+        self: Any, *args: Any, default_noise_model: Any = None, **kwargs: Any
+    ) -> None:
+        if default_noise_model is not None:
+            if kwargs.get("noise_model") is not None:
+                raise ValueError(
+                    "Cannot specify both 'noise_model' and 'default_noise_model'"
+                )
```
HGSilveri, line 827: Could you please add a simple test for this case?

jose-compu: adding test case here https://github.com/pasqal-io/Pulser/pull/1018/commits/776ad5e9209611e2e7721f79ea6e8b1e89c7b2f0

tests/test_backend.py
jose-compu: okey done that I understood properly here https://github.com/pasqal-io/Pulser/pull/1018/changes/c5aaf0483ae9127c3d03b5711bd8c92812b1e4d8

tests/test_backend.py
```diff
@@ -448,24 +448,24 @@ def run(self):
...
-        UserWarning,
-        match="'sequence.device.default_noise_model.runs=3' is being "
-        "ignored; 'config.n_trajectories=40' will be used instead",
-    ):
+    with pytest.warns((UserWarning, DeprecationWarning)) as record:
         _config = EmulationConfig(
             observables=(BitStrings(),), prefer_device_noise_model=True
         )
         assert _config.n_trajectories == 40
-        with pytest.deprecated_call():
```
HGSilveri, line 460: Are you not having to catch both `(UserWarning, DeprecationWarning)` because you removed this context manager? 
If you had kept it, the outer `pytest.warns()` could stay dedicated to the `UserWarning`

tests/test_backend.py
HGSilveri: Hum, I still see other changes that I think were needed because `with pytest.deprecated_call():` was deleted.

What I would like to make sure of is the following: if you take this test file in its original state (i.e. without any changes) and just rename `default_noise_model -> noise_model`, does everything pass without warnings? If my understanding is correct, it should

HGSilveri:
Hey @jose-blockchain , I see that this has been pending for a bit... 

From my point-of-view the PR is basically ready, it won't be a nit in a test that stops us from merging. I only insisted on that point because I felt there was some misunderstanding between us that I would rather not just leave unresolved if I could.

That being said, could you please let me know if you are still planning on picking it up? If not, I'll just go ahead and merge it, no problem

jose-compu:
@HGSilveri you can merge if you want or propose any changes in the PR and I will acept them, I dont recall much the final part of the conversation.

HGSilveri:
> @HGSilveri you can merge if you want or propose any changes in the PR and I will acept them, I dont recall much the final part of the conversation.

Sounds good, I'll test what I had in mind then. Thanks!

PR #1021, opened 2026-03-03
Fixes to `QutipConfig` and `QutipBackendV2`

### LINKED ISSUE #1020: Error when evaluation_times is an array
I am following the tutorial [Results and observables](https://pulser.readthedocs.io/en/latest/results.html) and I get an error if `evaluation_times` is a numpy array. This arises because of a boolean operztion `or`, which I think tries to check if `evaluation_times` is `None`, but `or` does not work well with arrays.

```python
ValueError                                Traceback (most recent call last)
Cell In[52], line 3
      1 dt = 0.001
      2 evaluation_times = np.linspace(0.0, 1.0, int(1 / dt + 1))
----> 3 emu_backend = emu_backend_class(seq, config=config)
      4 results = emu_backend.run()

File ~/.virtualenvs/qic/lib/python3.12/site-packages/pulser_simulation/qutip_backend.py:156, in QutipBackendV2.__init__(self, sequence, config, mimic_qpu)
    146 noise_model = noise_model or self._config.noise_model
    147 self._sim_obj = QutipEmulator.from_sequence(
    148     sequence,
    149     sampling_rate=self._config.sampling_rate,
   (...)
    153     n_trajectories=self._config.n_trajectories,
    154 )
    155 self._sim_obj.set_evaluation_times(
--> 156     self._config._get_legacy_evaluation_times(
    157         self._sim_obj.total_duration_ns
    158     ),

File ~/.virtualenvs/qic/lib/python3.12/site-packages/pulser_simulation/qutip_config.py:161, in QutipConfig._get_legacy_evaluation_times(self, total_duration_ns)
    159     return "Full"
    160 for obs in self.observables:
--> 161     extra_eval_times.update(obs.evaluation_times or [])
    163 rel_eval_times = self.default_evaluation_times
    164 if extra_eval_times:

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

HGSilveri:
- Support arrays as evaluation times in `QutipConfig.default_evaluation_times` and in its observables
- Exhaustively cast the `solver` argument to a `Solver` instance 
- [FIX] Rounding error in `QutipEmulator` causes repetition in evaluation times (#1022)


pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -152,7 +152,7 @@ def __init__(
         self._sampling_rate = sampling_rate
         device.validate_register(register)
         self._register = register
-        self.solver = solver
+        self.solver = Solver(solver)
```
murogrande, line 155: here if the solver is not in the `enum` options, the error will come from the enum but in `qutip_config` you raise a value error. I think the user is going to see a different error depending on the entry point.

HGSilveri: That is correct. `QutipEmulator` is not the official entry point and we want to deprecate it, so I figured that would be ok. The error message is actually quite informative already, I just made something custom in `QutipConfig` to inform the user of the available options

pulser-simulation/pulser_simulation/qutip_config.py
```diff
@@ -158,11 +167,12 @@ def _get_legacy_evaluation_times(
         if self.callbacks:
             return "Full"
         for obs in self.observables:
-            extra_eval_times.update(obs.evaluation_times or [])
+            if obs.evaluation_times is not None:
+                extra_eval_times.update(obs.evaluation_times)
```
murogrande, line 177: I think in `observables.py`, lines 102-104, you need to update the `evaluation_times`, doing the following:

`self.evaluation_times = self._validate_eval_times(evaluation_times) if evaluation_times is not
  None else None`

pulser-simulation/pulser_simulation/qutip_config.py
```diff
@@ -118,6 +118,15 @@ def __init__(
                 stacklevel=2,
             )
 
+        try:
```
elmerhej, line 127: I don't like the `try except` expression. This could be simply replaced by something more natural
`if not isinstance(solver, Solver):`
    `solver = Solver(solver)`

the downside is you lose the custom error message, so it's really up to you

pulser-simulation/pulser_simulation/qutip_config.py
```diff
@@ -158,11 +167,12 @@ def _get_legacy_evaluation_times(
...
             return "Full"
         for obs in self.observables:
-            extra_eval_times.update(obs.evaluation_times or [])
+            if obs.evaluation_times is not None:
+                extra_eval_times.update(obs.evaluation_times)
 
         rel_eval_times = self.default_evaluation_times
         if extra_eval_times:
-            if rel_eval_times == "Full":
+            if isinstance(rel_eval_times, str) and rel_eval_times == "Full":
```
a-quelle-pasqal, line 181: sure, I guess, but doesn't `EmulationConfig` already ensure that no strings other than `"Full"` go through?

pulser-simulation/pulser_simulation/qutip_config.py
a-quelle-pasqal: It's true that this would make the usage of `validate_eval_times` more uniform, since it does return a value for a reason.

pulser-simulation/pulser_simulation/qutip_config.py
```diff
@@ -118,6 +118,15 @@ def __init__(
                 stacklevel=2,
             )
 
+        try:
+            solver = Solver(solver)
```
a-quelle-pasqal, line 128: If I'm reading between the lines, the problem you're facing is that during deserialization, this code is called with a `str` for `solver`. Could you explain a bit via what code it ends up here? I'm wondering whether this is the best place to fix the issue. The type hint `solver : Solver` is quite explicit, and that the deserialization mechanism breaks the type hint is heinous.

pulser-simulation/pulser_simulation/qutip_config.py
HGSilveri: The issue with that is that string arguments (e.g. `solver="default"`) are valid but will fail the `isinstance` check (this was actually the issue with the old implementation). 
However, I can replace it with `solver in Solver` and that should work for all cases.

pulser-simulation/pulser_simulation/qutip_config.py
HGSilveri: Yes, but the `== "Full"` check is problematic when `rel_eval_times` is an array because it does an element-wise comparison

pulser-simulation/pulser_simulation/qutip_config.py
HGSilveri: To avoid having the user import the `Solver` class, I believe Kemal's idea was that we would support the corresponding strings as arguments (hence why Solver is a `(str, Enum)` subclass). 
I agree that it is still a type violation, but in that case we can instead relax the type hint of the argument to be
`solver: Solver | Literal["default", "MasterEquation", "MonteCarlo"]`

pulser-simulation/pulser_simulation/qutip_config.py
HGSilveri: No problem

pulser-simulation/pulser_simulation/qutip_config.py
HGSilveri: And you are right, the issue was that the `Solver` is turned into a `str` upon serialization and it was then not handled correctly in deserialization. 
I think it's nice to support the strings (+ it spares us from writing custom deserialization logic), so I went with that option.

pulser-simulation/pulser_simulation/qutip_config.py
HGSilveri: Ahh, looks like `solver in Solver` only works for Python >= 3.12. 
I'll have to go back to the try-expect block, sorry @elmerhej

elmerhej: No worries. Thanks for spending some time to check it! 
I'll resolve my comment 👍

a-quelle-pasqal (approved):
LGTM

PR #1022, opened 2026-03-05
[FIX] Rounding error in `QutipEmulator` causes repetition in evaluation times

HGSilveri:
- Rounding errors in [this line](https://github.com/pasqal-io/Pulser/blob/develop/pulser-simulation/pulser_simulation/simulation.py#L748) sporadically caused a repetition in the evaluation times of the results, resulting in an error from trying to save the same observable twice (see the test introduced in the  first commit)
- This changes the calculation to avoid the rounding error (second commit)

@pasqal-io/emulation-team We could also implement something more robust to make sure we never hit this `RuntimeError` again (e.g. keeping track of what evaluation times have been stored or add a `skip_if_existing` flag somewhere).

### DIFF
--- a/pulser-simulation/pulser_simulation/simulation.py
+++ b/pulser-simulation/pulser_simulation/simulation.py
@@ -738,7 +738,7 @@ class QutipEmulator:
                 self._meas_basis,
                 state,
                 self._meas_basis in self.basis_name,
-                evaluation_time=t / self._tot_duration * 1e3,
+                evaluation_time=t / (self._tot_duration * 1e-3),
             )
             for state, t in zip(result.states, self._eval_times_array)
         ]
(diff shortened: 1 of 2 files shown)


pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -738,7 +738,7 @@ def _run_solver(
                 self._meas_basis,
                 state,
                 self._meas_basis in self.basis_name,
-                evaluation_time=t / self._tot_duration * 1e3,
+                evaluation_time=t / (self._tot_duration * 1e-3),
```
a-quelle-pasqal, line 741: I'm not sure this fixes the issue. See
```
>>> a = 1 + 1e-15
>>> a
1.000000000000001
>>> a*1e-5*1e5 - a
2.220446049250313e-16
```
Or maybe I'm misunderstanding what the source of the problem is.

a-quelle-pasqal (approved):
LGTM

PR #1030, opened 2026-03-18
Use new noise trajectories on each QutipBackendV2 run

### LINKED ISSUE #1028: Recreate the noise trajectories on every call to `QutipBackendV2.run()`
Currently, the noise trajectories are created once in `QutipBackendV2.__init__()` and kept unchanged, leading to identical results on every call to `QutipBackendV2.run()`. This is error-prone and a mismatch with other emulator backends, so it should be changed.

a-quelle-pasqal:
Currently, the trajectories for noisy runs are computed on the initialization of `QutipEmulator`, which means that for `QutipEmulator`, `QutipBackend` and `QutipBackendV2`, calling run repeatedly on the same object will yield the same result each time. To get different noise trajectories, you have to recreate the backend. I've changed the code so that on subsequent calls to run, new noise trajectories are computed (if needed).

### DIFF
--- a/pulser-simulation/pulser_simulation/simulation.py
+++ b/pulser-simulation/pulser_simulation/simulation.py
@@ -204,6 +204,7 @@ class QutipEmulator:
         if not noise_model:
             noise_model = NoiseModel()
 
+        self.noise_trajectories_used = False
         self._hamiltonian_data = HamiltonianData(
             self.samples_obj,
             register,
@@ -379,6 +380,7 @@ class QutipEmulator:
         former_dim = self.dim
         former_basis = self.basis
         noise_model = cfg.to_noise_model()
+        self.noise_trajectories_used = False
         self._hamiltonian_data = HamiltonianData(
             self.samples_obj,
             self._register,
@@ -872,6 +874,18 @@ class QutipEmulator:
     ) -> Iterator[tuple[SimulationResults, int]]:
         n_trajectories = self.n_trajectories
         traj_nb = 0
+        # When run is called repeatedly, we want different noise trajectories
+        # to be used each time. After the first time, create new trajectories
+        if self.noise_trajectories_used:
+            noise_model = self._hamiltonian_data.noise_model
+            self._hamiltonian_data = HamiltonianData(
+                self.samples_obj,
+                self._register,
+                self.device,
+                noise_model,
+                self._get_n_trajectories(noise_model, check_value=True),
+            )
+        self.noise_trajectories_used = True
         for ham, reps in self._hamiltonians:
             if print_progress:
                 if reps == 1:
--- a/pulser-simulation/pulser_simulation/simulation.py
+++ b/pulser-simulation/pulser_simulation/simulation.py
@@ -204,7 +204,7 @@ class QutipEmulator:
         if not noise_model:
             noise_model = NoiseModel()
 
-        self.noise_trajectories_used = False
+        self._noise_trajectories_used = False
         self._hamiltonian_data = HamiltonianData(
             self.samples_obj,
             register,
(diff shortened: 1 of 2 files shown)


pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -204,6 +204,7 @@ def __init__(
         if not noise_model:
             noise_model = NoiseModel()
 
+        self.noise_trajectories_used = False
```
HGSilveri: Can we make this a private attribute?

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -872,6 +874,18 @@ def _noisy_runs(
...
+        # to be used each time. After the first time, create new trajectories
+        if self.noise_trajectories_used:
+            noise_model = self._hamiltonian_data.noise_model
+            self._hamiltonian_data = HamiltonianData(
+                self.samples_obj,
+                self._register,
+                self.device,
+                noise_model,
+                self._get_n_trajectories(noise_model, check_value=True),
+            )
```
HGSilveri, line 887: I think it might be preferable to encapsulate this in a `_generate_hamiltonian_data()` method and replace every explict assignment to `self._hamiltonian_data` with it. In this method, you also systematically set `self._noise_trajectories_used = False`.

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -872,6 +874,18 @@ def _noisy_runs(
...
+        self.noise_trajectories_used = True
```
HGSilveri: nit:
```suggestion
        # Make sure that the noise trajectories are refreshed next time run is called
        self.noise_trajectories_used = True
```

HGSilveri (commented):
Thanks for this @a-quelle-pasqal , I have only a couple suggestions

pulser-simulation/pulser_simulation/simulation.py
a-quelle-pasqal: I tried to do this but I ran into a slightly surprising issue, `register` and `device` are `@property`s and those apparently are not available in `init` yet. As a result, the only place where this utility function would be used is in `noisy_runs`, and in that case I think it's not worth it.

HGSilveri: How about making them arguments of the methods?

HGSilveri: Is `_generate_hamiltonian_data(seq_samples: SequenceSamples, register: Register, device: Device, noise_model: NoiseModel)` too much?

a-quelle-pasqal: It's not really shorter than just calling the constructor, is it? There is only one location where both the constructor and `self._noise_trajectories_used = False` needs to be called, in one other place it doesn't matter, and in a third place it has to be only the constructor, so the original idea was also not valid. At this point it's crossed the point of diminishing returns I think.

HGSilveri: I just wanted to make sure that whenever the `noise_trajectories` were refreshed, `_noise_trajectories_used` was set to False. 
On a related note, did you consider just updating the noise trajectories in place (i.e. not recreating the full `HamiltonianData` each time)?

a-quelle-pasqal: yes I did, but I feel it would reintroduce the same in-place updating behaviour from before that I strove so hard to get rid of. Ideally, I'd like to organize the `QutipEmulator` so that `HamiltonianData` is only created in `run`, but that would break some of the existing getters, and I felt this was the best compromise without taking on the burden of a huge refactor.

HGSilveri: Ok, fair enough, I give up 😅

HGSilveri (approved):

PR #1031, opened 2026-03-19
Normalize input state of observables in QutipBackendV2

### LINKED ISSUE #1029: `Fidelity` observable can return a value greater than 1
We should make sure that fidelity values are always `<=1`.

Here is a MWE of a case where fidelity comes out greater than one:

[fidely_above_1_mwe.py](https://github.com/user-attachments/files/26021704/fidely_above_1_mwe.py)

a-quelle-pasqal:
The ODE solver used by qutip does not preserve state norm analytically. As a result, the norm of the state can be observed to deviate from 1 by as much as 1e-4. Consequently, it is possible to measure `Fidelity` values larger than 1 (see the added unit test). I'm fixing the issue by feeding a normalized copy of the time-evolved state into the observables instead. I'm making a copy rather than normalizing in-place because I don't want the behaviour of the solver to depend on the evaluation times.

### DIFF
--- a/pulser-simulation/pulser_simulation/qutip_backend.py
+++ b/pulser-simulation/pulser_simulation/qutip_backend.py
@@ -223,6 +223,7 @@ class QutipBackendV2(EmulatorBackend):
                         state = QutipState(
                             qutip_res.state, eigenstates=eigenstates
                         )
+                        state._state /= state._state.norm()
                         ham = QutipOperator(
                             self._sim_obj._get_noiseless_hamiltonian(
                                 self._config.noise_model.with_leakage
--- a/pulser-simulation/pulser_simulation/qutip_backend.py
+++ b/pulser-simulation/pulser_simulation/qutip_backend.py
@@ -223,7 +223,7 @@ class QutipBackendV2(EmulatorBackend):
                         state = QutipState(
                             qutip_res.state, eigenstates=eigenstates
                         )
-                        state._state /= state._state.norm()
+                        state._state = state._state / state._state.norm()
                         ham = QutipOperator(
                             self._sim_obj._get_noiseless_hamiltonian(
                                 self._config.noise_model.with_leakage
--- a/pulser-simulation/pulser_simulation/qutip_backend.py
+++ b/pulser-simulation/pulser_simulation/qutip_backend.py
@@ -183,7 +183,9 @@ class QutipBackendV2(EmulatorBackend):
 
             for qutip_res in single_res:
                 t = qutip_res.evaluation_time
-                state = QutipState(qutip_res.state, eigenstates=eigenstates)
+                state = QutipState(
+                    qutip_res.state.unit(), eigenstates=eigenstates
+                )
                 ham: QutipOperator = QutipOperator(
                     self._sim_obj._get_noiseless_hamiltonian(
                         self._config.noise_model.with_leakage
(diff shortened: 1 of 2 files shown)


elmerhej (commented):
LGTM

pulser-simulation/pulser_simulation/qutip_backend.py
```diff
@@ -223,6 +223,7 @@ def run(self) -> Results:
                         state = QutipState(
                             qutip_res.state, eigenstates=eigenstates
                         )
+                        state._state = state._state / state._state.norm()
```
HGSilveri: Is this not an issue in the absence of stochastic noise?
Also, you could instead just call `qutip_res.state.unit()` when you create `QutipState` above

a-quelle-pasqal: thanks for noticing this. I had to fiddle with the test a bit to get the issue to reproduce without amp_sigma, but I managed. I also normalize on the noiseless path now, and I parametrized the new test to cover both cases.

tests/pulser_simulation/test_qutip_backend_v2.py
```diff
@@ -481,3 +481,53 @@ def test_rounding_error_eval_time_duplication():
...
+                total_duration,
+                U * np.array([-1, 0.0556, 0.332, 1]) * factor,
+                times=interp_pts,
+            ),
+            0,
+        ),
+        "rydberg_global",
+    )
+
+    # Start from the default_config and add the Fidelity observable
```
HGSilveri: You are not adding the Fidelity observable though

tests/pulser_simulation/test_qutip_backend_v2.py
```diff
@@ -481,3 +481,53 @@ def test_rounding_error_eval_time_duplication():
...
+    default_config = QutipBackendV2.default_config
+    np.random.seed(1234)
+    config = default_config.with_changes(
+        noise_model=pulser.NoiseModel(amp_sigma=amp_sigma)
+    )
+    qutip_bknd_custom = QutipBackendV2(seq, config=config)
+    results = qutip_bknd_custom.run()
+    assert results.final_state._state.norm() < 1 + 1e-8
```
HGSilveri, line 535: So you still can't guarantee that the norm is <= 1?

tests/pulser_simulation/test_qutip_backend_v2.py
```diff
@@ -481,3 +481,53 @@ def test_rounding_error_eval_time_duplication():
     # It works because the rounding error was fixed
     emu_backend = QutipBackendV2(seq, config=config)
     emu_backend.run()
+
+
+@pytest.mark.parametrize("amp_sigma", [0.0, 0.5])
+def test_output_state_normalization(amp_sigma):
+    factor = 1.2357175818662465 if not amp_sigma else 1.0
```
HGSilveri, line 490: I'm assuming that this factor makes a state with norm > 1 when not normalized, right? Can you add a comment stating that?

tests/pulser_simulation/test_qutip_backend_v2.py
a-quelle-pasqal: nope, floats are fun.

HGSilveri (commented):
The changes LGTM. However, since the original issues was that Fidelity values were sometimes > 1, it would be nice to have a test somewhere explicitly checking that no longer happens

tests/pulser_simulation/test_qutip_backend_v2.py
a-quelle-pasqal: I changed the comment

tests/pulser_simulation/test_qutip_backend_v2.py
a-quelle-pasqal: I added the norm that it gave before the fix in the comment. It is $1+10^{-5}$. Whether it should be a source of existential dread that a local error of $10^{-8}$ can blob up to a global error of $10^{-5}$, or possibly even larger, since this is just the norm, I will leave up to you. One possible thing we could consider elsewhere, is to make the local error configurable, it is an argument in the solvers I believe.

HGSilveri (approved):
Thanks for indulging me @a-quelle-pasqal , this LGTM

PR #1046, opened 2026-04-16
Updates to Local Addressability features

### LINKED ISSUE #1027: Add `DMM.min_avg_abs_detuning`
Analogously to [Channel.min_avg_amp](https://github.com/pasqal-io/Pulser/blob/develop/pulser-core/pulser/channels/base_channel.py#L112), we need to set an upper-bound (because the detuning on a DMM is always `<= 0`) on the _average_ detuning applied by the `DMM` on each atom (when different than zero). 

- [x] Add `DMM.top_avg_detuning: float = 0`
- [x] Update `DMM.validate_pulse()` to check that $(\min_i \epsilon_i) \cdot \overline{\delta_{DMM}} \leq$ `top_avg_detuning`  
- [x] Add parameter to the JSON schema and support its (de)serialization backwards-compatibly

### LINKED ISSUE #1040: Weigth Maps hash needs float precision for the weights
Currently, the hash in WeightMaps is computed from its position and weights. To compare two positions, its coordinates are rounded to 1e-6. We don't have this rounding for weights, which exposes two DetuningMaps to be identified as different due to floating point errors. Comparing weights with a precision of 1e-6 should be good enough.

HGSilveri:
- Hides irrelevant parameters from the `DMM` repr
- Changes `Sequence.config_detuning_map()` to default to choosing the first available DMM channel when `dmm_id` is not given
- Adds `DMM.min_avg_abs_detuning` (Closes #1027 )
- Improves checks on the DMM and DetuningMap
- Add WEIGHT_PRECISION to weights in a WeightMap (Closes #1040 )

### DIFF
--- a/pulser-core/pulser/channels/dmm.py
+++ b/pulser-core/pulser/channels/dmm.py
@@ -189,7 +189,9 @@ class DMM(Channel):
         min_non_zero_weight = np.min(
             # Can't be empty because the WeightMap enforces having at least one
             # non-zero weight
-            np.array(detuning_map.weights)[np.nonzero(detuning_map.weights)]
+            np.array(detuning_map.weights)[
+                np.nonzero(detuning_map.weights)  # type: ignore[arg-type]
+            ]
         )
         if (
             0
--- a/pulser-core/pulser/channels/dmm.py
+++ b/pulser-core/pulser/channels/dmm.py
@@ -25,7 +25,7 @@ from pulser.json.utils import get_dataclass_defaults
 from pulser.pulse import Pulse
 from pulser.register.weight_maps import DetuningMap
 
-OPTIONAL_ABSTR_DMM_FIELDS = ["total_bottom_detuning"]
+OPTIONAL_ABSTR_DMM_FIELDS = ["total_bottom_detuning", "min_avg_abs_detuning"]
 
 
 @dataclass(init=True, frozen=True)
@@ -46,10 +46,13 @@ class DMM(Channel):
         "no-delay".
 
     Args:
-        bottom_detuning: Minimum possible detuning per atom (in rad/µs),
-            must be below zero.
-        total_bottom_detuning: Minimum possible detuning distributed on all
-            atoms (in rad/µs), must be below zero.
+        bottom_detuning: Minimum possible detuning per detuning map spot
+            (in rad/µs); must be below zero.
+        total_bottom_detuning: Minimum possible detuning distributed over all
+            detuning map spots (in rad/µs); must be below zero.
+        min_avg_abs_detuning: The minimum acceptable value for the average
+            detuning (in absolute value and in rad/µs) applied on any detuning
+            map spot (when not 0). Defaults to 0.
         clock_period: The duration of a clock cycle (in ns). The duration of a
             pulse or delay instruction is enforced to be a multiple of the
             clock cycle.
@@ -62,6 +65,7 @@ class DMM(Channel):
 
     bottom_detuning: float | None = None
     total_bottom_detuning: float | None = None
+    min_avg_abs_detuning: float = 0.0
     addressing: Literal["Global"] = field(
         default="Global", init=False, repr=False
     )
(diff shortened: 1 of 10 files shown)


pulser-core/pulser/channels/dmm.py
```diff
@@ -46,45 +46,84 @@ class DMM(Channel):
...
 
     Args:
-        bottom_detuning: Minimum possible detuning per atom (in rad/µs),
-            must be below zero.
-        total_bottom_detuning: Minimum possible detuning distributed on all
-            atoms (in rad/µs), must be below zero.
+        bottom_detuning: Minimum possible detuning per detuning map spot
+            (in rad/µs); must be below zero.
+        total_bottom_detuning: Minimum possible detuning distributed over all
+            detuning map spots (in rad/µs); must be below zero.
```
a-corni, line 54: I see the point with detuning map spot, but I think we need to introduce it above (+ I am spotting a possible mistake in the main description):

```suggestion
"""Defines a Detuning Map Modulator (DMM) Channel.

   A Detuning Map Modulator can be used to define `Global` detuning Pulses
   (of zero amplitude and phase). These detuning Pulses are locally modulated by the
   weights of a `DetuningMap`, such that qubits experience a detuning map spot ie a detuning pulse equal to (detuning map weight on this qubit)*(detuning pulse value). That way, it provid a local control over the
   detuning. The detuning of the pulses added to a DMM has to be negative, such that the detuning map spots are
   between 0 and `bottom_detuning`, and that the sum of the detuning map spots is below `total_bottom_detuning`. This Channel targets
   the transition between the ground and rydberg states, thus encoding the
   'ground-rydberg' basis.

   Note:
       The protocol to add pulses to the DMM Channel is by default
       "no-delay".

   Args:
        bottom_detuning: Minimum possible detuning per detuning map spot
            (in rad/µs); must be below zero.
        total_bottom_detuning: Minimum possible sum of the detuning map spots (in rad/µs); must be below zero.
```

pulser-core/pulser/channels/dmm.py
```diff
@@ -46,45 +46,84 @@ class DMM(Channel):
...
+        min_avg_abs_detuning: The minimum acceptable value for the average
+            detuning (in absolute value and in rad/µs) applied on any detuning
+            map spot (when not 0). Defaults to 0.
         clock_period: The duration of a clock cycle (in ns). The duration of a
```
a-corni, line 58:
```suggestion
        min_avg_abs_detuning: The minimum acceptable value for the average
            absolute detuning (in rad/µs) applied on any detuning
            map spot (when not 0). Defaults to 0.
       clock_period: The duration of a clock cycle (in ns). The duration of a
```

pulser-core/pulser/channels/dmm.py
```diff
@@ -46,45 +46,84 @@ class DMM(Channel):
...
             pulse or delay instruction is enforced to be a multiple of the
             clock cycle.
         min_duration: The shortest duration an instruction can take.
         max_duration: The longest duration an instruction can take.
-        min_avg_amp: The minimum average amplitude of a pulse (when not zero).
```
a-corni, line 58: Nice catch ^^'

pulser-core/pulser/channels/dmm.py
```diff
@@ -46,45 +46,84 @@ class DMM(Channel):
...
+                f"(got {self.min_avg_abs_detuning})."
+            )
+        if (
+            self.bottom_detuning
+            and self.min_avg_abs_detuning >= -self.bottom_detuning
+        ):
+            bottom_detuning = self.bottom_detuning
+            raise ValueError(
+                f"'min_avg_abs_detuning' (got {self.min_avg_abs_detuning}) "
+                f"must be lower than or equal to {-bottom_detuning=}."
```
a-corni, line 127: Nice,  didn't know about this ;)

pulser-core/pulser/channels/dmm.py
```diff
@@ -122,25 +161,50 @@ def validate_pulse(
...
-            < self.bottom_detuning
+            and max_weight * min_round_detuning < self.bottom_detuning
         ):
             raise ValueError(
-                "The detunings on some atoms go below the local bottom "
+                f"For a detuning map with a maximum weight of {max_weight},"
+                f" a DMM pulse with minimum detuning {min_round_detuning} "
+                "rad/µs goes below the local bottom "
                 f"detuning of the DMM ({self.bottom_detuning} rad/µs)."
             )
```
a-corni, line 178: What about providing the minimal value above which the detuning should be ?

```suggestion
                f"For a detuning map with a maximum weight of {max_weight},"
                f" the DMM pulses should have a minimum detuning of {self.bottom_detuning/max_weight} rad/µs (got {min_round_detuning} rad/µs)."
)
```

pulser-core/pulser/channels/dmm.py
```diff
@@ -122,25 +161,50 @@ def validate_pulse(
...
-            < self.total_bottom_detuning
+            and sum_weight * min_round_detuning < self.total_bottom_detuning
         ):
             raise ValueError(
-                "The applied detuning goes below the total bottom detuning "
+                "For a detuning map with a total summed weight of "
+                f"{sum_weight}, the total applied detuning from a DMM pulse "
+                f"with minimum detuning {min_round_detuning} rad/µs goes below"
+                " the total bottom detuning "
                 f"of the DMM ({self.total_bottom_detuning} rad/µs)."
```
a-corni: same :)

pulser-core/pulser/channels/dmm.py
```diff
@@ -122,25 +161,50 @@ def validate_pulse(
...
             )
+        avg_abs_detuning = np.average(np.abs(round_detuning))
+        min_non_zero_weight = np.min(
+            # Can't be empty because the WeightMap enforces having at least one
+            # non-zero weight
+            np.array(detuning_map.weights)[
+                np.nonzero(detuning_map.weights)  # type: ignore[arg-type]
+            ]
```
a-corni: Shouldn't we have an assertion for this ?

pulser-core/pulser/channels/dmm.py
HGSilveri: Thanks, I adopted it with some modifications

pulser-core/pulser/channels/dmm.py
HGSilveri: I went for something similar

pulser-core/pulser/channels/dmm.py
```diff
@@ -183,16 +187,18 @@ def validate_pulse(
...
-            np.array(detuning_map.weights)[
-                np.nonzero(detuning_map.weights)  # type: ignore[arg-type]
-            ]
-        )
+
+        weights_arr = np.array(detuning_map.weights)
+        non_zero_weight_inds = np.nonzero(weights_arr)
+        # Can't be empty because the WeightMap enforces having at least one
+        # non-zero weight
+        assert len(non_zero_weight_inds) > 0
```
a-corni: np.non_zero returns a tuple with one array by dimension of the provided array, each of these arrays having the length of the number of zeros in the array.
Therefore, this assertion will always be True, but `assert len(non_zero_weight_inds) == 1 and len(non_zero_weight_inds[0]) >0` checks what we wanted to check :)

HGSilveri: Ahh, good catch!

```suggestion
        assert len(non_zero_weight_inds) == 1 and len(non_zero_weight_inds[0]) > 0
```

a-corni (approved):
LGTM !

PR #1047, opened 2026-04-17
Resolve inconsistencies in the XY Hamiltonian

### LINKED ISSUE #1042: Resolve inconsistencies in the XY Hamiltonian
There are currently some inconsistencies regarding the XY Hamiltonian, both in the documentation and in the implementation itself.

## What's wrong

1. **Contradicting state vector definitions in the [Conventions page](https://pulser.readthedocs.io/en/stable/conventions.html#driving-hamiltonian)**: The conventions page defines the XY basis state vectors as 

Image

At the same time, it states that

Image

- Combining these two statements, we must conclude that |0> is the higher energy state, but **this is not true**. 
- Therefore, we must change one of these to lift the contradiction.

2. **Incorrect driving Hamiltonian definition**: Regardless of the statevector definitions, the incorrect assumption that |0>=|u> is the higher energy state is reflected in the [construction of the driving Hamiltonian terms](https://github.com/pasqal-io/Pulser/blob/develop/pulser-simulation/pulser_simulation/hamiltonian.py#L339). This is incorrect and must be fixed.

3. **Confusing definitions of the $\sigma^+$ and $\sigma^-$ operators**: These operators appear (at least) [in the toggled section here](https://pulser.readthedocs.io/en/stable/programming.html#xy-hamiltonian), [here](https://pulser.readthedocs.io/en/stable/conventions.html#interaction-hamiltonian) and [here](https://pulser.readthedocs.io/en/stable/tutorials/xy_spin_chain.html#External-field-and-angular-dependency). Depending on the state vector definitions we fall on, the way these operators are defined is either incorrect or confusing. Since they can just be unambiguously replaced by projectors, we should just do that and get rid of them altogether.

## How to fix it

Points 2. and 3. are straightforward, so **let's focus on point 1**.

### Option A: Modify the XY basis state vectors

- We _could_ just say that $\ket{0}=(0,1)^T$ and  $\ket{1}=(1,0)^T$, but this goes against a very well established convention so it's a *bad idea*.
- The same argument holds for renaming $\ket{0} -> \ket{\uparrow}$, since there is also a strong expectation that  $\ket{\uparrow}$ is the +1 eigenstate of $\sigma_{z}$, so there is a strong expectation that $\ket{\uparrow}=(1,0)^T$ 
- The best way to do this might be to rename the basis states to something without any prior expectation on the state vector. For example, we could say that  $\ket{s}=(0,1)^T$ and  $\ket{p}=(1,0)^T$. 

#### Expected impact
- Widespread adjustment of the documentation to use the new nomenclature
- **Modification of the state vectors used in simulation**: The initial state vector (|00000...>) and the state vectors coming from simulations will have the order of their entries changed 

### Option B: Make an exception in the Pulser state vector convention for the XY Hamiltonian

- We keep the  $\ket{0}$ and  $\ket{1}$ state vector definitions intact and instead drop 

Image

#### Expected impact
- *Breaking of an established and uniform convention*:
    - Defining the $\ket{a}$ and $\ket{b}$ based on their relative energies will no longer be valid.
    - The definition of the driving Hamiltonian will now depend on the basis
    - Will require special treatment of the XY Hamiltonian _everywhere_. 
- Widespread adjustment of the documentation to reflect the new convention

HGSilveri:
- **The XY Hamiltonian in `pulser-simulation` is corrected to identify** $\ket{1}$ **as the higher energy state**
- The conventions page is adjusted to solve the contradiction outlined in #1042 

Closes #1042

### DIFF
--- a/docs/source/conventions.md
+++ b/docs/source/conventions.md
@@ -173,12 +173,14 @@ of $|a\rangle$ and $|b\rangle$.
 
 :::{important}
 **In `pulser-simulation`, this is the case for the `ground-rydberg` basis**, since
+
  $$
  E_r > E_g \rightarrow \ket{r} \equiv \ket{b}, \ket{g} \equiv \ket{a}
  $$
+
 :::
 
-When the basis vectors are defined in desceding energy order, we have
+When the basis vectors are defined in descending energy order, we have
 
 $$
 |b\rangle = (1, 0)^T,~~|a\rangle = (0, 1)^T
@@ -201,16 +203,18 @@ H^D(t) / \hbar = \frac{\Omega(t)}{2} \cos\phi(t) \hat{\sigma}^x
 - \delta(t) \hat{n}
 $$
 
-##### Case 2. Asceding energy order
+##### Case 2. Ascending energy order
 
 :::{important}
 **In `pulser-simulation`, this is the case for the `XY` basis**, since
+
  $$
  E_0 < E_1 \rightarrow \ket{0} \equiv \ket{a}, \ket{1} \equiv \ket{b}
  $$
+
 :::
 
-When the basis vectors are defined in asceding energy order, we have
+When the basis vectors are defined in ascending energy order, we have
 
 $$
 |a\rangle = (1, 0)^T,~~|b\rangle = (0, 1)^T
--- a/docs/source/conventions.md
+++ b/docs/source/conventions.md
@@ -235,6 +235,15 @@ H^D(t) / \hbar = \frac{\Omega(t)}{2} \cos\phi(t) \hat{\sigma}^x
 - \delta(t) \hat{n}
 $$
 
+:::{note}
+This is also the Hamiltonian one should use when trying to reconcile the basis states of the `ground-rydberg` basis
+with the computational-basis state-vector convention, i.e.
+
+$$
+|0\rangle = |g\rangle = |a\rangle = (1, 0)^T,~~|1\rangle = |r\rangle = |b\rangle = (0, 1)^T
+$$
+:::
+
 
 ### Interaction Hamiltonian
 
--- a/docs/source/conventions.md
+++ b/docs/source/conventions.md
@@ -169,10 +169,16 @@ A more conventional representation of the driving Hamiltonian uses Pauli operato
 instead of projectors. However, this form now **depends on the state vector definition**
 of $|a\rangle$ and $|b\rangle$.
 
-##### Pulser's state-vector definition
+##### Case 1. Descending energy order
 
-In Pulser, we consistently define the state vectors according to their relative energy.
-In this way we have, for any given basis, that
+:::{info}
+**In `pulser-simulation`, this is the case for the `ground-rydberg` basis**, since
+ $$
+ E_r > E_g \rightarrow \ket{r} \equiv \ket{b}, \ket{g} \equiv \ket{a}
+ $$
+:::
+
+When the basis vectors are defined in desceding energy order, we have
 
 $$
 |b\rangle = (1, 0)^T,~~|a\rangle = (0, 1)^T
@@ -195,17 +201,22 @@ H^D(t) / \hbar = \frac{\Omega(t)}{2} \cos\phi(t) \hat{\sigma}^x
 - \delta(t) \hat{n}
 $$
 
-##### Alternative state-vector definition
+##### Case 2. Asceding energy order
 
-Outside of Pulser, the alternative definition for the basis state
-vectors might be taken:
+:::{info}
+**In `pulser-simulation`, this is the case for the `XY` basis**, since
+ $$
+ E_0 < E_1 \rightarrow \ket{0} \equiv \ket{a}, \ket{1} \equiv \ket{b}
+ $$
+:::
+
+When the basis vectors are defined in asceding energy order, we have
 
 $$
 |a\rangle = (1, 0)^T,~~|b\rangle = (0, 1)^T
 $$
 
-This changes the operators and Hamiltonian definitions,
-as rewriten below with highlighted differences.
+This changes the operators and Hamiltonian definitions with respect to [Case 1](#case-1-descending-energy-order), as rewriten below with highlighted differences.
 
 $$
 \hat{\sigma}^x = |a\rangle\langle b| + |b\rangle\langle a|, \\
@@ -220,16 +231,6 @@ H^D(t) / \hbar = \frac{\Omega(t)}{2} \cos\phi(t) \hat{\sigma}^x
 - \delta(t) \hat{n}
 $$
 
-:::{note}
-A common case for the use of this alternative definition arises when
-trying to reconcile the basis states of the `ground-rydberg` basis
-(where $|r\rangle$ is the higher energy level) with the
-computational-basis state-vector convention, thus ending up with
-
-$$
-|0\rangle = |g\rangle = |a\rangle = (1, 0)^T,~~|1\rangle = |r\rangle = |b\rangle = (0, 1)^T
-$$
-:::
 
 ### Interaction Hamiltonian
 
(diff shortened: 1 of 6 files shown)


docs/source/conventions.md
```diff
@@ -169,10 +169,18 @@ A more conventional representation of the driving Hamiltonian uses Pauli operato
 instead of projectors. However, this form now **depends on the state vector definition**
 of $|a\rangle$ and $|b\rangle$.
 
-##### Pulser's state-vector definition
+##### Case 1. Descending energy order
 
-In Pulser, we consistently define the state vectors according to their relative energy.
-In this way we have, for any given basis, that
+:::{important}
+**In `pulser-simulation`, this is the case for the `ground-rydberg` basis**, since
```
a-corni, line 175:
```suggestion
**In `pulser-simulation`, this is the case for the `ground-rydberg` and `digital` basis**, since
```

docs/source/conventions.md
```diff
@@ -220,16 +235,6 @@ H^D(t) / \hbar = \frac{\Omega(t)}{2} \cos\phi(t) \hat{\sigma}^x
...
-:::{note}
-A common case for the use of this alternative definition arises when
-trying to reconcile the basis states of the `ground-rydberg` basis
-(where $|r\rangle$ is the higher energy level) with the
-computational-basis state-vector convention, thus ending up with
-
-$$
-|0\rangle = |g\rangle = |a\rangle = (1, 0)^T,~~|1\rangle = |r\rangle = |b\rangle = (0, 1)^T
-$$
-:::
```
a-corni, line 232: I agree that this seems less relevant now, even though it used to be important before. 
Perhaps we could keep it as a warning box at the end of the case 1 part ?

docs/source/programming.md
```diff
@@ -139,10 +139,10 @@ $$
 $$
 <details>
 
-  <summary>Interaction strength and entangling operator</summary>
+  <summary>Interaction strength in the XY Hamiltonian</summary>
 
 - The interaction strength is $\frac{C_3}{R_{ij}^3}$, with $C_3$ a coefficient that depends on the energy levels used to encode $\left|0\right>$ and $\left|1\right>$. 
-- The entangling operator between atom $i$ and $j$ is $\hat{\sigma}_i^{+}\hat{\sigma}_j^{-} + \hat{\sigma}_i^{-}\hat{\sigma}_j^{+} = |1\rangle\langle 0|_i |0\rangle\langle 1|_j + |0\rangle\langle 1|_i |1\rangle\langle 0|_j$.
```
a-corni, line 145: What was the issue with the entangling operator ?

pulser-core/pulser/channels/base_channel.py
```diff
@@ -45,7 +45,7 @@
     "propagation_dir",
 )
 
-# States ranked in decreasing order of their associated eigenenergy
+# States ordered as they appear in the state-vector representation
 States = Literal["u", "d", "r", "g", "h", "x"]
```
a-corni, line 49: Should we have a comment `# u->0, d->1` somewhere ? It could make sense to have it here ?

docs/source/conventions.md
HGSilveri: As we have done elsewhere in the docs, I would like to de-emphasize the `digital` basis, in part because I'm not even sure that this statement holds for the energy of its basis states.

docs/source/programming.md
HGSilveri: It was using the jump operators which we decided to just get rid of (see point 3. in #1042)

docs/source/conventions.md
HGSilveri: I can give it a shot but I'm afraid that it will cause more confusion rather than clarify anything

docs/source/programming.md
elmerhej, line 145: I agree here that writing `\sigma^{+}\sigma_{-}` is ambiguous, since its meaning depends on the state vector convention. However, the projector form

`|1\rangle\langle 0|_i |0\rangle\langle 1|_j + |0\rangle\langle 1|_i |1\rangle\langle 0|_j$. `

is still correct and unambiguous. We could leave this part I think

elmerhej:
The logic LGTM. I only have one minor comment in the docs

docs/source/programming.md
HGSilveri: Yeah, but then we are just repeating what's already clearly stated in the interaction term a few lines above, it seems redundant

pulser-core/pulser/channels/base_channel.py
HGSilveri: I think we already have it in some docstring but it doesn't hurt to put it here too

docs/source/conventions.md
```diff
@@ -221,16 +236,15 @@ H^D(t) / \hbar = \frac{\Omega(t)}{2} \cos\phi(t) \hat{\sigma}^x
 $$
 
 :::{note}
-A common case for the use of this alternative definition arises when
-trying to reconcile the basis states of the `ground-rydberg` basis
-(where $|r\rangle$ is the higher energy level) with the
-computational-basis state-vector convention, thus ending up with
+This is also the Hamiltonian one should use when trying to reconcile the basis states of the `ground-rydberg` basis
+with the computational-basis state-vector convention, i.e.
```
HGSilveri, line 240: @a-corni better?

a-corni (approved):
LGTM !

pulser-simulation/pulser_simulation/hamiltonian.py
```diff
@@ -337,7 +337,7 @@ def build_coeffs_ops(
             elif basis == "digital":
                 op_ids = ["sigma_hg", "sigma_gg"]
             elif basis == "XY":
-                op_ids = ["sigma_du", "sigma_uu"]
+                op_ids = ["sigma_ud", "sigma_dd"]
```
a-quelle-pasqal, line 340: Ok, so if I understand this correctly, you're changing the `delta` term in the Hamiltonian to now act on `d` rather than `u`. Together with all the comments it seems you're performing a swap of the states `u` and `d` everywhere. What is the reason for this? Are we modelling a different measurement protocol or something?

tutorials/quantum_simulation/Spin chain of 3 atoms in XY mode.ipynb
```diff
@@ -208,7 +208,7 @@
     "An external magnetic field can be added to the experiment, and will modify the hamiltonian. The XY Hamiltonian is then\n",
```
a-quelle-pasqal, line 208: <p>Isn't this the same? Are you doing this purely to be more explicit about the action of the interaction term? I agree for example that in this representation it's more explicit that the XY interaction migrates spins through the lattice.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pasqal-io/Pulser/pull/1047/discussion/#comment-3152314745'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='15'/>

pulser-simulation/pulser_simulation/hamiltonian.py
HGSilveri: The Hamiltonian that was being implemented assumed that `E_u > E_d` and that's apparently not the case - or rather, if we chose to keep it this way we would go against the established convention of which is the `u` and `d` states in XY mode. 
Since [in this form](https://pulser.readthedocs.io/en/stable/conventions.html#driving-hamiltonian) the `delta` term acts on the highest energy state, we had to correct the Hamiltonian.
More precisely, we changed which one is the highest and lowest energy state to match the established convention.

tutorials/quantum_simulation/Spin chain of 3 atoms in XY mode.ipynb
HGSilveri: It's the same, yes. 
The issue was that the possible definitions of the ladder operators were either wrong or confusing (e.g. `sigma^+` would bring |1> to |0>) so we decided to just write everything in terms of projectors (see point 3 in #1042 )

PR #1053, opened 2026-04-28
Clip InterpolatedWaveform samples to stay within the values' range

### LINKED ISSUE #1051: InterpolatedWaveform samples exceed specified maximum amplitude
Sampling an `InterpolatedWaveform` with values that include the device's maximum amplitude, produces values that exceed this maximum amplitude. This behaviour could lead to validation errors when the sequence is created.

To be assessed but likely caused by samples being rounded in the `_sample` method [here](https://github.com/pasqal-io/Pulser/blob/3ae06415543abfe99141990f108f54b45b64cb15/pulser-core/pulser/waveforms.py#L974)

## is this the expected behaviour?
- [ ] understand if the sampled values from an `InterpolatedWaveform` should not exceed the maximum value specified in the waveform definition, similar to how `ConstantWaveform` behaves.
- [ ] specific interpolators (scipy.interp1d, kind="quadradic" for example) produce functions that by design can exceed the input range. Is this desirable?
- [ ] understand why samples are rounded

## Actual Behaviour
`InterpolatedWaveform` produces sampled values that are greater than the specified maximum amplitude due to rounding.

```python
from pulser import ConstantWaveform, InterpolatedWaveform, DigitalAnalogDevice

max_amp = DigitalAnalogDevice.channels["rydberg_global"].max_amp  # 15.707963267948966

# InterpolatedWaveform case
interpolated_wf = InterpolatedWaveform(60.0, [0.0, 0.5, max_amp, 0.5, 0.0])
max_interpolated_sampled = interpolated_wf.samples.as_array().max()
print(f"InterpolatedWaveform max sampled value: {max_interpolated_sampled}")
print(f"Exceeds max_amp: {max_interpolated_sampled > max_amp}")  # True

# ConstantWaveform case
constant_wf = ConstantWaveform(60.0, max_amp)
max_constant_sampled = constant_wf.samples.as_array().max()
print(f"ConstantWaveform max sampled value: {max_constant_sampled}")
print(f"Exceeds max_amp: {max_constant_sampled > max_amp}")  # False
```

Output
```sh
InterpolatedWaveform max sampled value: 15.708981068448675
Exceeds max_amp: True
ConstantWaveform max sampled value: 15.707963267948966
Exceeds max_amp: False
```

# Suggested Solution
The interpolation algorithm should be constrained to ensure that sampled values never exceed the maximum value specified in the waveform definition, possibly by clamping interpolated values to the specified range.

# Environment
Pulser version: 1.7.2
Python version: 3.13

HGSilveri:
Fixes #1051.

### DIFF
--- a/pulser-core/pulser/waveforms.py
+++ b/pulser-core/pulser/waveforms.py
@@ -974,6 +974,10 @@ class InterpolatedWaveform(Waveform):
     def _samples(self) -> pm.AbstractArray:
         """The value at each time step that describes the waveform."""
         samples = self._interp_func(np.arange(self._duration))
+        if self._kwargs["interpolator"] == "PchipInterpolator" and set(
+            self._kwargs  # Checks if there are no custom kwargs
+        ) == {"times", "interpolator"}:
+            return pm.clip(samples, np.min(self._values), np.max(self._values))
         value_range = np.max(np.abs(samples))
         decimals = int(
             min(np.finfo(samples.dtype).precision - np.log10(value_range), 9)
--- a/tests/test_waveforms.py
+++ b/tests/test_waveforms.py
@@ -317,6 +317,13 @@ def test_interpolated():
     ):
         interp_wvf.build()
 
+    # Make sure Interpolated Waveform stays within the range of values
+    # (this used to fail), see #1051
+    max_amp = 2.5 * 2 * np.pi
+    interpolated_wf = InterpolatedWaveform(60.0, [0.0, 0.5, max_amp, 0.5, 0.0])
+    assert np.all(interpolated_wf.samples.as_array() <= max_amp)
+    assert np.all(interpolated_wf.samples.as_array() >= 0)
+
 
 def test_kaiser():
     duration: int = 40


sgrava (approved):
Nice! thanks for the fix

pulser-core/pulser/waveforms.py
```diff
@@ -974,6 +974,10 @@ def duration(self) -> int:
     def _samples(self) -> pm.AbstractArray:
         """The value at each time step that describes the waveform."""
         samples = self._interp_func(np.arange(self._duration))
+        if self._kwargs["interpolator"] == "PchipInterpolator" and set(
+            self._kwargs  # Checks if there are no custom kwargs
+        ) == {"times", "interpolator"}:
+            return pm.clip(samples, np.min(self._values), np.max(self._values))
         value_range = np.max(np.abs(samples))
```
a-corni, line 981: Why only clip for PChipInterpolator ? The only other interpolator accepted at the moment is interp1d, which should also enforce this :)

HGSilveri: Unlike PChipInterpolator, the `interp1d` is not designed to not overflow the input values, so if we clip on it we might get some weird non-smooth shapes. I don't know if anyone is using this interpolator, but I think it would constitute a breaking change.
In the case of the PChipInterpolator, this is really only making sure that we stay within the desired range (i.e. taking care of eventual rounding errors).

HGSilveri: Look what a cubic interpolator can give
image
If we clip it, it becomes
image

a-corni: I am confused
- you described PChipInterpolator (a cubic interpolator) as designed to not overflow (let's say it should be a convex interpolator). But in the plot you sent I would say that it overflows (it's not convex) and that clipping would constitute a breaking change. Was your plot not made with PChipInterpolator but another cubic interpolator ?
- interp1d is piecewise linear between two points to interpolate, so by design it is convex and it should not overflow, clipping should just enable to correct rounding errors

HGSilveri: Sorry, I was not clear enough. The plot above was done with
```
pulser.InterpolatedWaveform(duration=1000, times=[0., 0.5, 0.51, 1.], values=[0., 1., 0., 0.5], interpolator="interp1d", kind="cubic").draw()
```
This is what I meant by "cubic intepolator".
You can choose which kind of interpolator to use with `interp1d`. I agree that the linear one (the default) could be clipped, but not then cubic one, as shown.

a-corni: Oookay now it's clear, sorry I hadn't seen and forgotten about the `kind` argument ! I have the feeling that's why scipy has stopped development for this function (they said a better solution should exist). I approve then

a-corni (approved):
LGTM ! Thanks for the explanations

PR #1056, opened 2026-05-05
[FEAT] Add fastjsonschema as optional fast-path for JSON validation

stephenfreelance:
Add a new optional library to validate input data. Summary of changes:
- Introduces `fastjsonschema` as an optional dependency (`pulser-core[fast-validation]`) to speed up JSON schema validation of abstract representations
- When installed, validators are compiled at import time and used as the primary validation path; on failure, falls back to the standard `jsonschema` library to provide detailed error diagnostics
- Refactors shared constants (`ObjectType` literal type, `get_filename` helper) into `pulser/json/utils.py` to reduce duplication across `__init__.py` and `validation.py`
- Updates CI matrix (`ci.yml`, `test.yml`) to test both with and without `fastjsonschema` installed

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,6 +18,7 @@ jobs:
         os: [ubuntu-latest, macos-latest, windows-latest]
         numpy: ["numpy-v1", "numpy-v2"]
         with-torch: ["with-torch", "no-torch"]
+        fast-validation: ["with-fast-validation", "no-fast-validation"]
         python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
         exclude:
           # Numpy v1 does not support Python 3.13
@@ -40,5 +41,6 @@ jobs:
           extra-packages: pytest
           numpy: ${{ matrix.numpy }}
           with-torch: ${{ matrix.with-torch }}
+          with-fast-validation: ${{ matrix.fast-validation }}
       - name: Run the unit tests & generate coverage report
         run: pytest --cov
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -12,6 +12,10 @@ dev-install-core:
 dev-install-core-no-torch:
 	pip install -e ./pulser-core
 
+.PHONY: dev-install-fast-validation
```
HGSilveri: If we really want to add this, I would instead call it

```suggestion
.PHONY: dev-install-core-fast-validation
```

.github/workflows/test.yml
```diff
@@ -18,6 +18,7 @@ jobs:
         os: [ubuntu-latest, macos-latest, windows-latest]
         numpy: ["numpy-v1", "numpy-v2"]
         with-torch: ["with-torch", "no-torch"]
+        fast-validation: ["with-fast-validation", "no-fast-validation"]
```
HGSilveri: Do we really need to double the number of tests here? I feel like there will be a lot of redundancy

HGSilveri (commented):
Thanks for this @stephenfreelance , I only have a couple minor comments

.github/workflows/test.yml
stephenfreelance: Yes you're right, I added a variant for test only with: `numpy-v2`, `no-torch` and `with-fast-validation`, that's enough

HGSilveri (approved):
LGTM, thanks @stephenfreelance !

PR #1059, opened 2026-05-14
fix: clarify emulation config type errors

### LINKED ISSUE #1025: Explain better errors in EmulationConfig init
It would be helpful to give more information in errors thrown in `EmulationConfig.__init__` such that it is easier to identify which particular argument fails.

For example, I am getting the following error:
```python
  File "pulser/backend/config.py", line 276, in __init__
    raise TypeError(
TypeError: All entries in 'observables' must be instances of Observable. Instead, got instance of type <class 'str'>.
```
This requires a bit of work to identify the problematic observable. Changing the test on line 275 to:
```python
if not isinstance(obs, Observable):
    raise TypeError(
        "All entries in 'observables' must be instances of "
        f"Observable. Instead, got instance of type {type(obs)}. "
        f"for observable {obs}."
    )
```
makes it much easier to identify the problem. Similar improvements can be done to other tests.

nanookclaw:
Closes #1025.

This improves the `EmulationConfig` validation errors for invalid `callbacks` and `observables` entries. The previous errors only reported the failing type, which made it harder to find the bad value in a longer list.

The new messages include the failing index and `repr()` of the value while keeping the existing error prefixes intact. I also tightened the existing tests so they cover the added context for invalid observables, callbacks that are observables, and non-callback callback entries.

Verification:
- `python3 -m compileall -q pulser-core/pulser/backend/config.py tests/test_backend.py`
- `git diff --check`

I could not run the focused pytest locally because this environment is missing optional test/runtime dependencies (`evalcraft`, then `scipy` when disabling that plugin).

### DIFF
--- a/pulser-core/pulser/backend/config.py
+++ b/pulser-core/pulser/backend/config.py
@@ -260,22 +260,25 @@ class EmulationConfig(BackendConfig, Generic[StateType]):
                 " empty.",
                 stacklevel=2,
             )
-        for cb in callbacks:
+        for i, cb in enumerate(callbacks):
             if isinstance(cb, Observable):
                 raise TypeError(
                     "All entries in 'callbacks' must not be instances of "
-                    "Observable, since those go in 'observables'."
+                    "Observable, since those go in 'observables'. "
+                    f"Instead, got {cb!r} at index {i}."
                 )
             if not isinstance(cb, Callback):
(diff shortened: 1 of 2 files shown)


a-corni (approved):
Hello @nanookclaw !
Thanks for your contribution, this looks good to me !

PR #1060, opened 2026-05-22
Add Rydberg interaction in XY mode

### LINKED ISSUE #1035: Include the van der Waals interaction term in the XY Hamiltonian
The XY Hamiltonian still has the Ising van der Waals interaction terms, so we should include them in the Hamiltonian.

The terms are given by

$$ \frac{C_6}{R_{ij}^6} |0\rangle\langle 0|_i |0\rangle\langle 0|_j$$

These should also be added throughout the the documentation (see #1042 for the places where the XY Hamiltonian is mentioned).

a-quelle-pasqal:
The XY mode has a Rydberg interaction between the `d` states. This MR extends the `interaction_matrix` config parameter, to allow for 2 interaction matrices as a shape `(2,N,N)` tensor in the XY case, and adds the Rydberg term.

Closes #1035

### DIFF
--- a/pulser-core/pulser/_hamiltonian_data/hamiltonian_data.py
+++ b/pulser-core/pulser/_hamiltonian_data/hamiltonian_data.py
@@ -568,7 +568,12 @@ class HamiltonianData:
 
         Returns:
             The pairwise interaction coefficients, taking into account
-            Device specs and the Sequence type.
+            Device specs and the Sequence type. In XY mode,
+            it has shape (2, N, N), with N the number of qubits.
+            The first array of shape (N, N) is the C3 interaction,
+            the second the C6 interaction.
+            In Rydberg interaction, it has shape (1, N, N)
+            and corresponds to the C6 interaction only.
         """
         # SLM mask is not included, because it's time-dependent
         is_xy = self.basis_data.interaction_type == "XY"
@@ -624,6 +629,7 @@ class HamiltonianData:
         Returns:
             The pairwise interaction coefficients, taking into account
             Device specs, the Sequence type and missing atoms.
+            See the docstring for HamiltonianData._interaction_matrix.
         """
         mask = [False for _ in range(self.n_qudits)]
         for ind, value in enumerate(bad_atoms.values()):
--- a/pulser-core/pulser/_hamiltonian_data/hamiltonian_data.py
+++ b/pulser-core/pulser/_hamiltonian_data/hamiltonian_data.py
@@ -571,9 +571,11 @@ class HamiltonianData:
             Device specs and the Sequence type.
         """
         # SLM mask is not included, because it's time-dependent
+        is_xy = self.basis_data.interaction_type == "XY"
         d = _distances(register)
-        interactions = pm.zeros_like(d)._array
-        if self.basis_data.interaction_type == "XY":
+        interactions = pm.zeros_like(d.reshape((1,) + d.shape))._array
+        if is_xy:
+            interactions = pm.vstack([interactions, interactions])._array
             positions = list(register.qubits.values())
             assert self.samples._magnetic_field is not None
             assert self._device.interaction_coeff_xy is not None
@@ -588,17 +590,18 @@ class HamiltonianData:
                             [diff, pm.AbstractArray(np.array(0.0))]
                         )
                     cosine = pm.dot(diff, mag_arr) / (pm.norm(diff) * mag_norm)
-                    interactions[[i, j], [j, i]] = (
+                    interactions[[0, 0], [i, j], [j, i]] = (
                         self._device.interaction_coeff_xy  # type: ignore
                         * (1 - 3 * cosine._array**2)
                         / d._array[i, j] ** 3
                     )
-        else:
-            for i in range(self.n_qudits):
-                for j in range(i + 1, self.n_qudits):
-                    interactions[[i, j], [j, i]] = (
-                        self._device.interaction_coeff / d._array[i, j] ** 6
-                    )
+
+        for i in range(self.n_qudits):
+            for j in range(i + 1, self.n_qudits):
+                interactions[[-1, -1], [i, j], [j, i]] = (
+                    self._device.interaction_coeff / d._array[i, j] ** 6
+                )
+
         return interactions
 
     @property
@@ -630,13 +633,13 @@ class HamiltonianData:
             arr = np.array(mask)
             mask2 = arr.reshape(1, -1) | arr.reshape(-1, 1)
             mat = imat.copy()
-            mat[mask2] = 0.0
+            mat[:, mask2] = 0.0
             return pm.AbstractArray(mat)
         else:
             ten = pm.torch.tensor(mask, dtype=pm.torch.bool)
             mask3 = ten.reshape(1, -1) | ten.reshape(-1, 1)
             mat2 = imat.clone()
-            mat2[mask3] = 0.0
+            mat2[:, mask3] = 0.0
             return pm.AbstractArray(mat2)
 
     def _build_local_collapse_operators(
--- a/pulser-core/pulser/backend/config.py
+++ b/pulser-core/pulser/backend/config.py
@@ -182,11 +182,13 @@ class EmulationConfig(BackendConfig, Generic[StateType]):
             input or the expected output.
         interaction_matrix: An optional interaction matrix to replace the
             interaction terms in the Hamiltonian. For an N-qudit system,
-            must be an NxN symmetric matrix where entry (i, j) dictates
-            the interaction coefficient between qudits i and j, ie it replaces
-            the C_n/r_{ij}^n term. For XY, since there are two interactions
-            the input should be a tensor of shape (2,N,N), respresenting the
+            an interaction matrix is an NxN symmetric matrix,
+            where entry (i, j) dictates the interaction coefficient between
+            qudits i and j, ie it replaces the C_n/r_{ij}^n term.
+            In XY, there are the two interactions C3 and C6 so
+            the input should be a tensor of shape (2,N,N), representing the
             interaction matrix for the XY and the Rydberg term in that order.
+            Otherwise the input should be a tensor of shape (N, N) or (1,N,N).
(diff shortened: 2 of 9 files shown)


pulser-core/pulser/_hamiltonian_data/hamiltonian_data.py
```diff
@@ -571,9 +571,11 @@ def _interaction_matrix(
             Device specs and the Sequence type.
```
a-corni: We should rework this description to highlight the new format of the interaction matrix.

```suggestion
r"""C6 and eventual C3 Interactions between the qudits (in :math:`rad/\mu s`).

       Uses data from self to determine the C6 interaction and, if the XY mode is activated, C3 interactions.

       Args:
           register: The register for which to compute the interactions.

       Returns:
           The pairwise interaction coefficients, taking into account
           Device specs and the Sequence type. In XY mode, it has shape (2, N, N), with N the number of qubits. The first array of shape (N, N) is the C3 interaction, the second the C6 interaction. In Rydberg interaction, it has shape (1, N, N) and corresponds to the C6 interaction only.
```

pulser-core/pulser/math/abstract_array.py
```diff
@@ -322,5 +322,15 @@ def _to_abstract_repr(self) -> Any:
                 " losing the computational graph information."
             ) from e
 
+    def reshape(self, shape: tuple[int, ...]) -> AbstractArray:
+        """Return a new Abstractarray with the new shape.
```
a-corni: nit:
```suggestion
        """Returns a new AbstractArray with the new shape.
```

pulser-core/pulser/math/abstract_array.py
```diff
@@ -322,5 +322,15 @@ def _to_abstract_repr(self) -> Any:
...
+
+        See the reshape functions in torch and numpy respectively.
+
+        Args:
+            shape: The desired new shape. Must be compatible with the old.
```
a-corni:
```suggestion
            shape: The desired new shape. Must be compatible with the current shape.
```

pulser-core/pulser/math/abstract_array.py
```diff
@@ -322,5 +322,15 @@ def _to_abstract_repr(self) -> Any:
                 " losing the computational graph information."
             ) from e
 
+    def reshape(self, shape: tuple[int, ...]) -> AbstractArray:
+        """Return a new Abstractarray with the new shape.
+
+        See the reshape functions in torch and numpy respectively.
```
a-corni:
```suggestion
        See the reshape method of torch.Tensor and numpy.array respectively.
```

pulser-core/pulser/_hamiltonian_data/hamiltonian_data.py
a-corni, line 631: Here as well, the description should be changed, at least to define the shape of the returned AbstractArray

a-corni, line 623: Do you know if some of the "missing" qubits could still be interacting with a C6 interaction ? We are just saying they are "missing" because not in the Microwave basis, that doesn't say anything about their presence in the Rydberg states ?

a-corni, line 825: I am wondering if we shouldn't specify the shape of the interaction matrix in the description of `NoiseTrajectory` ? (1, N, N) or (2, N, N)

pulser-core/pulser/backend/config.py
```diff
@@ -184,7 +184,9 @@ class EmulationConfig(BackendConfig, Generic[StateType]):
             interaction terms in the Hamiltonian. For an N-qudit system,
             must be an NxN symmetric matrix where entry (i, j) dictates
             the interaction coefficient between qudits i and j, ie it replaces
-            the C_n/r_{ij}^n term.
+            the C_n/r_{ij}^n term. For XY, since there are two interactions
+            the input should be a tensor of shape (2,N,N), respresenting the
+            interaction matrix for the XY and the Rydberg term in that order.
```
a-corni, line 190:
```suggestion
       interaction_matrix: An optional interaction matrix to replace the
           interaction terms in the Hamiltonian. For an N-qudit system,
           an interaction matrix is an NxN symmetric matrix where entry (i, j) dictates
           the interaction coefficient between qudits i and j, ie it replaces
            the C_n/r_{ij}^n term. In XY, there are the two interactions C3 and C6 so
            the input should be a tensor of shape (2,N,N), representing the
            interaction matrix for the XY and the Rydberg term in that order. Without XY, the input should be a tensor of shape (N, N).
```

a-corni: Perhaps we should explain as well how a matrix of shape (1, N, N) or (N, N) is interpreted in the case of an XY sequence ?

pulser-core/pulser/backend/config.py
a-corni, line 339: Shouldn't you check `all([np.allclose(matrix_arr[i, :], matrix_arr[i, :].transpose()) for i in range(matrix_arr.shape[0])])` if `len(matrix_arr.shape)==3` ?

a-corni, line 388: I am wondering if the final interaction_matrix saved here shouldn't always be of shape (1, N, N) or (2, N, N) ? That would make your life easier afterwards, no ?

pulser-simulation/pulser_simulation/hamiltonian.py
```diff
@@ -281,11 +281,16 @@ def make_xy_term(q1: QubitId, q2: QubitId) -> qutip.Qobj:
...
+            U_xy = self.noise_trajectory.interaction_matrix.as_array(
+                detach=True
+            )[0, self._qid_index[q1], self._qid_index[q2]]
+            U_ryd = self.noise_trajectory.interaction_matrix.as_array(
+                detach=True
+            )[1, self._qid_index[q1], self._qid_index[q2]]
+            return U_xy * self.build_operator(
                 [("sigma_ud", [q1]), ("sigma_du", [q2])]
+            ) + 0.5 * U_ryd * self.build_operator(
+                [("sigma_dd", [q1]), ("sigma_dd", [q2])]
```
a-corni: I agree with sigma_dd, I will just wait for @HGSilveri to agree as well, since his initial proposition was to have sigma_uu.

a-corni (commented):
Thanks for tackling this @a-quelle-pasqal ! Are you planning on tackling the documentation in this MR as well ?

pulser-core/pulser/_hamiltonian_data/hamiltonian_data.py
a-quelle-pasqal: Doesn't the same thing hold for Rydberg sequences? So far we've treated missing atoms as actually gone, not just in a wrong state of some kind, and I wonder if we should change now.

pulser-core/pulser/backend/config.py
a-quelle-pasqal: I will change the text to your suggestion, but I think it's clear from the text that specifying anything other than `(2,N,N)` for XY is an error no? My only dissatisfaction is that the config has no access to the sequence, so you cannot check in the config that the user gave the correct shape and give a good error message. As it stands, it's only relevant for emu-mps, where the user will have to digest the runtime error when the Hamiltonian is built.

pulser-core/pulser/_hamiltonian_data/hamiltonian_data.py
a-quelle-pasqal: agreed. I've added descriptions for all the other properties as well, since I realized I didn't know them all by heart anymore.

HGSilveri: The "missing" qubits are either really gone or left in some leakage (ground) state (from state preparation errors). In either case, they should not interact at all

pulser-core/pulser/backend/config.py
```diff
@@ -182,9 +182,13 @@ class EmulationConfig(BackendConfig, Generic[StateType]):
             input or the expected output.
         interaction_matrix: An optional interaction matrix to replace the
```
HGSilveri, line 183: Instead of making this an argument with a union type, what about introducing a dedicated `xy_interaction_matrix` in the config and leave `interaction_matrix` reserved for the C6 interactions (specifying that in the docstring)?
I would find that more flexible and intuitive.

pulser-simulation/pulser_simulation/hamiltonian.py
```diff
@@ -269,7 +269,7 @@ def make_vdw_term(q1: QubitId, q2: QubitId) -> qutip.Qobj:
                 0.5
                 * self.noise_trajectory.interaction_matrix.as_array(
                     detach=True
-                )[self._qid_index[q1], self._qid_index[q2]]
+                )[0, self._qid_index[q1], self._qid_index[q2]]
```
HGSilveri: nit: It shouldn't matter, but I'd find it better to have this as
```suggestion
                )[-1, self._qid_index[q1], self._qid_index[q2]]
```

HGSilveri (commented):
I have a suggestion for how to handle `interaction_matrix` in `EmulationConfig`. Apologies if you have already considered it

pulser-core/pulser/backend/config.py
a-quelle-pasqal: I have considered it, but it doesn't mesh well with a code refactor I want to do in emu-mps to support this feature. Currently, what happens in emu-mps is that either the XY, or the Rydberg term is added to the Hamiltonian (which is somewhat of an involved process because you have to build the MPO), and on the off chance that something like this happens again, I want to properly implement the logic for taking a sequence of interaction matrices and the corresponding pairs of 1-body operators, and just build the full thing.

Things that could potentially happen in the future include
- adding more interaction terms to other operating modes
- allowing the user to also specify the sequence of interaction terms in the config to generalize the emulator.

In both these cases, to prevent the number of arguments in the config from exploding, the current format would work better. Also, the implementation I described above is simpler with the interaction matrices in the current packed format. I agree that for the user it's slightly less intuitive, but it doesn't feel unduly selfish to me.

HGSilveri (approved):

PR #1063, opened 2026-06-03
Replace references from pulser-pasqal to pasqal-cloud

Fenkiou:
We merged pulser-pasqal in pasqal-cloud package so we're updating repositories that referenced pulser-pasqal.


pulser-core/pulser/backends.py
```diff
@@ -50,11 +49,9 @@
     "QPUBackend": "pulser.backend",
     "QutipBackend": "pulser_simulation",
     "QutipBackendV2": "pulser_simulation",
-    "EmuFreeBackend": "pulser_pasqal",
-    "EmuFreeBackendV2": "pulser_pasqal",
-    "EmuTNBackend": "pulser_pasqal",
```
Fenkiou, line 55: is it considered a breaking change? 🤔 

If yes, I'm not sure how to handle that, I took the opportunity to remove them  (EmuFreeBackendV2 was was not deleted but renamed to EmuFreeBackend) from pasqal-cloud as they were deprecated in pulser-pasqal.

And we won't be able to have both pulser-pasqal (0.23.0) and pasqal-cloud (0.23.0) as pulser-pasqal 0.23.0 pinned pasqal-cloud to 0.22.0 due to many internal changes

docs/source/apidoc/pasqal.rst
```diff
@@ -18,15 +18,6 @@ Backends
...
-   ~pulser_pasqal.backends.EmuFreeBackendV2
+   ~pasqal_cloud.backends.EmuFreeBackend
    EmuMPSBackend
    EmuSVBackend
-
-Deprecated Backends
---------------------
-
-.. autosummary::
-   :toctree: _autosummary
```
Fenkiou, line 29: Should I have kept that? even if empty?

pulser-core/pulser/backends.py
Fenkiou: I suppose it's kind of breaking based on: https://github.com/search?q=repo%3Apasqal-io%2Fqoolqit%20EmuFreeBackendV2&type=code as docstrings/notebooks do `BackendType.EmuFreeBackendV2` and `BackendType` points to that module (see: https://github.com/pasqal-io/qoolqit/blob/main/qoolqit/execution/__init__.py#L5)

Fenkiou: @MohamedAminMallek @MatthieuMoreau0 @HGSilveri what do you think about that?

"Breaking" things in pasqal-cloud isn't problematic per se as we introduced new things in it. But as pulser-core re-exports things defined from pulser-pasqal and qoolqit/qubo-solver depending on those, I suppose we must re-export the same things as pulser-pasqal did..

MohamedAminMallek: Since EmuFreeBackend has been deprecated for a while, is it okay to remove it? @MatthieuMoreau0 @HGSilveri

If so, we could keep the deprecated EmuTNBackend in PasqalCloud(or remove it as well) and expose the new EmuFreeBackend with a V2 alias. In that case, the only breaking change would be the removal of the deprecated EmuFreeBackend.

Otherwise, to avoid that breaking change, we should also move the current EmuFreeBackend into PasqalCloud.

docs/source/apidoc/pasqal.rst
```diff
@@ -1,15 +1,15 @@
-﻿``pulser_pasqal``
+﻿``pasqal_cloud``
```
HGSilveri: I would argue that it's time we get rid of this API reference altogether. 
We can keep the page and just have a warning box saying the package was deprecated, that one should use `pasqal_cloud` instead and a link to redirect to the `pasqal_cloud` docs.

docs/requirements.txt
```diff
@@ -6,7 +6,7 @@ nbsphinx
 nbsphinx-link
 ipython >= 8.10 # Avoids bug with code highlighting
 myst-parser
-# To document pulser-pasqal classes
-pulser-pasqal >= 0.20.4
+# To document pasqal-cloud classes
+pasqal-cloud >= 0.23.0
```
HGSilveri: Let's see if we still need this dependency in the end

pulser-core/pulser/backends.py
HGSilveri: - You are free to remove the deprecated backends as you see fit. I would not keep `EmuTNBackend`, it doesn't make much sense to remove `EmuFreeBackend` but keep `EmuTNBackend`. What we can still do here is have a dedicated error message for when a user attempts to import one of the removed backends
- What we want to ensure is that previous non-deprecated imports lead to the same class. So, if you rename `EmuFreeBackendV2 -> EmuFreeBackend`, I would like `from pulser.backends import EmuFreeBackendV2` to now give me `pasqal_cloud.EmuFreeBackend` (along with a warning for the naming change). Feel free to modify `__getattr__()` to handle such special cases
- I realize it's prickly that a `from pulser.backends import EmuFreeBackend` silently goes from returning a deprecated backend to the latest one - and that adding a warning on this will just be noisy for whoever is using the latest version as intended. 
- However, keeping `EmuFreeBackendV2` is a missed opportunity to do some necessary clean-up. So let me make an alternative proposal (I know it's a bit disruptive, just hear me out and freely reject it if it doesn't fit your schedule or you just outright disagree):
    - It has been correctly pointed out that the backend names are confusing and not self-explanatory (e.g. I can't really tell if it's `MPSBackend` or `EmuMPSBackend` that is remote from the name alone). 
    - So what about we take the opportunity to rename them all to something better. _For instance_:
        - `EmuMPSBackend` -> `RemoteMPS(Backend)` (I think `Backend` could potentially be ommited
        - `EmuSVBackend` -> `RemoteSV(Backend)`
        - `EmuFreeBackend` -> (This one I'm not sure, could be `RemoteEmuFree`, `RemoteQutip`,... TBD)
    - This would have the added benefit of bypassing the `EmuFreeBackend` renaming issue

WDYT?

Fenkiou: >  I would like from pulser.backends import EmuFreeBackendV2 to now give me pasqal_cloud.EmuFreeBackend (along with a warning for the naming change). Feel free to modify __getattr__() to handle such special cases

Will do 👍 

> I realize it's prickly that a from pulser.backends import EmuFreeBackend silently goes from returning a deprecated backend to the latest one - and that adding a warning on this will just be noisy for whoever is using the latest version as intended.

Totally agree

> So what about we take the opportunity to rename them all to something better.

That's a very good idea! It clearly stands that they are remote backends and as you said will prevent the renaming issue. I'd tend to name them like their local counterpart, eg keeping `Backend` in the name, so:
- `EmuMPSBackend` -> `RemoteMPSBackend`
- `EmuSVBackend` -> `RemoteSVBackend`

For `EmuFreeBackendV2`, I would be more in favor of using `Qutip` instead of `Free` because "free" seems like a cloud pricing concern to me. So I would name it `RemoteQutipBackend`. No `V2` in the name despite the fact the local counter part is `QutipBackendV2` due to `QutipBackend` deprecation. Anyway, even if we want to keep `Free` because we use this term in lots of place (doc, user portal, etc..) or any other reason, `RemoteFreeBackend` will solve the issue of having `EmuFreeBackend` pointing to a previously deprecated class.

We didn't released pasqal-cloud yet so we still have the time to change that, I'll open a PR :)

---

Last thing is when you say

> You are free to remove the deprecated backends as you see fit. I would not keep EmuTNBackend

So you mean that if we remove those from the backends you won't consider this a breaking change on your side and you won't have to bump to a 2.0.0 because ? If yes that's a relief to me 😌

HGSilveri: > So you mean that if we remove those from the backends you won't consider this a breaking change on your side and you won't have to bump to a 2.0.0 because ? If yes that's a relief to me 😌

Indeed, I think it's reasonable to draw the line on removals from third parties. Let's just make sure the transition is properly signaled and documented and we should be good.

pulser-core/pulser/backends.py
```diff
@@ -50,24 +50,47 @@
...
+            new_name = _RENAMED_BACKENDS[name]
+
+            warnings.warn(
+                f"{name!r} was renamed to {new_name!r}. "
+                f"Please use {new_name!r} from now on.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+
+            name = new_name
```
Fenkiou, line 92: Works correctly using `from pulser import backends; backends.X`, see:

```
Python 3.10.19 (main, Feb 12 2026, 00:36:33) [Clang 21.1.4 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from pulser import backends
>>> backends.RemoteFreeBackend
...
ModuleNotFoundError: No module named 'pasqal_cloud'

During handling of the above exception, another exception occurred:

...
AttributeError: 'RemoteFreeBackend' requires the 'pasqal_cloud' package. To install it, run `pip install pasqal_cloud`.

>>> backends.EmuFreeBackend
...
AttributeError: 'EmuFreeBackend' was deprecated and is now removed from module 'pulser.backends'

>>> backends.EmuFreeBackendV2
<stdin>:1: DeprecationWarning: 'EmuFreeBackendV2' was renamed to 'RemoteFreeBackend'. Please use 'RemoteFreeBackend' from now on.
Traceback (most recent call last):...
ModuleNotFoundError: No module named 'pasqal_cloud'

During handling of the above exception, another exception occurred:

...
AttributeError: 'RemoteFreeBackend' requires the 'pasqal_cloud' package. To install it, run `pip install pasqal_cloud`.
```

But just so you know, it doesn't work when doing `from pulser.backends import X`

```
Python 3.10.19 (main, Feb 12 2026, 00:36:33) [Clang 21.1.4 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from pulser.backends import RemoteFreeBackend
ImportError: cannot import name 'RemoteFreeBackend' from 'pulser.backends' (/path/Pulser/pulser-core/pulser/backends.py)

>>> from pulser.backends import EmuFreeBackend
ImportError: cannot import name 'EmuFreeBackend' from 'pulser.backends' (/path/Pulser/pulser-core/pulser/backends.py)

>>> from pulser.backends import EmuFreeBackendV2
<stdin>:1: DeprecationWarning: 'EmuFreeBackendV2' was renamed to 'RemoteFreeBackend'. Please use 'RemoteFreeBackend' from now on.
ImportError: cannot import name 'EmuFreeBackendV2' from 'pulser.backends' (/path/Pulser/pulser-core/pulser/backends.py)
```

dev_requirements.txt
```diff
@@ -14,6 +14,6 @@ pre-commit
 # tutorials
 nbmake ~= 1.5
 notebook
-pulser-pasqal >= 0.20.8
+pasqal-cloud @ git+https://github.com/pasqal-io/pasqal-cloud@ap/1822-rename-classes#egg=pasqal-cloud&subdirectory=pasqal-cloud
```
Fenkiou: `>= 0.23.0` once released

setup.py
```diff
@@ -27,7 +27,7 @@
 
 # Pulser packages not pinned to __version__
 requirements = [
-    "pulser-pasqal >= 0.20.8",
+    "pasqal-cloud @ git+https://github.com/pasqal-io/pasqal-cloud@ap/1822-rename-classes#egg=pasqal-cloud&subdirectory=pasqal-cloud",
```
Fenkiou: `>= 0.23.0` once released

docs/source/noise_model.ipynb
```diff
@@ -123,7 +123,7 @@
...
+    "Noise parameters are specific to each QPU. To access a particular QPU's specs you'll need a **remote connection** (like [PasqalCloudConnection](./apidoc/_autosummary/pasqal_cloud.PasqalCloudConnection.rst#pasqal_cloud.PasqalCloudConnection)). With it, you can obtain the list of available QPUs through `connection.fetch_available_devices()`. Through this method, you can get the `Device` associated with the QPU. The [Device](./hardware.ipynb) optionally stores a `NoiseModel` in its `noise_model` attribute. When present, you can take into account the effect of its parameters during the design of your sequence. For this, have a look at its `summary()` or `get_noise_table()` methods. "
```
Fenkiou: make read the doc build fail: https://app.readthedocs.org/projects/pulser/builds/33003593/, I'll have to link to something else

tutorials/advanced_features/Execution on a QPU.ipynb
```diff
@@ -32,11 +32,11 @@
     "    give us a register from an automatically generated register layout that fits our desired register while obeying the device constraints. \n",
     "\n",
     "\n",
-    "Let us first create the `PasqalCloud` connection and use it to fetch the `FRESNEL_CAN1` `Device`.\n",
+    "Let us first create the `PasqalCloudConnection` connection and use it to fetch the `FRESNEL_CAN1` `Device`.\n",
     "\n",
     "<div class=\"alert alert-info\">\n",
     "\n",
-    "Note that other connections than `PasqalCloud` exist. Some of the connections maintained by Pasqal are documented as part of the `pulser-pasqal` package [here](../apidoc/pasqal.rst).\n",
+    "Note that other connections than `PasqalCloudConnection` exist. Some of the connections maintained by Pasqal are documented as part of the `pasqal-cloud` package [here](../apidoc/pasqal.rst).\n",
```
Fenkiou: I should point to pasqal-cloud doc instead now

pulser-core/pulser/backends.py
HGSilveri: That's unexpected... And unfortunately not something we can accept. I've been trying to figure out what could cause this but no luck so far

HGSilveri: One thing I found: I was also bumping into this `ImportError` with MPSBackend. When I tried to import directly I noticed that there was erroring too due to an issue with my `emu_mps` version and once I upgraded the issue went away.

If you just run e.g.
```
import importlib

getattr(importlib.import_module("pasqal_cloud"), "RemoteFreeBackend")
```
does it work?

Fenkiou: Behavior seems constant when you change the `AttributeError` below by any other exception (eg `ValueError`, `ImportError`, etc..)

Maybe that make sense, attribute not part of `_BACKENDS` -> `AttributeError`, otherwise, it is but package not installed -> `ImportError`

Fenkiou: @HGSilveri I managed to always get an `ImportError` with https://github.com/pasqal-io/Pulser/pull/1063/changes/6c739b4a0d97869406a0b70e313eceac2790c336, but this is not doable to always get an `AttributeError`, the importer somehow handles it to bubble up an `ImportError` (due to `from X import Y` statement).

As said, this is unrelated to my changes, so if you consider this a breaking change, we can handle that in another PR. I'm pretty sure there is a hacky way to detect if user is doing `from X import Y` vs `import X; X.Y` at runtime so we could raise an `AttributeError`/`ImportError` based on that but that's  the best I can do right now. :)

docs/source/apidoc/pasqal.rst
Fenkiou: done 👍

docs/requirements.txt
Fenkiou: Nope I removed it :)

pulser-core/pulser/backends.py
HGSilveri: I guess just changing the `AttributeError` to `ImportError` is technically a breaking change, so how about we revert that here and handle that later? I'm sorry for the time you wasted on this though

docs/source/apidoc/pasqal.rst
```diff
@@ -1,32 +1,6 @@
...
-
-Deprecated Backends
---------------------
-
-.. autosummary::
-   :toctree: _autosummary
-
-   EmuFreeBackend
-   EmuTNBackend
+    This package is deprecated and replaced by `pasqal-cloud <https://pypi.org/project/pasqal-cloud/>`_, see the `migration path <https://pasqal-io.github.io/pasqal-cloud/migrating-from-pulser-pasqal/>`_.
```
HGSilveri: - Wouldn't it be better to point to the `pasqal-cloud` docs instead of the package?
- Is the migration path going to be only on this page or in the docs portal too?

In general, I think that redirecting to the official Pasqal docs portal is a good practice to incentivize its use

pulser-core/pulser/backends.py
Fenkiou: No worries, i'll revert that 👍

docs/source/apidoc/pasqal.rst
Fenkiou: you're right I'll change, doc portal is deployed now so it's viewable there

pulser-core/pulser/backends.py
Fenkiou: Btw, in the doc: https://docs.pasqal.com/pulser/apidoc/_autosummary/pulser.backends/#_top we show the way that works :)

docs/source/apidoc/pasqal.rst
Fenkiou: changed

pulser-core/pulser/backends.py
Fenkiou: reverted :)

tutorials/advanced_features/Execution on an Emulator.ipynb
```diff
@@ -667,8 +667,8 @@
     "\n",
     "Now that we understand the behaviour of our sequence, and how to use emulator backends locally, let's run the sequence on a pair of remote backends\n",
     "\n",
-    "- `EmuFreeBackendV2` (from `pulser_pasqal`): Executes the sequence on `QutipBackendV2` remotely in the cloud. \n",
-    "- `EmuMPSBackend` (from `pulser_pasqal`): Executes the sequence on `MPSBackend` remotely in the cloud.\n",
+    "- `EmuFreeBackend` (from `pasqal_cloud`): Executes the sequence on `QutipBackendV2` remotely in the cloud. \n",
+    "- `EmuMPSBackend` (from `pasqal_cloud`): Executes the sequence on `MPSBackend` remotely in the cloud.\n",
```
HGSilveri:
```suggestion
    "- `RemoteEmuFreeBackend` (from `pasqal_cloud`): Executes the sequence on `QutipBackendV2` remotely in the cloud. \n",
    "- `RemoteMPSBackend` (from `pasqal_cloud`): Executes the sequence on `MPSBackend` remotely in the cloud.\n",
```
These should be changed in the tutorial further down too

HGSilveri (commented):
Almost there :)

tutorials/advanced_features/Execution on an Emulator.ipynb
Fenkiou: damn good catch, should be good now, thx!

HGSilveri (approved):
LGTM, thank you @Fenkiou !

PR #1068, opened 2026-06-17
[DOC] Register-noise dependency on detuning_map_spot_waist

elmerhej:
Adds documentation to the NoiseModel docstring clarifying that when register noise is active and the sequence uses a `DMM` channel, `detuning_map_spot_waist` must also be set.

### DIFF
--- a/pulser-core/pulser/noise_model.py
+++ b/pulser-core/pulser/noise_model.py
@@ -212,6 +212,9 @@ class NoiseModel:
       (2) Off plane standard deviation fluctuation given by:
       :math:`\sigma^z = \frac{\pi}{\lambda}\sqrt{2} w \sigma^{xy}`, where
       :math:`\lambda` is the trap wavelength with a constant value of 0.85 µm.
+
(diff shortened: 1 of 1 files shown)


pulser-core/pulser/noise_model.py
```diff
@@ -212,6 +212,9 @@ class NoiseModel:
       (2) Off plane standard deviation fluctuation given by:
       :math:`\sigma^z = \frac{\pi}{\lambda}\sqrt{2} w \sigma^{xy}`, where
       :math:`\lambda` is the trap wavelength with a constant value of 0.85 µm.
+
+      Note: when combined with a DMM laser channel, register noise also
```

```suggestion
      Note: when combined with a DMM channel, register noise also
```

HGSilveri (approved):
Thanks @elmerhej , I only have one nit

pulser-core/pulser/noise_model.py
elmerhej: Done :+1:

PR #1069, opened 2026-06-17
Relax requirement of non-zero weights in WeightMap to a warning

HGSilveri:
This requirement that at least one weight must be non-zero in a `WeightMap` was introduced in 1.8 and was unecessarily strict. A user requested it to be relaxed to a warning, so here it is


a-corni (approved):
LGTM !

PR #1070, opened 2026-06-18
Deprecate QutipEmulator.run() and add QutipBackendV2.run_from_sequence_samples()

### LINKED ISSUE #901: Deprecate QutipEmulator.run()
We should discourage direct usage of `QutipEmulator` in favour of `QutipBackendV2`. However, `QutipEmulator` still has useful features, so I propose we deprecate only `QutipEmulator.run()`

a-quelle-pasqal:
The new `QutipBackendV2.run_from_sequence_samples` is a static method because that was the easiest way to get everything to work. The core issue to circumvent is that `EmulatorBackend` requires a `Sequence` so I almost had to make it static.

### DIFF
--- a/pulser-simulation/pulser_simulation/qutip_backend.py
+++ b/pulser-simulation/pulser_simulation/qutip_backend.py
@@ -147,18 +147,16 @@ class QutipBackendV2(EmulatorBackend):
     ) -> None:
         """Initializes the backend."""
         super().__init__(sequence, config=config, mimic_qpu=mimic_qpu)
-        noise_model: None | NoiseModel = None
-        if self._config.prefer_device_noise_model:
-            noise_model = sequence.device.noise_model
-        noise_model = noise_model or self._config.noise_model
+
         self._sim_obj = QutipEmulator.from_sequence(
             sequence,
             sampling_rate=self._config.sampling_rate,
-            noise_model=noise_model,
+            noise_model=self._get_noise_model(self._config, sequence.device),
             with_modulation=self._config.with_modulation,
             solver=self._config.solver,
             n_trajectories=self._config.n_trajectories,
         )
+
         self._sim_obj.set_evaluation_times(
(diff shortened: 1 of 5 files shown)


pulser-simulation/pulser_simulation/qutip_backend.py
```diff
@@ -166,19 +171,88 @@ def __init__(
...
+        )
+
+    @staticmethod
+    def run_from_sequence_samples(
+        sequence_samples: SequenceSamples,
+        register: BaseRegister,
+        device: BaseDevice,
+        *,
+        config: EmulationConfig | None = None,
+    ) -> Results:
```
HGSilveri, line 201: I'm not a fan of how this goes against the Backend interface but I think I can live with it.

pulser-simulation/pulser_simulation/qutip_backend.py
```diff
@@ -166,19 +171,88 @@ def __init__(
...
+        }
+
+        return QutipBackendV2._run_raw(
+            sim_obj, register, config, qutip_options
+        )
+
+    @staticmethod
+    def _run_raw(
+        sim_obj: QutipEmulator,
+        register: BaseRegister,
```
HGSilveri: Why is the register an argument and not taken from `sim_obj`? What am I missing?

HGSilveri (commented):
I think this will do

pulser-simulation/pulser_simulation/qutip_backend.py
a-quelle-pasqal: I think you're missing my ability to overlook the obvious. I removed the argument.

HGSilveri (approved):
LGTM!

HGSilveri:
@a-quelle-pasqal Do you want someone from @pasqal-io/emulation-team to review or do we merge?

PR #1075, opened 2026-07-21
Prevent evaluation time collisions due to rounding

### LINKED ISSUE #1023: Make sure evaluation times are not lost nor duplicated in `QutipBackendV2`
#1022 outlined and fixed an issue where there was duplication of an evaluation time due to a rounding error. 

However, this fix merely eliminated this error source, it does not guarantee that errors of the same type won't appear from other sources or in different circumstances. For that effect, proper safeguards should be put in place instead.

a-quelle-pasqal:
Since `QutipBackendV2` uses `QutipEmulator` under the hood, it needs to convert `Obervable` evaluation times to the format required by `QutipEmulator`, and then back again at the end when storing the result in `Results`. Due to floating point arithmetic, the time stored in `Results` is not necessarily identical to that asked for in the result (it can differ by something of order machine precision). Since observable times have to be unique, this can cause collisions.

I solve the issue by requiring the evaluation times in an `Observable` or in `EmulationConfig.default_evaluation_times` to mutually differ by more than `1e-12`, which prevents colisions due to rounding, being orders of magnitude larger than machine precision.

Although it is possible to uniquely pair an obtained evaluation time to a requested one, I decided not to implement this, and allow for potential rounding errors to show in `Results`. The natural places to implement this would be in the `Obervable` or by post-processing the `Results` in `QutipBackendV2`. I'm afraid the first will have negative effects on the other emulators, and the second will lead to more code to maintain for a very marginal benefit.

Closes #1023

### DIFF
--- a/pulser-core/pulser/backend/observable.py
+++ b/pulser-core/pulser/backend/observable.py
@@ -31,7 +31,7 @@ if TYPE_CHECKING:
     from pulser.backend.results import Results
 
 
-def fuzzy_unique_sorted(sorted: np.ndarray, tolerance: float) -> bool:
+def _fuzzy_unique_sorted(sorted: np.ndarray, tolerance: float) -> bool:
     return not np.any(np.abs(sorted[:-1] - sorted[1:]) < tolerance)
 
 
@@ -214,7 +214,9 @@ class Observable(Callback):
             )
         # larger than machine precision, but effectively 0
         time_tolerance = 1e-12
-        unique_eval_times = fuzzy_unique_sorted(eval_times_arr, time_tolerance)
+        unique_eval_times = _fuzzy_unique_sorted(
+            eval_times_arr, time_tolerance
+        )
         if not unique_eval_times:
             raise ValueError(
                 f"Evaluation times must be unique up to {time_tolerance} but "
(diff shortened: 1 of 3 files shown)


pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -866,7 +866,7 @@ def run(
                 tuple(self._hamiltonian_data.register.qubits),
                 self._meas_basis,
                 total_count[ind],
-                evaluation_time=t / self._tot_duration * 1e3,
+                evaluation_time=t / (self._tot_duration * 1e-3,),
```
HGSilveri: I'm confused about how this division by a tuple is even working. Maybe it's numpy that's automatically casting to an array, but we should still avoid it

```suggestion
                evaluation_time=t / (self._tot_duration * 1e-3),
```

pulser-core/pulser/backend/observable.py
```diff
@@ -208,10 +212,14 @@ def _validate_eval_times(
                 "All evaluation times must be between 0. and 1. "
                 f"Instead, got {evaluation_times!r}."
             )
-        unique_eval_times = np.unique(eval_times_arr)
-        if unique_eval_times.size < eval_times_arr.size:
+        # larger than machine precision, but effectively 0
+        time_tolerance = 1e-12
```
HGSilveri: Let's make this a constant placed at the top of the file

pulser-simulation/pulser_simulation/simulation.py
a-quelle-pasqal: I am so, so sorry about this. The surprise of it made me laugh though.

HGSilveri (approved):

PR #1076, opened 2026-07-21
Add aggregators for (mean, std)

### LINKED ISSUE #1026: Natively support the calculation of the standard deviation alongside the mean in Results aggregation
As outlined in https://github.com/pasqal-io/Pulser/issues/969#issuecomment-4053945737, we currently don't have a built-in way to calculate the standard deviation of a given aggregated observable.

I think this is common enough that we should support it natively. A few things we could do are:

- Create a `AggregationMethod.MEAN_STD` and associated aggregator
- Allow for the desired aggregation method to be defined at the `Observable` instance level (rather than a `ClassVar` as is done currently) and include that information in the serialized observable

This way, a user could specify, for each observable, what aggregation method they desire. In the absence of an input, we would of course default to the current default aggregation method.

a-quelle-pasqal:
I've added inbuilt aggregation functions for the sample standard deviation of the list of results, and for a tuple containing both the mean and the std. I've also added tests.

To make the new functions easier to use, I've widened the supported types for `aggregation_functions`  in `Results.aggregate`. Before it was only a `Callable` but now it also takes the existing `AggregationMethod` enum. This makes it easier to use the added aggregation functions to override the default behaviour.

Closes #1026

### DIFF
--- a/pulser-core/pulser/backend/aggregators.py
+++ b/pulser-core/pulser/backend/aggregators.py
@@ -35,7 +35,7 @@ T = TypeVar(
 )
 
 
-def _validate_values(values: list[T]) -> None:
+def _assert_values_not_empty(values: list[T]) -> None:
     """Validate that ``values`` is a non-empty list.
 
     Args:
@@ -90,7 +90,7 @@ def _std_aggregator(
     Returns:
         The standard deviation over the first dimension of the given values.
     """
-    _validate_values(values)
+    _assert_values_not_empty(values)
 
     elt = values[0]
 
@@ -101,17 +101,15 @@ def _std_aggregator(
         return cast(np.ndarray, np.stack(values).std(axis=0, ddof=1))
 
     if isinstance(elt, float):
-        return cast(
-            float, np.std(values, ddof=1)
-        )  # this would have type np.floating
+        return float(np.std(values, ddof=1))  # instead of np.floating
 
     if isinstance(elt, complex):
-        return cast(
-            complex, np.std(values, ddof=1)
-        )  # this would have type np.complexfloating
+        return complex(np.std(values, ddof=1))  # np.complexfloating
 
     if not isinstance(elt, Sequence):
-        raise ValueError("Cannot process this type of data.")
+        raise ValueError(
+            f"Std aggregator cannot process data of type {type(elt)}."
+        )
 
     _validate_sequence_elements(elt)
 
@@ -131,7 +129,7 @@ def _mean_aggregator(
     Returns:
         The average over the first dimension of the provided results.
     """
-    _validate_values(values)
+    _assert_values_not_empty(values)
 
     elt = values[0]
 
@@ -142,15 +140,15 @@ def _mean_aggregator(
         return cast(np.ndarray, np.stack(values).mean(axis=0))
 
     if isinstance(elt, float):
-        return cast(float, np.mean(values))  # this would have type np.floating
+        return float(np.mean(values))  # this would have type np.floating
 
     if isinstance(elt, complex):
-        return cast(
-            complex, np.mean(values)
-        )  # this would have type np.complexfloating
+        return complex(np.mean(values))  # np.complexfloating
 
     if not isinstance(elt, Sequence):
-        raise ValueError("Cannot process this type of data.")
+        raise ValueError(
+            f"Mean aggregator cannot process data of type {type(elt)}."
+        )
 
     _validate_sequence_elements(elt)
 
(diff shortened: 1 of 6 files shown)


pulser-core/pulser/backend/aggregators.py
```diff
@@ -35,6 +35,89 @@
 )
 
 
+def _validate_values(values: list[T]) -> None:
```
HGSilveri: nit: Maybe `_check_values_not_empty()` would be a more explicit name?

pulser-core/pulser/backend/aggregators.py
```diff
@@ -35,6 +35,89 @@
...
+    if pm.AbstractArray.has_torch() and isinstance(elt, pm.torch.Tensor):
+        return pm.torch.stack(values).std(dim=0)
+
+    if isinstance(elt, np.ndarray):
+        return cast(np.ndarray, np.stack(values).std(axis=0, ddof=1))
+
+    if isinstance(elt, float):
+        return cast(
+            float, np.std(values, ddof=1)
+        )  # this would have type np.floating
```
HGSilveri: Shouldn't you truly cast instead?

```suggestion
        return float(np.std(values, ddof=1))
```

pulser-core/pulser/backend/aggregators.py
```diff
@@ -35,6 +35,89 @@
...
+
+    if isinstance(elt, complex):
+        return cast(
+            complex, np.std(values, ddof=1)
+        )  # this would have type np.complexfloating
```
HGSilveri: Same here

```suggestion
    if isinstance(elt, complex):
        return complex(np.std(values, ddof=1))
```

pulser-core/pulser/backend/aggregators.py
```diff
@@ -35,6 +35,89 @@
...
+
+    if not isinstance(elt, Sequence):
+        raise ValueError("Cannot process this type of data.")
```
HGSilveri: Our future selves will thank us:


```suggestion
        raise ValueError(f"The STD aggregator cannot process data of type {type(elt)}.")
```

pulser-core/pulser/backend/aggregators.py
```diff
@@ -63,31 +143,37 @@ def _mean_aggregator(
...
         return cast(float, np.mean(values))  # this would have type np.floating
+
     if isinstance(elt, complex):
         return cast(
             complex, np.mean(values)
         )  # this would have type np.complexfloating
 
     if not isinstance(elt, Sequence):
-        raise ValueError("Cannot average this type of data.")
+        raise ValueError("Cannot process this type of data.")
```
HGSilveri: nit: Same request for a more informative error message

pulser-core/pulser/backend/results.py
```diff
@@ -263,7 +263,7 @@ def from_abstract_repr(cls, repr: str) -> Results:
     def aggregate(
         cls,
         results_to_aggregate: typing.Sequence[Results],
-        **aggregation_functions: Callable[[Any], Any],
+        **aggregation_functions: Callable[[Any], Any] | AggregationMethod,
```
HGSilveri, line 266: Does this really "just work"? I was expecting some sort of type check inside the function

pulser-core/pulser/backend/aggregators.py
```diff
@@ -100,4 +186,6 @@ def _bag_union_aggregator(
 AGGREGATOR_MAPPING: dict[AggregationMethod, Callable] = {
     AggregationMethod.MEAN: _mean_aggregator,
     AggregationMethod.BAG_UNION: _bag_union_aggregator,
+    AggregationMethod.STD: _std_aggregator,
```
HGSilveri: Does it really make sense to have STD as a standalone aggregator? I feel like it's not really an aggregation of the results like bag union or the mean is. 
I would personally find it odd to get a `Results` instance that is giving me the standard deviation when I ask for some observable's result

pulser-core/pulser/backend/results.py
a-quelle-pasqal: It already worked because it had to already be able to take the default aggregator, which is specified as an enum, which it does roughly by merging the two variables into one with the union type anyway.

pulser-core/pulser/backend/aggregators.py
a-quelle-pasqal: It's kind of free to add it though, so while I entirely agree with you, I still thought I'd add it. While I'm usually in favour of not writing code just to keep the code smaller and simpler, in this case it felt like a freebie complexity wise due to the hierarchical implementation I did for the tuple.

HGSilveri: I get that... But if I put myself in the shoes of a user, trying to figure out from limited context what an "aggregation" should be and I see the options are [BAG_UNION, MEAN, STD], I'm not sure I would reach the conclusion that it is an aggregation of the various results into a single value, because to me an STD is not a proper representation of aggregated results, it is just a quantity you can extract from them.

a-quelle-pasqal: As you wish.

HGSilveri (approved):
Great, thanks!

PR #1077, opened 2026-07-22
Define `Results.from_final_bitstrings()` + Deprecate `Result` and its subclasses

### LINKED ISSUE #1073: Replace `SampledResult` with `Results.from_final_bitstrings()`
- Replicate the `SampledResult` specific logic in a `Results.from_final_bitstrings()` classmethod
- Deprecate `SampledResult()` and its custom methods

HGSilveri:
This is introduced in order to replicate the essential functions of `SampledResult`. It also deprecates `Result` and its subclasses, herein `SampledResult`.

Closes #1073.


AurelienNober:
LGTM (and thank you). One comment: should we add warnings for SampledResult deprecation?

__________________

Also, if we plan to include partial result as well, does it make sense to handle as well a kwarg `imperfect_bitstrings`? That we can ease a new serialization in the job api of pasqOS returning `Results` in the exact same format as emulators do.
```
results = Results.from_final_bitstrings(......, self.result["counter"], imperfect_bitstrings=self.result["partial_result"]).to_abstract_repr()
``` 
somewhere [here](https://gitlab.pasqal.com/software/pasqos/pasqos_main/-/blob/develop/backend/pasqos/pasqos_api/schemas/job.py#L37)
It would act as a validation layer and this is currently what we end up doing in all the connections/sdks anyway.

WDYT @MatthieuMoreau0 since I think you had that or something similar in mind?

HGSilveri:
> LGTM (and thank you). One comment: should we add warnings for SampledResult deprecation?

Yes, for sure. ~I will do this on separate PR.~ Edit: Ended up doing that here.

> Also, if we plan to include partial result as well, does it make sense to handle as well a kwarg `imperfect_bitstrings`? That we can ease a new serialization in the job api of pasqOS returning `Results` in the exact same format as emulators do.

These imperfect results are not yet properly supported in the `Results` instance, I would like to do that with a bit more care later on. They can safely be added to this method at that point so I refrained from rushing the implementation here.

pulser-simulation/pulser_simulation/simulation.py
```diff
@@ -861,15 +863,17 @@ def run(
...
-                self._meas_basis,
-                total_count[ind],
-                evaluation_time=t / self._tot_duration * 1e3,
-            )
-            for ind, t in enumerate(self._eval_times_array)
-        ]
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", category=DeprecationWarning)
+            results = [
+                SampledResult(
```
Fenkiou, line 869: why not using `Results.from_final_bitstrings` here?

tests/test_backends.py
```diff
@@ -61,11 +63,14 @@ def test_removed_deprecated_backends(backend_name):
     "backend_name", ["EmuFreeBackendV2", "EmuMPSBackend", "EmuSVBackend"]
 )
 def test_renamed_backends(backend_name):
+    has_pasqal_cloud = importlib.util.find_spec("pasqal_cloud") is not None
```
Fenkiou, line 66: I suppose it's because the test doesn't fail locally because you have the package installed right?

pulser-simulation/pulser_simulation/simulation.py
HGSilveri: Because `NoisyResults` (the legacy class this is given too) relies on the custom methods of `SampledResult`

tests/test_backends.py
HGSilveri: Yes, indeed

Fenkiou (approved):
Looking good, thanks!

MatthieuMoreau0 (approved):
LGTM

We discussed offline with Henrique. I will create a Notion page to dump everything.
Here are the immediate next steps to deprecate `SampledResult`.

If we switch from using `SampledResult` to `Results` for qpu results using this new `from_final_bitstrings`method then this could break some user scripts since SampledResults has some custom attributes and methods. 

We anticipate the impact to be low because it seems most of these are not directly used in our tutorials, but we need to confirm that. We found for example that `bitstring_counts` attribute from SampledResult is used in our OVH [tutorial](https://github.com/pasqal-io/pasqal-cloud/blob/main/pasqal-cloud/tutorials/ovh/AFM_state_preparation_OVH.ipynb). So we need a proper plan. 

Here is what we suggested:
- parse our docs and qoolqit to make sure custom SampledResult attributes/methods are not used largely. If they are, we will need to reassess the plan. Otherwise see next steps
- add a temporary alias of the `bitstring_counts` attribute on the `Results` class with a deprecation warning which nudges the user towards the `final_bitstrings` property instead. For other custom methods and attributes, intercept the call using `get_attr` to display a custom error message for the users.
- update the OVH notebook to use `final_bitstrings` in place of `bitstring_counts`. Notify OVH to update the notebook on their side
- fix other occurences in our docs and qoolqit/QApps
- update the different connections to parse QPU results with `Results.from_final_bitstrings`. At that point I guess we can consider SampledResult fully deprecated and we can plan getting rid of it. Some scripts may be disrupted but I expect the impact to be minimal at this stage

Medium term we should update the QPU to return serialised results which will include imperfect shots. I'll write this down on Notion and share it with you @AurelienNober

a-corni:
@MatthieuMoreau0 SampledResults is the current outcome of pulser-myqlm's [PulserQLMConnection](https://github.com/pasqal-io/Pulser-myQLM/blob/9462a2a9dec557c3c2fc2a1f17fe6a9791548139/pulser_myqlm/connection.py#L285C31-L289`). It seems that [some scripts](https://github.com/pasqal-io/Pulser-myQLM/blob/9462a2a9dec557c3c2fc2a1f17fe6a9791548139/tutorials/AFM_PulserQLMConnection.py#L59) that were delivered with pulser-myqlm use `SampledResults.bitstring_counts`, so it is possible that some non-regression tests uses this - and needs to be modified prior to proper deprecation.

a-corni (approved):
LGTM! As pointed out by @MatthieuMoreau0, more work is needed before properly deleting Result and its subclasses, but this is the first step towards this direction.

a-corni (approved):
LGTM

PR #1079, opened 2026-07-24
Let the user override the default aggregation method in an Observable

a-quelle-pasqal:
- Made `default_aggregation_method` a property, default value is the old hard-coded value
- Added the property to the (de)serialization.

### DIFF
--- a/pulser-core/pulser/backend/default_observables.py
+++ b/pulser-core/pulser/backend/default_observables.py
@@ -40,8 +40,8 @@ class StateResult(Observable):
         tag_suffix: An optional suffix to append to the tag. Needed if
             multiple instances of the same observable are given to the
             same EmulationConfig.
-        aggregation_method: How to combine the values of this observable
-            from multiple results.
+        default_aggregation_method: How to combine the values of this
+            observable from multiple results.
     """
 
     def __init__(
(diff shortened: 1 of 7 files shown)


pulser-core/pulser/backend/default_observables.py
```diff
@@ -58,8 +74,6 @@ def apply(self, *, state: StateType, **kwargs: Any) -> StateType:
         """Calculates the observable to store in the Results."""
         return copy.deepcopy(state)
 
-    default_aggregation_method = AggregationMethod.SKIP_WARN
```
HGSilveri, line 61: If you avoid the change from `default_aggregation_method` -> `aggregation_method`, you also avoid the breaking change from removing access to `.default_aggregation_method`.
I think keeping the `default_` is warranted too, since it can be overriden in the aggregation

a-quelle-pasqal: I did as you asked. As a small downside, black and flake8 got into a fight, where the constructor arg 
`default_aggregation_method: AggregationMethod  = AggregationMethod.BAG_UNION` is too long to fit on a single line, so flake8 would complain, but if I split it into multiple lines, black would undo it. In the end I added a tag for black to skip this line.

HGSilveri (approved):

PR #1080, opened 2026-07-27
Determine `interaction_coeff_xy` from the Rydberg level

### LINKED ISSUE #1036: Calculate the `interaction_coeff_xy` from the `rydberg_level`
Similarly to what is done for the Ising interaction coefficient, the XY interaction coefficient should be calculated from the `rydberg_level` rather than taken as a parameter of the `Device`.

HGSilveri:
- [x] Adds the file with the C3 coefficients
- [x] Turns `interaction_coeff_xy` into a property with a value based on `rydberg_level`
- [x] Deprecates `interaction_coeff_xy` as an argument but still allows it to be given. When provided, it overrides the value that would be automatically determined
- [x] Adapts existing UTs
- [x] Writes UTs for the changes functionalities  

Closes #1036 .


pulser-core/pulser/devices/interaction_coefficients/__init__.py
```diff
@@ -28,3 +32,9 @@
 ) as f:
     _json_dict = json.load(f)
 c6_dict = {int(key): value for key, value in _json_dict.items()}
+
+with open(
+    PurePath(__file__).parent / "C3_coeffs.json", "r", encoding="utf-8"
+) as f:
+    _json_dict = json.load(f)
+c3_dict = {int(key): value for key, value in _json_dict.items()}
```
a-corni, line 40: nit: Just a format question: Since this is a variable, should we name it C3_DICT ?

pulser-core/pulser/devices/_device_datacls.py
```diff
@@ -626,6 +630,7 @@ def _to_dict(self) -> dict[str, Any]:
     @abstractmethod
     def _to_abstract_repr(self) -> dict[str, Any]:
         defaults = get_dataclass_defaults(fields(self))
+        # Not init_only=True because of 'reusable_channels'
```
a-corni: That comment is not especially clear to me... wdym ?

pulser-core/pulser/devices/interaction_coefficients/__init__.py
HGSilveri: Yes, since they are constants, they should be upper cased. 
I think it's more important that the format is consistent with `c6_dict` though, so I'd go with renaming it to `C6_DICT` too and then we start getting into diminishing returns territory.
I'm inclined to just leave it like this

pulser-core/pulser/devices/_device_datacls.py
HGSilveri: You're right to be confused, it's a cryptic comment. `_params()` has an `init_only: bool = False` flag that selects only the parameters in the init of the class.  In this case, we don't set it to True because `reusable_channels` is not in the init of `BaseDevice` and `Device` but we want it on the abstract repr.
Because of that, we have to manually pop `_custom_interaction_coeff_xy` though

pulser-core/pulser/devices/interaction_coefficients/__init__.py
a-corni: I agree, we could keep having `c6_dict=C6_DICT` to keep `c6_dict` (I think it's used in pulser_myqlm) but that would be eventually too complex for a simple constant...

pulser-core/pulser/devices/_device_datacls.py
```diff
@@ -864,10 +865,12 @@ def wrapped_init(
                 category=DeprecationWarning,
                 stacklevel=2,
             )
-            if not isinstance(interaction_coeff_xy, float):
```
a-corni: Nice improvement 👍

a-corni (approved):
LGTM!

PR #1081, opened 2026-07-27
Fix legacy serialization of a parametrized object with only keyword args

HGSilveri:
I think the title is self explanatory. I'm surprised this went unnoticed all these years!


a-corni (approved):
LGTM! Indeed, impressive it never poped up earlier

PR #1083, opened 2026-07-29
Reveal the signature of Pulse and Waveform classes in auto completion

HGSilveri:
Although not urgent, this was something that had been bothering me for a while. 

For classes that we added a custom `__new__()` method to, before this change, we would get something like this

image

Now, we correctly get

image


a-corni (approved):
LGTM!

PR #1084, opened 2026-07-29
[FIX] flaky dict comparison test_qubit_weight_map

a-corni:
A comparison between two dictionaries was sometimes failing in test_dmm.py/test_qubit_weight_map, due to floating point precision (see an example: https://github.com/pasqal-io/Pulser/actions/runs/30455630995/attempts/1). This could be solved by simply re-running the test. This makes the comparison more robust using pytest.approx https://docs.pytest.org/en/stable/reference/reference.html#pytest-approx.


HGSilveri (approved):
Nice fix, LGTM!

PR #1085, opened 2026-07-29
Make changes to the config schema forwards compatible

HGSilveri:
We forgot to make the changes to the config schema forwards compatible, so that serialized configs with previous versions still pass validation under the latest schema.


a-quelle-pasqal (approved):
I looked at it carefully to really check that this is all that's needed. I think it is, so I approve.

PR #1086, opened 2026-07-30
Fix `$ref` resolution of the `config` and `results` schemas

HGSilveri:
Validating an emulation config or a results object against its schema with `jsonschema` was impossible: it always failed with
`Unresolvable: config-schema.json#/EmulationConfig` (resp.`/results-schema.json#/definitions/Results`) instead of a proper `ValidationError`. Since `validate_abstract_repr()` only falls back to `jsonschema` after `fastjsonschema` rejects an object, this masked the error message of *every* invalid config or results payload.

Two causes, both fixed here:

- `fastjsonschema.compile()` qualifies the `$ref`s of the schemas it compiles with their `$id` **in place**, so building `_FAST_VALIDATORS` mutated the contents of `SCHEMAS` (e.g. `#/EmulationConfig` became `config-schema.json#/EmulationConfig`). The `jsonschema` validators, built  afterwards, then had to resolve refs to documents that were not in the
  `REGISTRY`. It is now given deep copies, both of the schema it compiles and of the ones it pulls through the local `$ref` handler.

- `results-schema.json` declared `"$id": "/results-schema.json"`, with a leading slash that no other schema has, making it unresolvable under its own file name.

The `REGISTRY` is additionally built from all of `SCHEMAS` instead of a hardcoded subset. This is no longer required now that the schemas are left untouched (relative refs resolve against the root document), but it keeps resolution working if refs ever become `$id`-qualified again, and new schemas no longer need to be added by hand.

### DIFF
--- a/pulser-core/pulser/json/abstract_repr/schemas/results-schema.json
+++ b/pulser-core/pulser/json/abstract_repr/schemas/results-schema.json
@@ -1,7 +1,7 @@
 {
     "$ref": "#/definitions/Results",
     "$schema": "http://json-schema.org/draft-07/schema#",
-    "$id": "/results-schema.json",
+    "$id": "results-schema.json",
     "type": "object",
     "definitions":{
         "uuid": {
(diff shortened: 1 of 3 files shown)


a-corni (approved):
LGTM

PR #1087, opened 2026-07-30
Fix: Definition of InterpolatedWaveform with non-integer nanosecond times

a-corni:
Currently, when a user a list of fractional times, they are scaled by `duration-1` (since InterpolatedWaveform defines values at [0; duration-1]) and then rounded to 1. This rounding is wrong, as then the interpolation is not performed exactly on the users input, if some of the times are at a non-integer number of nanoseconds. Especially, if the user provides a list of values to interpolate at some times that might be below 1 nanosecond, the initialization of InterpolatedWaveform fails with a not-so-descriptive error message coming from the interpolating function:
```
InterpolatedWaveform(100, np.linspace(0, 10, 1000))
```
Raises: `ValueError: 'x' must be strictly increasing sequence`

This MR gets rid of the rounding, for users to get the exact behaviour of the Interpolator (the interpolating function is then sampled at integer nanosecond times).

### DIFF
--- a/pulser-core/pulser/waveforms.py
+++ b/pulser-core/pulser/waveforms.py
@@ -863,11 +863,11 @@ class InterpolatedWaveform(Waveform):
         duration: The waveform duration (in ns).
         values: Values of the interpolation points. Must be a list of castable
             to float or a parametrized object.
-        times: Fractions of the total duration (between 0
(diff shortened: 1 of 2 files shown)


HGSilveri (approved):
Good point, LGTM!

PR #1089, opened 2026-07-31
Revert "Determine `interaction_coeff_xy` from the Rydberg level (#1080)"

HGSilveri:
The C3 coefficients used in #1080 led to very dubious results in the XY tutorial.
 We don't have the means to properly vet them right now, so I'm reverting the changes and we can go back to them in the next release.


HGSilveri:
In fact, the issue seems to be in the tutorial itself. I'm adapting it instead

PR #1090, opened 2026-07-31
Adjust XY tutorial for the correct interaction strengths

### LINKED ISSUE #1036: Calculate the `interaction_coeff_xy` from the `rydberg_level`

HGSilveri:
The atom spacings in the XY tutorial were adapted to the previous value of `interaction_coeff_xy`, so they were giving incorrect results upon the addition of the C6 interaction to the XY hamiltionian and the new XY interaction coefficients. 

This tutorial adjusts the spacings to match the [original reference paper](https://arxiv.org/pdf/1408.1055), with seemingly good agreement.

Closes #1036 , which had been reopened in a rush.


a-corni (approved):
LGTM!

PR #1092, opened 2026-08-05
[FIX] Avoid timeouts and allow repetitions in publish.yml

HGSilveri:
Lately the `publish.yml` has been timing out at the "Confirm deployment" stage (cf. https://github.com/pasqal-io/Pulser/actions/runs/31026583078) 

I'm not sure why it's timing out now, but it seems that if we increase the timeout interval it should be ok. 

Also, when the job fails after publishing to `PyPI`, we can't rerun it because if fails at publishing to PyPI. The `skip-existing` option added here should fix that.


a-corni (approved):
LGTM

PR #1093, opened 2026-08-10
Report offending inputs in pulse module and AbstractArray

a-corni:
Finishing MR #1058 

Tackling partially #1057 

- Started tackling the report of offending inputs in pulse.py (3 offending inputs)
- Found some badly written tests. Improved test coverage.
- While doing so, got a hard-to-understand error message in AbstractArray, if the provided "array" can't generate an array or if its data type doesn't match the provided "dtype". Improved this error message. Impacted the error message raised in Variable._assign if the value isn't an array or doesn't have the dtype of the Variable.

Issue #1057, opened 2026-05-07
Improve information content in error messages


HGSilveri:
A lot of error messages have been written assuming that the user would hit it "directly" (i.e. when coding in some interactive environment like a Jupyter notebook). Take for example [this error](https://github.com/pasqal-io/Pulser/blob/develop/pulser-core/pulser/pulse.py#L108):

```
 ValueError("All samples of an amplitude waveform must be greater than or equal to zero.")
```
This is fine if the user just tried to create a `Pulse`. However, if it happens deep inside some function in the code base, potentially coming from the building of some parametrized sequence, it can be harder to tell what is going on. An immediate improvement would be to just add more information in the message, e.g.

```
 ValueError(f"All samples of an amplitude waveform must be greater than or equal to zero; got negative values for waveform {amp_wf}")
```

This alone might help narrow down where the error is coming from.

We should do this judiciously throughout the codebase.

Janvee31:
Hi! I'm new to contributing to Pulser and I'd like to work on improving the contextual information in error messages for this issue. If no one else is currently working on it, could this issue be assigned to me?

a-corni:
Hello @Janvee31,
Welcome! It's a pleasure to receive new contributions! The issue is assigned to you :)

jaewonyun1234:
Hi @HGSilveri, @a-corni — I'd like to help with this one if you're open to it.

Since #1058 tackles the `Pulse` side, would you be willing to assign me `pulser-core/pulser/waveforms.py` as a separate piece? Happy to take a different module if that fits better.

There are 7 raise sites in that file (6 distinct messages) that state a rule without reporting the offending input, e.g.:

- `"The step of the slice must be None or 1."` — doesn't report `s.step`
- `"Needs at least two waveforms to form a CompositeWaveform."` — doesn't report how many were given
- ``"`times` must be an array of non-repeating values."`` — doesn't report which values repeat

`"The maximum value and the area must have matching signs."` also appears in both `BlackmanWaveform.from_max_val` and `KaiserWaveform.from_max_val`. Since both are decorated with `@parametrize`, in a parametrized sequence this surfaces at `build()` time, where the message alone doesn't say which waveform produced it.

I'd follow the style already in the file (the duration and `beta` checks both interpolate the value), and append detail after the existing sentence rather than reword in place, so the `match=` assertions in `tests/test_waveforms.py` keep passing.

a-corni:
Hello @jaewonyun1234,
[...]

PR #1094, opened 2026-08-11
Report offending input in waveforms.py error messages

jaewonyun1234:
Adds the offending input to the error messages in `waveforms.py` that
previously stated only the rule, following the pattern from #1093.

The existing message text is left intact and the new details appended, so
the `match=` assertions in `tests/test_waveforms.py` keep passing without
modification.

`BlackmanWaveform.from_max_val` and `KaiserWaveform.from_max_val` shared an
identical message; both now report the class via `cls.__name__`, so the two
can be told apart.

Partially addresses #1057.

### DIFF
--- a/pulser-core/pulser/waveforms.py
+++ b/pulser-core/pulser/waveforms.py
@@ -335,7 +335,10 @@ class Waveform(ABC):
 
     def _check_slice(self, s: slice) -> slice:
         if s.step is not None and s.step != 1:
-            raise IndexError("The step of the slice must be None or 1.")
+            raise IndexError(
+                "The step of the slice must be None or 1; "
+                f"got {s} with step {s.step}."
+            )
 
         # Transform start and stop indexes into positive or null values
         # since they can be omitted (None) or negative (end-indexing)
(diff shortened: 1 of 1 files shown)


jaewonyun1234:
@a-corni ready for review whenever you have time — thanks for the guidance on the issue.

pulser-core/pulser/waveforms.py
```diff
@@ -335,7 +335,10 @@ def _check_index(self, i: int) -> int:
 
     def _check_slice(self, s: slice) -> slice:
         if s.step is not None and s.step != 1:
-            raise IndexError("The step of the slice must be None or 1.")
+            raise IndexError(
+                "The step of the slice must be None or 1; "
+                f"got {s} with step {s.step}."
```
a-corni, line 340: nit: Maybe worth mentioning explicitely "slice"?
```suggestion
                f"slice {s} has step {s.step}."
```
or 
```suggestion
                f"got slice {s} with step {s.step}."
```

a-corni (commented):
Thanks @jaewonyun1234 for your work! It looks good to me, I just have one improvement to suggest.

pulser-core/pulser/waveforms.py
jaewonyun1234: Thanks! Before applying — `f"{s}"` on a slice already renders as
`slice(0, 6, 2)`, so both suggestions end up repeating the word:

- current: `The step of the slice must be None or 1; got slice(0, 6, 2) with step 2.`
- suggestion 1: `The step of the slice must be None or 1; slice slice(0, 6, 2) has step 2.`
- suggestion 2: `The step of the slice must be None or 1; got slice slice(0, 6, 2) with step 2.`

So "slice" is already in there via the repr. Happy to apply either if you'd
still prefer it stated explicitly — just wanted to check you'd seen how it
comes out.

a-corni: Sorry, I hadn't checked, what you have is great then!

a-corni (approved):
LGTM! Thanks for the great work @jaewonyun1234!

PR #1095, opened 2026-08-14
Report offending input in register error messages

jaewonyun1234:
Same as #1094, applied to `base_register.py`.

Nine messages updated: `__init__`, `_init_kwargs`, `find_indices`,
`from_coordinates`, four in `_validate_layout`, and `define_detuning_map`.

Also fixed a missing space that printed as `Label length (3) does notmatch
number of coordinates (2)`.

I left `"Cannot create a Register with an empty qubit dictionary."` alone. The
dictionary is empty, so there's nothing to print.

One thing I noticed: `trap_coords[trap_id]` on line 229 isn't bounds checked, so
a bad trap id gives a numpy error instead of a Pulser one.

```
>>> Register(qubits, layout=layout, trap_ids=(0, 99))
IndexError: index 99 is out of bounds for axis 0 with size 4
```

`RegisterLayout.define_register` catches this but the `Register` constructor
skips that check. I didn't fix it here since it changes behaviour, not wording.

Partially addresses #1057.

### DIFF
--- a/pulser-core/pulser/register/base_register.py
+++ b/pulser-core/pulser/register/base_register.py
@@ -68,7 +68,8 @@ class BaseRegister(ABC, CoordsCollection):
         if not isinstance(qubits, dict):
             raise TypeError(
                 "The qubits have to be stored in a dictionary "
-                "matching qubit ids to position coordinates."
+                "matching qubit ids to position coordinates; "
+                f"got {type(qubits)}."
             )
         if not qubits:
             raise ValueError(
@@ -98,7 +99,7 @@ class BaseRegister(ABC, CoordsCollection):
             if kwargs.keys() != {"layout", "trap_ids"}:
                 raise ValueError(
                     "If specifying 'kwargs', they must only be 'layout' and "
-                    "'trap_ids'."
+                    f"'trap_ids'; got {sorted(kwargs)}."
                 )
             layout: RegisterLayout = kwargs["layout"]
             trap_ids: tuple[int, ...] = tuple(kwargs["trap_ids"])
@@ -149,7 +150,9 @@ class BaseRegister(ABC, CoordsCollection):
         if not set(id_list) <= set(self.qubit_ids):
             raise ValueError(
                 "The IDs list must be selected among the IDs of the register's"
-                " qubits."
+                " qubits; "
+                f"{sorted(set(id_list) - set(self.qubit_ids))} not in "
+                f"{list(self.qubit_ids)}."
             )
         return [self.qubit_ids.index(id_) for id_ in id_list]
 
@@ -190,14 +193,15 @@ class BaseRegister(ABC, CoordsCollection):
             if labels is not None:
                 raise NotImplementedError(
                     "It is impossible to specify a prefix and "
-                    "a set of labels at the same time"
+                    "a set of labels at the same time; "
+                    f"got prefix={prefix!r} and labels={list(labels)}."
                 )
 
         elif labels is not None:
             if len(coords_) != len(labels):
                 raise ValueError(
-                    f"Label length ({len(labels)}) does not"
-                    f"match number of coordinates ({len(coords_)})"
+                    f"Label length ({len(labels)}) does not "
+                    f"match number of coordinates ({len(coords_)})."
                 )
             qubits = dict(zip(cast(Iterable, labels), coords_))
         else:
@@ -212,15 +216,21 @@ class BaseRegister(ABC, CoordsCollection):
         if register_layout.dimensionality != self.dimensionality:
             raise ValueError(
                 "The RegisterLayout dimensionality is not the same as this "
-                "register's."
+                f"register's; layout is {register_layout.dimensionality}D "
+                f"and register is {self.dimensionality}D."
             )
         if len(set(trap_ids)) != len(trap_ids):
-            raise ValueError("Every 'trap_id' must be a unique integer.")
+            repeated = sorted({t for t in trap_ids if trap_ids.count(t) > 1})
+            raise ValueError(
+                "Every 'trap_id' must be a unique integer; "
+                f"found repeated ids {repeated} in {list(trap_ids)}."
+            )
 
         if len(trap_ids) != len(self._ids):
             raise ValueError(
                 "The amount of 'trap_ids' must be equal to the number of atoms"
-                " in the register."
+                f" in the register; got {len(trap_ids)} trap_ids "
+                f"for {len(self._ids)} atoms."
             )
 
         for reg_coord, trap_id in zip(
@@ -229,7 +239,9 @@ class BaseRegister(ABC, CoordsCollection):
             if np.any(reg_coord != trap_coords[trap_id]):
                 raise ValueError(
                     "The chosen traps from the RegisterLayout don't match this"
-                    " register's coordinates."
+                    f" register's coordinates; trap {trap_id} is at "
+                    f"{trap_coords[trap_id].tolist()} but the register "
+                    f"has {reg_coord.tolist()}."
                 )
 
     def define_detuning_map(
@@ -251,7 +263,9 @@ class BaseRegister(ABC, CoordsCollection):
         if not set(detuning_weights.keys()) <= set(self.qubit_ids):
             raise ValueError(
                 "The qubit ids linked to detuning weights have to be defined"
-                " in the register."
+                " in the register. Got "
+                f"{sorted(set(detuning_weights) - set(self.qubit_ids))}, "
+                f"which are not in {list(self.qubit_ids)}."
             )
         return DetuningMap(
             pm.vstack(
--- a/pulser-core/pulser/register/base_register.py
+++ b/pulser-core/pulser/register/base_register.py
@@ -70,7 +70,7 @@ class BaseRegister(ABC, CoordsCollection):
             raise TypeError(
                 "The qubits have to be stored in a dictionary "
                 "matching qubit ids to position coordinates; "
-                f"got {type(qubits)} ({qubits})."
+                f"got {type(qubits)}: {qubits}."
             )
         if not qubits:
             raise ValueError(
--- a/pulser-core/pulser/register/base_register.py
+++ b/pulser-core/pulser/register/base_register.py
@@ -18,6 +18,7 @@ from __future__ import annotations
 import json
 import warnings
 from abc import ABC, abstractmethod
+from collections import Counter
 from collections.abc import Iterable, Mapping
 from collections.abc import Sequence as abcSequence
 from typing import (
@@ -99,7 +100,7 @@ class BaseRegister(ABC, CoordsCollection):
             if kwargs.keys() != {"layout", "trap_ids"}:
                 raise ValueError(
                     "If specifying 'kwargs', they must only be 'layout' and "
-                    f"'trap_ids'; got {sorted(kwargs)}."
+                    f"'trap_ids'; got {list(kwargs)}."
                 )
             layout: RegisterLayout = kwargs["layout"]
             trap_ids: tuple[int, ...] = tuple(kwargs["trap_ids"])
@@ -151,7 +152,7 @@ class BaseRegister(ABC, CoordsCollection):
             raise ValueError(
                 "The IDs list must be selected among the IDs of the register's"
                 " qubits; "
-                f"{sorted(set(id_list) - set(self.qubit_ids))} not in "
+                f"{list(set(id_list) - set(self.qubit_ids))} not in "
                 f"{list(self.qubit_ids)}."
             )
         return [self.qubit_ids.index(id_) for id_ in id_list]
@@ -201,7 +202,8 @@ class BaseRegister(ABC, CoordsCollection):
             if len(coords_) != len(labels):
                 raise ValueError(
                     f"Label length ({len(labels)}) does not "
-                    f"match number of coordinates ({len(coords_)})."
+                    f"match number of coordinates ({len(coords_)}); "
+                    f"got coords {coords!r} and labels {labels!r}."
                 )
             qubits = dict(zip(cast(Iterable, labels), coords_))
         else:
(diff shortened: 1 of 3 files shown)


jaewonyun1234:
Hi @a-corni ready for review whenever you have time.

For the trap_ids bounds check I mentioned at the bottom — happy to fold a fix
into this PR, or open it as a separate issue. Whichever you prefer.

pulser-core/pulser/register/base_register.py
```diff
@@ -98,7 +99,7 @@ def _init_kwargs(self, **kwargs: Any) -> None:
             if kwargs.keys() != {"layout", "trap_ids"}:
                 raise ValueError(
                     "If specifying 'kwargs', they must only be 'layout' and "
-                    "'trap_ids'."
+                    f"'trap_ids'; got {sorted(kwargs)}."
```
a-corni: I am not sure the `sorted` is needed here. Since the target of this message is a software developer, I would rather return the kwargs in the order the program submitted them.

pulser-core/pulser/register/base_register.py
```diff
@@ -190,14 +193,15 @@ def from_coordinates(
...
+                    f"got prefix={prefix!r} and labels={list(labels)}."
                 )
 
         elif labels is not None:
             if len(coords_) != len(labels):
                 raise ValueError(
-                    f"Label length ({len(labels)}) does not"
-                    f"match number of coordinates ({len(coords_)})"
+                    f"Label length ({len(labels)}) does not "
+                    f"match number of coordinates ({len(coords_)})."
```
a-corni: Maybe report the two offending inputs: "Got coords {coords} and labels {labels}.".

pulser-core/pulser/register/base_register.py
```diff
@@ -212,15 +216,21 @@ def _validate_layout(
         if register_layout.dimensionality != self.dimensionality:
             raise ValueError(
                 "The RegisterLayout dimensionality is not the same as this "
-                "register's."
+                f"register's; layout is {register_layout.dimensionality}D "
+                f"and register is {self.dimensionality}D."
```
a-corni: I think what matters is knowing the input layout
. The information you are bringing has value, you can modify the error raised:
"The RegisterLayout dimensionality ({register_layout.dimensionality}D) is not the same as this "
f"register's ({self.dimensionality}D); Got layout {register_layout} on register {self.register}."

pulser-core/pulser/register/base_register.py
```diff
@@ -149,7 +150,9 @@ def find_indices(self, id_list: abcSequence[QubitId]) -> list[int]:
         if not set(id_list) <= set(self.qubit_ids):
             raise ValueError(
                 "The IDs list must be selected among the IDs of the register's"
-                " qubits."
+                " qubits; "
+                f"{sorted(set(id_list) - set(self.qubit_ids))} not in "
```
a-corni: I don't think the `sorted` is necessary here, and I would suggest to take it out as it might take unnecessary time to raise this error.

pulser-core/pulser/register/base_register.py
```diff
@@ -212,15 +216,21 @@ def _validate_layout(
         if register_layout.dimensionality != self.dimensionality:
             raise ValueError(
                 "The RegisterLayout dimensionality is not the same as this "
-                "register's."
+                f"register's; layout is {register_layout.dimensionality}D "
+                f"and register is {self.dimensionality}D."
             )
         if len(set(trap_ids)) != len(trap_ids):
-            raise ValueError("Every 'trap_id' must be a unique integer.")
+            repeated = sorted({t for t in trap_ids if trap_ids.count(t) > 1})
```
a-corni: I think it would be best to try to return the repeating elements in their order of appearance in the trap_ids. Therefore, I suggest to get rid of sorted, and use collections.Counter (offers a faster one-liner than your current solution)

```suggestion
            repeated = [t for t, freq in Counter(trap_ids).items() if freq > 1]
```

pulser-core/pulser/register/base_register.py
```diff
@@ -212,15 +216,21 @@ def _validate_layout(
...
+                "Every 'trap_id' must be a unique integer; "
+                f"found repeated ids {repeated} in {list(trap_ids)}."
+            )
 
         if len(trap_ids) != len(self._ids):
             raise ValueError(
                 "The amount of 'trap_ids' must be equal to the number of atoms"
-                " in the register."
+                f" in the register; got {len(trap_ids)} trap_ids "
+                f"for {len(self._ids)} atoms."
```
a-corni: You can modify the returned message to show the length of trap_ids and self._ids, but what matters is returning the incorrect input:
f"The amount of 'trap_ids' {len(trap_ids)} is not equal to the number of atoms {len(self._ids)}. Got trap ids {trap_ids} for atoms {self._ids}."

pulser-core/pulser/register/base_register.py
```diff
@@ -251,7 +263,9 @@ def define_detuning_map(
         if not set(detuning_weights.keys()) <= set(self.qubit_ids):
             raise ValueError(
                 "The qubit ids linked to detuning weights have to be defined"
-                " in the register."
+                " in the register. Got "
+                f"{sorted(set(detuning_weights) - set(self.qubit_ids))}, "
```
a-corni: Here as well, I don't think the `sorted` is necessary and suggest to take it of, for the sake of time.

pulser-core/pulser/register/base_register.py
a-corni, line 247: Indeed, it is missing a check that the provided trap_ids are a subset of `register_layout.traps_dict()`. You could do `register_layout.define_register(*trap_ids, qubit_ids=list(self.qubits))` to get all the checks from `RegisterLayout.defin_register`. You can include this test here, and properly test it in the tests, I would be fine with this. Otherwise, we can create a separate issue and tackle it separately, later. Let me know what you prefer.

a-corni (commented):
Thanks @jaewonyun1234 !
I have some comments, notably on the use of `sorted`. You can modify some error messages if you want. 
Regarding the missing check on _validate_layout. I think you are right, and I suggest we make a call to register_layout.define_register to check the inputs properly.

jaewonyun1234:
Hi @a-corni — thank you again for the careful review, I really appreciate the clear suggestions!

I’ve addressed your comments in the latest commit:
- removed the unnecessary `sorted` calls  
- used `Counter` for repeated `trap_ids`  
- enriched the error messages as you suggested  
- for the invalid `trap_ids` check, I inlined the `traps_dict` subset validation instead of calling `define_register`, to avoid recursion when the ids are valid, and added a test for that case  

Please let me know if anything still looks off — happy to adjust. Thanks again for your time!

pulser-core/pulser/register/base_register.py
```diff
@@ -211,25 +217,38 @@ def _validate_layout(
...
+                "Every 'trap_id' must be a unique integer; "
+                f"found repeated ids {repeated} in {list(trap_ids)}."
+            )
 
         if len(trap_ids) != len(self._ids):
             raise ValueError(
-                "The amount of 'trap_ids' must be equal to the number of atoms"
-                " in the register."
+                f"The amount of 'trap_ids' {len(trap_ids)} is not equal to "
+                f"the number of atoms {len(self._ids)}. Got trap ids "
```
a-corni: nit:
```suggestion
                f"the number of atoms {len(self._ids)} in the register. Got trap ids "
```

pulser-core/pulser/register/base_register.py
```diff
@@ -211,25 +217,38 @@ def _validate_layout(
         trap_coords = register_layout.coords
         if register_layout.dimensionality != self.dimensionality:
             raise ValueError(
-                "The RegisterLayout dimensionality is not the same as this "
-                "register's."
+                "The RegisterLayout dimensionality "
+                f"({register_layout.dimensionality}D) is not the same as this "
+                f"register's ({self.dimensionality}D); got layout "
+                f"{register_layout} on register {self}."
```
a-corni: `register_layout` will return the __repr__ of the register layout, that is just a string as "RegisterLayout_{hash}". I believe this is not enough speaking and returning the coordinates of the layout would be better


```suggestion
                f"register's ({self.dimensionality}D); Layout's "
                f"coordinates are {register_layout.coords} and register's {self.sorted_coords}."
```

tests/test_register_layout.py
```diff
@@ -98,16 +98,20 @@ def test_register_definition(layout, layout3d):
 
     reg2d = layout.define_register(0, 2)
     assert reg2d._layout_info == (layout, (0, 2))
-    with pytest.raises(ValueError, match="dimensionality is not the same"):
+    with pytest.raises(ValueError, match="dimensionality .* is not the same"):
```
a-corni: We can check that it is printing as fine:
```suggestion
    with pytest.raises(ValueError, match="dimensionality (3D) is not the same"):
```

pulser-core/pulser/register/base_register.py
```diff
@@ -211,25 +217,38 @@ def _validate_layout(
...
-                "The amount of 'trap_ids' must be equal to the number of atoms"
-                " in the register."
+                f"The amount of 'trap_ids' {len(trap_ids)} is not equal to "
+                f"the number of atoms {len(self._ids)}. Got trap ids "
+                f"{trap_ids} for atoms {self._ids}."
+            )
+        if not set(trap_ids).issubset(register_layout.traps_dict):
+            raise ValueError(
+                "All 'trap_ids' must correspond to the ID of a trap; "
+                f"got {trap_ids}."
```
a-corni: I agree with you there is no need to use RegisterLayout.define_register, but I think the error message has to be a bit improved:

```suggestion
                "All 'trap_ids' must correspond to the ID of a trap in the layout; "
                f"got {trap_ids}, must be among {register_layout.traps_dict}."
```

a-corni (commented):
Looking great! I have a few additional comments :)

jaewonyun1234:
Thanks @a-corni,
Applied all four, with two deviations:

**Test:** (3D) is a regex group, so the pattern resolves to dimensionality 3D is not the same and doesn't match. Used re.escape instead, consistent with test_dmm.py.

**Trap id message:** used list(register_layout.traps_dict) rather than the dict — the dict renders every coordinate (~2400 chars on SquareLatticeLayout(10, 10) vs ~400 for the ids), and only the valid ids are actionable here. Happy to include the coordinates if you'd rather.

Kept your dimensionality wording. Worth noting it reaches ~1300 chars / 100 lines on that same layout, since numpy breaks each row onto its own line — fine by me either way, but say the word if you'd prefer it bounded.

jaewonyun1234:
One more fix, pushed as 8bfe2c4.

Taking out `sorted` left the invalid-id lists in whatever order the set gave them, and that order changes every time you run it. The same wrong input printed differently each run:

```
run 1:  ['typo_a', 'typo_c', 'typo_b']
run 3:  ['typo_b', 'typo_a', 'typo_c']
```

Now it keeps the order the caller passed in, so the list matches what they wrote:

```python
invalid = [id_ for id_ in id_list if id_ not in valid_ids]
```

No `sorted` back — the set is still there just for the lookup. Same change in `find_indices` and `define_detuning_map`.

pulser-core/pulser/register/base_register.py
```diff
@@ -68,7 +69,8 @@ def __init__(
         if not isinstance(qubits, dict):
             raise TypeError(
                 "The qubits have to be stored in a dictionary "
-                "matching qubit ids to position coordinates."
+                "matching qubit ids to position coordinates; "
+                f"got {type(qubits)}."
```
a-corni: Ah damn I just realize that this message should return the provided qubits as well.

```suggestion
                "matching qubit ids to position coordinates; "
                f"got {type(qubits)} ({qubits})."
```

a-corni (commented):
It looks good to me! Very nice work! I just realized one error is missing is offending input. I suggest to include it, then we will be good to go :)

jaewonyun1234:
Good catch, thanks — applied in 0ee391c.

I also added assertions for it. While doing that I noticed the existing check was:

```python
with pytest.raises(TypeError):
    Register(coords)
    Register(ids)
```

`Register(ids)` never ran, since the line above it raises first, so the tuple case was never actually tested. Split into two blocks, each asserting its own message.

tests/test_register.py
```diff
@@ -32,15 +33,32 @@ def test_creation():
     coords = [(0, 0), (1, 0)]
     ids = ("q0", "q1")
     qubits = dict(zip(ids, coords))
-    with pytest.raises(TypeError):
+    with pytest.raises(
+        TypeError,
+        match=re.escape(
+            "The qubits have to be stored in a dictionary matching qubit ids "
+            "to position coordinates; got <class 'list'> ([(0, 0), (1, 0)])."
```
a-corni: nit: I am not 100% pleased by the way these errors are being shown. The parentheses were a bad suggestion, perhaps we should ditch them for "{type(qubits)}: {qubits}" ?

a-corni:
> Good catch, thanks — applied in [0ee391c](https://github.com/pasqal-io/Pulser/commit/0ee391c6ce9554d3eea855363f1a1c67ac20bae2).
> 
> I also added assertions for it. While doing that I noticed the existing check was:
> 
> ```python
> with pytest.raises(TypeError):
>     Register(coords)
>     Register(ids)
> ```
> 
> `Register(ids)` never ran, since the line above it raises first, so the tuple case was never actually tested. Split into two blocks, each asserting its own message.

Nice catch! Thanks for the rapid implementation :)

jaewonyun1234:
Both done in fa1944c — the colon reads better than the parentheses, agreed.

Apologies for the confusion on that thread: I'd pushed a cleanup that removed the test block seconds after you commented on it, which is why your note shows as outdated. The split is back, now asserting the colon format.

a-corni (approved):
LGTM! Thanks for the great work @jaewonyun1234 ! 🚀

jaewonyun1234:
Thanks for the thorough review @a-corni — the suggestions made the messages
much clearer, especially reporting the offending input consistently.

I have the rest of the register module ready as a follow-up (traps, layouts,
weight maps, mappable registers, Register/Register3D) along the same lines.
I'll open it as a separate PR shortly unless you'd prefer it batched
differently.

PR #1096, opened 2026-08-20
Preserve specialized layout types in abstract deserialization

### LINKED ISSUE #848: from_abstract_repr does not use specialized layout classes
It seems that `Device.from_abstract_repr` does not make use of specialized layout classes, which prevents from using some `*_register` methods.

For example, this works:
```python
from pulser import AnalogDevice as device

layout = device.calibrated_register_layouts[ "TriangularLatticeLayout(61, 5.0µm)" ]
register = layout.rectangular_register(rows=1, atoms_per_row=3, prefix="q")
```
but this does not:
```python
from pulser.devices import Device 

# get from TGCC
abstract_repr = "..."

device = Device.from_abstract_repr(abstract_repr)
layout = device.calibrated_register_layouts[ "TriangularLatticeLayout(61, 5.0µm)" ]
register = layout.rectangular_register(rows=1, atoms_per_row=3, prefix="q")
```

In the first case:
```python
>>> type(layout)
pulser.register.special_layouts.TriangularLatticeLayout
```
while in the second:
```python
>>> type(layout)
pulser.register.register_layout.RegisterLayout
```

danielgaskins:
## What changed

Abstract deserialization now reconstructs `RectangularLatticeLayout`, `SquareLatticeLayout`, and `TriangularLatticeLayout` when the serialized slug matches a built-in format and its coordinates match the generated layout.

Layouts with a similar slug but different geometry remain plain `RegisterLayout` objects. This avoids changing arbitrary user-defined layouts.

Tests cover all three specialized classes, generic fallback cases, and the `AnalogDevice` example from #848.

## Testing

- `python3 -m pytest tests/test_abstract_repr.py -q`
- `python3 -m pytest tests/test_abstract_repr.py tests/test_register_layout.py tests/test_devices.py tests/test_sequence.py -q`
- Black, isort, Flake8, and Ruff
- Mypy on the changed module

Fixes #848

### DIFF
--- a/pulser-core/pulser/json/abstract_repr/deserializer.py
+++ b/pulser-core/pulser/json/abstract_repr/deserializer.py
@@ -17,6 +17,7 @@ from __future__ import annotations
 
 import dataclasses
 import json
+import re
 from typing import TYPE_CHECKING, Any, Literal, Type, Union, cast, overload
 
 import jsonschema
@@ -47,6 +48,11 @@ from pulser.parametrized import ParamObj, Variable
 from pulser.pulse import Pulse
 from pulser.register.mappable_reg import MappableRegister
 from pulser.register.register_layout import RegisterLayout
+from pulser.register.special_layouts import (
+    RectangularLatticeLayout,
+    SquareLatticeLayout,
+    TriangularLatticeLayout,
+)
 from pulser.register.weight_maps import DetuningMap
 from pulser.waveforms import (
     BlackmanWaveform,
@@ -66,6 +72,18 @@ if TYPE_CHECKING:
 
 
 VARIABLE_TYPE_MAP = {"int": int, "float": float}
+
+_FLOAT_PATTERN = r"[+-]?(?:\d+(?:\.\d*)?|\.\d+)(?:[eE][+-]?\d+)?"
+_RECTANGULAR_LAYOUT_PATTERN = re.compile(
+    rf"RectangularLatticeLayout\((\d+)x(\d+), "
+    rf"({_FLOAT_PATTERN})x({_FLOAT_PATTERN})µm\)"
+)
+_SQUARE_LAYOUT_PATTERN = re.compile(
+    rf"SquareLatticeLayout\((\d+)x(\d+), ({_FLOAT_PATTERN})µm\)"
+)
+_TRIANGULAR_LAYOUT_PATTERN = re.compile(
+    rf"TriangularLatticeLayout\((\d+), ({_FLOAT_PATTERN})µm\)"
+)
 
 ExpReturnType = Union[int, float, list, ParamObj]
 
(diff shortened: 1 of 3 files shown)


pulser-core/pulser/json/abstract_repr/deserializer.py
```diff
@@ -67,6 +73,18 @@
...
+    rf"RectangularLatticeLayout\((\d+)x(\d+), "
+    rf"({_FLOAT_PATTERN})x({_FLOAT_PATTERN})µm\)"
+)
+_SQUARE_LAYOUT_PATTERN = re.compile(
+    rf"SquareLatticeLayout\((\d+)x(\d+), ({_FLOAT_PATTERN})µm\)"
+)
+_TRIANGULAR_LAYOUT_PATTERN = re.compile(
+    rf"TriangularLatticeLayout\((\d+), ({_FLOAT_PATTERN})µm\)"
+)
+
```
a-corni: I suggest to have these patterns in pulser-core/pulser/register/special_layouts.py or in a separate module in json/abstract_repr/slug.py
- I think we at least want to check that the slug of a special RegisterLayout matches its pattern (that way, we can track a modification of the pattern or of the slug easily).
- I think it could be great to be able to generate the slug and parse a given slug from the same object (have the functions you have below in an object).

tests/test_abstract_repr.py
```diff
@@ -151,7 +155,17 @@ def test_abstract_repr_encoder_torch():
...
+        RectangularLatticeLayout(3, 4, 2.5, 5),
         RegisterLayout([[10, 0], [1, 10]], slug="foo"),
+        RegisterLayout(
+            [[0, 0], [1, 1]],
+            slug="TriangularLatticeLayout(2, 1.0µm)",
+        ),
+        RegisterLayout(
+            [[0, 0], [0, 1], [1, 0], [1, 1]],
+            slug="SquareLatticeLayout(2x2, 0.0µm)",
+        ),
```
a-corni, line 172: Maybe add also 
- a test where the coordinates aim for a SquareLatticeLayout, with a wrong slug
- a test where the slug is the one of a RectangularLatticeLayout/TriangularLatticelayout/SquareLatticeLayout, but the coordinates don't match.

tests/test_abstract_repr.py
```diff
@@ -412,6 +427,15 @@ def _roundtrip(abstract_device):
 
         _roundtrip(abstract_device)
 
+    def test_special_layout_roundtrip(self):
+        device = Device.from_abstract_repr(AnalogDevice.to_abstract_repr())
+        layout = next(iter(device.pre_calibrated_layouts))
+        assert type(layout) is TriangularLatticeLayout
```
a-corni: It's already a good test! Could you also add it for RectangularLatticeLayout and SquareLatticeLayout? Since it's a different pattern for all of them, I think it would be better to test each of these special layouts :)

a-corni (commented):
Thanks for your proposal @danielgaskins ! It is already pretty good, here are some suggestions

pulser-core/pulser/json/abstract_repr/deserializer.py
danielgaskins: Sure thing, I moved the slug patterns into `special_layouts.py` and added a small helper that handles both formatting and parsing. Each specialized layout constructor now uses that helper, while deserialization only restores the specialized type when the reconstructed coordinates match.

tests/test_abstract_repr.py
danielgaskins: Sure thing, I added a case with square-lattice coordinates and an invalid square-layout slug. I also kept separate cases where a valid special-layout slug does not match the serialized coordinates.

tests/test_abstract_repr.py
danielgaskins: Sure thing, I expanded the device round-trip test to cover triangular, square, and rectangular layouts.

tests/test_abstract_repr.py
```diff
@@ -412,6 +431,20 @@ def _roundtrip(abstract_device):
...
+            TriangularLatticeLayout(61, 5),
+            SquareLatticeLayout(3, 4, 5),
+            RectangularLatticeLayout(3, 4, 5, 6),
+        ],
+    )
+    def test_special_layout_roundtrip(self, layout):
+        device = replace(AnalogDevice, pre_calibrated_layouts=(layout,))
+        device = Device.from_abstract_repr(device.to_abstract_repr())
+        deserialized_layout = next(iter(device.pre_calibrated_layouts))
+        assert type(deserialized_layout) is type(layout)
```
HGSilveri, line 446: Could you please also check the roundtrip when the layout is stored in `Register` (e.g. by de/serializing a register created through `RegisterLayout.define_register()`)?

Also, let's check that the layouts are equal too


```suggestion
        device = replace(AnalogDevice, pre_calibrated_layouts=(layout,))
        device = Device.from_abstract_repr(device.to_abstract_repr())
        deserialized_layout = next(iter(device.pre_calibrated_layouts))
        assert type(deserialized_layout) is type(layout)
        assert deserialized_layout == layout
```

HGSilveri (commented):
Thanks for your work on this @danielgaskins !
I have a small request and then we should be good to merge

tests/test_abstract_repr.py
danielgaskins: Sure thing, I added both checks for all three special layouts. The device round-trip now checks layout equality, and each layout also round-trips through a `Register` created with `define_register()`, checking both the concrete type and value. The abstract representation and register tests pass.

HGSilveri (approved):
LGTM, thanks a lot @danielgaskins !

PR #1097, opened 2026-08-23
Report offending input in register module error messages

jaewonyun1234:
Hi @a-corni,

Following on from #1095, this covers the rest of the register module.

| File | Messages | What they report now |
|---|---|---|
| `traps.py` | 3 | the coordinates given, their shape, the repeated ones |
| `weight_maps.py` | 2 | the weights given and which are out of `[0, 1]` |
| `register_layout.py` | 5 | repeated / invalid trap ids, repeated / mismatched qubit ids |
| `mappable_reg.py` | 4 | the qubit ids given and which were pre-declared |
| `register.py` | 3 | the dimensionality, the spacings, the device type |
| `register3d.py` | 1 | the dimensionality |
| **Total** | **18** | |

Two of them turned out to be more than wording:

- `Traps` raised the same message whether the coordinates couldn't be read as numbers at all or were merely the wrong shape, so `"abc"` and `[0, 1, 2]` were indistinguishable. Split into two.
- Weights outside `[0, 1]` were selected with `< 0 | > 1`. NaN satisfies neither, but does trigger the error — so a NaN weight raised an error whose list of offending weights was empty. Selecting by the negation of the accepted range fixes it.

PR #1098, opened 2026-08-24
Report offending input in sequence module error messages

jaewonyun1234:
Hi @a-corni, @HGSilveri,

Following on from #1095 and #1097, this covers the sequence module (38 messages: 35 in `sequence.py`, 2 in `_schedule.py`, 1 in `_seq_drawer.py`).

Error messages that only stated a rule now also report the value that broke it. `_schedule.py` reports the channel with no target and the duration that overflowed; `_seq_drawer.py` reports the channels that aren't in the rydberg basis. Qubit-id errors still list the declared ids — those sets are small and useful to see.

A few were more than wording:

- The `'Microwave'` clash in `declare_channel` interpolated the channel object, so it printed the full spec (~236 characters) instead of the name the user chose. It now reports the name and id, e.g. `got 'ryd' ('rydberg_global') with the declared {'mw': 'mw_global'}`.
- `estimate_added_delay` had two adjacent string literals with no space, so it read `...if sequence orpulse is parametrized.`
- `_schedule.get_samples` raises when `qubits` is missing for a DMM channel, but no test reached it. Added one.

Also aligned with the register review: `'device' must be an instance of 'BaseDevice'`; an unavailable DMM lists the still-available ids; a local channel that exceeds `max_targets` reports the limit, the count, and the given targets.

Messages that describe a state rather than a rejected input are left unchanged (`_in_ising` transitions, `qubit_info`, `register`, `get_measurement_basis`, SLM on a mappable register, `_target` with no qubits, empty-sequence drawing).

Existing `match=` assertions now cover the full message (`re.escape` where needed). Partially addresses #1057.

### DIFF
--- a/pulser-core/pulser/sequence/_schedule.py
+++ b/pulser-core/pulser/sequence/_schedule.py
@@ -258,7 +258,10 @@ class _ChannelSchedule:
         self, key: Union[int, slice]
     ) -> Union[_TimeSlot, list[_TimeSlot]]:
         if key == -1 and not self.slots:
-            raise ValueError("The chosen channel has no target.")
+            raise ValueError(
+                "The chosen channel has no target; channel "
+                f"{self.channel_id!r} has no time slots yet."
+            )
         return self.slots[key]
 
     def __iter__(self) -> Iterator[_TimeSlot]:
@@ -290,7 +293,7 @@ class _DMMSchedule(_ChannelSchedule):
         if qubits is None:
             raise ValueError(
                 "'qubits' must be defined when extracting the samples of a"
-                " DMM channel."
+                f" DMM channel; got None for channel {self.channel_id!r}."
             )
         return DMMSamples(
             **init_fields, detuning_map=self.detuning_map, qubits=qubits
@@ -714,7 +717,7 @@ class _Schedule(Dict[str, _ChannelSchedule]):
         if self.max_duration is not None and t > self.max_duration:
             msg = (
                 "The sequence's duration exceeded the maximum duration allowed"
-                f" by the device ({self.max_duration} ns)."
+                f" by the device ({self.max_duration} ns); got {t} ns."
             )
             if block_over_max_duration:
                 raise RuntimeError(msg)
--- a/pulser-core/pulser/sequence/_schedule.py
+++ b/pulser-core/pulser/sequence/_schedule.py
@@ -259,8 +259,7 @@ class _ChannelSchedule:
     ) -> Union[_TimeSlot, list[_TimeSlot]]:
         if key == -1 and not self.slots:
             raise ValueError(
-                "The chosen channel has no target; channel "
-                f"{self.channel_id!r} has no time slots yet."
+                f"The chosen channel ({self.channel_id!r}) has no target."
             )
         return self.slots[key]
 
--- a/pulser-core/pulser/sequence/_seq_drawer.py
+++ b/pulser-core/pulser/sequence/_seq_drawer.py
@@ -1018,8 +1018,14 @@ def _draw_qubit_content(
             for ch, ch_obj in sampled_seq._ch_objs.items()
         ]
     ):
+        wrong_basis = {
+            ch: ch_obj.basis
+            for ch, ch_obj in sampled_seq._ch_objs.items()
+            if ch_obj.basis != "ground-rydberg"
+        }
         raise NotImplementedError(
-            "Can only draw qubit contents for channels in rydberg basis."
+            "Can only draw qubit contents for channels in rydberg basis; "
+            f"got {wrong_basis}."
         )
     # Gather data per targeted qubits
     total_duration = data["total_duration"]
--- a/pulser-core/pulser/sequence/_seq_drawer.py
+++ b/pulser-core/pulser/sequence/_seq_drawer.py
@@ -1024,8 +1024,8 @@ def _draw_qubit_content(
             if ch_obj.basis != "ground-rydberg"
         }
         raise NotImplementedError(
-            "Can only draw qubit contents for channels in rydberg basis; "
-            f"got {wrong_basis}."
+            "Can only draw qubit contents for channels in the "
+            f"'ground-rydberg' basis; got {wrong_basis}."
         )
     # Gather data per targeted qubits
     total_duration = data["total_duration"]
--- a/pulser-core/pulser/sequence/sequence.py
+++ b/pulser-core/pulser/sequence/sequence.py
@@ -19,6 +19,7 @@ import copy
 import json
 import os
 import warnings
+from collections import Counter
 from collections.abc import Collection, Mapping
 from typing import (
     Any,
@@ -181,7 +182,9 @@ class Sequence(Generic[DeviceType]):
     @_in_ising.setter
     def _in_ising(self, value: bool) -> None:
         if not isinstance(value, bool):
-            raise TypeError("_in_ising must be a bool.")
+            raise TypeError(
+                f"_in_ising must be a bool; got {type(value)}: {value!r}."
+            )
         if self._in_ising == value:
             # If the value doesn't change, do nothing
             return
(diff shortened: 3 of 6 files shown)


pulser-core/pulser/sequence/sequence.py
```diff
@@ -351,7 +355,8 @@ def magnetic_field(self) -> np.ndarray:
         if not self._in_xy:
             raise AttributeError(
                 "The magnetic field is only defined when the "
-                "sequence is in 'XY Mode'."
+                "sequence is in 'XY Mode'; this sequence uses "
+                f"{self.get_addressed_bases()}."
```
HGSilveri:
```suggestion
                "sequence is in 'XY Mode'; this sequence addresses "
                f"{self.get_addressed_bases()}."
```

pulser-core/pulser/sequence/sequence.py
```diff
@@ -1553,9 +1633,15 @@ def estimate_added_delay(
         if isinstance(channel_obj, DMM):
             phase_ref = None
         elif len(ph_refs) != 1:
+            refs = {
+                q: float(self._basis_ref[basis][q].phase.last_phase)
+                for q in self._register.qubit_ids
+                if q in last.targets
```
HGSilveri, line 1639: You can just iterate over the `last.targets` here

pulser-core/pulser/sequence/sequence.py
```diff
@@ -2151,9 +2248,15 @@ def _add(
...
```
HGSilveri, line 2254: Same here

pulser-core/pulser/sequence/sequence.py
```diff
@@ -2245,9 +2361,15 @@ def _target(
                 for q in qubit_ids_set
             }
             if len(phase_refs) != 1:
+                refs = {
+                    q: float(self._basis_ref[basis][q].phase.last_phase)
+                    for q in self._register.qubit_ids
+                    if q in qubit_ids_set
```
HGSilveri, line 2367: And here

pulser-core/pulser/sequence/sequence.py
```diff
@@ -2387,22 +2516,34 @@ def _validate_channel(
...
+                n
+                for n, ch_obj in self.declared_channels.items()
+                if ch_obj.addressing == "Global"
+                and not isinstance(ch_obj, DMM)
+            ]
             raise ValueError(
                 "You should add a Pulse to a Global Channel prior to"
-                " modulating the DMM used for the SLM Mask."
+                " modulating the DMM used for the SLM Mask; got channel"
+                f" {channel!r}, declared global channels: {global_channels}."
```
HGSilveri:
```suggestion
                " modulating the DMM used for the SLM Mask; before adding a "
                f"pulse to {channel!r}, make sure there at least one of "
                f"{global_channels} already has a pulse."
```

pulser-core/pulser/sequence/_schedule.py
```diff
@@ -258,7 +258,10 @@ def __getitem__(
         self, key: Union[int, slice]
     ) -> Union[_TimeSlot, list[_TimeSlot]]:
         if key == -1 and not self.slots:
-            raise ValueError("The chosen channel has no target.")
+            raise ValueError(
+                "The chosen channel has no target; channel "
+                f"{self.channel_id!r} has no time slots yet."
```
HGSilveri: Time slots are an internal object so it's best not to mention them

```suggestion
                "The chosen channel ({self.channel_id!r}) has no target."
```

pulser-core/pulser/sequence/_seq_drawer.py
```diff
@@ -1018,8 +1018,14 @@ def _draw_qubit_content(
...
         ]
     ):
+        wrong_basis = {
+            ch: ch_obj.basis
+            for ch, ch_obj in sampled_seq._ch_objs.items()
+            if ch_obj.basis != "ground-rydberg"
+        }
         raise NotImplementedError(
-            "Can only draw qubit contents for channels in rydberg basis."
+            "Can only draw qubit contents for channels in rydberg basis; "
```
HGSilveri: Let's take the opportunity to be more precise in this message 

```suggestion
            "Can only draw qubit contents for channels in the  'ground-rydberg' basis; "
```

HGSilveri (commented):
Nice going @jaewonyun1234 !
Here are a few minor comments, keep up the good work :)

pulser-core/pulser/sequence/sequence.py
jaewonyun1234: Agree, I will apply this.

pulser-core/pulser/sequence/sequence.py
jaewonyun1234: Applied — dropped "there" so it reads "make sure at least one of [...] already has a pulse".

pulser-core/pulser/sequence/_schedule.py
jaewonyun1234: Applied, with an f prefix added so the channel id interpolates.

pulser-core/pulser/sequence/_seq_drawer.py
jaewonyun1234: Applied, with a single space between "the" and "'ground-rydberg'".

pulser-core/pulser/sequence/sequence.py
jaewonyun1234: `last.targets` is a `set[QubitId]`, so iterating it directly builds the dict in hash order, which changes from run to run. Going through `self._register.qubit_ids` and filtering is what keeps the printed order stable.

With your version, `test_phase` fails on 7 of 8 `PYTHONHASHSEED` values. On the 28-qubit fixture register the same error prints as:

```
seed 0: got {'q9': 0.0, 'q0': 1.0, 'q7': 0.0, 'q15': 0.0, ...}
seed 1: got {'q17': 0.0, 'q9': 0.0, 'q11': 0.0, ..., 'q0': 1.0, ...}
```

`q0` is the only qubit with a different reference, and it moves around each run.

That said, you're right that the loop doesn't show why it's written that way. Would you prefer a small helper, used here and at the two other spots you flagged, plus the same pattern in `config_slm_mask`?

```python
def _in_register_order(self, qubits: Collection[QubitId]) -> list[QubitId]:
    """Qubit ids in the register's order, so error messages are reproducible."""
    return [q for q in self._register.qubit_ids if q in qubits]
```

Separately on this message: on a global channel it lists every qubit, so 27 of the 28 entries are the same `0.0` and only `q0` differs. Would you rather it grouped them by reference value, or showed only the ones that stand out?

pulser-core/pulser/sequence/sequence.py
jaewonyun1234: Same reason as the thread above — these are sets, so the register loop is what keeps the printed order stable. Happy to switch all three to a helper if you prefer that.

pulser-core/pulser/sequence/sequence.py
jaewonyun1234: Same reason as the thread above — these are sets, so the register loop is what keeps the printed order stable. Happy to switch all three to a helper if you prefer that.

pulser-core/pulser/sequence/sequence.py
jaewonyun1234: Or, if a new method feels like too much for this, I can just add a one-line comment at each spot instead:

```python
# Iterate the register, not the target set, so the order is reproducible
```

Happy either way — whichever you prefer.

HGSilveri: How about iterating over `sorted(last.targets)`?

jaewonyun1234: I agree! thanks!

pulser-core/pulser/sequence/sequence.py
jaewonyun1234: Switched to sorted() here too, same as the thread above.

pulser-core/pulser/sequence/sequence.py
jaewonyun1234: Switched to sorted() here too, same as the thread above.

HGSilveri (approved):
LGTM, thanks a lot for seeing it through @jaewonyun1234 !

PR #1101, opened 2026-08-28
Support standalone DetuningMap serialization

### LINKED ISSUE #1065: Support independent serialization/deserialization of WeightMap
Currently, WeightMaps are serialized/deserialized inside the Sequence object. This is because the schema associated with the WeightMap is inside the Sequence schema.
In order to serialize/deserialize WeightMaps independently of the Sequence we need:
- to create a separate schema for the WeightMap, that the Sequence schema will refer to for validation/serialization/deserialization.
- to create a public method `to_abstract_repr()` to produce a serialized WeightMap.
- to create a public method `form_abstract_repr()` to create a WeightMap from a serialized WeightMap.

danielgaskins:
## Summary

- move the weight map definition into its own schema
- make the sequence schema reference the new schema
- add standalone serialization and deserialization methods to `DetuningMap`
- include the new schema in source distributions

Existing sequence JSON remains unchanged.

## Testing

- 244 abstract representation and detuning map tests
- Black 24.4.0
- isort 5.12.0
- Flake8 7.1.1
- Ruff 0.11.1
- JSON parsing and schema reference checks

Closes #1065

### DIFF
--- a/pulser-core/pulser/json/abstract_repr/deserializer.py
+++ b/pulser-core/pulser/json/abstract_repr/deserializer.py
@@ -581,6 +581,20 @@ def _deserialize_det_map(ser_det_map: dict) -> DetuningMap:
     )
 
 
+def deserialize_abstract_weight_map(obj_str: str) -> DetuningMap:
+    """Deserialize a weight map from an abstract JSON object.
+
+    Args:
+        obj_str: The JSON string representing the weight map encoded in the
+            abstract JSON format.
+
+    Returns:
+        The DetuningMap instance.
+    """
+    validate_abstract_repr(obj_str, "weight-map")
+    return _deserialize_det_map(json.loads(obj_str))
+
+
 def deserialize_abstract_sequence(obj_str: str) -> Sequence:
     """Deserialize a sequence from an abstract JSON object.
 
--- a/pulser-core/pulser/json/abstract_repr/deserializer.py
+++ b/pulser-core/pulser/json/abstract_repr/deserializer.py
@@ -581,11 +581,11 @@ def _deserialize_det_map(ser_det_map: dict) -> DetuningMap:
     )
 
 
-def deserialize_abstract_weight_map(obj_str: str) -> DetuningMap:
-    """Deserialize a weight map from an abstract JSON object.
+def deserialize_detuning_map(obj_str: str) -> DetuningMap:
+    """Deserialize a detuning map from an abstract JSON object.
 
     Args:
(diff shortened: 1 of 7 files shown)


pulser-core/pulser/json/abstract_repr/deserializer.py
```diff
@@ -581,6 +581,20 @@ def _deserialize_det_map(ser_det_map: dict) -> DetuningMap:
     )
 
 
+def deserialize_abstract_weight_map(obj_str: str) -> DetuningMap:
```
HGSilveri: I think we should call this
```
def deserialize_detuning_map(obj_str: str) -> DetuningMap:
```
to be consistent with the return type.

This would also warrant a change in the docstring

pulser-core/pulser/register/weight_maps.py
```diff
@@ -214,6 +218,38 @@ def _to_abstract_repr(self) -> dict[str, Any]:
...
+        detuning_map = (
+            pulser_abstract_repr.deserializer.deserialize_abstract_weight_map(
+                obj_str
+            )
+        )
+        return cls(
+            detuning_map.trap_coordinates,
+            detuning_map.weights,
+            detuning_map.slug,
+        )
```
HGSilveri: I find it awkward that we internally deserialize a `DetuningMap` only to turn it into a `WeightMap`. Given that we currently have no need for `WeightMap` de/serialization (just `DetuningMap` specifically), would you mind making these functions specific to `DetuningMap`?
We can always extend them to `WeightMap` later on if it's justified

HGSilveri (commented):
This is looking good overall, nice work @danielgaskins !
I think it's fine to keep the schema for the whole WeightMap but, looking at the code, I think perhaps it would be preferable to support de/serialization just for `DetuningMap`, at least for now. I realize this is not what I ask for in the issue description, my apologies for that

pulser-core/pulser/json/abstract_repr/deserializer.py
danielgaskins: Renamed the function to `deserialize_detuning_map()` and updated its docstring to describe a detuning map.

pulser-core/pulser/register/weight_maps.py
danielgaskins: Moved the public serialization methods to `DetuningMap` and return the deserialized `DetuningMap` directly. The tests now cover only `DetuningMap`.

HGSilveri (approved):
LGTM, great job @danielgaskins !

PR #1102, opened 2026-09-03
Report offending input in channels module error messages

jaewonyun1234:
Hi @a-corni, @HGSilveri,

Following on from #1095, #1097 and #1098, this covers the channels module: 11 messages across `base_channel.py` (7), `eom.py` (2), `dmm.py` (1) and `modulation.py` (1).

One was a bug rather than wording. `RydbergEOM` checks each beam in `chain((self.limiting_beam,), self.controlled_beams)`, but the message printed `self.limiting_beam` instead of `beam`. A bad value in `controlled_beams` was reported as the limiting beam, which is usually valid:

```
RydbergEOM(limiting_beam=RydbergBeam.RED,
           controlled_beams=(RydbergBeam.RED | RydbergBeam.BLUE,), ...)

before:  ... enumeration, not RydbergBeam.RED.        # valid, and on a different parameter
after:   ... enumeration, not RydbergBeam.BLUE|RED.
```

The existing assertion stopped before that value, so it passed either way. It now checks the beam.

Partially addresses #1057.

### DIFF
--- a/pulser-core/pulser/channels/base_channel.py
+++ b/pulser-core/pulser/channels/base_channel.py
@@ -221,7 +221,8 @@ class Channel(ABC):
             parameters += local_only
             if self.propagation_dir is not None:
                 raise NotImplementedError(
-                    "'propagation_dir' must be left as None in Local channels."
+                    "'propagation_dir' must be left as None in Local "
+                    f"channels; got {self.propagation_dir}."
                 )
 
         for param in parameters:
@@ -444,18 +445,20 @@ class Channel(ABC):
             _duration = int(duration)
         except (TypeError, ValueError):
             raise TypeError(
-                "duration needs to be castable to an int but "
-                "type %s was provided" % type(duration)
+                "'duration' needs to be castable to an int; got "
+                f"{duration!r} of type {type(duration)}."
             )
 
         if duration < self.min_duration:
             raise ValueError(
-                "duration has to be at least " + f"{self.min_duration} ns."
+                f"'duration' has to be at least {self.min_duration} ns; "
+                f"got {duration}."
             )
 
         if self.max_duration is not None and duration > self.max_duration:
             raise ValueError(
-                "duration can be at most " + f"{self.max_duration} ns."
+                f"'duration' can be at most {self.max_duration} ns; "
+                f"got {duration}."
             )
 
         if round_up and duration % self.clock_period != 0:
@@ -482,25 +485,26 @@ class Channel(ABC):
         amp_samples_np = pulse.amplitude.samples.as_array(detach=True)
         if self.max_amp is not None and np.any(amp_samples_np > self.max_amp):
             raise ValueError(
-                "The pulse's amplitude goes over the maximum "
-                "value allowed for the chosen channel."
+                "The pulse's amplitude goes over the maximum value allowed "
+                f"for the chosen channel ({self.max_amp}); got "
+                f"{amp_samples_np.max()}."
             )
-        if self.max_abs_detuning is not None and np.any(
-            np.round(
+        if self.max_abs_detuning is not None:
+            abs_detuning = np.round(
                 np.abs(pulse.detuning.samples.as_array(detach=True)),
                 decimals=6,
             )
-            > self.max_abs_detuning
-        ):
-            raise ValueError(
-                "The pulse's detuning values go out of the range "
-                "allowed for the chosen channel."
-            )
+            if np.any(abs_detuning > self.max_abs_detuning):
+                raise ValueError(
+                    "The pulse's detuning values go out of the range allowed "
+                    f"for the chosen channel ({self.max_abs_detuning}); got "
+                    f"a maximum absolute value of {abs_detuning.max()}."
+                )
         avg_amp = np.average(amp_samples_np)
         if 0 < avg_amp < self.min_avg_amp:
             raise ValueError(
                 "The pulse's average amplitude is below the chosen "
-                f"channel's limit ({self.min_avg_amp})."
+                f"channel's limit ({self.min_avg_amp}); got {avg_amp}."
             )
 
     @property
--- a/pulser-core/pulser/channels/dmm.py
+++ b/pulser-core/pulser/channels/dmm.py
@@ -161,7 +161,10 @@ class DMM(Channel):
         )
         # Check that detuning is negative
         if np.any(round_detuning > 0):
-            raise ValueError("The detuning in a DMM must not be positive.")
+            raise ValueError(
+                "The detuning in a DMM must not be positive; got a maximum "
+                f"of {round_detuning.max()}."
+            )
         # Check that detuning on each atom is above bottom_detuning
         min_round_detuning = np.min(round_detuning)
         max_weight = np.max(detuning_map.weights)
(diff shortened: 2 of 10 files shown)


pulser-core/pulser/channels/eom.py
```diff
@@ -192,7 +192,7 @@ def __post_init__(self) -> None:
             ):
                 raise TypeError(
                     "Every beam must be one of options of the `RydbergBeam`"
-                    f" enumeration, not {self.limiting_beam}."
+                    f" enumeration, not {beam}."
```
a-corni, line 195: Nice catch! Actually, I think it would be more interesting to say which attribute has an incorrect type. Perhaps we could use a dict instead of a chain, having {"limiting_beam":self.limiting_beam, "controlled_beams":self.controlled_beams} and 

```suggestion
                    f" enumeration. Got {beam} for attribute {dict_key}."
```

pulser-core/pulser/channels/base_channel.py
```diff
@@ -482,25 +485,26 @@ def validate_pulse(self, pulse: Pulse) -> None:
         amp_samples_np = pulse.amplitude.samples.as_array(detach=True)
         if self.max_amp is not None and np.any(amp_samples_np > self.max_amp):
             raise ValueError(
-                "The pulse's amplitude goes over the maximum "
-                "value allowed for the chosen channel."
+                "The pulse's amplitude goes over the maximum value allowed "
+                f"for the chosen channel ({self.max_amp}); got "
+                f"{amp_samples_np.max()}."
```
a-corni, line 490: I think we should specify the pulse that generated this error.

```suggestion
                "The pulse's amplitude goes over the maximum value allowed "
                f"for the chosen channel ({self.max_amp}); got "
                f"a maximum amplitude {amp_samples_np.max()} in pulse {pulse!r}."
```

pulser-core/pulser/channels/base_channel.py
```diff
@@ -482,25 +485,26 @@ def validate_pulse(self, pulse: Pulse) -> None:
...
-        ):
-            raise ValueError(
-                "The pulse's detuning values go out of the range "
-                "allowed for the chosen channel."
-            )
+            if np.any(abs_detuning > self.max_abs_detuning):
+                raise ValueError(
+                    "The pulse's detuning values go out of the range allowed "
+                    f"for the chosen channel ({self.max_abs_detuning}); got "
+                    f"a maximum absolute value of {abs_detuning.max()}."
```
a-corni, line 501: Likewise, it is important to return the pulse in this error message

```suggestion
                    f"a maximum absolute detuning of {abs_detuning.max()} in pulse {pulse!r}."
```

pulser-core/pulser/channels/base_channel.py
```diff
@@ -482,25 +485,26 @@ def validate_pulse(self, pulse: Pulse) -> None:
...
+                )
         avg_amp = np.average(amp_samples_np)
         if 0 < avg_amp < self.min_avg_amp:
             raise ValueError(
                 "The pulse's average amplitude is below the chosen "
-                f"channel's limit ({self.min_avg_amp})."
+                f"channel's limit ({self.min_avg_amp}); got {avg_amp}."
```
a-corni, line 507: Likewise, it is important to return the pulse here

```suggestion
                f"channel's limit ({self.min_avg_amp}); got average amplitude {avg_amp} in pulse {pulse!r}."
```

pulser-core/pulser/channels/dmm.py
```diff
@@ -161,7 +161,10 @@ def validate_pulse(
         )
         # Check that detuning is negative
         if np.any(round_detuning > 0):
-            raise ValueError("The detuning in a DMM must not be positive.")
+            raise ValueError(
+                "The detuning in a DMM must not be positive; got a maximum "
+                f"of {round_detuning.max()}."
```
a-corni, line 166: I believe what matters here is the detuning of the pulse (maybe not necessarily the pulse, as it's a DMM)

```suggestion
                f"of {round_detuning.max()} in detuning {pulse.detuning!r}."
```

a-corni (commented):
Hey @jaewonyun1234 ! Thanks for your contribution! Here are some suggestions :)

jaewonyun1234:
Applied all five, thanks @a-corni!

pulser-core/pulser/channels/base_channel.py
a-corni, line 277: Maybe include the failing `eom_config` here ?

a-corni: We should include the failing pulse here as well

pulser-core/pulser/channels/dmm.py
a-corni: We should report the offending pulse here as well. Reporting the values won't mean anything for a developer seeing this issue

pulser-core/pulser/channels/dmm.py
```diff
@@ -161,7 +161,10 @@ def validate_pulse(
         )
         # Check that detuning is negative
         if np.any(round_detuning > 0):
-            raise ValueError("The detuning in a DMM must not be positive.")
+            raise ValueError(
+                "The detuning in a DMM must not be positive; got a maximum "
+                f"of {round_detuning.max()} in detuning {pulse.detuning!r}."
```
a-corni: On these kind of errors, I am wondering if it's the maximum value that is really of interest. I think it would be better to report the times at which the detuning is greater than 0 strictly (likewise in channel.py, I guess it could be good to report the times at which the pulse's amplitude (or |detuning|) goes higher than channel's max amplitude (or max absolute detuning)

pulser-core/pulser/channels/dmm.py
a-corni, line 220: We should report the offending input here as well :)

a-corni (commented):
Hey @jaewonyun1234 ! Thanks for tackling my comments! I have two suggestions on the changes you have made, otherwise I think I have seen other error messages that need improving in this module :)

jaewonyun1234:
Applied all five, thanks @a-corni!

pulser-core/pulser/channels/dmm.py
```diff
@@ -162,8 +162,9 @@ def validate_pulse(
         # Check that detuning is negative
         if np.any(round_detuning > 0):
             raise ValueError(
-                "The detuning in a DMM must not be positive; got a maximum "
-                f"of {round_detuning.max()} in detuning {pulse.detuning!r}."
+                "The detuning in a DMM must not be positive; positive at "
```
a-corni: nit: I suggest to have "; it is positive at ..."

pulser-core/pulser/channels/base_channel.py
```diff
@@ -63,6 +63,18 @@ def get_states_from_bases(bases: Collection[str]) -> list[States]:
...
+    inds = np.flatnonzero(mask)
+    breaks = np.flatnonzero(np.diff(inds) > 1)
+    starts = np.concatenate(([inds[0]], inds[breaks + 1]))
+    ends = np.concatenate((inds[breaks], [inds[-1]]))
+    return ", ".join(
+        f"{start}-{end} ns" if start != end else f"{start} ns"
+        for start, end in zip(starts.tolist(), ends.tolist())
+    )
+
+
```
a-corni, line 77: I think it's good, may I suggest that you add additional cases? So far, only pulses with a amplitude > max_amp/detuning > max_abs_detuning has been tested. I would like to see the error message for a CustomWaveform([0, 0, max_amp+11, 0, max_amp, max_amp+10, max_amp+9, max_amp+8]). I expect to get '2 ns, 5-7 ns'.

jaewonyun1234:
Applied both, thanks @a-corni! The new case gives exactly the `2 ns, 5-7 ns` you expected — the sample equal to `max_amp` is not reported, only the ones above it.

a-corni (approved):
LGTM! Thanks a lot @jaewonyun1234 !

PR #1106, opened 2026-09-05
Report offending input in the simulation error messages

jaewonyun1234:
Hi @a-corni, @HGSilveri,

This covers `pulser_simulation/simulation.py` and `pulser/_hamiltonian_data/hamiltonian_data.py` together: they share eleven byte-identical error messages, so doing one without the other would leave the same error worded two ways depending on the entry point. 13 messages at 17 sites — 4 changed identically in both files, 7 only in `simulation.py`, 2 only in `hamiltonian_data.py`.

Three of them report qubit ids or bases that come from sets. The bases are listed in channel declaration order. The two qubit-id cases have no caller order to preserve — those ids are, by construction, the ones *not* in the register — so they are sorted, with a comment saying why.

Left unchanged: the ten that describe a state rather than a rejected input — empty samples, no declared channels, no instructions, `needs to be built`, and the configuration clashes (`noise_model` + `config`, SLM + modulation, state preparation + initial state, depolarizing in all-basis, relaxation without ground-rydberg).

Partially addresses #1057.

### DIFF
--- a/pulser-core/pulser/_hamiltonian_data/hamiltonian_data.py
+++ b/pulser-core/pulser/_hamiltonian_data/hamiltonian_data.py
@@ -217,14 +217,17 @@ class HamiltonianData:
         # Initializing the samples obj
         if not isinstance(samples, SequenceSamples):
             raise TypeError(
-                "The provided sequence has to be a valid "
-                "SequenceSamples instance."
+                "The provided samples must be an instance of "
+                f"'SequenceSamples', not {type(samples)}."
             )
         if samples.max_duration == 0:
             raise ValueError("SequenceSamples is empty.")
         # Check compatibility of register and device
         if not isinstance(device, BaseDevice):
-            raise TypeError("The device must be a Device or BaseDevice.")
+            raise TypeError(
+                "'device' must be an instance of 'BaseDevice', not "
+                f"{type(device)}."
+            )
         self._device = device
         self.device.validate_register(register)
         self._register = register
@@ -234,14 +237,32 @@ class HamiltonianData:
                 "Samples use SLM mask but device does not have one."
             )
         if not samples.used_bases <= self.device.supported_bases:
+            missing_bases = samples.used_bases - self.device.supported_bases
+            unsupported = [
+                b
+                for b in dict.fromkeys(
+                    ch.basis for ch in samples._ch_objs.values()
+                )
+                if b in missing_bases
+            ]
+            supported = list(
+                dict.fromkeys(ch.basis for ch in self.device.channel_objects)
+            )
             raise ValueError(
-                "Bases used in samples should be supported by device."
+                "Bases used in samples must be supported by the device; "
+                f"{unsupported} not in {supported}."
             )
         # Check compatibility of masked samples and register
         if not samples._slm_mask.targets <= set(self.register.qubits.keys()):
+            # The mask targets are a set, so there is no caller order to keep
+            missing_targets = sorted(
+                samples._slm_mask.targets - set(self.register.qubits.keys()),
+                key=str,
+            )
             raise ValueError(
-                "The ids of qubits targeted in SLM mask"
-                " should be defined in register."
+                "The ids of qubits targeted in the SLM mask must be defined "
+                f"in the register; {missing_targets} not in "
+                f"{list(self.register.qubit_ids)}."
             )
 
         self._samples = self._delocalize_samples(samples)
@@ -281,12 +302,19 @@ class HamiltonianData:
             if samples._ch_objs[ch].addressing == "Local":
                 # Check that targets of Local Channels are defined
                 # in register
-                if not set().union(
+                targets = set().union(
                     *(slot.targets for slot in ch_samples.slots)
-                ) <= set(self.register.qubits.keys()):
+                )
+                if not targets <= set(self.register.qubits.keys()):
+                    # Slot targets are sets: there is no caller order to keep
+                    missing_targets = sorted(
+                        targets - set(self.register.qubits.keys()), key=str
+                    )
                     raise ValueError(
-                        "The ids of qubits targeted in Local channels"
-                        " should be defined in register."
+                        "The ids of qubits targeted by Local channel "
+                        f"{ch!r} must be defined in the register; "
+                        f"{missing_targets} "
+                        f"not in {list(self.register.qubit_ids)}."
                     )
                 samples_list.append(ch_samples)
             else:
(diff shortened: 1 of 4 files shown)


pulser-core/pulser/_hamiltonian_data/hamiltonian_data.py
```diff
@@ -217,14 +217,17 @@ def __init__(
         # Initializing the samples obj
         if not isinstance(samples, SequenceSamples):
             raise TypeError(
-                "The provided sequence has to be a valid "
-                "SequenceSamples instance."
+                "The provided samples must be an instance of "
+                f"'SequenceSamples', not {type(samples)}."
```
a-corni, line 221: return the failing input

```suggestion
                f"'SequenceSamples', not {type(samples)}. Got {samples}."
```

pulser-core/pulser/_hamiltonian_data/hamiltonian_data.py
```diff
@@ -217,14 +217,17 @@ def __init__(
...
             )
         if samples.max_duration == 0:
             raise ValueError("SequenceSamples is empty.")
         # Check compatibility of register and device
         if not isinstance(device, BaseDevice):
-            raise TypeError("The device must be a Device or BaseDevice.")
+            raise TypeError(
+                "'device' must be an instance of 'BaseDevice', not "
+                f"{type(device)}."
```
a-corni, line 229: Return the failing device 
```suggestion
                f"{type(device)}. Got {device}."
```

pulser-core/pulser/_hamiltonian_data/hamiltonian_data.py
```diff
@@ -234,14 +237,32 @@ def __init__(
...
+                )
+                if b in missing_bases
+            ]
+            supported = list(
+                dict.fromkeys(ch.basis for ch in self.device.channel_objects)
+            )
             raise ValueError(
-                "Bases used in samples should be supported by device."
+                "Bases used in samples must be supported by the device; "
+                f"{unsupported} not in {supported}."
```
a-corni, line 253: It would be good to return the bases used by the samples as well :) 
```suggestion
                f"Samples uses {samples.used_bases} but basis {unsupported} is not among basis supported by the Device {supported}."
```

pulser-core/pulser/_hamiltonian_data/hamiltonian_data.py
```diff
@@ -234,14 +237,32 @@ def __init__(
...
             )
         # Check compatibility of masked samples and register
         if not samples._slm_mask.targets <= set(self.register.qubits.keys()):
+            # The mask targets are a set, so there is no caller order to keep
+            missing_targets = sorted(
+                samples._slm_mask.targets - set(self.register.qubits.keys()),
+                key=str,
+            )
```
a-corni, line 261: I would even say: there is no order to keep! You can delete the sorted ... key=str, just show the difference :)

pulser-core/pulser/_hamiltonian_data/hamiltonian_data.py
```diff
@@ -234,14 +237,32 @@ def __init__(
...
             raise ValueError(
-                "The ids of qubits targeted in SLM mask"
-                " should be defined in register."
+                "The ids of qubits targeted in the SLM mask must be defined "
+                f"in the register; {missing_targets} not in "
```
a-corni, line 264: Idem, we should show the list of all the provided samples._slm_mask.targets

pulser-core/pulser/_hamiltonian_data/hamiltonian_data.py
```diff
@@ -234,14 +237,32 @@ def __init__(
...
+            )
             raise ValueError(
-                "Bases used in samples should be supported by device."
+                "Bases used in samples must be supported by the device; "
+                f"{unsupported} not in {supported}."
             )
         # Check compatibility of masked samples and register
         if not samples._slm_mask.targets <= set(self.register.qubits.keys()):
+            # The mask targets are a set, so there is no caller order to keep
+            missing_targets = sorted(
```
a-corni, line 258: nit: it's not so much `missing_targets`, more like `unknown_targets`: the source of truth is the register, the error message is about the SLM mask targetting unknow qubit ids.

pulser-core/pulser/_hamiltonian_data/hamiltonian_data.py
```diff
@@ -281,12 +302,19 @@ def _delocalize_samples(self, samples: SequenceSamples) -> SequenceSamples:
...
-                if not set().union(
+                targets = set().union(
                     *(slot.targets for slot in ch_samples.slots)
-                ) <= set(self.register.qubits.keys()):
+                )
+                if not targets <= set(self.register.qubits.keys()):
+                    # Slot targets are sets: there is no caller order to keep
+                    missing_targets = sorted(
+                        targets - set(self.register.qubits.keys()), key=str
+                    )
```
a-corni, line 312: Same comments as above: it's more about unkown targets (register is source of truth) + no need to use sorted
```suggestion
                    # Slot targets are sets: there is no caller order to keep
                    unknown_targets = targets - set(self.register.qubits.keys())
```

pulser-core/pulser/_hamiltonian_data/hamiltonian_data.py
```diff
@@ -281,12 +302,19 @@ def _delocalize_samples(self, samples: SequenceSamples) -> SequenceSamples:
...
                     raise ValueError(
-                        "The ids of qubits targeted in Local channels"
-                        " should be defined in register."
+                        "The ids of qubits targeted by Local channel "
+                        f"{ch!r} must be defined in the register; "
+                        f"{missing_targets} "
```
a-corni, line 316:
```suggestion
                        f"Channel targets {targets}. Among them, {missing_targets} "
```

pulser-core/pulser/_hamiltonian_data/hamiltonian_data.py
```diff
@@ -359,8 +387,8 @@ def from_sequence(
         """
         if not isinstance(sequence, Sequence):
             raise TypeError(
-                "The provided sequence has to be a valid "
-                "pulser.Sequence instance."
+                "'sequence' must be an instance of 'Sequence', not "
+                f"{type(sequence)}."
```
a-corni, line 391:
```suggestion
                f"{type(sequence)}. Got {sequence}."
```

jaewonyun1234:
Thanks a lot for the review @a-corni, these were really helpful! I've applied all of them, in both `hamiltonian_data.py` and `simulation.py` so the shared messages stay identical.

One small note on removing the `sorted()`: with two or more unknown ids the set prints in a different order on each run, so a test pinning several of them would be flaky — I've written the two affected assertions to accept any order.

